PR #11, opened 2019-03-27
add client proxy and proxy server example

disksing:


proxy/httpproxy/handler.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// NewHTTPProxyHandler creates an http.Handler that serves as a TiKV client proxy.
+func NewHTTPProxyHandler() http.Handler {
+	router := mux.NewRouter()
+	rawkv := rawkvHandler{p: proxy.NewRaw()}
+	router.HandleFunc("/rawkv/new", rawkv.handlerFunc(rawkv.New))
+	router.HandleFunc("/rawkv/close", rawkv.handlerFunc(rawkv.Close))
+	router.HandleFunc("/rawkv/get", rawkv.handlerFunc(rawkv.Get))
+	router.HandleFunc("/rawkv/batch.get", rawkv.handlerFunc(rawkv.BatchGet))
```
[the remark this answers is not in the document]
disksing: Or `batch-get`?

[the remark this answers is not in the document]
rleungx: Yes. For now, both TiDB and PD use the hyphen.

proxy/httpproxy/handler.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/gorilla/mux"
+	"github.com/tikv/client-go/proxy"
+)
+
+// NewHTTPProxyHandler creates an http.Handler that serves as a TiKV client proxy.
+func NewHTTPProxyHandler() http.Handler {
+	router := mux.NewRouter()
+	rawkv := rawkvHandler{p: proxy.NewRaw()}
+
+	router.HandleFunc("/rawkv/client/new", rawkv.handlerFunc(rawkv.New))
```
siddontang: em, I prefer using raw and txn directly

disksing: ok

disksing: em, `/txn/txn/{id}/get` will be strange then, `txnkv` seems to still be the better option. @siddontang

siddontang: IMO, why not providing a gRPC support instead of Restful. For Restful, we have to care how to manage transaction ID, if the client forgets to rollback the transaction, we have to rollback it soon.

But in gRPC, we can manage the transaction in a session, it is more simple.

disksing: Not quite familiar with _grpc session_ you mentioned. But I think HTTP is a better idea than gRPC because it is supported by almost all modern languages. In general the developer of the client can introduce it with less effort.

proxy/txnkv.go
```diff
@@ -0,0 +1,274 @@
...
+		return 0, errors.WithStack(ErrTxnNotFound)
+	}
+	return txn.(*txnkv.Transaction).Size(), nil
+}
+
+// IterValid returns if the iterator is valid to use.
+func (p TxnKVProxy) IterValid(id UUID) (bool, error) {
+	iter, ok := p.iterators.Load(id)
+	if !ok {
+		return false, errors.WithStack(ErrTxnNotFound)
```
rleungx: ErrIterNotFound?

PR #13, opened 2019-04-16
add rawkv/ReverseScan

kamijin-fanta:
followed original code.

https://github.com/pingcap/tidb/blob/7666f688b3d91ea84fa6b0a20c2e784fff3c8945/store/tikv/rawkv.go#L313-L358


disksing:
@kamijin-fanta Thanks for your contribution!

PR #14, opened 2019-04-18
config: cleanup and pass config as parameter

disksing:


rpc/client.go
```diff
@@ -142,32 +144,33 @@ func (c *batchCommandsClient) batchRecvLoop() {
...
 	}
 }
 
-func newConnArray(maxSize uint, addr string, security config.Security) (*connArray, error) {
+func newConnArray(addr string, conf *config.RPC) (*connArray, error) {
 	a := &connArray{
+		conf:                 conf,
 		index:                0,
-		v:                    make([]*grpc.ClientConn, maxSize),
+		v:                    make([]*grpc.ClientConn, conf.MaxCallMsgSize),
```
rleungx: Maybe this variable name is not good?

txnkv/store/txn_committer.go
```diff
@@ -183,10 +186,10 @@ func (c *TxnCommitter) doActionOnKeys(bo *retry.Backoffer, action commitAction,
 		atomic.AddInt32(&c.detail.PrewriteRegionNum, int32(len(groups)))
 	}
 	// Make sure the group that contains primary key goes first.
-	batches = appendBatchBySize(batches, firstRegion, groups[firstRegion], sizeFunc, config.TxnCommitBatchSize)
+	batches = appendBatchBySize(batches, firstRegion, groups[firstRegion], sizeFunc, c.conf.Txn.CommitBatchSize)
```
rleungx: For such kind of things, how about using a variable to make it look more concise?

txnkv/kv/memdb_buffer.go
```diff
@@ -43,12 +42,15 @@ type memDbIter struct {
 }
 
 // NewMemDbBuffer creates a new memDbBuffer.
-func NewMemDbBuffer(cap int) MemBuffer {
+func NewMemDbBuffer(conf *config.Txn, cap int) MemBuffer {
+	if cap == 0 {
```
rleungx: How about `<=`?

config/raw.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2018 PingCAP, Inc.
```
rleungx: Do we need to change it to 2019?

txnkv/kv/memdb_buffer.go
disksing: ok.

config/raw.go
disksing: ok.

txnkv/store/txn_committer.go
disksing: ok.

rpc/client.go
disksing: How about `GrpcMaxCallMsgSize`?

rleungx: Hmmm...Actually, I mean the `v` here.

PR #23, opened 2019-06-17
Use context everywhere

### LINKED ISSUE #21: All methods that do IO should take context.Context as arguments
Some methods, such as `TiKVSnapshot.Get`, perform I/O using a `context.Background()`. This is non-desirable since if I/O is done in response to a client request, and the request is canceled/timesout then the TiKV operation will continue processing, and thus consuming up background resources.

disksing:
Fixes #21

### DIFF
--- a/examples/rawkv/rawkv.go
+++ b/examples/rawkv/rawkv.go
@@ -14,6 +14,7 @@
 package main
 
 import (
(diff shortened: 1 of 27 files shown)


disksing:
cc @shafreeck

PR #33, opened 2019-11-01
Delete some unused code

wangrzneu:
the check of kv.IsErrNotFound(err) is unused in Transaction.Get


disksing:
Thanks @wangrzneu

PR #35, opened 2019-11-21
Adding README

c4pt0r:


README.md
```diff
@@ -0,0 +1,352 @@
...
+## Raw Key-Value API
+
+To use the Raw Key-Value API in applications developed in the Go language, take the following steps:
+
+1. Install the necessary packages.
+
+    ```bash
+    export GO111MODULE=on
+    go mod init rawkv-demo
+    go get github.com/pingcap/tidb@master
```
disksing: why use tidb instead of `tikv/client-go`?

c4pt0r: Sorry, I just copy from the tikv website client-go page, it's out-of-date, I think we should update that too.

disksing: Actually, we have provided simple examples https://github.com/tikv/client-go/tree/master/examples. I think add some links should be enough.

Issue #18, opened 2019-04-23
proxy: support setting configurations


disksing:

wangrzneu:
I think the we can make config as arguments when client sends /rawkv/client/new or /txnkv/client/new requests.

disksing:
closing due to stale.

PR #40, opened 2019-12-17
support setting config after creating http proxy handler

wangrzneu:
support setting config after creating http proxy handler

Signed-off-by: renzheng.wang <renzheng.wang@ucloud.cn>


proxy/httpproxy/rawkv.go
```diff
@@ -49,7 +50,13 @@ type RawResponse struct {
 }
 
 func (h rawkvHandler) New(ctx context.Context, r *RawRequest) (*RawResponse, int, error) {
-	id, err := h.p.New(ctx, r.PDAddrs, config.Default())
+	var id proxy.UUID
+	var err error
+	if h.c == nil {
```
disksing: It seems it will never be nil.

disksing:
Thanks @wangrzneu . Your PR will fix https://github.com/tikv/client-go/issues/18 after get merged.

proxy/httpproxy/rawkv.go
wangrzneu: I think this check can make the code safe, although h.c cannot be nil for current usage of rawkvHandler.

disksing:
I think you did not add the method to allow client to update the config...

proxy/httpproxy/rawkv.go
disksing: em, how about add a `getConfig()` method to `rawkvHandler`, inside the method, when c is nil, return the default config. In this way, you will not have to pay extra attention in `newHTTPProxyHandler`.

wangrzneu:
The config can be updated after creating HTTPProxyHandler

wangrzneu:
Do we need to update the config through the http calling?

proxy/httpproxy/txnkv.go
```diff
@@ -269,3 +270,10 @@ func (h txnkvHandler) handlerFunc(f func(context.Context, *TxnRequest) (*TxnResp
 		w.Write(data)
 	}
 }
+
+func (h txnkvHandler) getConfig() config.Config {
+	if h.c == nil {
+		return config.Default()
+	}
+	return *h.c
+}
```
disksing: please end the file with an empty new line.

disksing:
I think I missunderstood your intention.. #18 means to update config by HTTP request. But I think we can merge your PR then handle #18 later.

wangrzneu:
> I think I missunderstood your intention.. #18 means to update config by HTTP request. But I think we can merge your PR then handle #18 later.

I can propose a PR after current PR is merged.

disksing:
Thanks for your contribution! @wangrzneu

PR #43, opened 2020-01-08
fix RowKVClient scan method can't rightly break loop

ldnvnbl:
Signed-off-by: Yushen Lin <ldnvnbl@gmail.com>

fix https://github.com/pingcap/tidb/issues/14363

### DIFF
--- a/rawkv/rawkv.go
+++ b/rawkv/rawkv.go
@@ -288,6 +288,10 @@ func (c *Client) Scan(ctx context.Context, startKey, endKey []byte, limit int) (
 		if len(startKey) == 0 {
 			break
 		}
+
+		if len(endKey) > 0 && bytes.Compare(startKey, endKey) > 0 {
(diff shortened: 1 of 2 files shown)


rawkv/rawkv.go
```diff
@@ -307,7 +307,7 @@ func (c *Client) ReverseScan(ctx context.Context, startKey, endKey []byte, limit
 		return nil, nil, errors.WithStack(ErrMaxScanLimitExceeded)
 	}
 
-	for len(keys) < limit {
+	for len(keys) < limit && bytes.Compare(startKey, endKey) > 0 {
```
shafreeck: Is it necessary to check len(endKey)?

ldnvnbl: No, if endKey is nil or empty slice, bytes.Compare always gt 0 except startKey is nil too.

Issue #44, opened 2020-04-03
Feature request: return found keys without values


davies:
Sometimes, we may just want the list of keys, then fetch the value for selected ones. The current API will return both keys and values, which is very slow if value is large.

disksing:
Hi @davies , if you are using the txn API, you can try the `KeyOnly` option: https://github.com/tikv/client-go/blob/a3ebdb020c83b86c36377d20d31e15db6168413d/txnkv/txn.go#L166

davies:
We are using the rawkv API, can we mixed them (only use txn API to scan) ?

disksing:
They cannot be used in same cluster together. It is possible to support some options for rawkv api, the problems is that is it not on our roadmap. I'd like to create a help-needed issue for now. You may consider lend a hand if too :-)

Connor1996:
Once https://github.com/tikv/tikv/issues/7614 finished, you can leverage it to reduce io cost.

disksing:
Looks it is cancelled on server side. closing the issue.

PR #47, opened 2020-04-29
Add Keyonly option for rawKv Scan

TommyCpp:
As mention in #44 and in tikv/tikv#7673. It would helpful to have some kind of options for RawKv.

However, in order to avoid the breaking change and maintain API stability. We cannot just add a options parameter in RawKv API functions. We can put options somewhere in client object
implement another version of RawKv API functions with Option parameters, like GetWithOptions.
I choose to leave the Option in Config in this PR, but feel free to let me know if you feel other choice is better.

### DIFF
--- a/config/raw.go
+++ b/config/raw.go
@@ -24,7 +24,7 @@ type Raw struct {
 	// BatchPairCount is the maximum limit for rawkv each batch get/delete request.
 	BatchPairCount int
 
-	// Return only keys when scan
+	// If true, the Server will only return keys when using Scan function
 	KeyOnlyScan bool
 }
 
--- a/config/raw.go
+++ b/config/raw.go
@@ -23,6 +23,9 @@ type Raw struct {
 
 	// BatchPairCount is the maximum limit for rawkv each batch get/delete request.
 	BatchPairCount int
+
+	// Return only keys when scan
+	KeyOnlyScan bool
 }
 
 // DefaultRaw returns default rawkv configuration.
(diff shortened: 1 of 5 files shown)


TommyCpp:
@Connor1996 PTAL

Connor1996:
PTAL @disksing

TommyCpp:
> Please add a related test

Test added. Also change the mockTikv to mimic the behavior in server end.

mockstore/mocktikv/rpc.go
```diff
@@ -476,6 +476,16 @@ func (h *rpcHandler) handleKvRawScan(req *kvrpcpb.RawScanRequest) *kvrpcpb.RawSc
 		endKey = req.EndKey
 	}
 	pairs := rawKV.RawScan(req.GetStartKey(), endKey, int(req.GetLimit()))
+	if req.KeyOnly {
+		//filter values when
```
Connor1996, line 480: please fix this comment

config/raw.go
```diff
@@ -23,6 +23,9 @@ type Raw struct {
 
 	// BatchPairCount is the maximum limit for rawkv each batch get/delete request.
 	BatchPairCount int
+
+	// If true, the Server will only return keys when using Scan function
+	KeyOnlyScan bool
```
Connor1996, line 28: seems it makes all scan key only. maybe it's better to add an argument for Scan API, though it breaks the compatibility. /cc @disksing

disksing: agree, I think option is better. Consider to use variable argument to avoid capatibility problem.

disksing: you can use a config structure as variable argument, they are not conflict.

Connor1996, line 28: please remove this

PR #54, opened 2020-09-02
init buffer when create buffer store

cxt90730:
When using txn client in multiple threads, transaction loss often occurs. Because lazyMemBuffer is not initialized during concurrent access, the transaction will be lost.

Signed-off-by: Cuixiaotian <cxt90730@126.com>

### DIFF
--- a/txnkv/kv/buffer_store.go
+++ b/txnkv/kv/buffer_store.go
@@ -35,8 +35,11 @@ type BufferStore struct {
 // NewBufferStore creates a BufferStore using r for read.
 func NewBufferStore(r Retriever, conf *config.Txn) *BufferStore {
 	return &BufferStore{
-		r:         r,
-		MemBuffer: &lazyMemBuffer{conf: conf},
+		r: r,
+		MemBuffer: &lazyMemBuffer{
+			mb:   NewMemDbBuffer(conf, 0),
+			conf: conf,
+		},
 	}
 }
 


disksing:
@cxt90730 Transactions are designed to be only used by a single thread, you need to wrap it with some locks in order to support multi-threading. (btw, clients can safely be accessed by multiple threads)

disksing:
Even if consider supporting multi-threading, this PR is an incomplete fix, so I'm going to close this PR.

cxt90730:
Maybe I didn't describe it clearly. Every time I just create one global txn, and then use this txn to concurrently execute operations ( such as txn.Delete(key) ), wait for all operations to complete, and finally txn.Commit(). I will find that several keys may  have not been deleted. @disksing

cxt90730:
The code like this:
c := NewTiKVClient()
// Put 1000 keys. k:0 -> k:999
for i := 0; i < 1000; i++ {
		k, v := "k:"+strconv.Itoa(i), "v:"+strconv.Itoa(i)
		c.TxPut(k, v)
}
//c.ScanResult()
tx, err := c.TxnCli.Begin(context.TODO())
if err != nil {
	panic(err)
}


wg := sync.WaitGroup{}
for i := 0; i < 1000; i++ {
	wg.Add(1)
	go func(i int) {
		wg.Done()
		k := "k:"+strconv.Itoa(i)
		err := tx.Delete([]byte(k))
		if err != nil {
			panic(err)
		}
	}(i)
}
wg.Wait()
if err == nil {
	err = tx.Commit(context.Background())
}
if err != nil {
	tx.Rollback()
}
fmt.Println("=====")
c.ScanResult()

########
You can see some keys are not deleted:
image

cxt90730:
I haven't found a better solution, at least the code in PR can resolve this problem. That's all :)

disksing:
Hi, @cxt90730 . You should use multiple goroutines to access the same Txn. The Txn is designed not to be used like that. To delete keys concurrently, you can create transaction for each goroutine, then it will be safe.

cxt90730:
Ok, I just want to delete 1000 keys as an atomic operation, no matter the `tx.Delete(key)` is executed asynchronously or synchronously.

The following is just to discuss the problem. Don't mind.

From the code level, the reason why goroutine cannot be used is because the `lazymembuffer`, the cache of transaction, is not initialized at the beginning. I want to know the reason of this design is to save memory or for other reasons?

cxt90730:
In addition, create transaction for each goroutine means to get the timestamp multiple times from pd. I found it may cause a warning: `get timestamp too slow`. I think as an atomic operation, it is better to only take the timestamp only once. Hope your feedback @disksing

disksing:
> From the code level, the reason why goroutine cannot be used is because the `lazymembuffer`, the cache of transaction, is not initialized at the beginning. I want to know the reason of this design is to save memory or for other reasons?

The `lazyMemBuffer` is not the only reason that a txn cannot support multi-goroutine. There are other places, such as `unionstore` makes it unable to be accessed by multiple goroutines.

I think to delete 1000 keys, it is unnecessary to introduce concurrency. Before a txn is commit, all writes are pure memory operations, which should be fast enough.

cxt90730:
@disksing  要不还是说中文吧，情况比较复杂，我英文编起来太慢。。。

先说下我这里相关的业务情况：
1. list 1000 keys
2. 对于每个key，伪代码操作如下：

// 1
 for i := range keys {
    go func() {
      // 2
      info_i = tx.Get( key[i] )
      tx.Delete( key[i] ) 
      tx.Put(  info_i )    // 业务可理解为增加删除记录
     // 3
    } 
 }
// 4

由于包含tx.Get( key[i] )操作，如果是串行，性能达不到要求，所以需要套用goroutine异步执行

之后问题来了！ 
如果tx.NewTrans()和tx.Commit() 放在for循环外(位置1和4)，也就是之前说的设置为全局Tx，就会出现之前和你说的少删key的情况； 
但如果tx.NewTrans()和tx.Commit() 放在goroutine里(位置2和3)，也就是像你之前说的每个goroutine新建一个Tx，就会报如下的WriteConflict错误。令我不解的是，无论是tx.Delete(key[i]) 还是tx.Put(info_i)，**我能确保每一个操作的key都是不一样的**，为什么会有写冲突？

```
[try again later]: WriteConflict: startTS=418174642014913561, conflictTS=418174642014913560, key="DeleteRecord/95170:1910775", primary="Key/node2_user6283842877737074242"

[try again later]: WriteConflict: startTS=418174642014913987, conflictTS=418174642014913985, key="Key/node2_user6285776918690718941", primary="Key/node2_user6285776918690718941
```

disksing:
严重怀疑你这个是 goroutine 写法的问题！

```
for i := range keys {
    go func() {
        print(i)
    }()
}
```

这么写会看到重复的 i，因为 goroutine 启动的时候，i 已经变了。

要这么写

```
for i := range keys {
    go func(i int) {
        print(i)
    }(i)
}
```

或者搞个临时变量

```
for i := range keys {
    x := i
    go func() {
        print(x)
    }()
}
```

cxt90730:
@disksing 吓得我赶紧回溯了下之前的commit，确定只是伪代码写的不对，实际的代码并没有闭包传参问题哈

目前是我回滚到之前的commit，WriteConflict的问题复现不出来了，只是报大量的 get timestamp too slow，至于当时怎么出来WriteConflict的，过了几个月了确实想不起来了。因为当时的认知是当并发的对同一个key进行写/删操作才会报该异常，但当时是能确定key肯定是不一样的，而由于赶时间就没深究了。

感谢你一直抽出宝贵的时间，并且耐心的反馈和指导，如果以后有机会能复现，我会及时cue你:)

disksing:
ok ok.

Issue #60, opened 2020-10-26
how about vendor pingcap/pd/client?


cosven:
I'm trying to use the client-go(rawkv client) to read/write data from/to **write CF(not only default CF)**. However, I run into some problems.

In my project, I use `tikv/client-go -- rawkv client` to read the write CF to get a key-value, which is written by TiDB server. I use the latest `pingcap/tidb -- tablecodec/codec` to encode and decode those keys and values. And the problem is that `pingcap/tidb` depends on a recent `pingcap/pd` and `client-go` depends on a old version `pingcap/pd`. We can't put two version pingcap/pd in one project because pd-client will register **unqiue** metrics collector during packge init.
https://github.com/tikv/pd/blob/fc997de0ae38c22dfaab0979dc0f64a66689b286/client/metrics.go#L47-L51
If we have two versions, it panic like this
```
panic: duplicate metrics collector registration attempted

goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*Registry).MustRegister(0xc000180be0, 0xc000076f40, 0x1, 0x1)
        /home/cosven/go/pkg/mod/github.com/prometheus/client_golang@v1.5.1/prometheus/registry.go:400 +0xad
github.com/prometheus/client_golang/prometheus.MustRegister(...)
        /home/cosven/go/pkg/mod/github.com/prometheus/client_golang@v1.5.1/prometheus/registry.go:177
github.com/tikv/client-go/pd.init.0()
        /home/cosven/coding/client-go/pd/metrics.go:35 +0x91
```
-----------------

I have tried to upgrade pingcap/pd in client-go, but it failed. It is really hard to upgrade pingcap/pd in client-go because the recent `pd.Client` interface in `pingcap/pd` has many changes(for example, many more methods to implement).

I find that client-go only depends `pd.Client`. I think client-go need not to depend on pingcap/pd and vendoring pingcap/pd/client maybe a good idea. pd/client is relative simple and vendoring pingcap/pd/client will bring much convenience for maintaintence.

PR #62, opened 2020-11-16
*: update pd client dependency

disksing:

related to https://github.com/tikv/client-go/issues/60

When someone needs to import tidb and go client at the same time, this PR should resolve to metrics conflict problem.

### DIFF
--- a/mockstore/mocktikv/pd.go
+++ b/mockstore/mocktikv/pd.go
@@ -18,9 +18,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/pkg/errors"
 	"github.com/pingcap/kvproto/pkg/metapb"
-	"github.com/pingcap/pd/client"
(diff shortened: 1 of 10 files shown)


mockstore/mocktikv/pd.go
```diff
@@ -46,6 +47,14 @@ func (c *pdClient) GetClusterID(ctx context.Context) uint64 {
 	return 1
 }
 
+func (c *pdClient) GetAllMembers(ctx context.Context) ([]*pdpb.Member, error) {
+	return nil, nil
```
cosven, line 51: how about just `panic` here like other unimplemented methods?

cosven (approved):
LGTM

mockstore/mocktikv/pd.go
disksing: good idea.

cosven:
LGTM

PR #71, opened 2021-02-20
Export TTL field

Connor1996:
Export the TTL field of PutOption


rawkv/rawkv.go
```diff
@@ -40,7 +40,7 @@ type ScanOption struct {
 
 // PutOptions is used to provide additional information for put operation.
 type PutOption struct {
-	ttl uint64
+	Ttl uint64
```
disksing, line 43: TTL?

disksing (approved):

PR #72, opened 2021-03-02
check region not nil to avoid nil pointer dereference panic

yaozongyou:
Signed-off-by: Yao Zongyou <yaozongyou@vip.qq.com>


yaozongyou:
according to https://github.com/tikv/pd/blob/master/client/client.go#L64, GetRegion may return nil region, so nil should be checked to avoid nil pointer dereference panic.

image

disksing (approved):

disksing:
Thanks @yaozongyou

PR #78, opened 2021-06-11
*: move go files in root directory to tikv

disksing:

- It can make root directory more clean.
- It is more easy to use. User can `import "github.com/tikv/client-go/v2/tikv"` instead of `import tikv "github.com/tikv/client-go/v2"`


AndreMouche (approved):
LGTM

Issue #97, opened 2021-06-21
Replace `github.com/pingcap/check` with `https://github.com/stretchr/testify`


disksing:
`testify` is more widely used and tools friendly.

- [x] /client
- [x] /config
- [x] /latch
- [x] /locate
- [x] /mockstore, /mockstore/deadlock
- [x] /mockstore/mocktikv
- [x] /oracle/oracles
- [x] /tikv
- [x] /unionstore/memdb{,_norace}_test.go
- [x] /integration_tests/1pc_test.go
- [x] /integration_tests/2pc{,_slow,_fail}_test.go
- [x] /integration_tests/async_commit{,_fail}_test.go
- [x] /integration_tests/{delete_range,isolation}_test.go
- [x] /integration_tests/lock_test.go
- [x] /integration_tests/range_task_test.go
- [x] /integration_tests/rawkv_test.go
- [x] /integration_tests/{safepoint,scan,scan_mock}_test.go
- [x] /integration_tests/snapshot{,_fail}_test.go
- [x] /integration_tests/split_test.go
- [x] /integration_tests/store{,_fail}_test.go
- [x] /integration_tests/ticlient{,_slow}_test.go

tisonkun:
That sounds good. I'd like to volunteer to this work.

From my perspective, we may first bring the testing dep and basic test framework to the project and later port all its tests to the new test framework.

Do you want to do the first step?

disksing:
Sure. I'll replace a subdirectory. Then we can move forward based on directories. 

I think to follow up we can leave a comment here before we are ready to do it in order to avoid conflicts.

tisonkun:
For "leave a comment", it would be better to clarify that a comment ref to subtask.

You may take a look at [FLINK-10392](https://issues.apache.org/jira/browse/FLINK-10392) and although GitHub doesn't have explicit subtasks supporting but we can use refer to achieve it. See also https://github.com/tikv/community/issues/121

disksing:
> For "leave a comment", it would be better to clarify that a comment ref to subtask.
> 
> You may take a look at [FLINK-10392](https://issues.apache.org/jira/browse/FLINK-10392) and although GitHub doesn't have explicit subtasks supporting but we can use refer to achieve it. See also [tikv/community#121](https://github.com/tikv/community/issues/121)

Good point. I'm going to break task into small units and create a set of issues.

tisonkun:
[...]

PR #103, opened 2021-06-21
util: replace pingcap/check with testify

disksing:

ref https://github.com/tikv/client-go/issues/97


TennyZhuang:
LGTM

disksing:
@TennyZhuang Thanks for reviewing, but we are using GitHub approve now.

PR #104, opened 2021-06-21
examples/rawkv: add example for rawkv

AndreMouche:
add example for rawkv


examples/rawkv/rawkv.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2019 PingCAP, Inc.
```
Xuanwo, line 1: Code should belong to tikv (aka CNCF) instead.

disksing: Copyright 2021 TiKV Authors

examples/rawkv/rawkv.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2021 PingCAP, Inc.
```
disksing, line 1: PingCAP, Inc.  -> TiKV Authors

examples/rawkv/rawkv.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+
+	"github.com/tikv/client-go/v2/config"
+	"github.com/tikv/client-go/v2/tikv"
+)
+
+func main() {
+	cli, err := tikv.NewRawKVClient([]string{"172.16.6.153:2379"}, config.DefaultConfig().Security)
```
disksing, line 24: better to use some IP like 127.0.0.1

disksing (approved):

Issue #105, opened 2021-06-21
Add PrefixNextKey API on v2 branch


gotoxu:
PrefixNextKey returns the next prefix key.

Assume there are keys like:
* rowkey1
* rowkey1_column1
* rowkey1_column2
* rowKey2

If we seek 'rowkey1' NextKey, we will get 'rowkey1_column1'.
If we seek 'rowkey1' PrefixNextKey, we will get 'rowkey2'.

PR #106, opened 2021-06-21
Add PrefixNextKey API (#105)

gotoxu:

fix issue: #105


disksing:
Any idea about this issue? https://github.com/pingcap/tidb/issues/12598

gotoxu:
> Any idea about this issue? [pingcap/tidb#12598](https://github.com/pingcap/tidb/issues/12598)
> Maybe return empty slice is more correct but I don't know if it can break anything.

Do you know why tidb still has not fixed this bug?

disksing:
I have no idea. Maybe we can fix it here and in the comment, we declare that it is different from TiDB.

gotoxu:
@disksing I fixed it and stated in the comments

gotoxu:
Please deal with this PR as soon as possible, we need to use this API, thank you!

/cc @disksing

disksing (approved):

disksing:
Thanks @gotoxu

Issue #108, opened 2021-06-21
Add BatchGet API for KVTxn 


gotoxu:
BatchGet gets kv from the memory buffer of statement and transaction, and the kv storage.

gotoxu:
Already done!

PR #109, opened 2021-06-21
tikv/txn: add the method BatchGet (#108)

gotoxu:

fix issue: #108


disksing:
Hi @gotoxu , would you like to add some test?

gotoxu:
Hi @disksing , I have added a test case for BufferBatchGetter!

disksing:
Hi @gotoxu , please add license headers to each new file. It looks that the code is mainly copied from tidb project, you will need to keep original TiDB license and a reference link to original file. Just like I have done in https://github.com/tikv/client-go/pull/110

tikv/txn.go
```diff
@@ -175,6 +175,18 @@ func (txn *KVTxn) Get(ctx context.Context, k []byte) ([]byte, error) {
 	return ret, nil
 }
 
+// BatchGet gets kv from the memory buffer of statement and transaction, and the kv storage.
+// Do not use len(value) == 0 or value == nil to represent non-exist.
+// If a key doesn't exist, there shouldn't be any corresponding entry in the result map.
+func (txn *KVTxn) BatchGet(ctx context.Context, keys [][]byte) (map[string][]byte, error) {
+	if span := opentracing.SpanFromContext(ctx); span != nil && span.Tracer() != nil {
```
disksing, line 182: It seems that other methods don't use opentracing. I think we can remove it and only record it in tidb side.

disksing:
Another question, do you have plan to update TiDB code to use the new method?

tikv/batch_getter_test.go
```diff
@@ -0,0 +1,104 @@
...
+	s.value = append(s.value, v)
+	return nil
+}
+
+func (s *mockBatchGetterStore) Delete(k []byte) error {
+	s.Set(k, []byte{})
+	return nil
+}
+
+func (s *mockBatchGetterStore) Iter(k []byte, upperBound []byte) (unionstore.Iterator, error) {
```
AndreMouche, line 98: Could we remove this function?

gotoxu: This method is implemented because the parameters of the `NewUnionStore` method need to implement the `uSnapshot` interface:
```Go
// uSnapshot defines the interface for the snapshot fetched from KV store.
type uSnapshot interface {
	// Get gets the value for key k from kv store.
	// If corresponding kv pair does not exist, it returns nil and ErrNotExist.
	Get(ctx context.Context, k []byte) ([]byte, error)
	// Iter creates an Iterator positioned on the first entry that k <= entry's key.
	// If such entry is not found, it returns an invalid Iterator with no error.
	// It yields only keys that < upperBound. If upperBound is nil, it means the upperBound is unbounded.
	// The Iterator must be Closed after use.
	Iter(k []byte, upperBound []byte) (Iterator, error)

	// IterReverse creates a reversed Iterator positioned on the first entry which key is less than k.
	// The returned iterator will iterate from greater key to smaller key.
	// If k is nil, the returned iterator will be positioned at the last key.
	// TODO: Add lower bound limit
	IterReverse(k []byte) (Iterator, error)
}
```
In fact, I just need to create a `MemDB` object. Can we consider exposing the `newMemDB` method?

tikv/txn.go
gotoxu: @disksing done!

gotoxu:
> Another question, do you have plan to update TiDB code to use the new method?

We do not use TiDB, we just use TiKV as the storage layer of our entire system, so we need a separate Go client project.

disksing:
> > Another question, do you have plan to update TiDB code to use the new method?
> 
> We do not use TiDB, we just use TiKV as the storage layer of our entire system, so we need a separate Go client project.

Got it. We will create a linked issue at TiDB side and handle it later.

disksing:
LGTM. Please help to take another look. /cc @AndreMouche

tikv/batch_getter_test.go
AndreMouche: How about using `mockBatchGetterStore` as the first argument for the function `NewBufferBatchGetter` since the `mockBatchGetterStore` already implement the interface `BatchBufferGetter`, and also this would make the test easier to understand(the reviewer needn't understand what is `UnionStore` which could not be related to this test).

AndreMouche (commented):
Rest LGTM

tikv/batch_getter_test.go
gotoxu: This is a great idea @AndreMouche

disksing:
Sorry, @gotoxu I have missed one problem. As we are working on replacing `pingcap/check` with `testify` in all tests. Please rewrite the test using `testify`, thanks! You can find some examples in PRs linked to this issue: https://github.com/tikv/client-go/issues/97

gotoxu:
> Sorry, @gotoxu I have missed one problem. As we are working on replacing `pingcap/check` with `testify` in all tests. Please rewrite the test using `testify`, thanks! You can find some examples in PRs linked to this issue: #97

done!

disksing (approved):

disksing:
Thanks @gotoxu , I'll update your branch before merging.

PR #110, opened 2021-06-21
update license

disksing:

Add TiKV Authors header and the references to the original files in TiDB.


AndreMouche (approved):
LGTM

PR #133, opened 2021-06-21
ci: add compatibility test

disksing:

Check if library users can build against the latest `client-go`.


AndreMouche (approved):
LGTM

PR #134, opened 2021-06-21
latch: replace pingcap/check with testify

### LINKED ISSUE #114: Replace check with testify: /latch
subtask of https://github.com/tikv/client-go/issues/97

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>

closes #114 

cc @disksing 

I'm a bit confused by `scheduler_test.go#TestWithConcurrency` since it has no assertion but just validate code executed. We may take a closer look at the tests.

### DIFF
--- a/latch/latch_test.go
+++ b/latch/latch_test.go
@@ -14,99 +14,90 @@
 package latch
 
 import (
-	"github.com/stretchr/testify/suite"
-	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/tikv/client-go/v2/oracle"
+	"go.uber.org/atomic"
 )
 
-func TestT(t *testing.T) {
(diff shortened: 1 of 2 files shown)


disksing:
This is probably just to test that there are no data races. (It also reminded me that I need to add CI test with -race on)

disksing:
Another issue that might be worth discussing is whether to avoid using a test suite when the associated setup code is simple enough, e.g. in our case, the `LatchSuite` setup process is simple enough (only one line) that we can choose not to use a test suite while copying that one line of code to each test.

IMO the big downside to using test suite is that it makes Go believe that there is only one test in the file, which makes it not well compatible with the original go toolchain or other derivatives tools.

tisonkun:
@disksing Yep. It's later last night. Let me give it another pass.

disksing (approved):

disksing:
Thanks @tisonkun

Issue #117, opened 2021-06-21
Replace check with testify: /mockstore/mocktikv


disksing:
subtask of https://github.com/tikv/client-go/issues/97

Issue #116, opened 2021-06-21
Replace check with testify: /mockstore, /mockstore/deadlock


disksing:
subtask of https://github.com/tikv/client-go/issues/97

PR #138, opened 2021-06-22
mocktikv: replace pingcap/check with testify

disksing:

Fix https://github.com/tikv/client-go/issues/117

Also resolve part of https://github.com/tikv/client-go/issues/116

cc @tisonkun


mockstore/mocktikv/mock_tikv_test.go
```diff
@@ -128,502 +100,551 @@ func (s *testMockTiKVSuite) mustDeleteOK(c *C, key string, startTS, commitTS uin
...
-	s.mustRangeScanOK(c, start, "", limit, ts, expect...)
+func mustScanOK(t *testing.T, store MVCCStore, start string, limit int, ts uint64, expect ...string) {
+	mustRangeScanOK(t, store, start, "", limit, ts, expect...)
 }
 
-func (s *testMockTiKVSuite) mustRangeScanOK(c *C, start, end string, limit int, ts uint64, expect ...string) {
-	pairs := s.store.Scan([]byte(start), []byte(end), limit, ts, kvrpcpb.IsolationLevel_SI, nil)
-	c.Assert(len(pairs)*2, Equals, len(expect))
+func mustRangeScanOK(t *testing.T, store MVCCStore, start, end string, limit int, ts uint64, expect ...string) {
+	assert := assert.New(t)
```
tisonkun, line 116: This kind of code introduce the problem `Variable 'assert' collides with imported package name` while it is not quite more convenient than `assert.Xxx(t, )`. I'd prefer avoid this kind of code but not insist.

mockstore/mocktikv/mock_tikv_test.go
```diff
@@ -128,502 +100,551 @@ func (s *testMockTiKVSuite) mustDeleteOK(c *C, key string, startTS, commitTS uin
...
+
 	key := "key"
 	value := "value"
-	s.mustPutOK(c, key, value, 5, 10)
+	mustPutOK(t, store, key, value, 5, 10)
 	mutations := []*kvrpcpb.Mutation{{
 		Op:  kvrpcpb.Op_Lock,
 		Key: []byte(key),
 	},
 	}
```
tisonkun, line 212:
```suggestion
	}}
```

tisonkun (changes requested):
We repeat these lines of code many times.

```go
store, err := NewMVCCLevelDB("")
require.Nil(t, err)
```

We'd better move it to a test suite so that when we try to update initialization logic, we don't touch them lines by line.

It is a blocker because part of tests in `mock_tikv_test.go` doesn't require a `MVCCLevelDB` store, we may separate them into new test file.

* `func TestMarshalmvccLock(t *testing.T)`
* `func TestMarshalmvccValue(t *testing.T)`
* `func TestErrors(t *testing.T)`

It's better to have a series of dedicate test files than one large "'common" test file.

mockstore/mocktikv/mock_tikv_test.go
disksing: Normally it is a bad idea. But for this special case, I guess it is ok because methods/functions of them are almost the same. In this way, we create a function overloading like effect. By the way, readme of `stretchr/testify` also suggests use `assert`.

disksing:
@tisonkun As I mentioned in a previous issue https://github.com/tikv/client-go/pull/134#issuecomment-865203089, I actually purposely avoid using suites. Here, the initialization code isn't too complicated or special - we're testing the `MVCCStore`, so creating an `MVCCStore` for testing, which I don't think is a big deal to create in each test.

tisonkun:
> @tisonkun As I mentioned in a previous issue [#134 (comment)](https://github.com/tikv/client-go/pull/134#issuecomment-865203089), I actually purposely avoid using suites. Here, the initialization code isn't too complicated or special - we're testing the `MVCCStore`, so creating an `MVCCStore` for testing, which I don't think is a big deal to create in each test.

Make sense. We can postpone the decision when the setup logic really touched and changes frequently.

mockstore/mocktikv/mock_tikv_test.go
tisonkun: Fine. I don't have a strong feeling.

mockstore/mocktikv/errors_test.go
```diff
@@ -0,0 +1,47 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NOTE: The code in this file is based on code from the
```
tisonkun, line 14: We don't have to retain this header for new files I think.

mockstore/mocktikv/errors_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestErrors(t *testing.T) {
+	assert := assert.New(t)
+	assert.Equal((&ErrKeyAlreadyExist{}).Error(), `key already exist, key: ""`)
+	assert.Equal(ErrAbort("txn").Error(), "abort: txn")
+	assert.Equal(ErrAlreadyCommitted(0).Error(), "txn already committed")
+	assert.Equal((&ErrConflict{}).Error(), "write conflict")
```
tisonkun, line 46: Why do we test this things...It seems like we test getter/setter.

Does the error string should keep compatible?

disksing: Good point. I did some code search, at least in TiDB there is no usage rely on the specific strings. I think we'd better to remove this test.

tisonkun (approved):
LGTM.

Issue #120, opened 2021-06-21
Replace check with testify: /unionstore/memdb{,_norace}_test.go


disksing:
subtask of https://github.com/tikv/client-go/issues/97

PR #141, opened 2021-06-22
unionstore: replace pingcap/check with testify

disksing:

Fix https://github.com/tikv/client-go/issues/120

/cc @tisonkun


tisonkun (approved):
LGTM

disksing:
Found an unstable test https://github.com/tikv/client-go/issues/144 which is not related.

PR #142, opened 2021-06-22
[WIP] return error early when stores are unavailable

sticnarf:
This PR adds a new backoff type `regionStoreUnavailable`.

If an RPC fails to send, in this PR, the client will try to request the liveness state of all stores of the region. If a majority of stores of the region are unreachable, a `regionStoreUnavailable` backoff is triggered.

The max allowed `maxRegionStoreUnavailable` count is set to `20` here. So if a region keeps not having a majority of working peers, it seems to encounter a permanent failure (probably a disaster). So in this case, an error is returned to the user early.


disksing:
closing for inactivity. Please reopen later when it is ready.

PR #150, opened 2021-06-23
all tests run in serial

disksing:

As we are migrating to `testify`, later test suites cannot run in parallel. This PR makes all tests run in serial to test if there are any problems.

Test result shows that it runs faster in serial. I have no idea why.

```
                    norace+parallel       norace+serial        race+parallel        race+serial
unit-test           35.974s               25.274s              9.268s               4.315s
integration-test    47.216s               37.393s              54.462s              46.847s
```

### DIFF
--- a/client/client_fail_test.go
+++ b/client/client_fail_test.go
@@ -47,16 +47,6 @@ import (
 )
 
 type testClientFailSuite struct {
-	OneByOneSuite
-}
-
-func (s *testClientFailSuite) SetUpSuite(_ *C) {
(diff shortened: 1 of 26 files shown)


tisonkun (approved):
LGTM.

OneByOneSuite runs as SerialSuite when `with-tikv=true`, with extra overhead to wait on locks. The suite only "speedup" when run with unistore but still has overhead to manipulate RLock. It seems the downside over the upside and we actually create a regression.

Anyway, it is a good news we improve testing time.

Issue #132, opened 2021-06-21
Replace check with testify: /integration_tests/ticlient{,_slow}_test.go


disksing:
subtask of https://github.com/tikv/client-go/issues/97

PR #151, opened 2021-06-23
ticlient_test: replace pingcap/check with testify

disksing:

Fix https://github.com/tikv/client-go/issues/132

/cc @tisonkun


AndreMouche (approved):
LGTM

Issue #129, opened 2021-06-21
Replace check with testify: /integration_tests/snapshot{,_fail}_test.go


disksing:
subtask of https://github.com/tikv/client-go/issues/97

PR #154, opened 2021-06-23
snap_test: replace pingcap/check with testify

disksing:

Fix https://github.com/tikv/client-go/issues/129


tisonkun (approved):

AndreMouche (approved):
LGTM

Issue #121, opened 2021-06-21
Replace check with testify: /integration_tests/1pc_test.go


disksing:
subtask of https://github.com/tikv/client-go/issues/97

PR #162, opened 2021-06-24
1pc_tesst, async_commit_test: replace pingcap/check with testify

disksing:

Fix https://github.com/tikv/client-go/issues/123
Fix https://github.com/tikv/client-go/issues/121


AndreMouche (approved):
LGTM

Issue #169, opened 2021-06-24
Use `goleak` to replace `util/testleak`


disksing:
Pros:

- it is more commonly used
- no need for scripts and code generation
- we can enable it in `TestMain` so don't have to intrusive every test

Ref: https://github.com/uber-go/goleak

PR #171, opened 2021-06-25
remove testleak utilities

disksing:

Remove testleak as we are going to use goleak https://github.com/tikv/client-go/issues/169
`pingcap/check` dependency is also finally resolved. Fix https://github.com/tikv/client-go/issues/97


unionstore/memdb_test.go
```diff
@@ -699,14 +698,12 @@ func mustGet(t *testing.T, buffer *MemDB) {
 }
 
 func TestKVGetSet(t *testing.T) {
-	defer testleak.AfterTestT(t)()
```
tisonkun, line 702: whether we just replace with `defer goleak.VerifyNone(t)` in this pass?

disksing: Adding in each Test is a bit tedious, I am going to add goleak in each package's `TestMain`. Please wait for my subsequent PR.

tisonkun: Well. Given leak tests don't in used actually. I'm ok with this.

tisonkun (approved):

PR #172, opened 2021-06-25
introduce goleak

disksing:

Add goleak for each package and fix leak problems.

Note: need to add goleak for `integration_tests` before closing https://github.com/tikv/client-go/issues/169


tisonkun (approved):
LGTM

PR #173, opened 2021-06-25
integration: add goleak

disksing:
There is a `ttlManager` issue that need to fix. I've added it to ignore list now https://github.com/tikv/client-go/issues/174

Fix https://github.com/tikv/client-go/issues/169


disksing:
Still have some goroutine leak when tested against real tikv. Investigating.

AndreMouche (approved):
LGTM

PR #175, opened 2021-06-25
tikv: stop async commit goroutines before store close

disksing:

In the recent leak test https://github.com/tikv/client-go/pull/173 I found that the async commit goroutine can linger after test ended. 

It does not make sense from a life cycle perspective. 

This PR guarantees that all created asynchronous commit goroutines will exit before the store is closed.

Note: Since the `Store` is usually not closed in a production environment, this PR mainly affects testing and should have a limited impact on online services.


tikv/kv.go
```diff
@@ -466,13 +475,14 @@ func (s *KVStore) getMinSafeTSByStores(stores []*locate.Store) uint64 {
 }
 
 func (s *KVStore) safeTSUpdater() {
+	defer s.wg.Done()
 	t := time.NewTicker(time.Second * 2)
 	defer t.Stop()
 	ctx, cancel := context.WithCancel(context.Background())
```
sticnarf, line 481: Does it need to be derived from `s.ctx`?

sticnarf (commented):
The rest lgtm.

tikv/kv.go
disksing: Good point!

sticnarf (approved):

Issue #145, opened 2021-06-23
Imporve failpoint performance


disksing:
In order to move code here. We migrated failpoints to a fashion that need no code generation, which means it brings more performance cost. We can improve it by introducing a new global bool flag. After that  in most cases the only cost is to check the boolean.

tiancaiamao:
![Lark20210625-104927](https://user-images.githubusercontent.com/1420062/123362367-23af2b00-d5a3-11eb-9dc7-1414ca1f1753.png)

It is a problem, failpoint is in the hot code path, and if the path does not exist, the failpoint throw a not exist error.
The error include the stack trace, and the error is frequent, so it impact the performance a lot!

PR #176, opened 2021-06-26
util: add flag to control all failpoints

disksing:

Fix https://github.com/tikv/client-go/issues/145

cc @tiancaiamao @tisonkun

### DIFF
--- a/util/failpoint.go
+++ b/util/failpoint.go
@@ -36,30 +36,25 @@ import (
 	"errors"
 
 	"github.com/pingcap/failpoint"
-	"go.uber.org/atomic"
 )
 
 const failpointPrefix = "tikvclient/"
 
 var (
-	failpointsEnabled atomic.Bool
+	failpointsEnabled bool
 	errDisabled       = errors.New("failpoints are disabled")
 )
 
 // EnableFailpoints enables use of failpoints.
+// It should be called before using client to avoid data race.
 func EnableFailpoints() {
-	failpointsEnabled.Store(true)
-}
-
-// DisableFailpoints turns failpoints off all.
-func DisableFailpoints() {
-	failpointsEnabled.Store(false)
+	failpointsEnabled = true
 }
 
 // EvalFailpoint injects code for testing. It is used to replace `failpoint.Inject`
 // to make it possible to be used in a library.
 func EvalFailpoint(name string) (interface{}, error) {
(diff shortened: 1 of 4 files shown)


util/failpoint.go
```diff
@@ -33,13 +33,34 @@
...
+	"errors"
+
 	"github.com/pingcap/failpoint"
+	"go.uber.org/atomic"
 )
 
 const failpointPrefix = "tikvclient/"
 
+var (
+	failpointsEnabled atomic.Bool
```
tisonkun, line 45: Can we use a global flag and setup to `true` from `go` cli when testing? `atomic.Bool` seems another significant overhead.

tisonkun (commented):
Logically it is true. But I wonder whether we can reduce the overhead further.

util/failpoint.go
disksing: That is a very good point. The good side of using `bool` is performance, while the down side is that we can only decide whether to enable failpoints before starting using client-go. It basically means we need to set up it in `init()` or `main` or there could be race conditions.

The situation is that the `tidb-server` can provide services through HTTP and can be controlled in a dynamic way (of course, the tidb-server should be compiled with `failpoint-ctl enable` for code generation).

I think a better option could be using `bool + atomic.Bool` way. `bool` should be set at initial phase, `atomic.Bool` can be set dynamically. The failpoints only work when both values are `true`.

Looking forward to your feedback :)

tisonkun: cc @tiancaiamao

I'd like to understand the previous code generate way first. In that way how we decide whether or not a failpoint is enabled?

disksing: Previously, if build without failpoint, there is no cost at all. If build after failpoint code generation, the cost is about the same as client-go before this PR.

tisonkun: @disksing yep. I think it may be a bit overkill we have two options because one option keeps the same as previous flexibility. But I'm ok with two options for now since we can deprecated the inner `atomic.Bool` if it is proven to be redundant. Also, users enable failpoint should be already aware of the cost.

disksing: After careful consideration, I think what you said makes sense. I decided to use a single bool variable for now, and it can't be too late to add atomic if there is a subsequent need.

tisonkun (approved):
LGTM

Issue #178, opened 2021-06-26
Move `unionstore` to internal package


disksing:
- Move the directory to internal package.
- Export interfaces in `tikv` for out use.
- Keep compatibility test pass.

PR #179, opened 2021-06-26
move unionstore package to internal

disksing:

Fix https://github.com/tikv/client-go/issues/178


AndreMouche (approved):
LGTM

PR #181, opened 2021-06-26
move latch to internal package

disksing:

Fix https://github.com/tikv/client-go/issues/180


AndreMouche (approved):
LGTM

PR #183, opened 2021-06-26
*: move locate to internal package

disksing:

Fix https://github.com/tikv/client-go/issues/182


AndreMouche (approved):
LGTM

PR #187, opened 2021-06-29
clear RPC error only if prewrite succeeds

sticnarf:
Same as https://github.com/pingcap/tidb/pull/25803.

First, this PR reverts #25120. Receiving a Write Conflict does not mean the transaction fails deterministically.

This PR also reverts a change in #25040. If there is no region error, the rpcErr should not be cleared. Because even if the retry fails, the previous prewrite can still make the transaction succeed. Instead, this PR removes the rpcErr only if the prewrite request has no error.


disksing:
It looks to me that when async commit is enabled, maintaining a single undetermined error in the committer is not enough. Instead, the undetermined status should be maintained by mutationBatch. The undetermined status of a batch can be reset when we received a deterministic result or every sub batches all have received a deterministic result. And the committer's result is undetermined if any mutationBatch is undetermined.

I'll merge this PR now. We may want to record an issue afterwards.

disksing (approved):

Issue #200, opened 2021-07-01
Move RawKVClient to separated directory


disksing:
The task has 2 objectives:
- move RawKV API to standalone package to make it easier to use. At the same time, the autogenerated documents will be more clear.
- we need to add context parameters to public APIs, using a new package can ensure that the API update process does not break the compatibility test.

The task will be broke into 3 steps:
- add new API in rawkv directory, provides APIs with context parameters
- upgrade library users to use new API
- move code from `tikv` to `rawkv`.

If it can go well, we will implement the same process for txnkv.

disksing:
Fixed by https://github.com/tikv/client-go/pull/227

PR #202, opened 2021-07-01
Add rawkvclient with context parameters

disksing:

Step 1 of https://github.com/tikv/client-go/issues/200

### DIFF
--- a/rawkv/rawkv.go
+++ b/rawkv/rawkv.go
@@ -1,0 +1,121 @@
+// Copyright 2021 TiKV Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


sticnarf (approved):

rawkv/rawkv.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// ClusterID returns the TiKV cluster ID.
+func (c *Client) ClusterID() uint64 {
+	return c.client.ClusterID()
+}
+
+// Get queries value with the key. When the key does not exist, it returns `nil, nil`.
+func (c *Client) Get(ctx context.Context, key []byte) ([]byte, error) {
+	return c.client.Get(key)
```
iosmanthus, line 70: The `ctx` is not used in the function. How can user cancel this execution using this `ctx`?

sticnarf: I think it's remained for future work...

iosmanthus: Well then...

iosmanthus: Could we add some TODO to the function signatures in case we forget to refactor these codes? @disksing

disksing: won't forget but I think it is better to add TODO :)

PR #205, opened 2021-07-01
tikv: extract batch util to kvrpc

disksing:

The Batch utility is used by both rawkv and split_region. As we are moving rawkv to anathor directory, we'd to extract Batch utility first.


internal/kvrpc/batch.go
```diff
@@ -0,0 +1,76 @@
...
+	Values   [][]byte
+}
+
+// BatchResult wraps a Batch request's server response or an error.
+type BatchResult struct {
+	*tikvrpc.Response
+	Error error
+}
+
+// AppendBatches divides the mutation to be requested into Batches, ensuring that the total key-value length
```
iosmanthus, line 34: This comment declares that the size of each `Batch` will not larger than `limit`, but the code in L40 tells us a new `Batch` is created only if the previous `Batch` is larger or equal to `limit`.

iosmanthus: The comments sematic make sense. We may modify the code itself.

disksing: I think the code is correct because we need to handle the case when a single key-value pair exceeds the limit. I'll update the comment.

iosmanthus: OK.

iosmanthus (approved):
LGTM

PR #214, opened 2021-07-06
etcd and grpc update

franklinkim:
Hi, 

this PR update `google.golang.org/grpc` to v1.38.0 and `go.etcd.io/etcd` to `go.etcd.io/etcd/api/v3` & `go.etcd.io/etcd/client/v3` v3.5.0. 

The reason why it's needed is because if you use the grpc >v1.27.1 you get this error:

```
$ go mod tidy

go: finding module for package google.golang.org/grpc/naming
integration_tests tested by
	integration_tests.test imports
	github.com/pingcap/tidb/store/driver/txn imports
	github.com/pingcap/tidb/sessionctx/binloginfo imports
	github.com/pingcap/tidb-tools/tidb-binlog/node tested by
	github.com/pingcap/tidb-tools/tidb-binlog/node.test imports
	go.etcd.io/etcd/integration imports
	go.etcd.io/etcd/proxy/grpcproxy imports
	google.golang.org/grpc/naming: module google.golang.org/grpc@latest found (v1.39.0), but does not contain package google.golang.org/grpc/naming
```

Please note that the `intregration_tests` are *not* yet updated will not work since they are pulling in `github.com/tikv/pd` which itself relies on the older versions. They would need to be updated as well.


disksing:
Thanks @franklinkim . Does it mean we should update `tikv/pd` first before merging this PR?

franklinkim:
Yes, that would be great! 

You would need to update `pingcap/tidb-dashboard` as well as it relies on `tikv/pd`.

I haven't digged to  deep into all the pingcap repos but I fear the update on`pingcap/tidb-dashboard` might cascaded down to require a lot of repos to be updated as well e.g. `pingcap/br`, `pingcap/tidb`

disksing:
Yes, because of the dependencies between the different pingcap/tikv projects, this problem can be far more difficult than it seems. We'll spend some time sorting this out.

But before that, I wourld want to make it clear that the problem you are currently trying to fix is that you are using a higher version of grpc in your project and there is a conflict after importing client-go, so you need to upgrade client-go's grpc to a compatible version. Is that an accurate description?

franklinkim:
Yes, that would be perfectly accurate 👍.

disksing:
Hi @franklinkim , I tried to upgrade etcd this week, and as expected I encountered some problems caused by project dependencies (specifically pd and tidb-dashboard dependencies). We already have a solution for one of the main issues, which involves splitting the go module for pd, but it's not expected to be done anytime soon. You may follow https://github.com/tikv/pd/issues/3877 to keep track of the latest progress.

I'm going to close this PR for now and wait until after finish of upgrading on PD side. For the grpc conflict issue, I think we can only lock the grpc version to 1.27 in the application for now.

If you have any other questions, please feel free to leave a comment and continue to discuss.

franklinkim:
@disksing thanks for looking into it. Will keep an eye on #3877.

PR #217, opened 2021-07-08
use ioutil.ReadFile to release dependency of Go 1.16

Sandy4999:
This is the only line that forces users to use Go 1.16+, and it brings few benefit


tisonkun:
@Sandy4999 thanks for bringing this topic, we don't officially announce go version compatibility, though.

@disksing maybe we should setup a line declare the compatibility with go version.

disksing:
I thought `go 1.16` in go.mod already states compatible go version.

Sandy4999:
> I thought `go 1.16` in go.mod already states compatible go version.

Yes, it does. The problem here IMHO is that it doesn't have to be so high since this project can actually be compatible with lower versions. You may bump it to 1.16+ later when new features are required and when they bring real improvements.

disksing:
Okay you've convinced me, I think you've made a good point. Would you like to update the version number in the go.mod as well by the way?
Also we need to update CI config to make sure 1.15 can be compiled without problems. But it is fine to leave this issue for me to fix later.

/cc @tisonkun do you have any more opinions?

Sandy4999:
I found another snippet using io.WriteFile (probably added after the v2.0.0-alpha release). Now all unit tests (`go test ./...`) passed with Go 1.14. I'm not sure if everything compiles and works well so please update the version number in go.mod by yourself after all CI tests succeed with a lower version :D

tisonkun:
> Okay you've convinced me, I think you've made a good point. Would you like to update the version number in the go.mod as well by the way?
> Also we need to update CI config to make sure 1.15 can be compiled without problems. But it is fine to leave this issue for me to fix later.
> 
> /cc @tisonkun do you have any more opinions?

From a [previous discussion](https://internals.tidb.io/t/topic/121) we said go version in ci will be upgraded every six months. We may write a line in README about the go version requirement.

Changes like this one which cannot be guarded by ci is less maintainable. What if a series of following prs require to be compatible with more previous go version?

@Sandy4999 different from java software, as go binary are runtime free. What is the problem you meet with higher go version? For managing different go version in one env, you can refer to [gvm](https://github.com/moovweb/gvm)

sticnarf:
@tisonkun The latest two major versions are maintained by the Go team. So 1.15 is a reasonable target now because people who cares about stability may stay at one major release older than the latest.

dragonly:
IMHO, It's really a bad idea to revert the go version back to 1.13 [which has its end of life by 03 Sep 2019](https://endoflife.date/go).

TennyZhuang:
TiDB is a binary instead of a library, we only need a consistent version and use the latest feature. Keep the compatibility with multiple versions will increase developer’s work with few benefits. Currently our CI is based on go 1.16, so we can only promise that it will work on go 1.16, make it compilable with go 1.15 is unnecessary.

Sandy4999:
> @Sandy4999 different from java software, as go binary are runtime free. What is the problem you meet with higher go version? For managing different go version in one env, you can refer to [gvm](https://github.com/moovweb/gvm)

It forces users who want to import this module to use higher version, while they may be conservative and want a mature version in their project (in fact we are using 1.15, and may bump to 1.16 after 1.17 released, similar strategy as described in [this thread](https://internals.tidb.io/t/topic/121)). Anyway of course I can just fork this repo to maintain a customized version, so it's totally up to you guys to make the final decision.

tisonkun:
> major

Make sense client-go as a library to be as compatible as possible.

@Sandy4999 would you mind to downgrade our ci go version to 1.15 and ensure it is compatible with that version? I'll approve for the ci to run but request changes to finish the whole version work.

@disksing may you make the final version.

disksing:
I think 1.15 should be better than 1.14 as 1.14 is already marked archived on the golang website.

sticnarf (approved):
LGTM if we come to a conclusion that the minimum supported version of this library is Go 1.15 now.

disksing:
Hi @Sandy4999 we decide to support go1.15, would you like to update the version in go.mod?

Sandy4999:
> Hi @Sandy4999 we decide to support go1.15, would you like to update the version in go.mod?

Okay

Sandy4999:
> Error: ../../../../go/pkg/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20210702041526-83acf6a4ac28/config/config_util.go:112:12: undefined: os.WriteFile

It seems to be complicated since it depends on tidb repo to run integration tests, which requires 1.16.

sticnarf:
> > Error: ../../../../go/pkg/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20210702041526-83acf6a4ac28/config/config_util.go:112:12: undefined: os.WriteFile
> 
> It seems to be complicated since it depends on tidb repo to run integration tests, which requires 1.16.

I think it's possible to only use Go 1.16 for tidb integration tests.

disksing:
Hi @Sandy4999 We can use 1.16 in integration test. As long as the version of `github.com/tikv/client-go/v2` is go1.15, the library can be used by applications that are using go1.15+.

Sandy4999:
Let's try again then

disksing:
I'll update your branch before merging. Thanks for contributing.

Sandy4999:
Sure, thank you for understanding

PR #218, opened 2021-07-08
util: make reduce tsset object allocation

tiancaiamao:
TSSet object allocation ranks Top3 during the point get operation.

It's a burden for Go GC as we observed in https://github.com/pingcap/tidb/issues/25573

This PR make the object allocation lazy, because not every transaction would meet lock conflict.
After this change, this object allocation is totally eliminated.

And the benchmark result in tidb shows

```
cd session
go test -test.run XXX -bench BenchmarkPreparedPointGet  -test.benchmem -test.benchtime 30s  -test.memprofile mem.out
```

Before vs After:

```
BenchmarkPreparedPointGet-16    	 3750292	      9442 ns/op	    7830 B/op	     101 allocs/op
BenchmarkPreparedPointGet-16    	 3824121	      9293 ns/op	    7702 B/op	      97 allocs/op
```

### DIFF
--- a/util/ts_set.go
+++ b/util/ts_set.go
@@ -61,7 +61,7 @@ func (s *TSSet) Put(tss ...uint64) {
 func (s *TSSet) GetAll() []uint64 {
 	s.RLock()
 	defer s.RUnlock()
-	if s.m == nil || len(s.m) == 0 {
+	if len(s.m) == 0 {
 		return nil
 	}
 	ret := make([]uint64, 0, len(s.m))
--- a/util/ts_set.go
+++ b/util/ts_set.go
@@ -40,17 +40,18 @@ type TSSet struct {
 	m map[uint64]struct{}
 }
 
-// NewTSSet creates a set to store timestamps.
-func NewTSSet(capacity int) *TSSet {
-	return &TSSet{
-		m: make(map[uint64]struct{}, capacity),
-	}
-}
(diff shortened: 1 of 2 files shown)


sticnarf:
Why the git history is too long... Could you try to rebase master again?

tiancaiamao:
> Why the git history is too long... Could you try to rebase master again?

Done

util/ts_set.go
```diff
@@ -60,7 +61,7 @@ func (s *TSSet) Put(tss ...uint64) {
 func (s *TSSet) GetAll() []uint64 {
 	s.RLock()
 	defer s.RUnlock()
-	if len(s.m) == 0 {
+	if s.m == nil || len(s.m) == 0 {
```
sticnarf, line 64: This change looks unnecessary because `len(nil map)`  also returns 0

sticnarf (approved):
LGTM except a tiny optional change

tiancaiamao:
The TiDB part will be fixed by https://github.com/pingcap/tidb/pull/26045

util/ts_set.go
tiancaiamao: Done

disksing:
I'll merge it. Please don't forget to merge TiDB's PR to recover the compatibility test. @tiancaiamao

PR #222, opened 2021-07-09
add raw key ttl feature

iosmanthus:

This pull request introduces a new integration test kind that depends only on TiKV instances, you could run them by:

```sh
go test -with-tikv -test.run TestTTL
```

- [x] add put with TTL
- [x] add get key TTL
- [x] add integration tests


andylokandy:
@iosmanthus Suggestiion:

add get with ttl -> get key ttl

integration_tests/ttl_test.go
```diff
@@ -0,0 +1,82 @@
+package tikv_test
```
disksing, line 1: Please add license header.

disksing (commented):
the rest LGTM

iosmanthus:
@disksing PTAL

iosmanthus:
@andylokandy PTAL

integration_tests/ttl_test.go
```diff
@@ -0,0 +1,95 @@
...
+	s.Nil(v)
+}
+
+func (s *ttlTestSuite) mustGetKeyTTL(key []byte) *uint64 {
+	ttl, err := s.client.GetKeyTTL(context.TODO(), key)
+	s.Nil(err)
+	return ttl
+}
+
+// TODO: It's an unstable test, we may mock this feature in unistore.
```
andylokandy, line 65: Is it unstable now? What about increasing the sleep time.

integration_tests/ttl_test.go
```diff
@@ -0,0 +1,95 @@
...
+	s.mustPutWithTTL(key, value, ttl)
+	time.Sleep(time.Second * time.Duration(ttl/2))
+
+	rest := s.mustGetKeyTTL(key)
+	s.NotNil(rest)
+	s.LessOrEqual(*rest, ttl/2)
+
+	time.Sleep(time.Second * time.Duration(ttl/2))
+	s.mustNotExist(key)
+
```
andylokandy, line 88:
```suggestion
```

integration_tests/ttl_test.go
iosmanthus: No need. There are very few chances that this test might fail because we can't count on the time of TiKV flow as the same as the clients. The TODO is to TODO for `unistore`.

andylokandy: If so, I'd like to remove `It's an unstable test`.

disksing (approved):

andylokandy (approved):

disksing:
@iosmanthus Please resolve the conflicts before merge.

PR #225, opened 2021-07-12
export get all tiflash stores information

hanfei1991:
This PR changes two things:
1. add a kvproto interface: mpp.IsAlive, which is used to check if this tiflash is available or not.
2. extend the GetAllTiFlashStores method in order to check the availablities of each node.

Signed-off-by: Han Fei <hanfei19910905@gmail.com>


internal/locate/region_cache.go
```diff
@@ -1603,17 +1603,17 @@ func (c *RegionCache) PDClient() pd.Client {
 	return c.pdClient
 }
 
-// GetTiFlashStoreAddrs returns addresses of all tiflash nodes.
-func (c *RegionCache) GetTiFlashStoreAddrs() []string {
+// GetTiFlashStores returns the information of all tiflash nodes.
+func (c *RegionCache) GetTiFlashStores() []*Store {
```
disksing, line 1607: I wonder if there will be some data race if we access `Store` later without acquiring locks.

hanfei1991: Yeah. But I checked every store access is protected by rwlock, why questioning that?

disksing: okay.

disksing (approved):

PR #227, opened 2021-07-12
rawkv: move rawkv client to rawkv

disksing:
Signed-off-by: disksing <i@disksing.com

Move rawkv client code to standalone directory.


AndreMouche (approved):
LGTM

PR #228, opened 2021-07-12
do not check sessionID for async commit and 1PC

sticnarf:
`SessionID` was used to prevent internal transactions in TiDB from committing with async commit or 1PC.

As a separate client, we don't need this check any more.


integration_tests/1pc_test.go
```diff
@@ -62,27 +61,26 @@ func (s *testOnePCSuite) TearDownTest() {
...
+	err = txn.Commit(context.Background())
 	s.Nil(err)
 	s.True(txn.GetCommitter().IsOnePC())
 	s.Equal(txn.GetCommitter().GetOnePCCommitTS(), txn.GetCommitter().GetCommitTS())
 	s.Greater(txn.GetCommitter().GetOnePCCommitTS(), txn.StartTS())
 	// ttlManager is not used for 1PC.
 	s.True(txn.GetCommitter().IsTTLUninitialized())
 
-	// 1PC doesn't work if sessionID == 0
+	// 1PC doesn't work if system variable not set
```
AndreMouche, line 78: I think we do not need this test anymore?

AndreMouche (commented):
Rest LGTM

integration_tests/1pc_test.go
sticnarf: We still need it. It's to check the default committing mode of a transaction. Now I change the comment to be more appropriate for this separated client.

disksing:
https://github.com/tikv/client-go/blob/ec820e01f18a6443069fdb539b61767dbb9dd57c/tikv/2pc.go#L1217

Does this reference needs to update too?

sticnarf:
> https://github.com/tikv/client-go/blob/ec820e01f18a6443069fdb539b61767dbb9dd57c/tikv/2pc.go#L1217
> 
> Does this reference needs to update too?

I think it's unrelated. A pure kv user does not set `schemaAmender`, so it will not affect users, I think.

disksing (approved):

PR #230, opened 2021-07-12
tikv/txn: add new pub function LockKeysWithWaitTime

AndreMouche:
This PR adds a new public function `LockKeysWithWaitTime`, with this function the caller needn't learn about the LockCtx anymore.


tikv/txn.go
```diff
@@ -514,8 +514,27 @@ func (txn *KVTxn) onCommitted(err error) {
 	}
 }
 
-// LockKeys tries to lock the entries with the keys in KV store.
+// LockKeysWithWaitTime tries to lock the entries with the keys in KV store.
 // lockWaitTime in ms, except that kv.LockAlwaysWait(0) means always wait lock, kv.LockNowait(-1) means nowait lock
```
sticnarf, line 518: The public API may be polished to not use these magic numbers...

disksing:
LGTM. @sticnarf please take another look.

sticnarf (approved):
LGTM

PR #231, opened 2021-07-13
examples/*: rm go.sum

AndreMouche:
This PR removes `go.sum` in each directory of examples since `go.sum` may change frequently, so we'd like to generate it in real time.


tisonkun:
shall we add a ci workflow to guard examples build successfully?

disksing (approved):

PR #232, opened 2021-07-14
examples/txnkv: add example for delete_range

AndreMouche:
This PR adds an example for `delete_range`.


MyonKeminta (commented):
The code looks fine to me. And by the way, do you think it would be better to check the remaining keys after deleting range, in order to avoid mistake in the example?

disksing (approved):

AndreMouche:
> The code looks fine to me. And by the way, do you think it would be better to check the remaining keys after deleting range, in order to avoid mistake in the example?

I think it is better to put the check logic in the test.

PR #233, opened 2021-07-14
reset resolvedLocks when setting snapshot ts

sticnarf:
This PR intends to fix the jepsen failure https://github.com/tikv/tikv/issues/10552#issuecomment-879531922.

If the `minCommitTS` of some transaction was advanced before, and the snapshot timestamp is reset, we must also reset the `resolvedLocks`. Otherwise, the reader can read a stale result if only the primary lock is committed. (See the test as an example)

### DIFF
--- a/integration_tests/snapshot_fail_test.go
+++ b/integration_tests/snapshot_fail_test.go
@@ -258,3 +258,47 @@ func (s *testSnapshotFailSuite) TestRetryPointGetResolveTS() {
 	s.Nil(err)
 	s.Equal(v, []byte("v2"))
 }
+
+func (s *testSnapshotFailSuite) TestResetSnapshotTS() {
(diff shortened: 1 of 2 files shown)


sticnarf:
PTAL @cfzjywxk @tiancaiamao

cfzjywxk:
> PTAL @cfzjywxk @tiancaiamao

@tiancaiamao @sticnarf 
Maybe we need to check other places in which the initialization logic is changed or some objects are resued, as they could affect the correctness because of some implicit dependencies or connections.

jackysp (approved):

tiancaiamao (approved):

sticnarf:
There is already a https://github.com/pingcap/br/pull/1354 fixing the BR compatibility failure, fix anyway.

PR #234, opened 2021-07-14
txnkv/*: init txnkv/rangetask package

AndreMouche:
This PR inits package `txnkv/rangetask`, and it also adds a new function `DeleteRange` for `KVStore`


examples/txnkv/delete_range/delete_range.go
```diff
@@ -85,19 +85,16 @@ func main() {
 	// delete range
 	// Be careful while using this API. This API doesn't keep recent MVCC versions, but will delete all versions of all keys
 	// in the range immediately. Also notice that frequent invocation to this API may cause performance problems to TiKV.
```
MyonKeminta, line 87: I think you also need to mention that this API does not even guarantee the atomicity.

MyonKeminta (approved):

PR #235, opened 2021-07-15
[internal]: improve RegionRequestRuntimeStats

ClSlaid:
I've been working on improving TiDB's Point-Get performance, and I found out there is space for improvement in client-go.
图片
So I merely replaced that `fmt.Sprintf()` with `strings.Builder`. By doing so, time cost of the function should be reduced to a half.

Signed-off-by: ClSlaid <cailue@bupt.edu.cn>

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -132,6 +132,7 @@ func (r *RegionRequestRuntimeStats) String() string {
 		if builder.Len() > 0 {
 			builder.WriteByte(',')
 		}
+		// append string: fmt.Sprintf("%s:{num_rpc:%v, total_time:%s}", k.String(), v.Count, util.FormatDuration(time.Duration(v.Consume))")
 		builder.WriteString(k.String())
 		builder.WriteString(":{num_rpc:")
 		builder.WriteString(strconv.FormatInt(v.Count, 10))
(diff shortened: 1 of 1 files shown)


disksing (approved):

disksing:
Hi @ClSlaid the file is not well formatted. You needs to run `gofmt -s`.

sticnarf:
@ClSlaid 
I tried to profile point get at my place, but I don't see this sprintf cost...
What is the caller of `(*SnapshotRuntimeStats) String()`? Is it an optimization only at some features enabled?

ClSlaid:
> @ClSlaid
> I tried to profile point get at my place, but I don't see this sprintf cost...
> What is the caller of `(*SnapshotRuntimeStats) String()`? Is it an optimization only at some features enabled?

It's caller is tidb/planner/core/encode.go: func (pn *planEncoder) encodePlan().

It's my fault that I didn't give out the full testing configuration... The SysBench is set to benching TiDB's performance under such situation:

```MySQL
set @@global.tidb_stmt_summary_max_stmt_count=1;
```

With a very limited size of statements summary, TiDB's point-get performance almost cut to half, due to the statements summary's cache miss. Besides simply scaling statements summary table's default size to a larger number, we're also working on reducing time cost by `executor.getEncodePlan()`.

sticnarf (approved):
OK, I see

ClSlaid:
> I am actually a bit worried that replacing `Sprintf` with `strings.Builder` lowers the readability of the code.
> So I am wondering where and how often the code is called, to see whether the optimization is really worth.

I'll add a comment in the code.

PR #237, opened 2021-07-16
*: solve cloud tikv disk full problem

### LINKED ISSUE #243: Solve cloud TiKV disk full problem - Client Go part

tier-cap:

Issue Number：closes #243

When the disk is full, no matter one or two or the majority or all, what we expected will be:

1. TiKV not crash directly.
2. Support restoring storage through drop table/delete from table, expansion is also allowed.
3. Provide slightly better reading ability.

###  What is changed and how it works?

1. Introduce a two-level threshold defense mechanism
2. Different flow control algorithms are used under each threshold to provide different business support capabilities and spatial recovery capabilities

### Check List

Test

UserDoc

### Release note

- No release note.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,3 +44,4 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	honnef.co/go/tools v0.2.0 // indirect
 )
+replace github.com/pingcap/kvproto => /home/xiaojin.zxj/010.tidb/kvproto
--- a/tikv/2pc.go
+++ b/tikv/2pc.go
@@ -135,6 +135,9 @@ type twoPhaseCommitter struct {
 
 	storeWg  *sync.WaitGroup
 	storeCtx context.Context
+
+	// allowed when tikv disk full happened.
+	allowedOnDiskFull bool
 }
 
 type memBufferMutations struct {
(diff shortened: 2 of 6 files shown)


tikv/txn.go
```diff
@@ -307,6 +309,10 @@ func (txn *KVTxn) SetKVFilter(filter KVFilter) {
 	txn.kvFilter = filter
 }
 
+func (txn *KVTxn) SetAllowed() {
```
sticnarf, line 312: ditto

tikv/2pc.go
```diff
@@ -755,6 +759,10 @@ func (c *twoPhaseCommitter) keySize(key, value []byte) int {
 	return len(key)
 }
 
+func (c *twoPhaseCommitter) SetAllowed() {
```
sticnarf, line 762: Please use the full name `SetAllowedOnDiskFull`

tikv/txn.go
sticnarf, line 312: ditto

go.mod
```diff
@@ -44,3 +44,4 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	honnef.co/go/tools v0.2.0 // indirect
 )
+replace github.com/pingcap/kvproto => /home/xiaojin.zxj/010.tidb/kvproto
```
iosmanthus, line 47: Please use the repo of your own instead of the local path. The other couldn't test it with the local path.

disksing:
Looks no progress for a long time. @tier-cap I'll close it for now to avoid distraction. You can reopen the PR when it is ready.

PR #239, opened 2021-07-19
add raw cas feature

iosmanthus:

### DIFF
--- a/integration_tests/cas_test.go
+++ b/integration_tests/cas_test.go
@@ -1,0 +1,81 @@
+package tikv_test
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"github.com/tikv/client-go/v2/config"
+	"github.com/tikv/client-go/v2/rawkv"
+)
+
+func TestCAS(t *testing.T) {
+	suite.Run(t, new(casTestSuite))
+}
+
+type casTestSuite struct {
+	suite.Suite
+	client *rawkv.Client
+}
+
+func (s *casTestSuite) SetupTest() {
+	t := s.T()
+	if !*withTiKV {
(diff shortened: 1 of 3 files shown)


integration_tests/cas_test.go
```diff
@@ -0,0 +1,81 @@
+package tikv_test
```
disksing, line 1: Please add license header.

rawkv/rawkv.go
```diff
@@ -71,6 +71,13 @@ type Client struct {
 	regionCache *locate.RegionCache
 	pdClient    pd.Client
 	rpcClient   client.Client
+	atomic      bool
```
disksing, line 74: What's this flag used for? why does it only affect `RawBatchGet` and `RawBatchPut` but not `CAS`?

iosmanthus: This `atomic` flag is supplied for raw requests related to `write` operations like `RawPut` and `RawDelete`. It won't affect to read operation like `RawGet`. In TiKV, the `CAS` operation is implemented by the mechanism of `latches` of the transaction module. While CAS using `latch` to write, the other write operations need to use this mode either, or It will break the linearizability of TiKV.

iosmanthus:
@disksing PTAL

disksing:
LGTM. please fix test before merge @iosmanthus

iosmanthus:
@disksing PTAL

disksing (approved):

rawkv/rawkv.go
neicunlideyu: Can we use RawCASRequest in RawBatchPut？

disksing:
Apparently, we do not support batch CAS operations. I will create an issue
to discuss if this is a generic requirement.

rawkv/rawkv.go
iosmanthus: @neicunlideyu For now We can't.

PR #246, opened 2021-07-22
test: fix overflow in ttl equal compare

you06:
- Fix the overflow of usigned integer calculation
- Fix the elapsed time calculation, it should be calculated before prewrite phase

The details where overflow and UB happens.

```
func (s *testLockSuite) ttlEquals(x, y uint64) {
	...
	s.LessOrEqual(int(math.Abs(float64(x-y))), 2)
```

- If `x` is less than `y`, `x - y` leads to the first overflow, we'll got an extreme large number. After it, the abs function takes no effect.
- Parse the large number to signed int leads to the second overflow and the **undefined behaviod**, there we'll get a negative number in some platforms. A negative number is absolutely less then 2. That's the reason this test works "fine" with incorrect elapsed time calculation.

~~I don't have a `ppc64le` machine, so I left a TODO on it.~~
Remove the `ppc64le` branch.


integration_tests/lock_test.go
```diff
@@ -445,9 +445,13 @@ func (s *testLockSuite) ttlEquals(x, y uint64) {
 	// NOTE: On ppc64le, all integers are by default unsigned integers,
 	// hence we have to separately cast the value returned by "math.Abs()" function for ppc64le.
 	if runtime.GOARCH == "ppc64le" {
+		// TODO: fix the value compare
 		s.LessOrEqual(int(-math.Abs(float64(x-y))), 2)
 	} else {
```
sticnarf, line 450: It seems we needn't keep this code. The problem in the note should not happen after doing a swap first.

you06: Removed.

sticnarf (approved):
Though `int(x-y)` still has the risk of overflow, lgtm for just test usage.

you06:
The time estimate would make it an unstable test :(

you06:
> Though `int(x-y)` still has the risk of overflow, lgtm for just test usage.

There is no risk of overflow now :)

disksing:
Hi, @you06 thanks for contributing! It seems there is a test failure in integration test.

integration_tests/lock_test.go
```diff
@@ -442,17 +442,17 @@ func (s *testLockSuite) mustGetLock(key []byte) *txnkv.Lock {
...
+		x, y = y, x
 	}
-
+	s.LessOrEqual(x-y, uint64(5))
 }
 
 func (s *testLockSuite) TestLockTTL() {
+	managedLockTTL := atomic.LoadUint64(&tikv.ManagedLockTTL)
+	atomic.StoreUint64(&tikv.ManagedLockTTL, 20000)                // set to 20s
+	defer atomic.StoreUint64(&tikv.ManagedLockTTL, managedLockTTL) // restore value
```
you06, line 454: `ManagedLockTTL` will be set to 3000 in test. The calculated TTL which is great than `ManagedLockTTL` will be cut to `ManagedLockTTL`, this strategy fails the test.

you06:
@disksing the test has been fixed.

disksing:
@you06 Thanks. I'll update your branch before merge.

PR #253, opened 2021-07-26
*: make grpc request duration more accurate

jackysp:
Signed-off-by: Shuaipeng Yu jackysp@gmail.com
In `SendRequest`, the metrics of `tidb_tikvclient_request_seconds` will include the duration of batch client recycle and reconnection, which may make it not accurate for the real RPC duration.
Move it before the send RPC request and add a new metrics for the batch client recycle and reconnection.


lysu (approved):

lysu:
how about add new metric for recycleIdleConnArray?- -

disksing:
Hi, @jackysp please add DCO for your commits.

jackysp:
> how about add new metric for recycleIdleConnArray?- -

Maybe next one.

disksing:
Thanks @jackysp , I'll update your branch before merge.

PR #257, opened 2021-07-29
txnkv/transaction: init package

AndreMouche:
This PR init package `txnkv/transaction`.


disksing:
IMO it is better to move files to `txnkv`, should be more easy to use.

AndreMouche:
> IMO it is better to move files to `txnkv`, should be more easy to use.

I think export some public types to `txnkv` would be better

disksing (approved):

PR #261, opened 2021-08-03
Start heartbeat immediately after locking the primary key

sticnarf:
See https://github.com/pingcap/tidb/issues/26801 for what this PR wants to solve.

This PR starts the `keepAlive` goroutine after locking the primary key. So even if locking other keys takes a long time, the TTL of the primary lock will not expire.

However, before all keys are locked successfully, the primary lock may be determined. So if some keys fail to be locked and , we need to stop the heartbeat goroutine.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -360,9 +360,6 @@ func newTwoPhaseCommitter(txn *KVTxn, sessionID uint64) (*twoPhaseCommitter, err
 		startTS:       txn.StartTS(),
 		sessionID:     sessionID,
 		regionTxnSize: map[uint64]int{},
-		ttlManager: ttlManager{
-			ch: make(chan struct{}),
-		},
 		isPessimistic: txn.IsPessimistic(),
 		binlog:        txn.binlog,
 	}, nil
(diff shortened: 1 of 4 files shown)


txnkv/transaction/txn.go
```diff
@@ -621,14 +621,12 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 				}
 			}
 			if assignedPrimaryKey {
-				// unset the primary key if we assigned primary key when failed to lock it.
+				// unset the primary key and stop heartbeat if we assigned primary key when failed to lock it.
 				txn.committer.primaryKey = nil
+				txn.committer.ttlManager.close()
```
youjiali1995, line 626: Seems it can't run any more after closing.

txnkv/transaction/2pc.go
```diff
@@ -871,6 +873,14 @@ func (tm *ttlManager) close() {
 	close(tm.ch)
 }
 
+func (tm *ttlManager) reset() {
+	if !atomic.CompareAndSwapUint32((*uint32)(&tm.state), uint32(stateRunning), uint32(stateUninitialized)) {
+		return
+	}
+	close(tm.ch)
+	tm.ch = make(chan struct{})
```
youjiali1995, line 881: I think it has the risk of race condition. Is it possible the `keepAlive` goroutine uses the new `ch`?

sticnarf: Ah, yes... I change to copy the channel into the `keepAlive` goroutine instead of using the one in `ttlManager`.

sticnarf:
Oops, the primary key is also a potential race. orz.

youjiali1995 (approved):

youjiali1995 (approved):

PR #262, opened 2021-08-06
WIP: txnkv/txnsnapshot: Support snapshot set ranges for memory scan

lcwangchao:
Related to issue: https://github.com/pingcap/tidb/issues/26952

Introduce method `SetSortedCustomKeyRetrievers` to read custom ranges. For example, we can do it like below:

```
var snapshot *KVSnapshot
...
retrievers := []*RangeRetriever{NewRangeRetriever(tblPrefix, tblKeyEnd, NewMemDBRetriver(memDB))}
snapshot.SetSortedCustomKeyRetrievers(retrievers)
```

Then the key in range `[tblPrefix, tblKeyEnd)` will be fetched from memDB

### DIFF
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -582,6 +582,16 @@ type Scanner = txnsnapshot.Scanner
 // KVSnapshot implements the tidbkv.Snapshot interface.
 type KVSnapshot = txnsnapshot.KVSnapshot
 
+// MemDBRetriever implements unionstore.Retriever
+type MemDBRetriever = txnsnapshot.MemDBRetriever
+
+// EmptyRetriever implements unionstore.Retriever
+type EmptyRetriever = txnsnapshot.EmptyRetriever
+
+type RangeRetriever = txnsnapshot.RangeRetriever
+
+var NewRangeRetriever = txnsnapshot.NewRangeRetriever
+
 // SnapshotRuntimeStats records the runtime stats of snapshot.
 type SnapshotRuntimeStats = txnsnapshot.SnapshotRuntimeStats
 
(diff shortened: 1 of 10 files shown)


tikv/kv.go
```diff
@@ -582,6 +582,16 @@ type Scanner = txnsnapshot.Scanner
 // KVSnapshot implements the tidbkv.Snapshot interface.
 type KVSnapshot = txnsnapshot.KVSnapshot
 
+// MemDBRetriever implements unionstore.Retriever
+type MemDBRetriever = txnsnapshot.MemDBRetriever
+
+// EmptyRetriever implements unionstore.Retriever
+type EmptyRetriever = txnsnapshot.EmptyRetriever
```
djshow832, line 589: Why do we need `EmptyRetriever`? For the integration of the interfaces?

lcwangchao: `EmptyRetriever` is used to return an empty data set, i.e, the global temporary table.

disksing:
@lcwangchao Hi, want to continue?

lcwangchao:
> @lcwangchao Hi, want to continue?

Still under review. It is not determined if we should do it yet.

tikv/kv.go
djshow832: I didn't see the usage of `EmptyRetriever` in the other PR. Is it used?

lcwangchao: It is used in https://github.com/pingcap/tidb/pull/26948 . I think it is also ok to move it to tidb

lcwangchao:
I think it's time to continue

PR #268, opened 2021-08-14
client: cleanup RPCClient options

disksing:

use optional security argument so we don't have to pass empty security config every time.

### DIFF
--- a/internal/client/client_fail_test.go
+++ b/internal/client/client_fail_test.go
@@ -59,9 +59,8 @@ func TestPanicInRecvLoop(t *testing.T) {
 	defer server.Stop()
 
 	addr := fmt.Sprintf("%s:%d", "127.0.0.1", port)
-	rpcClient := NewRPCClient(config.Security{}, func(c *RPCClient) {
-		c.dialTimeout = time.Second / 3
-	})
+	rpcClient := NewRPCClient()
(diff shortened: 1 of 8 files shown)


internal/client/client_fail_test.go
```diff
@@ -59,9 +59,8 @@ func TestPanicInRecvLoop(t *testing.T) {
 	defer server.Stop()
 
 	addr := fmt.Sprintf("%s:%d", "127.0.0.1", port)
-	rpcClient := NewRPCClient(config.Security{}, func(c *RPCClient) {
-		c.dialTimeout = time.Second / 3
-	})
+	rpcClient := NewRPCClient()
+	rpcClient.dialTimeout = time.Second / 3
```
sticnarf, line 63: I'm a bit curious about the `opts` design. If the options can be set by this way, why need we pass opts in `NewRPCClient`?

disksing: opts can be used by library users that do not have access to private fields. For `dialTimeout`, we only need to update it in the same package (at least for now), so don't have to use opts.

sticnarf: Ah, I see. If any field that needs to be modified by the library user, we provide a function returning `Opt` like `WithSecurity`.

sticnarf (approved):
LGTM

PR #270, opened 2021-08-19
internal/client: fix recycle idle connection block sending request

tiancaiamao:
Move the `recycleIdleConnArray()` operation to a separate goroutine.

In our internal OnCall issue 3520 and 3381, I observed sending request blocked here when recycling idle connection.

recycle idle connection need to obtain the `recycleMu` write lock
when some connections are sending batch request, the `recycleMu` read lock is hold.
If the send batch request blocks for some reason (like TiFlash down), recycle idle connection will also be blocked by that, and then it blocks `SendRequest()` function.

I also consider some other options, like remove the recycle idle connection logic ... we introduce it because when some TiKV server are gone, there will be resource leak here. 

Anyway, this change is small and should fix the problem, the risk of side effect is low.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -365,9 +365,7 @@ func (c *RPCClient) SendRequest(ctx context.Context, addr string, req *tikvrpc.R
 
 	start := time.Now()
 	if atomic.CompareAndSwapUint32(&c.idleNotify, 1, 0) {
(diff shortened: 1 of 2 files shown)


tiancaiamao:
PTAL @crazycs520  @lysu @disksing

disksing (approved):

PR #272, opened 2021-08-19
txnkv: `RangedKVRetriever` struct to provide custom range data operations

lcwangchao:
Refer: https://github.com/pingcap/tidb/issues/26952

To make a better implement for temporary table. It is required that snapshot can read data from other sources like memory for some specified ranges. So we expect `KVSnapshot` can provide some methods to override data fetch.

The draft has been proposed here: #262 . This pr is the first step, introduce a struct to provide custom range data operations. 

Signed-off-by: Chao Wang <cclcwangchao@hotmail.com>

### DIFF
--- a/internal/unionstore/union_store.go
+++ b/internal/unionstore/union_store.go
@@ -61,12 +61,17 @@ type Getter interface {
 // Retriever is the interface to get or scan key/values
 type Retriever interface {
 	Getter
-	// Scan creates an Iterator positioned on the first entry that startKey <= entry's key.
+	// Iter creates an Iterator positioned on the first entry that k <= entry's key.
 	// If such entry is not found, it returns an invalid Iterator with no error.
-	// It yields only keys that < endKey. If upperBound is nil, it means the upperBound is unbounded.
-	// If reverse is specified, the returned iterator will iterate from greater key to smaller key
+	// It yields only keys that < upperBound. If upperBound is nil, it means the upperBound is unbounded.
 	// The Iterator must be Closed after use.
-	Scan(startKey []byte, endKey []byte, reverse bool) (Iterator, error)
+	Iter(k []byte, upperBound []byte) (Iterator, error)
+
+	// IterReverse creates a reversed Iterator positioned on the first entry which key is less than k.
+	// The returned iterator will iterate from greater key to smaller key.
+	// It yields only keys that >= lowerBound. If lowerBound is nil, it means the lowerBound is unbounded.
+	// The Iterator must be Closed after use.
+	IterReverse(k []byte, lowerBound []byte) (Iterator, error)
 }
 
 // uSnapshot defines the interface for the snapshot fetched from KV store.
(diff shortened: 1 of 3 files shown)


disksing:
It seems the new method
```
Scan(startKey []byte, endKey []byte, reverse bool) (Iterator, error)
```
is almost the same like the combination of the two old methods (except for ReverseIter does not have bound)
```
Iter(k []byte, upperBound []byte) (Iterator, error)
IterReverse(k []byte) (Iterator, error)
```

Is it possible to merge them together?

lcwangchao:
> It seems the new method
> 
> ```
> Scan(startKey []byte, endKey []byte, reverse bool) (Iterator, error)
> ```
> 
> is almost the same like the combination of the two old methods (except for IterReverse does not have bound)
> 
> ```
> Iter(k []byte, upperBound []byte) (Iterator, error)
> IterReverse(k []byte) (Iterator, error)
> ```
> 
> Is it possible to merge them together?

They are a little different for iter reverse. For `Scan` it can support both start and end are not nil, it is a required feature. But current `IterReverse` do not support it.

disksing:
> > It seems the new method
> > ```
> > Scan(startKey []byte, endKey []byte, reverse bool) (Iterator, error)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is almost the same like the combination of the two old methods (except for IterReverse does not have bound)
> > ```
> > Iter(k []byte, upperBound []byte) (Iterator, error)
> > IterReverse(k []byte) (Iterator, error)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is it possible to merge them together?
> 
> They are a bit different for iter reverse. For `Scan` it can support both start and end are not nil, it is a required feature. But current `IterReverse` do not support it.

Can we add `lowerBound` to `IterReverse`?

lcwangchao:
> > > It seems the new method
> > > ```
> > > Scan(startKey []byte, endKey []byte, reverse bool) (Iterator, error)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > is almost the same like the combination of the two old methods (except for IterReverse does not have bound)
> > > ```
> > > Iter(k []byte, upperBound []byte) (Iterator, error)
> > > IterReverse(k []byte) (Iterator, error)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Is it possible to merge them together?
> > 
> > 
> > They are a bit different for iter reverse. For `Scan` it can support both start and end are not nil, it is a required feature. But current `IterReverse` do not support it.
> 
> Can we add `lowerBound` to `IterReverse`?

I think it's a good idea. I searched the current code, only `uSnapshot` delecars a method `IterReverse` but its a internal interface. Adding `lowerBound` to `IterReverse` for `Retriever` is fine.
```
Iter(k []byte, upperBound []byte) (Iterator, error)
IterReverse(k []byte, lowerBound []byte) (Iterator, error)
```
If one day

lcwangchao:
Done

lcwangchao:
@disksing  PTAL again

lcwangchao:
It should moved to tidb if it is currently only used by temporary table

PR #273, opened 2021-08-19
[WIP]support set assertion in 2pc mutation

lysu:
[WIP] ref https://github.com/pingcap/tidb/issues/26833

### DIFF
--- a/kv/keyflags.go
+++ b/kv/keyflags.go
@@ -35,7 +35,7 @@
 package kv
 
 // KeyFlags are metadata associated with key
-type KeyFlags uint8
+type KeyFlags uint16
 
 const (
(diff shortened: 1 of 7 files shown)


disksing:
Please reopen when it is ready.

kv/keyflags.go
```diff
@@ -45,10 +45,27 @@ const (
 	flagNeedCheckExists
 	flagPrewriteOnly
 	flagIgnoredIn2PC
+	flagAssertExist
+	flagAssertNotExist
```
MyonKeminta, line 49: 🤔 How about making a 2-bit flag rather than two 1-bit flags so that it can represent up to 4 different choices

PR #277, opened 2021-08-20
internal/client: remove the lock for idle connection recycle

tiancaiamao:
After mock testing, I find https://github.com/tikv/client-go/pull/270 can not sufficient to solve the problem.
To my surprise, **RLock can block RLock in [some case](https://www.zenlife.tk/go-read-mutex-block.md)**

The final decision is to reduce the granularity of the lock.
Just lock the modification of the `map`...

There is a risk that an idle connection become active again, since `recycleMu` is removed to protect such case.
In very rare case, someone is maybe using the connection and the connection is closed by the recycle operation.

I think the outcome is acceptable, comparing to the recycle operation blocks the `SendRequest()` function.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -261,9 +261,7 @@ type RPCClient struct {
 	security config.Security
 
 	idleNotify uint32
-	// recycleMu protect the conns from being modified during a connArray is taken out and used.
-	// That means recycleIdleConnArray() will wait until nobody doing sendBatchRequest()
-	recycleMu sync.RWMutex
+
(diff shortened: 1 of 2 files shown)


tiancaiamao:
Damn ... the DCO check ... I open a new one https://github.com/tikv/client-go/pull/278

disksing:
> Damn ... the DCO check ... I open a new one #278

you can reset + commit with DCO + push -f
you can also run [this script](https://gist.github.com/disksing/4122718e910e33138032a72a45688480) + push -f

PR #278, opened 2021-08-20
internal/client: remove the lock for idle connection recycle

tiancaiamao:
After mock testing, I find https://github.com/tikv/client-go/pull/270 is not sufficient to solve the problem.
To my surprise, **RLock can block RLock in [some case](https://www.zenlife.tk/go-read-mutex-block.md)**

The final decision is to reduce the granularity of the lock.
Just lock the modification of the `map`...

There is a risk that an idle connection become active again.
Since `recycleMu` is removed to protect such case, in very rare cases, someone maybe using the connection and the connection is closed by the recycle operation.

I think the outcome is acceptable, comparing to the recycle operation blocks the `SendRequest()` function.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -298,7 +298,7 @@ func (c *RPCClient) getConnArray(addr string, enableBatch bool, opt ...func(cfg 
 
 	// An idle connArray will not change to active again, this avoid the race condition
 	// that recycling idle connection close an active connection unexpectedly (idle -> active).
-	if array.isIdle() {
+	if array.batchConn != nil && array.isIdle() {
 		return nil, errors.Errorf("rpcClient is idle")
 	}
 
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -288,7 +288,9 @@ func (c *RPCClient) getConnArray(addr string, enableBatch bool, opt ...func(cfg 
 	}
 	array, ok := c.conns[addr]
 	c.RUnlock()
-	if !ok {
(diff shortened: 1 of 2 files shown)


tiancaiamao:
Update: `getConnArray` will check `isIdle` first, this can avoid the recycle goroutine remove active connections unexpectedly.

internal/client/client.go
```diff
@@ -297,6 +295,13 @@ func (c *RPCClient) getConnArray(addr string, enableBatch bool, opt ...func(cfg
 			return nil, err
 		}
 	}
+
+	// An idle connArray will not change to active again, this avoid the race condition
+	// that recycling idle connection close an active connection unexpectedly (idle -> active).
+	if array.batchConn != nil && array.isIdle() {
```
tiancaiamao, line 301: I confirmed return error will not make TiDB unavailable.
This error will be recognized as a RPC error, and TiDB will refresh the region cache ... and retry the request.

disksing (approved):

internal/client/client.go
ngaut: > I confirmed return error will not make TiDB unavailable.
> This error will be recognized as a RPC error, and TiDB will refresh the region cache ... and retry the request.

This behavior is a magic,  it's hard to figure out that tidb will retry the request while reading the source code. Could you add some notes here and file an issue? Also, we should do refactor to make sure we know when should return a "retry" error.

tiancaiamao:
TiDB issue traced in https://github.com/pingcap/tidb/issues/32500

PR #283, opened 2021-08-23
txnsnapshot: provide oneByOneIter to combine multiple iterator together

lcwangchao:
Signed-off-by: Chao Wang <cclcwangchao@hotmail.com>

Refer: pingcap/tidb#26952

For scanning multiple retrievers, provide oneByOneIter to combine multiple iterator together


lcwangchao:
It should moved to tidb if it is currently only used by temporary table

PR #285, opened 2021-08-26
select replica for replica read requests using ReplicaSelector

youjiali1995:

Added a `accessFollower` state that sends requests to followers or all replicas with mixed store option. The client uses `replicaReadSeed` to switch to different replicas for replica reads before. Now since the `replicaSelector` hold the context for requests, I remove the seed strategy and will remove more codes after this PR is merged.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -99,13 +99,13 @@ func LoadShuttingDown() uint32 {
 // For other region errors, since region range have changed, the request may need to
 // split, so we simply return the error to caller.
 type RegionRequestSender struct {
-	regionCache           *RegionCache
-	client                client.Client
-	storeAddr             string
-	rpcError              error
-	leaderReplicaSelector *replicaSelector
-	failStoreIDs          map[uint64]struct{}
-	failProxyStoreIDs     map[uint64]struct{}
+	regionCache       *RegionCache
+	client            client.Client
+	storeAddr         string
+	rpcError          error
+	replicaSelector   *replicaSelector
+	failStoreIDs      map[uint64]struct{}
+	failProxyStoreIDs map[uint64]struct{}
 	RegionRequestRuntimeStats
 }
 
@@ -240,6 +240,14 @@ type replica struct {
 	attempts int
 }
 
+func (r *replica) isEpochStale() bool {
+	return r.epoch != atomic.LoadUint32(&r.store.epoch)
+}
+
+func (r *replica) isExhausted(maxAttempt int) bool {
+	return r.attempts >= maxAttempt
+}
+
 type replicaSelector struct {
 	regionCache *RegionCache
 	region      *Region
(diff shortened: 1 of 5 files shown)


internal/locate/region_request.go
```diff
@@ -514,7 +587,7 @@ func (state *invalidLeader) next(_ *retry.Backoffer, _ *replicaSelector) (*RPCCo
 	return nil, nil
 }
 
-func newReplicaSelector(regionCache *RegionCache, regionID RegionVerID) (*replicaSelector, error) {
+func newReplicaSelector(regionCache *RegionCache, regionID RegionVerID, reqType kv.ReplicaReadType, opts ...StoreSelectorOption) (*replicaSelector, error) {
```
sticnarf, line 590: I think it is worth commentting that `opts` is currently only effective for follower read.

internal/locate/region_request.go
```diff
@@ -599,8 +681,9 @@ func (s *replicaSelector) refreshRegionStore() {
 	}
 
 	// If leader has changed, it means a recent request succeeds an RPC on the new
-	// leader. Give the leader an addition chance.
-	if oldRegionStore.workTiKVIdx != newRegionStore.workTiKVIdx {
+	// leader. Give the leader an addition chance if the request is not a leader request.
```
sticnarf, line 684: Need the `leaderIdx` in `accessFollower` be updated?

sticnarf (commented):
Please also fix the lint

internal/locate/region_request.go
```diff
@@ -922,7 +983,17 @@ func (s *RegionRequestSender) SendReqCtx(
...
-			retry, err = s.onRegionError(bo, rpcCtx, req, regionErr, &opts)
+			// Stale Read request will retry the leader or next peer on error,
+			// if txnScope is global, we will only retry the leader by using the WithLeaderOnly option,
+			// if txnScope is local, we will retry both other peers and the leader by the strategy of replicaSelector.
+			if tryTimes < 1 && req != nil && req.TxnScope == oracle.GlobalTxnScope && req.GetStaleRead() {
+				accessFollowerState, ok := s.replicaSelector.state.(*accessFollower)
+				if ok {
+					WithLeaderOnly()(&accessFollowerState.option)
+				}
+			}
```
sticnarf, line 994: As stale read has some special logics, and `storeSelectorOp` is only for stale read now, actually I am considering if it's better to give it a special state in the `replicaSelector`...

youjiali1995: How about adding a field `isGlobalStaleRead` to the `accessFollower` state to distinguish it since the strategy of local stale read is same as follower read? If the `isGlobalStaleRead` is true, the second `next` call will try the leader.

sticnarf (approved):
LGTM

disksing:
`integration-tikv` cannot pass now. It could be related to https://github.com/tikv/tikv/pull/10800 . I found many `can't sched txn cmd(Prewrite) with TTL enabled` in logfile.

youjiali1995:
Let me see see.

PR #293, opened 2021-08-31
Multi ttl

daimashusheng:
batch_put add ttl for every key


rawkv/rawkv.go
```diff
@@ -249,13 +249,16 @@ func (c *Client) BatchPut(ctx context.Context, keys, values [][]byte) error {
 	if len(keys) != len(values) {
 		return errors.New("the len of keys is not equal to the len of values")
 	}
+	if nil != ttls && len(keys) != len(ttls) {
```
disksing, line 252:
```suggestion
	if len(ttls) > 0 && len(keys) != len(ttls) {
```
`nil != ttls` is not very easy to read.

internal/kvrpc/batch.go
```diff
@@ -24,6 +24,7 @@ type Batch struct {
 	RegionID locate.RegionVerID
 	Keys     [][]byte
 	Values   [][]byte
+	Ttls     []uint64
```
disksing, line 27: `TTLs` would be more in line with Go conventions

internal/kvrpc/batch.go
```diff
@@ -34,25 +35,30 @@ type BatchResult struct {
...
 		}
 		key := groupKeys[start]
 		value := keyToValue[string(key)]
+		ttl := keyToTtl[string(key)]
 		keys = append(keys, key)
 		values = append(values, value)
+		ttls = append(ttls, ttl)
 		size += len(key)
 		size += len(value)
+		size += 8
```
disksing, line 58: I think we can ignore the size of TTL as protobuf uses varint to encode integers. Typically it won't cost 8 bytes.

rawkv/rawkv.go
```diff
@@ -652,10 +655,16 @@ func (c *Client) sendDeleteRangeReq(ctx context.Context, startKey []byte, endKey
...
+func (c *Client) sendBatchPut(bo *retry.Backoffer, keys, values [][]byte, ttls []uint64) error {
 	keyToValue := make(map[string][]byte, len(keys))
+	keyTottl := make(map[string]uint64, len(keys))
 	for i, key := range keys {
 		keyToValue[string(key)] = values[i]
+		if nil != ttls {
+			keyTottl[string(key)] = ttls[i]
+		} else {
+			keyTottl[string(key)] = 0
+		}
```
disksing, line 667:
```suggestion
		if len(ttls) > 0 {
			keyTottl[string(key)] = ttls[i]
		}
```
1. `len(tts) > 0` should be more easy to read.
2. you don't need to set map value to the default value.

disksing:
@daimashusheng  Please update the pull request title to make it more specific.

marsishandsome:
https://github.com/pingcap/kvproto/pull/806
https://github.com/tikv/client-go/pull/298
https://github.com/tikv/tikv/pull/10891

Issue #299, opened 2021-09-07
Add function to check if an error is undeterminedError


disksing:
When the commit result of a transaction is undetermined, the returned error will be a wrapped error of `terror.ErrUndetermined`. In TiDB, we compare the error with `terror.ErrUndetermined` to check if the result is unknown. However, we have planned to remove the dependency of `terror` later, it is better to add a function to check the error type now to avoid breaking compatibility in the future.

PR #300, opened 2021-09-07
error: add function to check if error is undetermined

disksing:

Fix https://github.com/tikv/client-go/issues/299


sticnarf (approved):
LGTM

PR #302, opened 2021-09-09
*:rechange to retry when tikv disk full

tier-cap:
Signed-off-by: tier-cap <zhengxiaojin@pingcap.com>
1. When TiKV is disk full, the original design is just return the disk full error to user without any retry.
2. Currently, need change to retry before return to user.

1. Use backoff to retry it when TiKV is disk full.

### DIFF
--- a/internal/retry/config.go
+++ b/internal/retry/config.go
@@ -99,6 +99,11 @@ func (c *Config) String() string {
 	return c.name
 }
 
+// SetErrors sets a more detailed error instead of the default bo config.
+func (c *Config) SetErrors(err error) {
+	c.err = err
+}
+
 const txnLockFastName = "txnLockFast"
 
 // Backoff Config variables.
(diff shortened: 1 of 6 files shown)


internal/unionstore/memdb.go
```diff
@@ -166,10 +165,10 @@ func (db *MemDB) Reset() {
...
 }
 
-// SetDiskFullOpt is used by TiDB test case.
-func (db *MemDB) SetDiskFullOpt(level kvrpcpb.DiskFullOpt) {
-	// Nothing to do.
-}
+// // SetDiskFullOpt is used by TiDB test case.
+// func (db *MemDB) SetDiskFullOpt(level kvrpcpb.DiskFullOpt) {
+// 	// Nothing to do.
+// }
```
sticnarf, line 171: Please remove the code.

tier-cap: Right, tidb has already removed this func.
Done!

internal/retry/config.go
```diff
@@ -99,6 +99,10 @@ func (c *Config) String() string {
 	return c.name
 }
 
+func (c *Config) SetErrors(err error) {
```
disksing, line 102: please add comment. the rest LGTM

tier-cap: OK, Done

disksing (approved):

PR #309, opened 2021-09-23
add `skipSwitchPeerLog` in `OnSendFailForTiFlash`

windtalker:
In `OnSendFailForTiFlash`, it always record "switch peer" log, however, in most scenarios, TiFlash will batch all the regions in one TiFlash store into one request, so when meet send failure, this function is called repeatedly for all the regions, since one TiFlash store might contain thousands of regions, we need a way to avoid generating too much useless log.


fzhedu (approved):
/LGTM

disksing (approved):

PR #311, opened 2021-09-24
Support setting assertions and receiving assertion errors for Prewrite requests

MyonKeminta:
This PR is part of https://github.com/pingcap/tidb/issues/26833 and is a  continuation of https://github.com/tikv/client-go/pull/273 .


txnkv/transaction/2pc.go
```diff
@@ -249,6 +265,8 @@ type PlainMutations struct {
 	keys              [][]byte
 	values            [][]byte
 	isPessimisticLock []bool
+	isAssertExist     []bool
+	isAssertNotExist  []bool
```
sticnarf, line 269: It looks a bit wasteful to me... Maybe also define "flags" here as the lowest 3 bits of `UserData`?

txnkv/transaction/2pc.go
```diff
@@ -209,25 +209,39 @@ func (m *memBufferMutations) GetValue(i int) []byte {
...
 func (m *memBufferMutations) Slice(from, to int) CommitterMutations {
 	return &memBufferMutations{
 		handles: m.handles[from:to],
 		storage: m.storage,
 	}
 }
 
-func (m *memBufferMutations) Push(op kvrpcpb.Op, isPessimisticLock bool, handle unionstore.MemKeyHandle) {
-	aux := uint16(op) << 1
+func (m *memBufferMutations) Push(op kvrpcpb.Op, isPessimisticLock bool, assertExist, assertNotExist bool, handle unionstore.MemKeyHandle) {
```
sticnarf, line 234: I think it is better to add comments about the format of `aux` and add unit tests for it.

txnkv/transaction/2pc.go
```diff
@@ -1222,6 +1280,16 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
...
 
 	if err != nil {
+		if strings.Contains(err.Error(), "assertion fail") {
+			ts, err1 := c.store.GetTimestampWithRetry(retry.NewBackofferWithVars(ctx, TsoMaxBackoff, c.txn.vars), c.txn.GetScope())
+			if err1 == nil {
+				_, _, err2 := c.checkSchemaValid(ctx, ts, c.txn.schemaVer, false)
+				if err2 != nil {
+					err = err2
+				}
+			}
```
sticnarf, line 1290: I guess it is to avoid some false positives with online DDL? Could you add some comments?

txnkv/transaction/2pc.go
MyonKeminta: Ah.. sorry, this PR is not ready yet and it was supposed to be a draft PR, and currently this PR only did some minor changes based on  https://github.com/tikv/client-go/pull/273 .

disksing:
sorry, I misunderstood the time of last commit

txnkv/transaction/2pc.go
```diff
@@ -1222,6 +1335,16 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 	err = c.prewriteMutations(bo, c.mutations)
 
 	if err != nil {
+		if strings.Contains(err.Error(), "assertion fail") {
```
ekexium, line 1338: I prefer not to use string comparison to compare errors

txnkv/transaction/txn.go
```diff
@@ -656,14 +663,19 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
 	}
 	for _, key := range keys {
 		valExists := tikv.SetKeyLockedValueExists
 		// PointGet and BatchPointGet will return value in pessimistic lock response, the value may not exist.
 		// For other lock modes, the locked key values always exist.
-		if lockCtx.ReturnValues {
+		if lockCtx.ReturnValues || checkedExistence {
 			val := lockCtx.Values[string(key)]
-			if len(val.Value) == 0 {
+			// TODO: Check if it's safe to use `val.Exists` instead of assuming empty value.
```
ekexium, line 677: I think for TiDB they are equivalent. Maybe `len(val.Value) == 0 || !val.Exists` in favor of safety?

ekexium (commented):
Assertion_unknown is not allowed to be changed by design. Shall we handle it in client-go?

txnkv/transaction/txn.go
MyonKeminta: It doesn't seem to work if ReturnValue is false but CheckExistence is true?

txnkv/transaction/txn.go
```diff
@@ -666,15 +673,23 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
-			if len(val.Value) == 0 {
-				valExists = tikv.SetKeyLockedValueNotExists
+		if lockCtx.ReturnValues || checkedExistence {
+			// If ReturnValue is disabled and CheckExistence is requested, it's still possible that the TiKV's version
+			// is too old and CheckExistence is not supported.
+			if val, ok := lockCtx.Values[string(key)]; ok {
+				// TODO: Check if it's safe to use `val.Exists` instead of assuming empty value.
+				if !val.Exists {
+					valExists = tikv.SetKeyLockedValueNotExists
+				}
```
MyonKeminta, line 692: @cfzjywxk  Is this safe? Previously when a key is locked without `ReturnValue` option, we assume it must exist since TiDB will only lock the key after it's found in scanning. However this changed the behavior. I'm thinking if I'd better add two new flags for storing whether it exists or not,  that the new flags will only be used in assertion, to avoid changing the normal behavior.

MyonKeminta: I decide to add new flags for it. Even it's safe, it will introduce difficulties when we meet problem about the LockedValueExists flag (which affects the behavior about "duplicated entry") and try to diagnose it.

cfzjywxk: @MyonKeminta 
The meanings seems to be the same, use pessimistic lock response to record existing key value, I think it'ok if the judging condition is handled correctly.

ekexium:
I think the keys and values in error messages can be encoded for readability.

cfzjywxk:
> I think the keys and values in error messages can be encoded as hex for readability.

I usually use the hex encoded format printing keys and values.

MyonKeminta:
> I think the keys and values in error messages can be encoded as hex for readability.

Did you mean `ErrAssertionFailed` defined in error/errors.go? The `Error` method is implemented in the same way as other kinds of errors defined in this file.

However, by the way, I'm afraid things about redacting is questionable here, for example, the not-redacted keys might be printed when printing these kinds of errors in logs.

ekexium:
> Did you mean ErrAssertionFailed defined in error/errors.go? The Error method is implemented in the same way as other kinds of errors defined in this file.

Yes. It comes from the protobuf marshaller. I agree on leaving it for now.

> However, by the way, I'm afraid things about redacting is questionable here, for example, the not-redacted keys might be printed when printing these kinds of errors in logs.

Then the problem also exists for other error types? client-go doesn't care about log redaction at all.

MyonKeminta:
> > However, by the way, I'm afraid things about redacting is questionable here, for example, the not-redacted keys might be printed when printing these kinds of errors in logs.
> 
> Then the problem also exists for other error types? client-go doesn't care about log redaction at all.

Yes. I mean, it's possible that TiDB prints an error from client-go, and the error might be of these kinds, and as a result it will introduce the risk of leaking sensitive information to the log. And I think even if there's no such case in TiDB, it can still be easily introduced.
It's not related to this PR though. I'm just thinking there might be a problem.

txnkv/transaction/txn.go
```diff
@@ -666,15 +674,23 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
 		valExists := tikv.SetKeyLockedValueExists
 		// PointGet and BatchPointGet will return value in pessimistic lock response, the value may not exist.
 		// For other lock modes, the locked key values always exist.
-		if lockCtx.ReturnValues {
-			val := lockCtx.Values[string(key)]
-			if len(val.Value) == 0 {
-				valExists = tikv.SetKeyLockedValueNotExists
+		if lockCtx.ReturnValues || checkedExistence {
+			// If ReturnValue is disabled and CheckExistence is requested, it's still possible that the TiKV's version
+			// is too old and CheckExistence is not supported.
```
sticnarf, line 688: I'm not clear about what this comment mean... TiDB should not be used with an old version of TiKV, so I don't think we should care about it. And I don't think a client-go user will set assertions (hopefully...)

MyonKeminta: It's not necessary if we assume TiDB is never used with an old version of TiKV. I'm feeling afraid since we are supporting downgrading, and I'm not sure if there's any special cases about that, so I added this...🤔

sticnarf: Emm, downgrading... I'm thinking when downgrading, the order should also reverse? First TiDB, then TiKV and PD?

sticnarf: I'm ok with the comment. It shows a requirement for minimal TiKV version.

MyonKeminta: > Emm, downgrading... I'm thinking when downgrading, the order should also reverse? First TiDB, then TiKV and PD?

I think so, but I'm not sure how it really be.

internal/mockstore/mocktikv/mvcc_leveldb.go
```diff
@@ -725,6 +734,9 @@ func checkConflictValue(iter *Iterator, m *kvrpcpb.Mutation, forUpdateTS uint64,
 		return nil, err
 	}
 	if !ok {
+		if m.Assertion == kvrpcpb.Assertion_Exist && assertionLevel != kvrpcpb.AssertionLevel_Off {
+			logutil.BgLogger().Error("ASSERTION FAIL!!! non-exist for must exist key", zap.Stringer("mutation", m))
+		}
 		return nil, nil
```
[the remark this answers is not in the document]
MyonKeminta: Ah, this is old code before my modification and assertion is not supported in mocktikv (so there are only tests based on unistore). Since there's now too much difference between mocktikv and tikv, it might be complicated to implement assertion in mocktikv. I think there changes can be removed. How do you think?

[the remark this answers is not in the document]
MyonKeminta: Wait a minute, I need to double check it

txnkv/transaction/2pc.go
```diff
@@ -211,29 +211,74 @@ func (m *memBufferMutations) GetValue(i int) []byte {
...
 		storage: m.storage,
 	}
 }
 
-func (m *memBufferMutations) Push(op kvrpcpb.Op, isPessimisticLock bool, handle unionstore.MemKeyHandle) {
-	aux := uint16(op) << 1
+func (m *memBufferMutations) Push(op kvrpcpb.Op, isPessimisticLock, assertExist, assertNotExist bool, handle unionstore.MemKeyHandle) {
+	// The format to put to the handle.UserData:
+	// MSB                                                                            LSB
+	// [13 bits: Op][1 bit: assertNotExist][1 bit: assertExist][1 bit: isPessimisticLock]
```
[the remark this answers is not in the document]
MyonKeminta: The format is defined by the user of memdb instead of memdb itself. I'll move the comment to near the definition of `memBufferMutations.handles` field.

internal/mockstore/mocktikv/mvcc_leveldb.go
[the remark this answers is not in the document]
MyonKeminta: Sorry I forgot it. To write tests for pessimistic lock returning existence, the related assertion logic is actually already half-implemented. Let's fix this return value for completeness.

internal/mockstore/mocktikv/mvcc_leveldb.go
```diff
@@ -739,7 +751,8 @@ func checkConflictValue(iter *Iterator, m *kvrpcpb.Mutation, forUpdateTS uint64,
 	}
 
 	needGetVal := getVal
-	needCheckAssertion := m.Assertion == kvrpcpb.Assertion_NotExist
+	needCheckShouldNotExistForPessimisticLock := m.Assertion == kvrpcpb.Assertion_NotExist && m.Op == kvrpcpb.Op_PessimisticLock
+	needCheckAssertionForPrewerite := m.Assertion != kvrpcpb.Assertion_None && m.Op != kvrpcpb.Op_PessimisticLock && assertionLevel != kvrpcpb.AssertionLevel_Off
```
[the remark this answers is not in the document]
MyonKeminta: No. Unknown will not be passed via TiKV RPC, so Assertion_Unknown even doesn't exist in kvproto.

txnkv/transaction/2pc.go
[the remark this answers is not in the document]
ekexium: Makes sense. It's just that when I read the usage of the type, I would jump to its definition and hope to find some explanations there.

ekexium (approved):
The rest LGTM

txnkv/transaction/pessimistic.go
```diff
@@ -182,10 +183,23 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
 				c.run(c, action.LockCtx)
 			}
 
-			if action.ReturnValues {
+			// Handle the case that the TiKV's version is too old and doesn't support `CheckExistence`.
+			// If `CheckExistence` is set, `ReturnValues` is not set and `CheckExistence` is not supported, skip
+			// retrieving value totally (indicated by `skipRetrievingValue`) to avoid panicking.
+			skipRetrievingValue := !action.ReturnValues && action.CheckExistence && len(lockResp.NotFounds) == 0
```
cfzjywxk, line 189: Could the `len(lockResp.NotFounds) == 0` represent such situation accurately?

kv/kv.go
```diff
@@ -93,9 +95,19 @@ func NewLockCtx(forUpdateTS uint64, lockWaitTime int64, waitStartTime time.Time)
...
 	ctx.ReturnValues = true
-	ctx.Values = make(map[string]ReturnedValue, valueLen)
+	if ctx.Values == nil {
+		ctx.Values = make(map[string]ReturnedValue, capacity)
+	}
+}
+
+// InitCheckExistence creates the map to store whether each key exists or not.
+func (ctx *LockCtx) InitCheckExistence(capacity int) {
+	ctx.CheckExistence = true
```
cfzjywxk, line 107: This field will be set to true only in tests?

txnkv/transaction/pessimistic.go
MyonKeminta: The previous version of TiKV already supports setting the `NotFounds` field when `ReturnValues` is set. So when `ReturnValues` is set, both new and old versions of TiKV should work. But if `CheckExistence` is set and `ReturnValues` is unset, since old TiKV cannot recognize the new field `CheckExistence`, the `NotFounds` returned from TiKV will be empty. We cannot retrieve existence in this case, and it should be the only case.

kv/kv.go
MyonKeminta: It's exported by TiDB (https://github.com/pingcap/tidb/blob/e7bd1faf5526921badda00ed96d98fd4c2ac2bf3/executor/executor.go#L1011)

cfzjywxk (approved):

PR #321, opened 2021-10-09
Add read throughput metric for SLI

jyz0309:


disksing (approved):

disksing:
@jyz0309 I think you need to update test to make CI pass.

PR #322, opened 2021-10-10
error: replace usages of terror.Log

disksing:

Copy `terror.Log` utility function to package `error` so we can remove dependency on `terror` later.


iosmanthus (approved):
LGTM

PR #324, opened 2021-10-11
*: remove parser dependency

disksing:

pingcap/parser is merged to tidb. We should avoid using it.


iosmanthus (approved):
LGTM

PR #327, opened 2021-10-11
integration_tests: remove use of errors.Trace

disksing:

Fix https://github.com/tikv/client-go/issues/326


iosmanthus (approved):
LGTM

PR #337, opened 2021-10-12
txn: add newly inserted flag for memdb key node

cfzjywxk:
Solution for https://github.com/pingcap/tidb/issues/27564.

Add a new flag for key in `memdb`, when the key node is allocated by a `PUT` operation, add this flag to the key. This mark is used to identify the inserted keys generated from the transaction iteself, so that the delete operations on these unlocked keys could be skipped committing this transaction, see the comments for details in the issue.

This change could be risky, need to think it over.


internal/unionstore/memdb.go
```diff
@@ -328,6 +328,12 @@ func (db *MemDB) set(key []byte, value []byte, ops ...kv.FlagsOp) error {
 		return nil
 	}
 
+	// If this node is newly allocated and the value length is greater than zero, add
+	// newly inserted flag to it.
+	if len(value) > 0 && isNewNode {
+		x.setKeyFlags(kv.ApplyFlagsOps(x.getKeyFlags(), kv.SetNewlyInserted))
```
sticnarf, line 334: I notice that the flags in memdb is not rollbackable. Then, is it possible that persistent flags may be reset incorrectly?

For example, `SetNewlyInserted` is set already. Then, we set the key again, but just in a staging state. Later, we rollback the staging state, will `SetNewlyInserted` be cleared unexpectedly?

longfangsong (approved):

you06:
Point get and batch point get use `PUT` operation for locking keys, not sure if there is a risk. Maybe a `PUT` which is enable  to bypasses the newly-inserted flag is safer.

disksing:
can we merge this? cc @cfzjywxk

cfzjywxk:
> can we merge this? cc @cfzjywxk

@disksing 

It could not solve the issue completely, I think we could hold it by now.

cfzjywxk:
@disksing 
I've updated the local branch but could not find the reopen button, need help~

MyonKeminta:
> @disksing I've updated the local branch but could not find the reopen button, need help~

@cfzjywxk  It seems reopening is only available if the branch have not been updated since when the PR is closing...

PR #340, opened 2021-10-18
tikv: remove use of errors.Trace

### LINKED ISSUE #338: Remove use of `errors.Trace` for package `tikv`

feitian124:
Fix #338


tikv/safepoint.go
```diff
@@ -137,7 +137,7 @@ func (w *EtcdSafePointKV) Put(k string, v string) error {
 	_, err := w.cli.Put(ctx, k, v)
 	cancel()
 	if err != nil {
-		return errors.Trace(err)
+		return errors.WithStack(err)
```
disksing, line 140: we can remove the `if` and `return errors.WithStack(err)` directly.

tikv/split_region.go
```diff
@@ -327,11 +327,9 @@ func (s *KVStore) CheckRegionInScattering(regionID uint64) (bool, error) {
 		}
 		if err != nil {
 			err = bo.Backoff(retry.BoRegionMiss, errors.New(err.Error()))
+			return true, err
```
disksing, line 330: This change breaks original logic. Before the patch, It did not return when `Backoff` returns `nil`.

disksing: Don't panic 😄  Thanks for you contribution.

PR #362, opened 2021-11-01
Add thread safety explanation to RegoinCache's comment

MyonKeminta:
So that users and new developers of RegionCache can know what they should or should not do.


MyonKeminta:
PTAL @disksing @sticnarf 
cc @overvenus

sticnarf (approved):
Looks good. I don't find any method incomformant with the comment at the first glance.

Maybe we can add concurrent tests and run them with race enabled in the future.

Issue #359, opened 2021-10-29
Can KVStore.BeginWithOption pass StartTSOption parameter by reference


gotoxu:
When we pass the Options parameter to a method, we usually write like this:
```Go
ops := &StartTSOption{}
ops.SetStartTS(startTS)

store.BeginWithOption(ops)
```
But when StartTSOption is implemented as pass-by-value, then more code must be written:
```Go
ops := StartTSOption{}
ops = ops.SetStartTS(startTS)

store.BeginWithOption(ops)
```
And this is also prone to errors, such as forgetting to assign the return value of `SetStartTS` to the variable `ops`:
```Go
ops := StartTSOption{}
ops.SetStartTS(startTS)

// An error occurred here 
store.BeginWithOption(ops)
```
So, can we implement StartTSOption like this:
```Go
// StartTSOption indicates the option when beginning a transaction
// `TxnScope` must be set for each object
// Every other fields are optional, but currently at most one of them can be set
type StartTSOption struct {
	TxnScope string
	StartTS  *uint64
}

// SetStartTS returns a new StartTSOption with StartTS set to the given startTS
func (to *StartTSOption) SetStartTS(startTS uint64) {
	to.StartTS = &startTS
}

// SetTxnScope returns a new StartTSOption with TxnScope set to txnScope
func (to *StartTSOption) SetTxnScope(txnScope string) {
	to.TxnScope = txnScope
}
```
And `BeginWithOption` will be:
```Go
func (s *KVStore) BeginWithOption(options *StartTSOption) (*transaction.KVTxn, error)
```

Any thoughts?

disksing:
@gotoxu Thanks for bringing this up. I totally agree with you :)  Do you have any concerns to make this change? cc @Yisaer @JmPotato

JmPotato:
If I remember correctly, every `StartTSOption` is only used inside one session, so there is no data race problem here. I think it's ok to make this change.

gotoxu:
@disksing Sorry I've overlooked your PR. But I think my way is better. What do you say?

disksing:
> @disksing Sorry I've overlooked your PR. But I think my way is better. What do you say?

I think it is a good idea to use the option idiom. I have left you a comment. Please check.

PR #363, opened 2021-11-01
Use a interface method to update the transaction option

gotoxu:


gotoxu:
close #359 

cc @disksing

tikv/kv.go
```diff
@@ -281,27 +279,33 @@ func (s *KVStore) runSafePointChecker() {
...
 func (s *KVStore) Begin() (*transaction.KVTxn, error) {
-	return s.BeginWithOption(DefaultStartTSOption())
+	return s.BeginWithOption()
 }
 
 // BeginWithOption begins a transaction with the given StartTSOption
-func (s *KVStore) BeginWithOption(options StartTSOption) (*transaction.KVTxn, error) {
-	if options.TxnScope == "" {
-		options.TxnScope = oracle.GlobalTxnScope
+func (s *KVStore) BeginWithOption(options ...StartTSOption) (*transaction.KVTxn, error) {
```
disksing, line 286: For flexibility and extensibility, maybe use more generic name like `BeginOption`. By the way, since the arguments are optional, it seems that we can merge it with `Begin` function.
Also, for upgrading TiDB smoothly, we can first add options to `Begin` function and keep old `BeginWithOption` unchanged. After TiDB upgrades successfully, we can then remove the old method.

disksing, line 286: For flexibility and extensibility, maybe use more generic name like `BeginOption`. By the way, since the arguments are optional, it seems that we can merge it with `Begin` function.
Also, for upgrading TiDB smoothly, we can first add options to `Begin` function and keep old `BeginWithOption` unchanged. After TiDB upgrades successfully, we can then remove the old method.

gotoxu: I'll fix it later.

disksing (approved):

disksing:
Thanks @gotoxu , would you like to continue to update tidb?

gotoxu:
@disksing OK!

PR #366, opened 2021-11-02
test: remove ticdc compatibility test

disksing:

TiCDC often becomes incompatible due to TiDB modifications (usually unrelated to client-go). Remove it to avoid noises.


disksing:
no risk. Merging.

PR #368, opened 2021-11-03
txnkv: add callback for setting ResourceGroupTag

mornyx:

Use callback function to generate `ResourceGroupTag` instead of passing it directly. In this way, we can expand `ResourceGroupTag` more flexibly on demand in TiDB, and be transparent to `client-go`.

- Added `ResourceGroupTagger`, which is called to generate tag when `ResourceGroupTag` is `nil`.


mornyx:
PTAL, Thanks~
/cc @breeswish @zhongzc @crazycs520

Upstream PR: [pingcap/tidb#29044](https://github.com/pingcap/tidb/pull/29044).

crazycs520 (commented):
LGTM

breeswish (commented):
How about something like this (feel free to discuss):

```go
type ResourceGroupTagger interface {
    tagCleanup(c *twoPhaseCommitter, bo *retry.Backoffer, batch batchMutations) []byte
    tagCommit(...)
    tagPrewrite(...)
    tagScan(...)
    tagGet(...)
    ....
}
```

or this:

```go
type ResourceGroupTagger func(r *Request)
```

or even this:

```go
type RequestPostBuildFn func(r *Request)
```

@disksing What do you think?

zhongzc:
> How about something like this (feel free to discuss):
> 
> ```go
> type ResourceGroupTagger interface {
>     tagCleanup(c *twoPhaseCommitter, bo *retry.Backoffer, batch batchMutations) []byte
>     tagCommit(...)
>     tagPrewrite(...)
>     tagScan(...)
>     tagGet(...)
>     ....
> }
> ```
> 
> or this:
> 
> ```go
> type ResourceGroupTagger func(r *Request)
> ```
> 
> or even this:
> 
> ```go
> type RequestPostBuildFn func(r *Request)
> ```
> 
> or:
> 
> ```go
> type RequestPreSendFn func(r *Request)
> ```
> 
> @disksing What do you think?

`ResourceGroupTagger` voted.

breeswish:
Generally LGTM. However I believe there should be at least some unit tests to ensure that:

1. When ResourceGroup is set, it appears in the final request.
2. When ResourceGroupTagger is set, its modification take effects in the final request.
3. When ResourceGroup and ResourceGroupTagger are both set, ResourceGroupTagger does not take effect (according to your logic).

breeswish (approved):

zhongzc (approved):

breeswish:
@mornyx I opened an empty PR and seems that integration tests work fine https://github.com/tikv/client-go/pull/372

mornyx:
> @mornyx I opened an empty PR and seems that integration tests work fine #372

Got, seems that problem occurred in the use of mock store in the new integration tests, I will fix it.

mornyx:
> @mornyx I opened an empty PR and seems that integration tests work fine #372

The problem did appear in my own integration test. Because of the incorrect use of other component, the error message did not occur in my code. Now I have fixed it (at least it works on my local machine, Github actions need to wait too long...), and next I need to find a way to fix `Compatibility Test`.

breeswish:
Yes. I've checked with a last merged PR https://github.com/tikv/client-go/commit/8a4ce38c2ac83aec62b6827cb64eb60d8f95a8cd the compatibility test also failed and it is merged anyway. I think we can ignore it and focus on other tests.

breeswish:
@disksing Please take a look~ Thanks!

PR #369, opened 2021-11-08
txn: add newly inserted flag for memdb key node

cfzjywxk:
Signed-off-by: cfzjywxk <lsswxrxr@163.com>

Solution for pingcap/tidb#27564.

Add a new flag for key in memdb, when the key node is allocated by a PUT operation, add this flag to the key. This mark is used to identify the inserted keys generated from the transaction iteself, so that the delete operations on these keys could be skipped committing this transaction, see the comments for details in the issue. There are mainly two situations:
- The key node is newly generated and it has value length greater than 0. This means the node is generated by a insert statement, the existency check mechanism will ensure the key dose not exist in storage.
- The key node is set by a `PUT` opertion and it existing node in `memdb` is a node generated by `LOCK`. This means the insert is done followed a `select for update`.

This change could be risky, need to think it over.


cfzjywxk:
@MyonKeminta @you06 @longfangsong 
PTAL

PR #379, opened 2021-11-18
Bump go version to 1.16

SandyXSD:
Go 1.15 is EOL and 1.17 has been released for about 3 months (https://endoflife.date/go). It should be reasonable to bump required go version to 1.16. I'd like to revert the previous PR (#217).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.15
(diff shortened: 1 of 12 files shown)


SandyXSD:
@disksing Hi, please consider it if the upgrade is not in your schedule, thanks!

sticnarf:
LGTM.

By the way, I think we can define the policy explicitly. For example, we can always drop support of an EOL go version. cc @disksing

disksing:
Thanks @SandyXSD

PR #385, opened 2021-11-26
bugfix:os.ReadFile, module requires Go 1.16

sotowang:
client-go Compatible with go1.16 and below


disksing:
Thanks @sotowang , but we have decided to require minimal version of Go 1.16.

PR #388, opened 2021-12-06
txnkv: decrease some max and base backoff time

youjiali1995:

1. Decrease some `MaxBackoff` to its old value.
2. Decrease the base backoff time of `BoTxnLockFast` from 100ms to 10ms, because 100ms is too big to endure for a small read request.
3. Decrease the base backoff time of `BoTxnLock` from 200ms to 100ms as the same reason for read. Because `BoTxnLock` is seldom used, I didn't decrease it a lot.


sticnarf (approved):
It looks good to decrease base of TxnLockFast to 10ms

PR #392, opened 2021-12-10
txnkv/transaction: add a kv option for cached table lock lease

tiancaiamao:
This is used by the cached table https://github.com/pingcap/tidb/pull/30206

A brief description of the [cached table](https://github.com/pingcap/tidb/issues/25293):

We cache the table data directly in TiDB.
The cached table need to maintain a READ lock and renew the lock lease for caching.
For write operation, it first wait the READ lock lease to expire, then it changes to WRITE lock mode, and maintain the write lock lease for writing (prevent caching).

When transaction commit, we must guarantee **the final commitTS is less than the write lock lease** of a cached table,
otherwise the transaction should abort.

This PR add the transaction option.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -1019,6 +1019,13 @@ func (c *twoPhaseCommitter) checkAsyncCommit() bool {
 		return false
 	}
 
+	// Don't use async commit when writing on cached table.
+	// The actual restriction is (cachedTableWriteLease < commitTS), so using async commit is possible.
+	// But that's unnecessary, because cached table is not designed for write performance.
+	if c.txn.cachedTableWriteLease != nil {
+		return false
+	}
+
 	asyncCommitCfg := config.GetGlobalConfig().TiKVClient.AsyncCommit
 	// TODO the keys limit need more tests, this value makes the unit test pass by now.
 	// Async commit is not compatible with Binlog because of the non unique timestamp issue.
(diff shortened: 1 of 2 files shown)


tiancaiamao:
PTAL  @disksing  @lcwangchao

txnkv/transaction/2pc.go
```diff
@@ -1019,6 +1019,13 @@ func (c *twoPhaseCommitter) checkAsyncCommit() bool {
 		return false
 	}
 
+	// Don't use async commit when writing on cached table.
+	// The actual restriction is (cachedTableWriteLease < commitTS), so using async commit is possible.
+	// But that's unnecessary, because cached table is not designed for write performance.
+	if c.txn.cachedTableWriteLease != nil {
```
disksing, line 1025: Is it possible to make it more abstract and avoid using the term `table`? It does not have a clear meaning at kv layer.

tiancaiamao: `writeLease` seems not a good idea.
Any suggestion about the naming?

lcwangchao: How about expose a method `SetCommitTSValidFunction` instead of `SetCachedTableWriteLease` to custom validation

tiancaiamao: Now it's changed to `SetCommitTSUpperBoundCheck` which is much more general...

disksing (approved):

PR #394, opened 2021-12-16
client: fix metrics of batch client requests

jackysp:
Signed-off-by: Shuaipeng Yu <jackysp@gmail.com>
The requests are always 0, because it reseted before metrics observe.


disksing:
Hi @jackysp would like to remove the debug logs?

PR #395, opened 2021-12-17
txnkv: add shortcut to get global timestamp

disksing:

Make it easier to get current global timestamp. No need to specify scope.


iosmanthus (approved):
LGTM

PR #396, opened 2021-12-18
fix data race and panic when exiting

hawkingrei:

When you exit the KVStore, Some task will still call ```WaitGroup.Add```. It is a dangerous way to make DATA RACE and panic between ```WaitGroup.Wait``` and ```WaitGroup.Add```. 

https://github.com/pingcap/tidb/issues/30658
https://github.com/pingcap/tidb/issues/30651

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -778,6 +778,9 @@ func (c *twoPhaseCommitter) doActionOnGroupMutations(bo *retry.Backoffer, action
 	if actionIsCommit && !actionCommit.retry && !c.isAsyncCommit() {
 		secondaryBo := retry.NewBackofferWithVars(c.store.Ctx(), CommitSecondaryMaxBackoff, c.txn.vars)
 		if c.store.IsClose() {
+			logutil.Logger(bo.GetCtx()).Warn("the store is closed",
+				zap.Uint64("startTS", c.startTS), zap.Uint64("commitTS", c.commitTS),
+				zap.Uint64("sessionID", c.sessionID))
 			return nil
 		}
 		c.store.WaitGroup().Add(1)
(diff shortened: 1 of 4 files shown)


hawkingrei:
@disksing @sticnarf PTAL

txnkv/transaction/2pc.go
```diff
@@ -1100,6 +1105,12 @@ const (
 )
 
 func (c *twoPhaseCommitter) cleanup(ctx context.Context) {
+	if c.store.IsClose() {
+		logutil.Logger(ctx).Error("twoPhaseCommitter fail to cleanup because the store exited",
```
sticnarf, line 1109: I think we can lower the log level to warning. It's not severe that some locks are not cleaned up on exit.

txnkv/transaction/2pc.go
```diff
@@ -775,6 +777,9 @@ func (c *twoPhaseCommitter) doActionOnGroupMutations(bo *retry.Backoffer, action
 	// Already spawned a goroutine for async commit transaction.
 	if actionIsCommit && !actionCommit.retry && !c.isAsyncCommit() {
 		secondaryBo := retry.NewBackofferWithVars(c.store.Ctx(), CommitSecondaryMaxBackoff, c.txn.vars)
+		if c.store.IsClose() {
+			return tikverr.ErrStoreIsExiting
```
sticnarf, line 781: I don't think we should return an error. Failing to commit secondary locks does not mean the transaction fails.
If we return an error here, the user may receive a commit failure when the transaction succeeds to commit.

txnkv/transaction/2pc.go
```diff
@@ -1401,6 +1412,9 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		logutil.Logger(ctx).Debug("2PC will use async commit protocol to commit this txn",
 			zap.Uint64("startTS", c.startTS), zap.Uint64("commitTS", c.commitTS),
 			zap.Uint64("sessionID", c.sessionID))
+		if c.store.IsClose() {
+			return tikverr.ErrStoreIsExiting
```
sticnarf, line 1416: Similar to committing secondary locks, we shouldn't return an error here. Otherwise, the user may receive commit error but the transaction succeeds to commit.

sticnarf (changes requested):

txnkv/transaction/2pc.go
hawkingrei: Done

txnkv/transaction/2pc.go
hawkingrei: Done

txnkv/transaction/2pc.go
hawkingrei: Done

sticnarf (approved):
LGTM

PR #403, opened 2021-12-24
fix RawBatchPutRequest forward compatibility

iosmanthus:

RawBatchPutRequest has forward compatibility issues. This pull request merge this two version into one message and update the request constructor to support the older version of TiKV (< 5.3.0)

Old version used in TiKV < 5.3.0:
```protobuf
message RawBatchPutRequest {
    Context context = 1;
    repeated KvPair pairs = 2;
    string cf = 3;
    uint64 ttl = 4;
    bool for_cas = 5;
}
```
New version used in TiKV 5.3.0:
```protobuf
message RawBatchPutRequest {
    Context context = 1;
    repeated KvPair pairs = 2;
    string cf = 3;
    // The time-to-live for each keys in seconds, and if the length of `ttls`
    // is exactly one, the ttl will be applied to all keys. Otherwise, the length
    // mismatch between `ttls` and `pairs` will return an error.
    repeated uint64 ttls = 4;
    bool for_cas = 5;
}
```

Now, to be compatible with these two version, we modify this message into:

```protobuf
message RawBatchPutRequest {
    Context context = 1;
    repeated KvPair pairs = 2;
    string cf = 3;
    bool for_cas = 5;
	uint64 ttl = 4 [deprecated = true];
    // The time-to-live for each keys in seconds, and if the length of `ttls`
    // is exactly one, the ttl will be applied to all keys. Otherwise, the length
    // mismatch between `ttls` and `pairs` will return an error.
    repeated uint64 ttls = 6;
```

related: 
- [x] https://github.com/tikv/client-java/pull/441
- [x] https://github.com/tikv/client-java/issues/419


sticnarf:
The new proto in this PR description seems not correct? `ttls` should be assigned a new number according to https://github.com/tikv/client-go/pull/403?

sticnarf:
Is there any test for RawKV TTL in this client?

iosmanthus:
> Is there any test for RawKV TTL in this client?

Yes, running tests with `-with-tikv` will trigger the TTL test. But no tests to cover TiKV compatibility for now.

iosmanthus:
> The new proto in this PR description seems not correct? `ttls` should be assigned a new number according to #403?

Sorry, seems the PR description has ambiguity, I update to clarify the versions of `proto` files.

sticnarf (approved):
LGTM.

sticnarf:
@disksing PTAL when you're free because the new format is suggested by you.

Issue #265, opened 2021-08-11
Replica read won't access failed stores until the cached region is invalidated


youjiali1995:
Region cache filters all failed replicas in replica read, and the store information won't be updated until the cached region is invalidated. BTW, seems the load balance strategy isn't very reasonable. It uses an auto-increment seed to choose a replica, but the seed is a global seed.

youjiali1995:
Fixed by new region request sender.

PR #410, opened 2021-12-30
Try to fix issue265. When 'store' is filtered, resolve agagin

zeminzhou:
Try to fix [issue265](https://github.com/tikv/client-go/issues/265).
When the 'store' is filtered, save to 'filteredStore'. At the end of function, call 'reResolve' function for 'store' which save to 'filteredStore'.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -174,13 +174,13 @@ func (r *regionStore) clone() *regionStore {
 
 // return next follower store's index
 func (r *regionStore) follower(seed uint32, op *storeSelectorOp) AccessIndex {
-	filteredStore := []*Store{}
-	defer func(filteredStore []*Store) {
+	var filteredStore []*Store
+	defer func() {
 		for _, s := range filteredStore {
 			_, err := s.reResolve(r.regionCache)
 			tikverr.Log(err)
 		}
-	}(filteredStore)
+	}()
 
 	l := uint32(r.accessStoreNum(tiKVOnly))
 	if l <= 1 {
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -132,6 +132,7 @@ type regionStore struct {
 	stores         []*Store             // stores in this region
 	storeEpochs    []uint32             // snapshots of store's epoch, need reload when `storeEpochs[curr] != stores[cur].fail`
 	accessIndex    [numAccessMode][]int // AccessMode => idx in stores
+	regionCache    *RegionCache
 }
 
 func (r *regionStore) accessStore(mode accessMode, idx AccessIndex) (int, *Store) {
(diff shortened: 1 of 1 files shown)


zeminzhou:
@youjiali1995 PTAL

internal/locate/region_cache.go
```diff
@@ -172,6 +174,14 @@ func (r *regionStore) clone() *regionStore {
 
 // return next follower store's index
 func (r *regionStore) follower(seed uint32, op *storeSelectorOp) AccessIndex {
+	filteredStore := []*Store{}
+	defer func(filteredStore []*Store) {
+		for _, s := range filteredStore {
+			_, err := s.reResolve(r.regionCache)
+			tikverr.Log(err)
+		}
+	}(filteredStore)
```
disksing, line 183:
```suggestion
	var filteredStore []*Store
	defer func() {
		for _, s := range filteredStore {
			_, err := s.reResolve(r.regionCache)
			tikverr.Log(err)
		}
	}()
```

sticnarf:
It seems the `follower` method can be only called from the legacy `GetTiKVRPCContext`. But we have switched to `replicaSelector` for all TiKV requests. `GetTiKVRPCContext` is completedly unused now. cc @youjiali1995

zeminzhou:
> It seems the `follower` method can be only called from the legacy `GetTiKVRPCContext`. But we have switched to `replicaSelector` for all TiKV requests. `GetTiKVRPCContext` is completedly unused now. cc @youjiali1995
> 
> So, I think to improve the similar case in the new implementation, we need to re-resolve the followers skipped by `accessFollower`?

zeminzhou:
> It seems the `follower` method can be only called from the legacy `GetTiKVRPCContext`. But we have switched to `replicaSelector` for all TiKV requests. `GetTiKVRPCContext` is completedly unused now. cc @youjiali1995
> 
> So, I think to improve the similar case in the new implementation, we need to re-resolve the followers skipped by `accessFollower`?

New implementation means region_request.go ?

sticnarf:
> New implementation means region_request.go ?

Yes, it's in `region_request.go`.

But it seems to me that the exact problem does not exist now... When the store epoch is updated, it will inform `asyncCheckAndResolveLoop` to update store info. @youjiali1995 can help confirm if I understand the original issue correctly...

youjiali1995 (commented):
> But it seems to me that the exact problem does not exist now... When the store epoch is updated, it will inform asyncCheckAndResolveLoop to update store info. @youjiali1995 can help confirm if I understand the original issue correctly...

This issue doesn't exist now. 

And `reResolve` stores can't solve the issue in the old implementation. A peer can be accessed only if its `regionStore.storeEpochs` is equal to the corresponding `Store.epoch`. `regionStore.storeEpochs` won't be changed until a region is invalidated in the old implementation.

Issue #415, opened 2022-01-11
What error type should be used when max sleep time of backoff exceeded?


sticnarf:
Now we return the error of the first backoff type:

https://github.com/tikv/client-go/blob/d2281509972049b341132b08f9c6eae921f1c456/internal/retry/backoff.go#L154

But it is possible that after some rounds of retries, the error leading to the final timeout is different. Reporting the first backoff error may confuse the user. Is it better to return the last backoff or the most frequent backoff type?

cc @cfzjywxk @disksing

cfzjywxk:
Returning the last backoff type seems to be fine, the detailed backoff information including the most frequent backoff type could be printted in the error log.

sticnarf:
At the very beginning, the last backoff is used, while https://github.com/pingcap/tidb/issues/6793 claims using the first backoff is sometimes better.

So this leads me to thinking that the most frequent backoff may be a better idea.

Of course I think the log should include more information instead of only the latest backoff cfg.

cfzjywxk:
Make sense, the most frequent backoff may help more to reveal the root cause for issues https://github.com/pingcap/tidb/issues/6793 describes, to answer what takes so long to backoff seems to be more useful to answer what makes the backoff timeout finally.

sticnarf:
In #416, I choose the backoff that sleeps longest, considering different backoffs may vary a lot in their sleep base and cap, so it may be not so fair to use the frequency.

PR #416, opened 2022-01-12
internal/retry: return the longest-sleeping type on backoff timeout

sticnarf:
Closes https://github.com/tikv/client-go/issues/415

Before this PR, when `maxSleep` of a backoff is exceeded, the first error causing the backoff is returned. However, the first error may be just an accidental and temporary error.

For example, the first error can be a `txnLockFast` that is resolved very soon but later backoffs are all `regionMiss`. In this case, we expect the error returned should be "Region is unavailable", but the actual returned error is "resolve lock timeout", which is the error corresponding to `txnLockFast`.

We will encounter the similar problem if we use the last backoff type. The last error may be a temporary error, but most errors that causing the timeout can be something else. Then, the user may receive an unrepresentative error.

So this PR changes to return the error which sleeps for the longest time. Not using the most frequent backoff is because different backoff types have different base and caps. Those errors with shorter base and caps may be much easier to be the most frequent one, but these events may only contributes a small portion to the timeout.

### DIFF
--- a/internal/retry/backoff.go
+++ b/internal/retry/backoff.go
@@ -138,10 +138,11 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 		return errors.WithStack(err)
 	default:
 	}
-
-	b.errors = append(b.errors, errors.Errorf("%s at %s", err.Error(), time.Now().Format(time.RFC3339Nano)))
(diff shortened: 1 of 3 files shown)


sticnarf:
cc @cfzjywxk @youjiali1995

cfzjywxk (approved):
LGTM

youjiali1995 (approved):

Issue #284, opened 2021-08-24
Using tikv client-go@v2.0.0-alpha has some error


1026237416:
![image](https://user-images.githubusercontent.com/6825312/130616632-2be122da-07ec-4add-8a0e-468d2a24d32c.png)

disksing:
Hi @1026237416 , do you import etcd in your project?

cmsax:
I have the same issue when using `/v2/tikv`, there is an compiler error:

```
../../../../go/pkg/mod/github.com/tikv/client-go/v2@v2.0.0-alpha/tikv/txn.go:47:2: missing go.sum entry for module providing package github.com/dgryski/go-farm (imported by github.com/tikv/client-go/v2/tikv); to add:
        go get github.com/tikv/client-go/v2/tikv@v2.0.0-alpha
```

after ran `go get`, there was another error:

```
go get github.com/tikv/client-go/v2/tikv@v2.0.0-alpha
# go.etcd.io/etcd/clientv3/balancer/picker
../../../../go/pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200824191128-ae9734ed278b/clientv3/balancer/picker/err.go:25:9: cannot use &errPicker{...} (type *errPicker) as type Picker in return argument:
        *errPicker does not implement Picker (wrong type for Pick method)
                have Pick(context.Context, balancer.PickInfo) (balancer.SubConn, func(balancer.DoneInfo), error)
                want Pick(balancer.PickInfo) (balancer.PickResult, error)
../../../../go/pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200824191128-ae9734ed278b/clientv3/balancer/picker/roundrobin_balanced.go:33:9: cannot use &rrBalanced{...} (type *rrBalanced) as type Picker in return argument:
        *rrBalanced does not implement Picker (wrong type for Pick method)
                have Pick(context.Context, balancer.PickInfo) (balancer.SubConn, func(balancer.DoneInfo), error)
                want Pick(balancer.PickInfo) (balancer.PickResult, error)
```

I've imported `trpc` which is a framework like `gRPC` and use `etcd` internally, following is my `go.mod`:

![image](https://user-images.githubusercontent.com/18530762/131433469-cdfb48eb-26f7-4c4b-92d3-50e59f5bb7d8.png)

disksing:
[...]

Issue #63, opened 2020-12-15
Upgrade etcd/grpc version


strahe:
tidb current version: https://github.com/pingcap/tidb/blob/master/go.mod#L70

There may be some incompatibilities in the current version：
![image](https://user-images.githubusercontent.com/7932644/102166404-f46a8380-3ec6-11eb-8bf3-d090e459cce1.png)

omaskery:
We encountered this while trying to upgrade one of our projects to use the newest versions of the Go gRPC libraries, `protoc` versions, etc.

I had a brief look into creating a PR to address this but found quite quickly that this is currently all blocked on some stuff going on in the `etcd` Go client library: https://github.com/etcd-io/etcd/pull/12398

disksing:
@omaskery Thank you for raising this issue, I have added it to the list and managed to put the preceding issue to the roadmap schedule of the PD project. Hopefully, we will upgrade it after the PD upgrade is completed before the release.

PR #421, opened 2022-02-10
Update etcd to latest version

killzoner:
The recent update of tidb (cea0c4c6e67c4dfd92ba71a4cb543ac06225ab39) made it possible to update painlessly `etcd` to latest version.

In my tests, it solves conflict with more recent versions of `grpc` (I successfully used `grpc` 1.44 in my project using the fork) and issues like the one mentioned here : https://github.com/tikv/client-go/issues/63#issue-767123043 or https://github.com/tikv/client-go/issues/284

Unless there is some unknown constraint to me, that would be great to have this merged, which is preventing me so far to use at all this library in a monorepo having latest `grpc` dependency.

Thanks!

EDIT: in the same spirit, I created https://github.com/pingcap/tidb/pull/32242 to remove completely `go.etcd.io/etcd v0.5.0-alpha.5.0.20210512015243-d19fbe541bf9` from dependency path. 
Might also be able to get rid of the `grpc` replace with that merged


killzoner:
Hey @maoueh, my pleasure, I think the tests failing are all linked to `go.etcd.io/etcd v0.5.0-alpha.5.0.20210512015243-d19fbe541bf9` being still in the dependency tree (hence the proto definitions registered twice).

This PR on `tidb` should help remove it completely (https://github.com/pingcap/tidb/pull/32242#event-6044161593) but it relies itself on 3 other PRs on pd/tidb-tools/tidb-dashboard, so it will probably take a while to be done

disksing:
Hi @killzoner thanks for your work. The main problem that currently affects the upgrade process is that `tidb-dashboard` does not use multiple release branches like PD, i.e. all PD branches use the master branch of dashboard. Therefore, if you upgrade the etcd on which tidb-dashboard depends, the other release branches of PD will no longer be able to update the dashboard.

I'm now trying another approach: splitting `pd/client` into separate mods (and making them not dependent on etcd), so that client-go (including tidb) may be free to choose a version regardless of the version of etcd used by pd. Of course, we need to make sure that they are compatible at the API level.

Currently the `pd/client` is using a standalone go mod (https://github.com/tikv/pd/pull/4569). I'm trying to update client-go (https://github.com/tikv/client-go/pull/423) and tidb. If it goes well, this should solve the compatibility problems caused by the old etcd.

killzoner:
Hey @disksing I think it's a better approach, because I gave it a try updating all the other projects with tied dependency, and it's a mess. Your way should be much smoother and ease maintenance for future.

Will probably close all my PRs then.

Thanks

Issue #428, opened 2022-02-24
Unexpected result when DeleteRange at range [nil, nil)


Jy-poi:
It does not delete all data in tikv when i called `DeleteRange(ctx, nil, nil, 10)`

Reproduce code:
```go
package main

import (
	"context"
	"fmt"

	"github.com/alecthomas/kong"
	"github.com/sirupsen/logrus"
	"github.com/tikv/client-go/v2/tikv"
)

// clean Tikv command clean all data in tikv, be careful to use.

type CleanTikvCmd struct {
	Concurrency  int32    `short:"c" default:"10" help:"clean concurrency"`
	PDEndpoints  []string `required:"" help:"endpoints of tikv's pd"`
	UseMaxEndkey bool     `default:"false" help:"use []byte{255...} as end key when DeleteRange"`
}

func CountKey(tikvClient *tikv.KVStore, begin, end []byte) error {
	numberOfKeys := 0
	txn, err := tikvClient.Begin()
	if err != nil {
		return err
	}
	txn.GetSnapshot().SetKeyOnly(true)

	// Here use [nil, nil] can count all of keys.
	iter, err := txn.Iter(begin, end)
	if err != nil {
		return err
	}

	for iter.Valid() {
		numberOfKeys += 1
		if err := iter.Next(); err != nil {
			return err
		}
	}

	logrus.Infof("tikv: number of keys in range [%v, %v] is %v", begin, end, numberOfKeys)
	return nil
}

func PutKeys(ctx context.Context, tikvClient *tikv.KVStore, count int) error {
	txn, err := tikvClient.Begin()
	if err != nil {
		return err
	}
	for i := 0; i < 10; i++ {
		err := txn.Set([]byte(fmt.Sprint(i)), []byte("balala"))
		if err != nil {
			return err
		}
	}
	err = txn.Commit(ctx)
	return err
}

func (cmd *CleanTikvCmd) Run() error {
	ctx := context.Background()
	tikvClient, err := tikv.NewTxnClient(cmd.PDEndpoints)
	if err != nil {
		return err
	}

	if err := PutKeys(ctx, tikvClient, 10); err != nil {
		return err
	}
	var (
		begin []byte = nil
		end   []byte = nil
	)
	if cmd.UseMaxEndkey {
		for i := 0; i < 32; i++ {
			end = append(end, 255)
		}
	}

	logrus.Infof("count key before DeleteRange")
	if err := CountKey(tikvClient, begin, end); err != nil {
		return err
	}

[...]

PR #429, opened 2022-02-24
txnkv: fix the issue that deleteRange cannot use nil as endkey

disksing:
See https://github.com/tikv/client-go/issues/428

Note: it only fix the issue on client side. We need to update tikv to resolve it completely.

### DIFF
--- a/internal/mockstore/mocktikv/mvcc_leveldb.go
+++ b/internal/mockstore/mocktikv/mvcc_leveldb.go
@@ -1538,7 +1538,11 @@ func (mvcc *MVCCLevelDB) GC(startKey, endKey []byte, safePoint uint64) error {
 
 // DeleteRange implements the MVCCStore interface.
 func (mvcc *MVCCLevelDB) DeleteRange(startKey, endKey []byte) error {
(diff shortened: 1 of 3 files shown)


internal/mockstore/mocktikv/mvcc_leveldb.go
```diff
@@ -1538,7 +1538,11 @@ func (mvcc *MVCCLevelDB) GC(startKey, endKey []byte, safePoint uint64) error {
 
 // DeleteRange implements the MVCCStore interface.
 func (mvcc *MVCCLevelDB) DeleteRange(startKey, endKey []byte) error {
-	return mvcc.doRawDeleteRange(codec.EncodeBytes(nil, startKey), codec.EncodeBytes(nil, endKey))
+	var end []byte
+	if len(endKey) > 0 {
+		end = codec.EncodeBytes(nil, endKey)
+	}
```
youjiali1995, line 1544: Why not do it for start key too?

disksing: It is innecessary. There will be no key located before `""` or `encode("")` so encode or not encode both are fine.

youjiali1995 (approved):

Jy-poi:
LGTM for me

PR #439, opened 2022-03-04
region_cache: support buckets

youjiali1995:

Region cache now saves buckets information which is fetched from PD by `GetRegion()` or `GetRegionByID()`(https://github.com/tikv/pd/pull/4699):
- buckets has a version which is the local unix timestamp and it contains range of each buckets.
- buckets are stored in the `RegionStore` because it can change without region change.
- TiDB splits copr requests by buckets in https://github.com/pingcap/tidb/pull/32818.

There are some TODOs left, I will file PRs for them:
- `BatchLoadRegionsWithKeyRange()` which is used by GC worker populates the region cache now.
- Reduce `GetRegion()` traffic flow by `singleflight`.
- Optimize some update policy, e.g., on epoch-not-match error.
- Replace the TiDB dependency. Now it depends on a develop branch.


youjiali1995:
@cfzjywxk PTAL too, if you have time.

internal/locate/region_cache.go
```diff
@@ -734,6 +745,48 @@ func (l *KeyLocation) String() string {
...
+// If the region doesn't contain buckets, returns 0.
+func (l *KeyLocation) GetBucketVersion() uint64 {
+	if l.Buckets == nil {
+		return 0
+	}
+	return l.Buckets.GetVersion()
+}
+
+// LocateBucket returns the bucket the key is located.
+func (l *KeyLocation) LocateBucket(key []byte) *Bucket {
```
sticnarf, line 758: Maybe it can be simplified like:
```go
keys := l.Buckets.GetKeys()
searchLen := len(keys) - 1
i := sort.Search(searchLen, func(i int) bool {
	return bytes.Compare(key, keys[i]) < 0
})

// buckets contains region's start/end key, so i==0 means it can't find a suitable bucket
// which can happen if the bucket information is stale.
if i == 0 ||
	// the key isn't located in the last range.
	(i == searchLen && len(keys[searchLen]) != 0 && bytes.Compare(key, keys[searchLen]) >= 0) {
	return nil
}
return &Bucket{keys[i-1], keys[i]}
```

internal/locate/region_cache.go
```diff
@@ -1027,6 +1084,8 @@ func (c *RegionCache) BatchLoadRegionsWithKeyRange(bo *retry.Backoffer, startKey
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	// TODO(youjiali1995): scanRegions always fetch regions from PD and these regions don't contain buckets information
```
sticnarf, line 1087: Does it mean that we are going to add option to the `ScanRegions` interface, so it can return buckets in the same response?

internal/locate/region_cache.go
```diff
@@ -1533,17 +1623,29 @@ func (c *RegionCache) OnRegionEpochNotMatch(bo *retry.Backoffer, ctx *RPCContext
...
 			var err error
+			// Can't modify currentRegions in this function because it can be shared by
+			// multiple goroutines, refer to https://github.com/pingcap/tidb/pull/16962.
 			if meta, err = decodeRegionMetaKeyWithShallowCopy(meta); err != nil {
 				return false, errors.Errorf("newRegion's range key is not encoded: %v, %v", meta, err)
 			}
 		}
-		region, err := newRegion(bo, c, &pd.Region{Meta: meta})
+		// TODO(youjiali1995): new regions inherit old region's buckets now. Maybe we should make EpochNotMatch error
+		// carry buckets information. Can it bring much overhead?
```
sticnarf, line 1647: Is is true that bucket information usually doesn't change greatly just after a region merge or split? Then, an asynchronous `UpdateBucketsIfNeeded` from PD may be ok.

internal/locate/region_cache.go
```diff
@@ -734,6 +745,48 @@ func (l *KeyLocation) String() string {
...
+	// which can happen if the bucket information is stale.
+	if i == 0 ||
+		// the key isn't located in the last range.
+		(i == searchLen && !isMaxEndKey) {
+		return nil
+	}
+	return &Bucket{keys[i-1], keys[i]}
+}
+
+// Bucket contains the bucket version, bucket ranges of a region.
```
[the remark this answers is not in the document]
youjiali1995: I forgot to change it.

internal/locate/region_cache.go
youjiali1995: No, too much network traffic. A 10GB region can contains 100 buckets, so the payload of a single region can be several KBs. I will make `BatchLoadRegionsWithKeyRange()` use `scanRegionsFromCache()` or not insert regions to cache.

internal/locate/region_cache.go
youjiali1995: Agree. Buckets is generated asynchronously after region leader or range changed. Querying PD immediately may not get the newer buckets, so not handling it and depending on https://github.com/pingcap/tidb/pull/32818/files#diff-23eb1043649369d5956b123b173b799ded8c6b7cbf4519bb32cb3caa01928d75R936 is also a choice.

sticnarf (approved):
LGTM

PR #453, opened 2022-03-21
backoff: deep copy the needed fields for backoffer type

cfzjywxk:
Close: https://github.com/pingcap/tidb/issues/33265

Related changes:
- Avoid possible race changing the global backoff type config.
- Check nil before using the `longestSleepCfg` result.
- Copy the `configs`, `backoffSleepMS` and `backoffTimes` fields more deeply but not complete deeply.

The `Backoffer` is not very robust by now and it's easy to make mistakes, maybe we need to open a enhancement task for it in the future to avoid data races and related risks.

Signed-off-by: cfzjywxk <lsswxrxr@163.com>

### DIFF
--- a/internal/retry/backoff.go
+++ b/internal/retry/backoff.go
@@ -150,10 +150,14 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 				errMsg += "\n" + err.Error()
 			}
 		}
-		errMsg += fmt.Sprintf("\nlongest sleep type: %s, time: %dms", longestSleepCfg.String(), longestSleepTime)
+		returnedErr := err
+		if longestSleepCfg != nil {
+			errMsg += fmt.Sprintf("\nlongest sleep type: %s, time: %dms", longestSleepCfg.String(), longestSleepTime)
+			returnedErr = longestSleepCfg.err
+		}
 		logutil.BgLogger().Warn(errMsg)
 		// Use the backoff type that contributes most to the timeout to generate a MySQL error.
-		return errors.WithStack(longestSleepCfg.err)
+		return errors.WithStack(returnedErr)
 	}
 	b.errors = append(b.errors, errors.Errorf("%s at %s", err.Error(), time.Now().Format(time.RFC3339Nano)))
 	b.configs = append(b.configs, cfg)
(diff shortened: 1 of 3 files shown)


internal/retry/backoff.go
```diff
@@ -219,32 +223,51 @@ func (b *Backoffer) String() string {
...
-		errors:        b.errors,
-		vars:          b.vars,
-		parent:        b.parent,
+		ctx:            b.ctx,
+		maxSleep:       b.maxSleep,
+		totalSleep:     b.totalSleep,
+		excludedSleep:  b.excludedSleep,
+		vars:           b.vars,
+		errors:         append([]error{}, b.errors...),
+		configs:        append([]*Config{}, b.configs...),
```
youjiali1995, line 247:
```suggestion
		errors:         append(make([]error, len(b.errors)), b.errors...),
		configs:        append(make([]*Config, len(b.errors)), b.configs...),
```

internal/retry/backoff.go
```diff
@@ -219,32 +223,51 @@ func (b *Backoffer) String() string {
...
-		excludedSleep: b.excludedSleep,
-		errors:        b.errors,
-		vars:          b.vars,
-		parent:        b,
+		ctx:            ctx,
+		maxSleep:       b.maxSleep,
+		totalSleep:     b.totalSleep,
+		excludedSleep:  b.excludedSleep,
+		errors:         append([]error{}, b.errors...),
+		configs:        append([]*Config{}, b.configs...),
```
youjiali1995, line 266:
```suggestion
		errors:         append(make([]error, len(b.errors)), b.errors...),
		configs:        append(make([]*Config, len(b.configs)), b.configs...),
```

internal/retry/backoff.go
cfzjywxk: Could `append(make([]*Config, 0, len(b.errors)), b.configs...)` be a faster way? I'm not quite sure about it.

youjiali1995: Oh, yes. Should be `make([]*Config, 0, len(b.errors))`. It should be faster, I think.

sticnarf: I doubt it too. I try these two, and they perform the same.

```go
func BenchmarkAppend1(b *testing.B) {
	source := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
	for i := 0; i < b.N; i++ {
		_ = append([]int{}, source...)
	}
}

func BenchmarkAppend2(b *testing.B) {
	source := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
	for i := 0; i < b.N; i++ {
		_ = append(make([]int, 0, len(source)), source...)
	}
}
```

```
BenchmarkAppend1-24     49070558                40.73 ns/op           80 B/op          1 allocs/op
BenchmarkAppend2-24     38036925                39.76 ns/op           80 B/op          1 allocs/op
```

sticnarf (approved):
LGTM

internal/retry/backoff.go
youjiali1995: >  It should be faster, I think.

To be more accurate, it can't be slower.

youjiali1995: I tried some benchmarks and found it's not faster because it still use `append()` to copy data. If we use `copy()` here, I think it should be faster.

youjiali1995 (approved):

PR #458, opened 2022-03-29
*: use global variable instead of config for `TxnCommitBatchSize`

jackysp:
Signed-off-by: Jack Yu <jackysp@gmail.com>

As title says. Asked by @morgo.


disksing:
What is the difference to use a global atomic variable?

jackysp:
See https://github.com/pingcap/tidb/pull/33507#issuecomment-1081352558 @disksing .
It is easy to modify online, but currently the semantics of `set global` are confusing here. As a result, `set global` for the different variables are not in the same scope.

disksing (approved):

PR #459, opened 2022-03-31
tikv:move split region request to pd

qidi1:
Signed-off-by: qidi1 <1083369179@qq.com>

Issue Number: tikv/pd/issues/4095
Proposal PR:  pingcap/tidb/pull/33664
The original region split and scatter are handled by tikv and pd respectively, in order to improve efficiency, the split is also changed to be handled and controlled by PD.

 In SplitRegions function,Using split api provide by PD instead of TiKV

Code changes based on this [work](https://github.com/hzh0425/pd/tree/add_split_and_scatter_region)
Thanks to [hzh0435](https://github.com/hzh0425)

### Check List

Tests

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [x] No code

Code changes

- [ ] Has configuration change
- [ ] Has HTTP API interfaces change (Don't forget to [add the declarative for API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- [ ] Has persistent data change

Side effects

- [ ] Possible performance regression
- [ ] Increased code complexity
- [ ] Breaking backward compatibility

### DIFF
--- a/tikv/split_region.go
+++ b/tikv/split_region.go
@@ -40,18 +40,13 @@ import (
 	"fmt"
 	"math"
 
-	"github.com/pingcap/kvproto/pkg/kvrpcpb"
-	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pkg/errors"
 	tikverr "github.com/tikv/client-go/v2/error"
-	"github.com/tikv/client-go/v2/internal/client"
 	"github.com/tikv/client-go/v2/internal/kvrpc"
-	"github.com/tikv/client-go/v2/internal/locate"
 	"github.com/tikv/client-go/v2/internal/logutil"
 	"github.com/tikv/client-go/v2/internal/retry"
 	"github.com/tikv/client-go/v2/kv"
-	"github.com/tikv/client-go/v2/tikvrpc"
 	"github.com/tikv/client-go/v2/txnkv/rangetask"
 	"github.com/tikv/client-go/v2/util"
 	pd "github.com/tikv/pd/client"
@@ -64,7 +59,7 @@ func equalRegionStartKey(key, regionStartKey []byte) bool {
 	return bytes.Equal(key, regionStartKey)
 }
 
-func (s *KVStore) splitBatchRegionsReq(bo *Backoffer, keys [][]byte, scatter bool, tableID *int64) (*tikvrpc.Response, error) {
+func (s *KVStore) splitBatchRegionsReq(bo *Backoffer, keys [][]byte, scatter bool, tableID *int64) (*pdpb.SplitRegionsResponse, error) {
 	// equalRegionStartKey is used to filter split keys.
 	// If the split key is equal to the start key of the region, then the key has been split, we need to skip the split key.
 	groups, _, err := s.regionCache.GroupKeysByRegion(bo, keys, equalRegionStartKey)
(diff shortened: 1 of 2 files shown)


disksing:
would you like to explain why do you want to move it to using PD?

qidi1:
> would you like to explain why do you want to move it to using PD?

we are working in this [issue](https://github.com/tikv/pd/issues/4095),In Tikv,In tikv, it does not have the ability to check the current condition of the region, leading to the problem of unstable splitting. We move it to pd which having the global state of the region will reduce the probability of the problem in the above issue

qidi1:
> would you like to explain why do you want to move it to using PD?

I updated the description of the PR,PTAL

disksing:
It seems that the integration tests (run with tikv) are failed.

qidi1:
Sorry, I can't run integration tests locally，Because the tidb package is not available to me

qidi1:
> Sorry, I can't run integration tests locally，Because the tidb package is not available to me

I change pd.RegionSplit like tikv.RegionSplit,please try to run it

disksing:
Hi @qidi1 , would like to fix the lint warnings? (some of them can be fixed by merging master)

tikv/split_region.go
```diff
@@ -227,10 +191,9 @@ func (s *KVStore) SplitRegions(ctx context.Context, splitKeys [][]byte, scatter
 	bo := retry.NewBackofferWithVars(ctx, int(math.Min(float64(len(splitKeys))*splitRegionBackoff, maxSplitRegionsBackoff)), nil)
 	resp, err := s.splitBatchRegionsReq(bo, splitKeys, scatter, tableID)
 	regionIDs = make([]uint64, 0, len(splitKeys))
-	if resp != nil && resp.Resp != nil {
-		spResp := resp.Resp.(*kvrpcpb.SplitRegionResponse)
-		for _, r := range spResp.Regions {
-			regionIDs = append(regionIDs, r.Id)
+	if resp != nil && resp.RegionsId != nil {
```
disksing, line 194:
```suggestion
	if resp != nil {
```

tikv/split_region.go
```diff
@@ -141,80 +137,48 @@ func (s *KVStore) batchSendSingleRegion(bo *Backoffer, batch kvrpc.Batch, scatte
...
+
 	logutil.BgLogger().Info("batch split regions complete",
 		zap.Uint64("batch region ID", batch.RegionID.GetID()),
 		zap.String("first at", kv.StrKey(batch.Keys[0])),
-		zap.String("first new region left", newRegionLeft),
-		zap.Int("new region count", len(spResp.Regions)))
-
+		zap.Int("new region count", len(regionIDs)))
+	if resp.FinishedPercentage != 100 {
+		err = errors.Errorf("Fail to batch split regions, finishedPercentage : %d, batch region ID : %d",
```
disksing, line 159: where is the `err` be used?

tikv/split_region.go
```diff
@@ -141,80 +137,49 @@ func (s *KVStore) batchSendSingleRegion(bo *Backoffer, batch kvrpc.Batch, scatte
...
+			resp.FinishedPercentage, batch.RegionID.GetID())
+		resp.Header.Error = &pdpb.Error{Message: err.Error()}
+	}
```
qidi1, line 162: I forget to use before,I want to put it in here.

tikv/split_region.go
qidi1: I forget to use before,I want to use it like this
```
	if resp.FinishedPercentage != 100 {
		err = errors.Errorf("Fail to batch split regions, finishedPercentage : %d, batch region ID : %d",
			resp.FinishedPercentage, batch.RegionID.GetID())
		resp.Header.Error = &pdpb.Error{Message: err.Error()}
	}
```

qidi1: > where is the `err` be used?
I forget to use before,I want to use it like this
```
	if resp.FinishedPercentage != 100 {
		err = errors.Errorf("Fail to batch split regions, finishedPercentage : %d, batch region ID : %d",
			resp.FinishedPercentage, batch.RegionID.GetID())
		resp.Header.Error = &pdpb.Error{Message: err.Error()}
	}
```

disksing:
Hi, it is nearly mergable! but there is still a lint warning:
```
Error: S1011: should replace loop with `regionIDs = append(regionIDs, resp.RegionsId...)` (gosimple)
```

disksing (approved):

disksing:
Thanks @qidi1

PR #465, opened 2022-04-06
client: fix the race between connArray.Close() and connArray.Get()

youjiali1995:

Make `v` immutable after initialization.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -137,7 +137,7 @@ func newConnArray(maxSize uint, addr string, security config.Security, idleNotif
 func (a *connArray) Init(addr string, security config.Security, idleNotify *uint32, enableBatch bool) error {
 	a.target = addr
 
-	opt := grpc.WithInsecure() //nolint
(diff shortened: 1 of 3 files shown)


internal/client/client.go
```diff
@@ -237,11 +237,10 @@ func (a *connArray) Close() {
 		a.batchConn.Close()
 	}
 
-	for i, c := range a.v {
+	for _, c := range a.v {
```
disksing, line 240: - if `c` can never be `nil`, we can remove following `if`
- will it cause problem if someone call `Close` multiple times?

youjiali1995: - if `c` can never be `nil`, we can remove following `if`
Of course.

- will it cause problem if someone call `Close` multiple times?
`connArray.Close()` isn't able to be called multiple times even without this change, and we protect it by lock and it can't be called mutilple times.

disksing: you are right.

disksing (approved):

PR #473, opened 2022-04-15
txn: record resolving locks

longfangsong:


txnkv/txnlock/lock_resolver.go
```diff
@@ -76,13 +78,21 @@ type LockResolver struct {
 	asyncResolveCancel func()
 }
 
+type LockInfo struct {
```
disksing, line 81: There are many lock types. Maybe call it `ResolvingLock`

txnkv/txnlock/lock_resolver.go
```diff
@@ -425,6 +449,29 @@ func (lr *LockResolver) resolveLocks(bo *retry.Backoffer, callerStartTS uint64,
...
+	lr.mu.Unlock()
+}
+
+func (lr *LockResolver) Resolving() []LockInfo {
+	result := []LockInfo{}
+	lr.mu.RLock()
+	defer lr.mu.RUnlock()
+	for txnID, item := range lr.mu.resolving {
+		for _, lock := range item {
+			result = append(result, LockInfo{
```
disksing, line 464: any need of dedup?

txnkv/txnlock/lock_resolver.go
```diff
@@ -319,6 +337,13 @@ func (lr *LockResolver) ResolveLocks(bo *retry.Backoffer, callerStartTS uint64,
 // Read operations needn't wait for resolve secondary locks and can read through(the lock's transaction is committed
 // and its commitTS is less than or equal to callerStartTS) or ignore(the lock's transaction is rolled back or its minCommitTS is pushed) the lock .
 func (lr *LockResolver) ResolveLocksForRead(bo *retry.Backoffer, callerStartTS uint64, locks []*Lock, lite bool) (int64, []uint64 /* canIgnore */, []uint64 /* canAccess */, error) {
+	resolving := []Lock{}
+	for _, lock := range locks {
+		resolving = append(resolving, *lock)
+	}
+	lr.mu.Lock()
+	lr.mu.resolving[callerStartTS] = resolving
+	lr.mu.Unlock()
```
disksing, line 346: for DRY reason, better to extract a new function for it.

txnkv/transaction/prewrite.go
```diff
@@ -365,6 +365,7 @@ func (action actionPrewrite) handleSingleBatch(c *twoPhaseCommitter, bo *retry.B
 		}
 		start := time.Now()
 		msBeforeExpired, err := c.store.GetLockResolver().ResolveLocks(bo, c.startTS, locks)
+		defer c.store.GetLockResolver().ResolveLocksDone(c.startTS)
```
sticnarf, line 368: Why is the `defer` written in prewrite instead of in `ResolveLocks`? `ResolveLocksDone` looks missing if resolve lock is triggered by other commands.

longfangsong: > Why is the `defer` written in prewrite instead of in `ResolveLocks`?

It's because we need to keep this infomation during backoffing.

> ResolveLocksDone looks missing if resolve lock is triggered by other commands.

I've added them now.

txnkv/transaction/pessimistic.go
```diff
@@ -227,6 +227,7 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
 		// tikv default will wait 3s(also the maximum wait value) when lock error occurs
 		startTime = time.Now()
 		msBeforeTxnExpired, err := c.store.GetLockResolver().ResolveLocks(bo, 0, locks)
+		defer c.store.GetLockResolver().ResolveLocksDone(c.startTS)
```
MyonKeminta, line 230: It's in a retry loop here. It seems we'd better protect it from being deferred more than once or do the call before entering the next loop? The same question to the prewrite part.

txnkv/txnlock/lock_resolver.go
```diff
@@ -319,9 +331,20 @@ func (lr *LockResolver) ResolveLocks(bo *retry.Backoffer, callerStartTS uint64,
 // Read operations needn't wait for resolve secondary locks and can read through(the lock's transaction is committed
 // and its commitTS is less than or equal to callerStartTS) or ignore(the lock's transaction is rolled back or its minCommitTS is pushed) the lock .
 func (lr *LockResolver) ResolveLocksForRead(bo *retry.Backoffer, callerStartTS uint64, locks []*Lock, lite bool) (int64, []uint64 /* canIgnore */, []uint64 /* canAccess */, error) {
+	lr.saveResolvingLocks(locks, callerStartTS)
 	return lr.resolveLocks(bo, callerStartTS, locks, true, lite)
 }
 
+func (lr *LockResolver) saveResolvingLocks(locks []*Lock, callerStartTS uint64) {
```
sticnarf, line 338: What if `ResolveLocks` happens concurrently for multiple regions?

For example, a batch get involves multiple regions and it encounters locks in more than one regions. But the `resolving` array always only includes the information of one region.

MyonKeminta:
In this implementation, it requires all callers to `ResolveLocks`/`ResolveLocksForRead` to call `ResolveLocksDone` when they finish. Besides the usages within client-go, there are also some invocations in tidb, br, etc. I'm afraid about that. If we cannot put saving and removing logic both within `ResolveLocks`, I suggest making `saveResolvingLocks` a public method, and only who called `SaveResolvingLocks` explicitly need to call `ResolveLocksDone`.

sticnarf:
Looks working but it seems to create more complexity to the code than I expected, though I don't have a good idea yet...

MyonKeminta (approved):

sticnarf (approved):

PR #476, opened 2022-04-24
Revert "tikv:move split region request to pd (#459)"

disksing:
This patch will break tests in TiDB and currently there is no corresponding PR in TiDB side. We should revert it now, maybe consider do it when it is all ready.


AndreMouche (approved):
LGTM

PR #481, opened 2022-04-27
support find stores with tiflash_mpp label

guo-shaoge:
Signed-off-by: guo-shaoge <shaoge1994@163.com>

close： #498

Related tidb pr: https://github.com/pingcap/tidb/pull/33535

1. add `GetTiFlashMPPRPCContextByConsistentHash` and its related functions.
2. handle tiflash_mpp in `onSendFail()`.
3. change some if branch from `ctx.Store.storeType == tikvrpc.TiFlash` `to ctx.Store.storeType.IsTiFlashRelatedType()`.


internal/locate/region_cache.go
```diff
@@ -1723,6 +1730,59 @@ func (c *RegionCache) GetTiFlashStores() []*Store {
...
+	for _, s := range stores {
+		tmpStore := &Store{
+			storeID:   s.GetId(),
+			addr:      s.GetAddress(),
+			saddr:     s.GetStatusAddress(),
+			storeType: tikvrpc.GetStoreTypeByMeta(s),
+			labels:    s.GetLabels(),
+			state: uint64(resolved),
+		}
+		if tmpStore.IsLabelsMatch(mppLabels) {
```
XuHuaiyu, line 1766: Will this be better?
``` go
if s.IsLableMatch(mppLabels) {
    tmpStore = &Store{}
    res = append(res, tmpStore)
}
```

internal/locate/region_cache.go
```diff
@@ -1723,6 +1730,59 @@ func (c *RegionCache) GetTiFlashStores() []*Store {
...
+
+func (c *RegionCache) reloadTiFlashMPPStores(bo *retry.Backoffer) (res []*Store, _ error) {
+	stores, err := c.pdClient.GetAllStores(bo.GetCtx())
+	if err != nil {
+		return nil, err
+	}
+	mppLabels := []*metapb.StoreLabel{
+		{
+			Key:   "engine",
+			Value: "tiflash_mpp",
```
you06, line 1754: Better to use constants, I find some [similar usage](https://github.com/tikv/client-go/blob/79f010d806110cb4bccf9218ceae5e5db54951fe/tikvrpc/endpoint.go#L62-L67) here.

you06, line 1754: Better to use some constants, I find some [similar usage](https://github.com/tikv/client-go/blob/79f010d806110cb4bccf9218ceae5e5db54951fe/tikvrpc/endpoint.go#L62-L67) here.

internal/locate/region_cache.go
```diff
@@ -389,6 +394,8 @@ func NewRegionCache(pdClient pd.Client) *RegionCache {
 	c.mu.latestVersions = make(map[uint64]RegionVerID)
 	c.mu.sorted = btree.New(btreeDegree)
 	c.storeMu.stores = make(map[uint64]*Store)
+	c.tiflashMPPStoreMu.needReload = true
+	c.tiflashMPPStoreMu.stores = make([]*Store, 1)
```
you06, line 398: It'll be reloaded when reading for the first time, why initiate it with a non-zero-length array?

you06, line 398: It'll be reloaded when reading for the first time, why initiate it with a non-zero-length array?

internal/locate/region_cache.go
```diff
@@ -1723,6 +1730,59 @@ func (c *RegionCache) GetTiFlashStores() []*Store {
...
 }
 
+// GetTiFlashMPPStores returns all stores with lable <engine, tiflash_mpp>.
+func (c *RegionCache) GetTiFlashMPPStores(bo *retry.Backoffer) (res []*Store, err error) {
+	c.tiflashMPPStoreMu.RLock()
+	needReload := c.tiflashMPPStoreMu.needReload
+	stores := c.tiflashMPPStoreMu.stores
+	c.tiflashMPPStoreMu.RUnlock()
+
+	if needReload {
```
you06, line 1740: So in a bad case, there may be many threads calling this function and turned to reload, however one reload task is enough.

you06, line 1740: So in a bad case, there may be many threads calling this function and turned to reload, however one reload task is enough.

internal/locate/region_cache.go
guo-shaoge: Add `isStoreContainLabel`.

internal/locate/region_cache.go
guo-shaoge: Done

internal/locate/region_cache.go
guo-shaoge: done

internal/locate/region_cache.go
guo-shaoge: Yes. But looks like `RegionCache` is the same behavior?

you06: Ok, this looks good to me.

you06:
What about adding [counter and duration metrics](https://github.com/tikv/client-go/blob/5d3195ce1174d3fd918c00281e77fdae078db81c/metrics/shortcuts.go#L193-L208) for the cache invalidation and loading.

guo-shaoge:
> What about adding [counter and duration metrics](https://github.com/tikv/client-go/blob/5d3195ce1174d3fd918c00281e77fdae078db81c/metrics/shortcuts.go#L193-L208) for the cache invalidation and loading.

To keep this pr clean, I will update metrics in the next pr(https://github.com/tikv/client-go/issues/509)~

you06 (approved):

XuHuaiyu (approved):

sticnarf (approved):

PR #482, opened 2022-04-28
internal/unionstore: avoid panic and add log for tidb issue/34307

tiancaiamao:


tiancaiamao:
PTAL @disksing

disksing (approved):

PR #485, opened 2022-05-06
Add backoff for error recovery in progress

Connor1996:
`RecoveryInProgress` error occurs when a region is being recovered online, during which any write and read is forbidden.

Part of https://github.com/tikv/tikv/issues/10483


sticnarf:
How long does a region in the recovery progress last for? I want to confirm this because if the typical recovery time exceeds the backoff timeout, then the backoff for this error may be unnecessary.

Connor1996:
> How long does a region in the recovery progress last for? I want to confirm this because if the typical recovery time exceeds the backoff timeout, then the backoff for this error may be unnecessary.

Should be within several seconds at most. Though if there is something wrong with the recovery progress(e.g. the network is isolated), the region may be in the recovery state for a long time.

sticnarf:
> Should be within several seconds at most. Though if there is something wrong with the recovery progress(e.g. the network is isolated), the region may be in the recovery state for a long time.

OK. It makes sense now.

What if the region size increases to 10GB? How would that affect the recovery time of a region?

Connor1996:
> > Should be within several seconds at most. Though if there is something wrong with the recovery progress(e.g. the network is isolated), the region may be in the recovery state for a long time.
> 
> OK. It makes sense now.
> 
> What if the region size increases to 10GB? How would that affect the recovery time of a region?

Doesn't make any difference. It doesn't related to the data size, but num of store.

internal/locate/region_request.go
```diff
@@ -1395,6 +1397,16 @@ func (s *RegionRequestSender) onRegionError(bo *retry.Backoffer, ctx *RPCContext
...
 	}
 
+	if regionErr.GetRecoveryInProgress() != nil {
+		s.regionCache.InvalidateCachedRegion(ctx.Region)
+		logutil.BgLogger().Debug("tikv reports `RecoveryInProgress`", zap.Stringer("ctx", ctx))
+		err = bo.Backoff(retry.BoRegionRecoveryInProgress, errors.Errorf("region recovery in progress, ctx: %v", ctx))
+		if err != nil {
+			return false, err
+		}
+		return true, nil
```
sticnarf, line 1407: It is a bit strange to both invalidate region cache and return `retry = true`.

If retry is true, in the next retry, it will still use the same `replicaSelector` with the region info in the last time.

So, it depends how the region will change after it get recovered.

If the region epoch or the region range does not change, maybe it's not necessary to invalidate the cache, because a `NotLeader` error will also update the leader.

If the epoch will change, here `retry` should be false. The request will be retried from more upper.

Connor1996: The reason to invalidate the cache is that the region epoch in the client cache may be bigger than the recovered region due to lost-updates suffering from majority failure.

sticnarf: If you hope that the region cache should refresh in the retry, please return `false, nil` here.

sticnarf (approved):
LGTM

Connor1996:
The tidb-compatibility test failure is irrelevant, merge it directly

PR #486, opened 2022-05-07
region_cache: in some cases, LocateBucket should not return nil.

SpadeA-Tang:
When  key is in [region startkey, first bucket key) or [last bucket key, region endkey), it's better not to return nil, as they can be considered to be `Bucket`.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -755,6 +755,40 @@ func (l *KeyLocation) GetBucketVersion() uint64 {
 	return l.Buckets.GetVersion()
 }
 
+// LocateBucketV2 will not return nil if the key is in the region.
+// LocateBucketV2 is similar with LocateBucket. The difference is that when the key is in [KeyLocation.StartKey, first Bucket key)
+// it will return Bucket{KeyLocation.StartKey, first Bucket key} rather than nil --- it's reasonable to assume that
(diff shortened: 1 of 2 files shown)


youjiali1995 (approved):

internal/locate/region_cache.go
```diff
@@ -755,6 +755,40 @@ func (l *KeyLocation) GetBucketVersion() uint64 {
 	return l.Buckets.GetVersion()
 }
 
+// LocateBucketV2 will not return nil if the key is in the region.
+// LocateBucketV2 is similar with LocateBucket. The difference is that when the key is in [KeyLocation.StartKey, first Bucket key)
+// it will return Bucket{KeyLocation.StartKey, first Bucket key} rather than nil --- it's reasonable to assume that
+// Bucket{KeyLocation.StartKey, first Bucket key} is a bucket belonging to the region. Key in [last Bucket key, KeyLocation.EndKey)
+// is handled similarly.
+func (l *KeyLocation) LocateBucketV2(key []byte) *Bucket {
```
cfzjywxk, line 763: What does `V2` suppose to mean? Does that mean there would be two versions of the `LocateBucket` function and they would be used in different situations? If so it's better to change the name to some meaningful name but not use `V2`.

cfzjywxk: /cc @SpadeA-Tang

PR #490, opened 2022-05-09
unionstore: add checkpoint API to support TiDB savepoint feature

crazycs520:


you06 (approved):

internal/unionstore/memdb.go
```diff
@@ -155,6 +155,20 @@ func (db *MemDB) Cleanup(h int) {
 	db.stages = db.stages[:h-1]
 }
 
+type MemCheckpoint = memdbCheckpoint
```
bb7133, line 158: How about changing `memdbCheckpoint` to public?

bb7133:
Please sign the DCO.

internal/unionstore/memdb.go
crazycs520: ok, done.

PR #492, opened 2022-05-10
fix the vulnerability of atomicity in async commit

ekexium:
To fix https://github.com/pingcap/tidb/issues/33641

- For a pessimistic async commit txn, set its TTL to be at least (max commit ts - start ts)


ekexium:
@sticnarf @MyonKeminta PTAL

sticnarf (commented):
Could you add a unit test?

sticnarf (approved):

PR #494, opened 2022-05-12
Introduce API v2

iosmanthus:

Introducing API V2 for client-go, all the interfaces of RawKV have been upgraded, and their tests are migrated to detect the API version of the cluster.

- [ ] Migrate RawKV tests to auto-detect API version.


kv/api_version.go
```diff
@@ -0,0 +1,57 @@
...
+		return ApiV2TxnEndKey
+	}
+	panic("unreachable")
+}
+
+func BuildV2RequestKey(mode Mode, key []byte) []byte {
+	return append(GetV2Prefix(mode), key...)
+}
+
+func DecodeV2StartKey(mode Mode, key []byte) []byte {
```
pingyu, line 43: Suggest to combine the following methods as `DecodeV2KeyRange`. Then it would be easier to be used as `startKey, endKey := DecodeV2KeyRange(startKey, endKey)`.

rawkv/rawkv.go
```diff
@@ -176,7 +198,7 @@ func (c *Client) Get(ctx context.Context, key []byte, options ...RawOption) ([]b
 	req := tikvrpc.NewRequest(
 		tikvrpc.CmdRawGet,
 		&kvrpcpb.RawGetRequest{
-			Key: key,
+			Key: c.buildRequestKey(key),
```
pingyu, line 201: Then we will need to encode keys twice (one for request PD & another for request TiKV) ? How about do the encoding before here. And then it also seems that we don't need the `CodecPDClientV2`.

iosmanthus: Yes, we will encode keys twice. I keep the encoding happening only in the request sending phase because I don't want to change the existing logic of the interface like `Scan`/`RawScan`. The current commit doesn't delay this encoding late enough, in the next one, I am trying to encode the key and set the API version in `kvrpcpb.Context` only in package `tikvrpc`. This might keep the encoding transparent from the upper code, reducing the appearance of bugs.

iosmanthus: Besides, the API version is hard to pass to some request constructors like in struct `Scanner`. Keeping this information only in `RegionRequestSender` might save us from the code review hell.

internal/client/api_version.go
```diff
@@ -0,0 +1,257 @@
...
+
+type Mode int
+
+const (
+	ModeRaw = iota
+	ModeTxn
+)
+
+var (
+	ApiV2RawKeyPrefix = []byte("r")
```
sunxiaoguang, line 19: We need to add prefix reserved for keyspace now.

internal/locate/region_cache.go
```diff
@@ -483,6 +484,15 @@ func (c *RegionCache) SetPDClient(client pd.Client) {
 	c.pdClient = client
 }
 
+func (c *RegionCache) GetApiVersion() kvrpcpb.APIVersion {
```
sunxiaoguang, line 487: why not store api version when creating RegionCache?

rawkv/rawkv.go
```diff
@@ -756,7 +777,7 @@ func (c *Client) sendBatchPut(bo *retry.Backoffer, keys, values [][]byte, ttls [
 	var batches []kvrpc.Batch
 	// split the keys by size and RegionVerID
 	for regionID, groupKeys := range groups {
-		batches = kvrpc.AppendBatches(batches, regionID, groupKeys, keyToValue, keyTottl, rawBatchPutSize)
+		batches = kvrpc.AppendBatches(batches, regionID, groupKeys, keyToValue, keyToTTL, rawBatchPutSize)
```
sunxiaoguang, line 780: We can combine keyToValue and keyToTTL to a single map to save memory and lookup

internal/locate/region_cache.go
iosmanthus: `ApiVersion` should be removed from  `RegionCache`, and moved to `RegionRequestSender`. The API version could be derived while constructing `RegionRequestSender`.

internal/client/client.go
```diff
@@ -121,21 +121,22 @@ type connArray struct {
 	done chan struct{}
 }
 
-func newConnArray(maxSize uint, addr string, security config.Security, idleNotify *uint32, enableBatch bool, dialTimeout time.Duration) (*connArray, error) {
+func newConnArray(maxSize uint, addr string, security config.Security,
+	idleNotify *uint32, enableBatch bool, dialTimeout time.Duration, opts ...grpc.DialOption) (*connArray, error) {
```
pingyu, line 125: Suggest to use `opts []grpc.DialOption` instead, in case we will have more argument later.

rawkv/rawkv.go
```diff
@@ -131,20 +166,32 @@ func (c *Client) SetColumnFamily(columnFamily string) *Client {
 }
 
 // NewClient creates a client with PD cluster addrs.
-func NewClient(ctx context.Context, pdAddrs []string, security config.Security, opts ...pd.ClientOption) (*Client, error) {
+func NewClient(ctx context.Context, pdAddrs []string, opts ...ClientOpt) (*Client, error) {
```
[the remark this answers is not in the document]
iosmanthus: How about creating a new function called `NewClientWithOpts`

pingyu (commented):
The `NewClientWithOpts` LGTM~

tikv/region.go
```diff
@@ -91,6 +91,17 @@ type RPCRuntimeStats = locate.RPCRuntimeStats
 // CodecPDClient wraps a PD Client to decode the encoded keys in region meta.
 type CodecPDClient = locate.CodecPDClient
 
+type CodecPDClientV2 = locate.CodecPDClientV2
```
disksing, line 94: need some comment.

internal/locate/region_request.go
```diff
@@ -1414,7 +1425,11 @@ func (s *RegionRequestSender) onRegionError(bo *retry.Backoffer, ctx *RPCContext
 	}
 
 	if regionErr.GetKeyNotInRegion() != nil {
-		logutil.BgLogger().Debug("tikv reports `KeyNotInRegion`", zap.Stringer("ctx", ctx))
+		var key []byte
+		if req.Type == tikvrpc.CmdRawPut {
+			key = req.RawPut().Key
+		}
```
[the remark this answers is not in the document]
iosmanthus: Yes. It should be removed.

internal/client/api_version.go
```diff
@@ -0,0 +1,190 @@
...
+	case CmdRawCompareAndSwap:
+		r := *req.RawCompareAndSwap()
+		r.Key = EncodeV2Key(ModeRaw, r.Key)
+		newReq.Req = &r
+	}
+
+	return &newReq, nil
+}
+
+func DecodeKey(mode Mode, key []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
iosmanthus: Make sense

tikvrpc/tikvrpc.go
```diff
@@ -699,12 +700,12 @@ func SetContext(req *Request, region *metapb.Region, peer *metapb.Peer) error {
...
 		req.RawScan().Context = ctx
-	case CmdUnsafeDestroyRange:
-		req.UnsafeDestroyRange().Context = ctx
 	case CmdGetKeyTTL:
 		req.RawGetKeyTTL().Context = ctx
 	case CmdRawCompareAndSwap:
 		req.RawCompareAndSwap().Context = ctx
+	case CmdUnsafeDestroyRange:
+		req.UnsafeDestroyRange().Context = ctx
 	case CmdRegisterLockObserver:
```
haojinming, line 709: As `checksum` has supported in master branch, maybe apiv2 also need support `checksum` cmd here?

internal/locate/region_cache.go
```diff
@@ -1666,12 +1666,18 @@ func (c *RegionCache) OnRegionEpochNotMatch(bo *retry.Backoffer, ctx *RPCContext
 	newRegions := make([]*Region, 0, len(currentRegions))
 	// If the region epoch is not ahead of TiKV's, replace region meta in region cache.
 	for _, meta := range currentRegions {
-		if _, ok := c.pdClient.(*CodecPDClient); ok {
-			var err error
+		var err error
+		oldMeta := meta
+		switch c.pdClient.(type) {
```
sunxiaoguang, line 1671: Why not move `decodeRegionMetaKeyWithShallowCopy` into interface and use interface to dispatch.

internal/locate/region_request.go
```diff
@@ -189,8 +190,18 @@ func RecordRegionRequestRuntimeStats(stats map[tikvrpc.CmdType]*RPCRuntimeStats,
 
 // NewRegionRequestSender creates a new sender.
 func NewRegionRequestSender(regionCache *RegionCache, client client.Client) *RegionRequestSender {
+	var v kvrpcpb.APIVersion
+
+	switch regionCache.pdClient.(type) {
```
sunxiaoguang, line 195: Type assertion looks verbose, we can cache the version somewhere in RegionCache or simply have a method in client to return API version.

internal/locate/region_cache.go
iosmanthus: Because c.pdClient is a `pd.Client` interface already, if we change the interface, we need to modify the PD.

internal/locate/region_request.go
iosmanthus: We could store the API version in the region cache.

iosmanthus:
@sticnarf @sunxiaoguang @disksing PTAL

sunxiaoguang (approved):
LGTM

sunxiaoguang:
Compatibility test will be fixed in a separate PR.

PR #495, opened 2022-05-13
metric: Add memdb size and key count metric

longfangsong:
Add size and keys count metric for memdb


internal/unionstore/memdb.go
```diff
@@ -358,6 +361,7 @@ func (db *MemDB) setValue(x memdbNodeAddr, value []byte) {
 	}
 	x.vptr = db.vlog.appendValue(x.addr, x.vptr, value)
 	db.size = db.size - len(oldVal) + len(value)
+	metrics.MemdbSize.Set(float64(db.size))
```
cfzjywxk, line 364: Should `Add` be used for the `MemdbSize` as it's a server metric?

ekexium:
There could be multiple `MemDB`s. I think Gauge is inappropriate.

ekexium:
I'm wondering about the purpose of these metrics. It answers "how much data is in memdb" but cannot tell us a lot about "how much memory is used by memdb"

ekexium:
We don't have a clear idea of the relationship between `db.size` and the memory allocated. I'm thinking could we export both of them so that we will be able to estimate the memory usage of certain kinds of jobs, e.g.`insert into select` a large table?

internal/unionstore/memdb.go
```diff
@@ -563,6 +567,8 @@ func (db *MemDB) deleteNode(z memdbNodeAddr) {
 
 	db.count--
 	db.size -= int(z.klen)
+	metrics.MemdbCount.Dec()
```
disksing, line 570: I prefer `Set(db.count)`

internal/unionstore/memdb.go
cfzjywxk: @longfangsong 
PTAL

internal/unionstore/memdb.go
```diff
@@ -163,6 +164,8 @@ func (db *MemDB) Reset() {
 	db.vlogInvalid = false
 	db.size = 0
 	db.count = 0
+	metrics.MemdbSize.Set(0.0)
+	metrics.MemdbCount.Set(0.0)
```
MyonKeminta, line 168: Sorry I didn't understand it well. The metrics looks like global gauges. May there be many MemDB instances in a single process and they all `Set` values to the single gauge by themselves?

cfzjywxk: I've mentioned it above and discussed it with @longfangsong , it's needed to be changed I think.

ekexium:
#520 tracks the memory allocated to the arena. That could be reused here.

cfzjywxk:
@longfangsong 
Please address the above comments.

disksing:
@cfzjywxk @MyonKeminta do we want to continue this?

disksing:
closing due to inactivity. Please reopen it if need.

PR #502, opened 2022-05-16
metrics: add loading region cache duration

you06:
Signed-off-by: you06 <you1474600@gmail.com>

This PR record the latency when loading the region cache.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1365,6 +1365,7 @@ func (c *RegionCache) loadRegion(bo *retry.Backoffer, key []byte, isEndKey bool)
 				return nil, errors.WithStack(err)
 			}
 		}
+		start := time.Now()
 		var reg *pd.Region
 		var err error
 		if searchPrev {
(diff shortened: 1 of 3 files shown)


sticnarf (approved):

internal/locate/region_cache.go
```diff
@@ -1365,17 +1365,19 @@ func (c *RegionCache) loadRegion(bo *retry.Backoffer, key []byte, isEndKey bool)
...
 		}
+		start := time.Now()
 		var reg *pd.Region
 		var err error
 		if searchPrev {
 			reg, err = c.pdClient.GetPrevRegion(ctx, key, pd.WithBuckets())
 		} else {
 			reg, err = c.pdClient.GetRegion(ctx, key, pd.WithBuckets())
 		}
+		metrics.LoadRegionCacheHistogramWhenCacheMiss.Observe(time.Since(start).Seconds())
```
TonsnakeLin, line 1376: LoadRegionCacheHistogramWhenCacheMiss observes the time for every round of getting region cache from PD. It is meaningful.
Do we need observe the total backoff time of getting region cache? The total time may be more clearly for users.

internal/locate/region_cache.go
```diff
@@ -1416,7 +1418,9 @@ func (c *RegionCache) loadRegionByID(bo *retry.Backoffer, regionID uint64) (*Reg
 				return nil, errors.WithStack(err)
 			}
 		}
+		start := time.Now()
 		reg, err := c.pdClient.GetRegionByID(ctx, regionID, pd.WithBuckets())
+		metrics.LoadRegionCacheHistogramWithRegionByID.Observe(time.Since(start).Seconds())
```
TonsnakeLin, line 1423: Ditto

internal/locate/region_cache.go
you06: Users may have interest in the total loading region cache time, and I think such time can be recorded in execution details and printed in slow log.

Because loading region cache can be concurrent, the total time does not stand for the wall time. I don't think it's a clear metric.

PR #503, opened 2022-05-17
Avoid sending requests to unhealthy leader

sticnarf:
It's an improvement for recovery time when hibernate region is enabled.

When we detect a TiKV store is unreachable, we only clear the region cache of regions on it. But it's not enough.

Considering one TiKV is unhealthy, all requests sending to that store will timeout. But due to hibernate region, the regions whose leaders are on the unhealthy store will not be re-elected until a request is sent to the follower.

If there are thousands of such regions, every time we are accessing a new region, it will send a request to the original leader on the unhealthy TiKV because the leader definitely does not change. And then, the user will always experience very long request duration until all regions are touched and elect new leaders.

So, this PR simply prevents sending requests to the leader that are on unhealthy TiKVs. To achieve this, the check loop is always enabled instead of only when forwarding is enabled.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -322,7 +322,16 @@ type accessKnownLeader struct {
 
 func (state *accessKnownLeader) next(bo *retry.Backoffer, selector *replicaSelector) (*RPCContext, error) {
 	leader := selector.replicas[state.leaderIdx]
-	if leader.isExhausted(maxReplicaAttempt) {
+	liveness := leader.store.getLivenessState()
+	if liveness == unreachable && selector.regionCache.enableForwarding {
+		selector.state = &tryNewProxy{leaderIdx: state.leaderIdx}
+		return nil, stateChanged{}
+	}
+	// If hibernate region is enabled and the leader is not reachable, the raft group
+	// will not be wakened up and re-elect the leader until the follower receives
+	// a request. So, before the new leader is elected, we should not send requests
+	// to the unreachable old leader to avoid unnecessary timeout.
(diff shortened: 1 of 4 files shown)


internal/locate/region_request.go
```diff
@@ -322,7 +322,19 @@ type accessKnownLeader struct {
 
 func (state *accessKnownLeader) next(bo *retry.Backoffer, selector *replicaSelector) (*RPCContext, error) {
 	leader := selector.replicas[state.leaderIdx]
-	if leader.isExhausted(maxReplicaAttempt) {
+	liveness := reachable
+	if atomic.LoadInt32(&leader.store.unreachable) != 0 {
+		liveness = selector.checkLiveness(bo, leader)
```
youjiali1995, line 327: Seems it's not necessary. `store.unreachable` means the liveness is `unreachable`.

internal/locate/region_request.go
```diff
@@ -322,7 +322,19 @@ type accessKnownLeader struct {
...
+	}
+	if liveness == unreachable {
+		selector.state = &tryNewProxy{leaderIdx: state.leaderIdx}
```
youjiali1995, line 330: Why `tryNewProxy` here? I think it should try followers.

sticnarf: Sorry, I missed the forwarding condition. The corresponding test is also fixed.

sticnarf:
@youjiali1995 @MyonKeminta PTAL

MyonKeminta (approved):

youjiali1995 (approved):

Issue #518, opened 2022-06-02
RawKV checksum is not supported. 


haojinming:
Checksum is not supported in `rawkv.Client`.

haojinming:
Close this issue as checksum is supported in master.

PR #519, opened 2022-06-02
support checksum

haojinming:
Signed-off-by: haojinming <jinming.hao@pingcap.com>
Support `checksum` in rawkv.Client.
Fix issue #518

### DIFF
--- a/rawkv/rawkv.go
+++ b/rawkv/rawkv.go
@@ -504,6 +504,47 @@ func (c *Client) ReverseScan(ctx context.Context, startKey, endKey []byte, limit
 	return
 }
 
+// Checksum do checksum of continuous kv pairs in range [startKey, endKey).
+// If endKey is empty, it means unbounded.
+// If you want to exclude the startKey or include the endKey, push a '\0' to the key. For example, to scan
+// (startKey, endKey], you can write:
+// `Scan(ctx, push(startKey, '\0'), push(endKey, '\0'), limit)`.
+func (c *Client) Checksum(ctx context.Context, startKey, endKey []byte, options ...RawOption,
+) (crc64Xor, totalKvs, totalBytes uint64, err error) {
+
+	start := time.Now()
+	defer func() { metrics.RawkvCmdHistogramWithRawChecksum.Observe(time.Since(start).Seconds()) }()
+
+	crc64Xor = 0
+	totalKvs = 0
+	totalBytes = 0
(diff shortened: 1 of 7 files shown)


rawkv/rawkv.go
```diff
@@ -504,6 +504,47 @@ func (c *Client) ReverseScan(ctx context.Context, startKey, endKey []byte, limit
 	return
 }
 
+// Checksum do checksum of continuous kv pairs in range [startKey, endKey).
+// If endKey is empty, it means unbounded.
+// If you want to exclude the startKey or include the endKey, push a '\0' to the key. For example, to scan
+// (startKey, endKey], you can write:
+// `Scan(ctx, push(startKey, '\0'), push(endKey, '\0'), limit)`.
+func (c *Client) Checksum(ctx context.Context, startKey, endKey []byte, options ...RawOption,
+) (crc64Xor, totalKvs, totalBytes uint64, err error) {
```
iosmanthus, line 513: Prefer to use a struct to wrap all these infos.

rawkv/rawkv.go
```diff
@@ -504,6 +504,47 @@ func (c *Client) ReverseScan(ctx context.Context, startKey, endKey []byte, limit
 	return
 }
 
+// Checksum do checksum of continuous kv pairs in range [startKey, endKey).
+// If endKey is empty, it means unbounded.
+// If you want to exclude the startKey or include the endKey, push a '\0' to the key. For example, to scan
+// (startKey, endKey], you can write:
+// `Scan(ctx, push(startKey, '\0'), push(endKey, '\0'), limit)`.
```
iosmanthus, line 511: Seem it's the comments from `ReverseScan`.

haojinming: yes, they accept the same inpu parameters.

iosmanthus:
```suggestion
// Checksum do checksum of continuous kv pairs in range [startKey, endKey).
// If endKey is empty, it means unbounded.
// If you want to exclude the startKey or include the endKey, push a '\0' to the key. For example, to checksum
// (startKey, endKey], you can write:
// `Checksum(ctx, push(startKey, '\0'), push(endKey, '\0'), limit)`.
```

rawkv/rawkv.go
haojinming: Done, PTAL~

disksing (approved):

PR #527, opened 2022-06-15
resolve-lock: try resolve legacy locks

joccau:
Signed-off-by: joccau <zak.zhao@pingcap.com>

Close https://github.com/tikv/tikv/issues/12582

- [ ] Uint test
- [x] Integration test
- [x] Manual test

### Manual test
- begin a transaction
- update a big transaction about 540000 line records.
- during committing this transaction, kill the `tidb-server`

image

image

as above figure, after kill `tidb-server`, the legacy locks keep the checkpoint-ts in `backup-log` from advancing for a while. After `tidb-server` resolves these legacy locks, the checkpoint-ts advanced normally.


txnkv/txnlock/lock_resolver.go
```diff
@@ -219,9 +219,45 @@ func (lr *LockResolver) getResolved(txnID uint64) (TxnStatus, bool) {
...
+	if len(locks) == 0 {
+		return true, nil
+	}
+
+	logutil.BgLogger().Info("TryBatchResolveLocks", zap.Uint64("safepoint", safepoint), zap.Uint64("target-resolve-ts", targetResolveTS))
+
+	locksBeforeSafePoint := make([]*Lock, 0, len(locks))
+	locksAfterSafePoint := make([]*Lock, 0, len(locks))
+	for _, l := range locks {
+		if l.TxnID < safepoint {
```
MyonKeminta, line 239:
```suggestion
		if l.TxnID <= safepoint {
```

To be consistent with the current implementation of scan_lock in tikv, here it's inclusive. It shouldn't matter much though.

sticnarf:
TiDB compatibility test is allowed to fail. It only reminds that there will be additional changes due to API changes when updating client-go in TiDB.

txnkv/txnlock/lock_resolver.go
```diff
@@ -220,9 +220,55 @@ func (lr *LockResolver) getResolved(txnID uint64) (TxnStatus, bool) {
...
+// 1. If the tts of lock is equal with or smaller than safepoint, it will
+// rollback the txn, no matter the lock is expired of not.
+// 2. If the tts of lock is larger than safepoint, it will check status of the txn.
+// Resolve the lock if txn is expired, Or do nothing.
+func (lr *LockResolver) BatchResolveLegacyLocks(
+	bo *retry.Backoffer,
+	locks []*Lock,
+	loc locate.RegionVerID,
+	safepoint uint64,
+	lowResolveTS uint64,
```
3pointer, line 234: maybe unified the name: lowResolveTS -> lowResolutionTS

3pointer (approved):

txnkv/txnlock/lock_resolver.go
```diff
@@ -220,9 +220,55 @@ func (lr *LockResolver) getResolved(txnID uint64) (TxnStatus, bool) {
...
+		}
+	}
+
+	logutil.BgLogger().Info("BatchResolveLegacyLocks",
+		zap.Uint64("safepoint", safepoint),
+		zap.Uint64("low-resolution-ts", lowResolutionTS),
+		zap.Int("before-safepoint-count", len(locksBeforeSafePoint)),
+		zap.Int("after-safepoint-count", len(locksAfterSafePoint)))
+
+	// Use math.MaxUint64 as usedCheckTTL, which means rollback the txn, no matter the lock is expired or not!
```
sticnarf, line 256: Is `usedCheckTTL` an old name that you forgot to change?

txnkv/txnlock/lock_resolver.go
```diff
@@ -266,11 +311,13 @@ func (lr *LockResolver) BatchResolveLocks(bo *retry.Backoffer, locks []*Lock, lo
 		}
 
 		if status.ttl > 0 {
-			logutil.BgLogger().Error("BatchResolveLocks fail to clean locks, this result is not expected!")
-			return false, errors.New("TiDB ask TiKV to rollback locks but it doesn't, the protocol maybe wrong")
+			if checkTxnExpired == math.MaxUint64 {
+				logutil.BgLogger().Error("BatchResolveLocks fail to clean locks, this result is not expected!")
+				return false, errors.New("TiDB ask TiKV to rollback locks but it doesn't, the protocol maybe wrong")
+			}
+		} else {
```
sticnarf, line 318: If some locks are not expired and not resolved, do you still consider `BatchResolveLocks` as success?

txnkv/txnlock/lock_resolver.go
joccau: This field `usedCheckTTL` is used for check whether is the lock with TTL expired.
Maybe a better name, what about you ?

txnkv/txnlock/lock_resolver.go
joccau: Yes, it justly resolves the expired lock and some lock, ts of which is smaller than safepoint.

sticnarf:
Well, my biggest question is that all the changes in this PR seem not really necessary for **client-go** to me...

From the caller, you can scan locks before `now - 5min` and choose the locks before safepoint. And then, use `BatchResolveLocks` to handle the locks before safepoint and use normal `ResolveLocks` for the rest.

Then, the orignal logic in client-go does not need to be touched.

PR #528, opened 2022-06-16
[DNM] Support not-atomic pessimistic lock request and lock with conflict

MyonKeminta:
Requires: https://github.com/tikv/tikv/pull/12749

This PR adapts to the new behavior of acquire_pessimistic_lock requests in the PR above.
This PR is experimental and still under development.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -238,6 +238,16 @@ func (m *memBufferMutations) Slice(from, to int) CommitterMutations {
 	}
 }
 
+func (m *memBufferMutations) GetMutation(i int) PlainMutation {
+	mutation := PlainMutation{
+		KeyOp: m.GetOp(i),
+		Key:   m.GetKey(i),
+		Value: m.GetValue(i),
+		Flags: makeMutationFlags(m.IsPessimisticLock(i), m.IsAssertExists(i), m.IsAssertNotExist(i)),
+	}
+	return mutation
+}
+
 func (m *memBufferMutations) Push(op kvrpcpb.Op, isPessimisticLock, assertExist, assertNotExist bool, handle unionstore.MemKeyHandle) {
 	// See comments of `m.handles` field about the format of the user data `aux`.
 	aux := uint16(op) << 3
@@ -293,6 +303,7 @@ type CommitterMutations interface {
 	Slice(from, to int) CommitterMutations
 	IsAssertExists(i int) bool
 	IsAssertNotExist(i int) bool
+	GetMutation(i int) PlainMutation
 }
 
 // PlainMutations contains transaction operations.
@@ -417,6 +428,17 @@ func (c *PlainMutations) AppendMutation(mutation PlainMutation) {
 	c.keys = append(c.keys, mutation.Key)
 	c.values = append(c.values, mutation.Value)
 	c.flags = append(c.flags, mutation.Flags)
+}
+
+// GetMutation returns the mutation at the specified index.
+func (c *PlainMutations) GetMutation(i int) PlainMutation {
+	mutation := PlainMutation{
+		KeyOp: c.ops[i],
+		Key:   c.keys[i],
+		Value: c.values[i],
+		Flags: c.flags[i],
+	}
+	return mutation
 }
 
 // newTwoPhaseCommitter creates a twoPhaseCommitter.
@@ -1854,9 +1876,10 @@ func (c *twoPhaseCommitter) shouldWriteBinlog() bool {
 }
 
 type batchMutations struct {
-	region    locate.RegionVerID
-	mutations CommitterMutations
-	isPrimary bool
+	region       locate.RegionVerID
+	mutations    CommitterMutations
+	isPrimary    bool
+	primaryIndex int
 }
 
 func (b *batchMutations) relocate(bo *retry.Backoffer, c *locate.RegionCache) (bool, error) {
@@ -1894,6 +1917,8 @@ func (b *batched) appendBatchMutationsBySize(region locate.RegionVerID, mutation
 
 	var start, end int
 	for start = 0; start < mutations.Len(); start = end {
+		isPrimary := false
+		primaryIndexInBatch := -1
 		var size int
 		for end = start; end < mutations.Len() && size < limit; end++ {
 			var k, v []byte
@@ -1902,11 +1927,15 @@ func (b *batched) appendBatchMutationsBySize(region locate.RegionVerID, mutation
 			size += sizeFn(k, v)
 			if b.primaryIdx < 0 && bytes.Equal(k, b.primaryKey) {
 				b.primaryIdx = len(b.batches)
+				isPrimary = true
+				primaryIndexInBatch = end - start
 			}
 		}
 		b.batches = append(b.batches, batchMutations{
-			region:    region,
-			mutations: mutations.Slice(start, end),
+			region:       region,
+			mutations:    mutations.Slice(start, end),
+			isPrimary:    isPrimary,
+			primaryIndex: primaryIndexInBatch,
 		})
 	}
 }
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -135,6 +135,8 @@ type twoPhaseCommitter struct {
 	primaryKey  []byte
 	forUpdateTS uint64
 
+	maxLockWithConflictTS uint64
+
 	mu struct {
 		sync.RWMutex
 		undeterminedErr error // undeterminedErr saves the rpc error we encounter when commit primary key.
--- a/txnkv/transaction/pessimistic.go
+++ b/txnkv/transaction/pessimistic.go
@@ -84,18 +84,23 @@ func (actionPessimisticRollback) tiKVTxnRegionsNumHistogram() prometheus.Observe
 }
 
 func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *retry.Backoffer, batch batchMutations) error {
-	m := batch.mutations
-	mutations := make([]*kvrpcpb.Mutation, m.Len())
-	for i := 0; i < m.Len(); i++ {
-		mut := &kvrpcpb.Mutation{
-			Op:  kvrpcpb.Op_PessimisticLock,
-			Key: m.GetKey(i),
+	convertMutationsToPb := func(committerMutations CommitterMutations) []*kvrpcpb.Mutation {
+		mutations := make([]*kvrpcpb.Mutation, committerMutations.Len())
+		for i := 0; i < committerMutations.Len(); i++ {
+			mut := &kvrpcpb.Mutation{
+				Op:  kvrpcpb.Op_PessimisticLock,
+				Key: committerMutations.GetKey(i),
+			}
+			if c.txn.us.HasPresumeKeyNotExists(committerMutations.GetKey(i)) || (c.doingAmend && committerMutations.GetOp(i) == kvrpcpb.Op_Insert) {
+				mut.Assertion = kvrpcpb.Assertion_NotExist
+			}
+			mutations[i] = mut
 		}
-		if c.txn.us.HasPresumeKeyNotExists(m.GetKey(i)) || (c.doingAmend && m.GetOp(i) == kvrpcpb.Op_Insert) {
-			mut.Assertion = kvrpcpb.Assertion_NotExist
-		}
-		mutations[i] = mut
+		return mutations
 	}
+
+	m := batch.mutations
+	mutations := convertMutationsToPb(m)
 	req := tikvrpc.NewRequest(tikvrpc.CmdPessimisticLock, &kvrpcpb.PessimisticLockRequest{
 		Mutations:      mutations,
 		PrimaryLock:    c.primary(),
@@ -106,6 +111,7 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
 		ReturnValues:   action.ReturnValues,
 		CheckExistence: action.CheckExistence,
 		MinCommitTs:    c.forUpdateTS + 1,
+		WaitLockMode:   action.LockWaitMode,
 	}, kvrpcpb.Context{Priority: c.priority, SyncLog: c.syncLog, ResourceGroupTag: action.LockCtx.ResourceGroupTag,
 		MaxExecutionDurationMs: uint64(client.MaxWriteExecutionTime.Milliseconds())})
 	if action.LockCtx.ResourceGroupTag == nil && action.LockCtx.ResourceGroupTagger != nil {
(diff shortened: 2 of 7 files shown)


txnkv/transaction/pessimistic.go
```diff
@@ -147,152 +165,384 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
+	failedMutations := NewPlainMutations(0)
+	keyErrs := lockResp.GetErrors()
+
+	if batch.isPrimary && len(lockResp.Results) > 0 && lockResp.Results[batch.primaryIndex].Type != kvrpcpb.PessimisticLockKeyResultType_Failed {
+		// After locking the primary key, we should protect the primary lock from expiring
+		// now in case locking the remaining keys take a long time.
+		c.run(c, action.LockCtx)
+	}
+
+	for i, res := range lockResp.Results {
```
cfzjywxk, line 374: As we've limited the use case to be a single key lock, seems these extra checks could be ignored by now?

txnkv/transaction/pessimistic.go
```diff
@@ -147,152 +165,384 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
+			finished, err := action.handlePessimisticLockResponseRetryFirstMode(c, bo, &batch, mutations, resp, &diagCtx)
+			if err != nil {
+				return err
+			}
+			if finished {
+				return nil
 			}
-			same, err := batch.relocate(bo, c.store.GetRegionCache())
+		} else if action.lockWaitMode == kvrpcpb.PessimisticWaitLockMode_LockFirst {
+			finished, retryMutations, err := action.handlePessimisticLockResponseLockFirstMode(c, bo, &batch, mutations, resp, &diagCtx)
```
cfzjywxk, line 187: Maybe we could record how many pessimistic RPC requests are sent before finishing, and also how many resolving operations are triggered.
If the whole processing takes several seconds and tikv keeps telling the kv-client it needs not to do the resolving, does it mean that this request is always at the back of the waiting queue which could be unexpected?

txnkv/transaction/pessimistic.go
MyonKeminta: Yes.

txnkv/transaction/pessimistic.go
MyonKeminta: Yes. But since TiKV uses a priority queue based on the transaction's start_ts, I think it's impossible.

MyonKeminta:
The tests look not sufficient to me now, but this PR is large enough. Considering that it's not expected to affect the original logic, I think I can add more tests in further PRs.

txnkv/transaction/txn.go
```diff
@@ -656,10 +861,11 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 				return txn.committer.extractKeyExistsErr(e)
 			}
 		}
+		keyStr := string(key)
```
cfzjywxk, line 864: Would this introduce allocation in the recent golang version? Seems it still needs to https://github.com/golang/go/issues/25484?

txnkv/transaction/txn.go
```diff
@@ -704,19 +910,68 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
 			assignedPrimaryKey = true
+			txn.selectPrimaryForPessimisticLock(keys)
+		}
+
+		txn.committer.forUpdateTS = lockCtx.ForUpdateTS
+
+		allKeys := keys
+		// If aggressive locking is enabled and we don't need to update the primary for all locks, we can avoid sending
+		// RPC to those already locked keys.
+		if txn.aggressiveLockingContext != nil {
```
cfzjywxk, line 922: How about wrapping this part into a function like `filterAggresiveLockedKeys` since the `LockKeys` is already a little bit long?

txnkv/transaction/txn.go
```diff
@@ -704,19 +910,68 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
+			//   unlikely to be resolved by other transactions.
+
+			// In case primary is not assigned in this phase, or primary is set but unchanged, we don't need to update
+			// the locks.
+			canTrySkip := !txn.aggressiveLockingContext.assignedPrimaryKey || bytes.Equal(txn.aggressiveLockingContext.lastPrimaryKey, txn.aggressiveLockingContext.primaryKey)
+
+			// Do not preallocate since in most cases the keys need to lock doesn't change during pessimistic-retry.
+			keys = make([][]byte, 0)
+			for _, k := range allKeys {
+				keyStr := string(k)
```
cfzjywxk, line 938: Same question as above, would it introduce allocation?

txnkv/transaction/txn.go
```diff
@@ -791,30 +1063,93 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
+		txn.committer.primaryKey = sortedKeys[0]
+	}
+}
+
+type aggressiveLockingContext struct {
+	lastRetryUnnecessaryLocks map[string]tempLockBufferEntry
+	lastPrimaryKey            []byte
+	lastAttemptStartTime      time.Time
+
+	currentLockedKeys       map[string]tempLockBufferEntry
```
cfzjywxk, line 1146: How to track the `tempLockBufferEntry` related using the new memory statistic mechanism for a transaction? Is it necessary? 
@ekexium PTAL

txnkv/transaction/2pc.go
```diff
@@ -708,7 +741,10 @@ func (c *twoPhaseCommitter) initKeysAndMutations(ctx context.Context) error {
 
 func (c *twoPhaseCommitter) primary() []byte {
 	if len(c.primaryKey) == 0 {
-		return c.mutations.GetKey(0)
+		if c.mutations != nil {
+			return c.mutations.GetKey(0)
```
cfzjywxk, line 745: As the `selectPrimaryForPessimisticLock` is introduced, maybe adding a check here to assert pessimistic transactions MUST have a valid primary key to make it safer?

txnkv/transaction/pessimistic.go
```diff
@@ -150,156 +168,378 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
+			return true, nil, c.extractKeyExistsErr(e)
+		}
+		if deadlock := keyErr.Deadlock; deadlock != nil {
+			return true, nil, errors.WithStack(&tikverr.ErrDeadlock{Deadlock: deadlock})
+		}
+
+		// Extract lock from key error
+		lock, err1 := txnlock.ExtractLockFromKeyErr(keyErr)
+		if err1 != nil {
+			return true, nil, err1
```
cfzjywxk, line 441: After the key was forcibly locked in tikv, would it still be a `WriteConflictError` returned here to the upper layer to inform the statement must be retried?

txnkv/transaction/pessimistic.go
```diff
@@ -150,156 +168,378 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
+		return true, nil, errors.New("Pessimistic lock response corrupted")
+	}
+
+	if regionErr != nil || len(locks) != 0 {
+		return true, nil, errors.New("Pessimistic lock response corrupted")
+	}
+
+	// If the Results in response, there must be either some unretryable error in keyErrs or some region error, therefore
+	// the function must have returned.
+	if len(lockResp.Results) == 0 {
```
cfzjywxk, line 522: This comment is a little bit confusing, why the `len(lockResp.Results)` must be zero here?

txnkv/transaction/pessimistic.go
```diff
@@ -150,156 +168,378 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
+		}
+		if action.LockCtx.PessimisticLockWaited != nil {
+			atomic.StoreInt32(action.LockCtx.PessimisticLockWaited, 1)
+		}
+	}
+
+	return false, nil
+}
+
+func (action actionPessimisticLock) handlePessimisticLockResponseForceLockMode(c *twoPhaseCommitter, bo *retry.Backoffer, batch *batchMutations, mutationsPb []*kvrpcpb.Mutation, resp *tikvrpc.Response, diagCtx *diagnosticContext) (finished bool, retryMutations CommitterMutations, err error) {
```
cfzjywxk, line 354: If a key is locked forcibly, how does the caller know about it considering the returned results and retry the whole statement if necessary?

txnkv/transaction/txn.go
ekexium: Currently we only track the memory in memdb via callback, which should be the most significant part.
In the future we could consider introducing an individual tracker for each transaction which contains child trakcers for memdb, `aggressiveLockingContext` , snapshot cache, or anything necessary.

txnkv/transaction/txn.go
MyonKeminta: It's the simplest way, as it's impossible in golang to use slices as the builtin map's key. Here the purpose is to index the lockCtx.Value, which already exists before and uses string as the key too.

txnkv/transaction/pessimistic.go
MyonKeminta: Sorry the comment is mistakenly imcomplete. This should be an unreachable path. I'll update it.

txnkv/transaction/pessimistic.go
MyonKeminta: The LockedWithConflictTS information will be stored in the LockCtx like how we store returned values in it, and TiDB will check it.

internal/mockstore/mocktikv/mvcc_leveldb.go
```diff
@@ -532,11 +532,16 @@ type lockCtx struct {
 	ttl         uint64
 	minCommitTs uint64
 
-	returnValues     bool
-	checkExistence   bool
-	values           [][]byte
-	keyNotFound      []bool
+	returnValues   bool
+	checkExistence bool
+	//values         [][]byte
```
cfzjywxk, line 537: Need to remove?

txnkv/transaction/pessimistic.go
```diff
@@ -150,156 +168,375 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
+		}
+		if same {
+			return false, nil
+		}
+		err = c.pessimisticLockMutations(bo, action.LockCtx, action.wakeUpMode, batch.mutations)
+		return true, err
+	}
+	if resp.Resp == nil {
+		return true, errors.WithStack(tikverr.ErrBodyMissing)
+	}
```
cfzjywxk, line 253: There seem to be some duplicates with the implementation in `handlePessimisticLockResponseForceMode`, could it be abstracted to a single function?

txnkv/transaction/pessimistic.go
```diff
@@ -150,156 +168,375 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
+			return true, errors.WithStack(&tikverr.ErrDeadlock{Deadlock: deadlock})
+		}
+
+		// Extract lock from key error
+		lock, err1 := txnlock.ExtractLockFromKeyErr(keyErr)
+		if err1 != nil {
+			return true, err1
+		}
+		locks = append(locks, lock)
+	}
```
cfzjywxk, line 311: There seem to be some duplicates with the implementation in `handlePessimisticLockResponseForceMode`, could it be abstracted to a single function?

txnkv/transaction/txn.go
```diff
@@ -594,6 +650,185 @@ func (txn *KVTxn) LockKeysWithWaitTime(ctx context.Context, lockWaitTime int64,
 	return txn.LockKeys(ctx, lockCtx, keysInput...)
 }
 
+// StartAggressiveLocking makes the transaction enters aggressive locking state.
```
cfzjywxk, line 653: The aggressive locked keys processing functions are complex, we may need more SQL tests to verify it's working as expected.

txnkv/transaction/pessimistic.go
MyonKeminta: Ok. Btw I just realize that some of the code to handle the partially-successful cases is not removed. I'll remove them.

txnkv/transaction/pessimistic.go
```diff
@@ -150,156 +168,375 @@ func (action actionPessimisticLock) handleSingleBatch(c *twoPhaseCommitter, bo *
...
 			}
-			if same {
-				continue
+			if finished {
+				return nil
+			}
+
+			if retryMutations != nil && retryMutations.Len() < batch.mutations.Len() {
+				// The request may be partially succeeded. Replace the request with remaining mutations.
+				// The primary lock (if it's in this request) must have been succeeded.
```
sticnarf, line 200: Where is it ensured? (Though it's not important when we only enable this mode for single key)

MyonKeminta: TiKV should ensure it. But since this kind of code is useless in current version, I'm now removing them.

txnkv/transaction/txn.go
```diff
@@ -791,30 +1077,89 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 	}
 	skipedLockKeys := 0
 	for _, key := range keys {
-		valExists := tikv.SetKeyLockedValueExists
+		valExists := true // tikv.SetKeyLockedValueExists
```
cfzjywxk, line 1080:
```suggestion
		valExists := true
```

txnkv/transaction/txn.go
```diff
@@ -791,30 +1077,89 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
-		if lockCtx.ReturnValues || checkedExistence {
-			// If ReturnValue is disabled and CheckExistence is requested, it's still possible that the TiKV's version
-			// is too old and CheckExistence is not supported.
-			if val, ok := lockCtx.Values[string(key)]; ok {
-				// TODO: Check if it's safe to use `val.Exists` instead of assuming empty value.
+		keyStr := string(key)
+		var val tikv.ReturnedValue
+		var ok bool
+		if val, ok = lockCtx.Values[keyStr]; ok {
+			// TODO: Check if it's safe to use `val.Exists` instead of assuming empty value.
```
cfzjywxk, line 1087: Should we keep some old comments here too? Seems they are all removed.

cfzjywxk (commented):
Rest LGTM

cfzjywxk (approved):

cfzjywxk:
@sticnarf PTAL
I think there is still quite some work to do in the tidb execution path.

PR #530, opened 2022-06-18
client: add AllocID interface for mock client

3pointer:
ref https://github.com/pingcap/tidb/issues/35489

Signed-off-by: 3pointer <luancheng@pingcap.com>


disksing (approved):

D3Hunter:
we change `getAllocId` to admin api(http), no need this change

disksing:
Thanks, I'm closing it now. You can open it if realize still need it later.

PR #534, opened 2022-06-26
fix:move tikv.NewTxnClient to txnk.NewClient

583373805:
Signed-off-by: 583373805 <583373805@qq.com>


sticnarf (approved):
Thanks for fixing the examples!

PR #540, opened 2022-07-06
add replica read adjuster function in KVSnapshot

glorv:
https://github.com/pingcap/tidb/pull/35927 introduced a new follower read type `closest-adaptive` to allow send request to   the nearest replica based on the response size. This can help minimize the overhead of the extra readIndex raft propose when the request is small. This PR add a new checker function in KVSnapshot so we can also apply this logic to Get/BatchGet requests.


glorv:
@disksing @youjiali1995 PTAL

youjiali1995 (approved):

disksing (approved):

PR #556, opened 2022-07-26
Support postponed conflict check

ekexium:
ref: pingcap/tidb#36579

What's changed:
Add a flag NeedConstraintCheckInPrewrite. It's a temporary flag, which means any read or write to a key with this flag will result in unsetting the flag. It's implemented by copying a new value log in the latest stage, so that later `InspectStage` could find the key. While it brings side effects to `getValue`, so we need to figure out where the side effect is needed.
An alternative way could be recording the changed value log addresses in a separate map, and resetting them every time stage changes. I chose not to do so for better compatibility (e.g. with savepoints).

I suppose it is unnecessary to encode flags in `memBufferMutations`, and `CommitterMutationFlags` could be removed. We can just read from the mem buffer for flags.
But for simplicity, I followed the routine.


sticnarf (commented):
Considering the possible impact to `getValue`, do you think it is an option to create a new variant of `memDB` (which may be a wrapper of the current one)? And the type of `memDB` is decided by the session variable when initializing?

ekexium:
> Considering the possible impact to `getValue`, do you think it is an option to create a new variant of `memDB` (which may be a wrapper of the current one)? And the type of `memDB` is decided by the session variable when initializing?

Great idea!

internal/unionstore/memdb_arena.go
```diff
@@ -331,6 +337,47 @@ func (l *memdbVlog) getValue(addr memdbArenaAddr) []byte {
...
+	// only process if current access is on the latest version, so that we do it at most once.
+	if flags.HasNeedConstraintCheckInPrewrite() && node.vptr == valueAddr {
+		flags = kv.ApplyFlagsOps(flags, kv.DelNeedConstraintCheckInPrewrite)
+		node.setKeyFlags(flags)
+
+		// if this is not in the latest stage, we need to copy the vlog entry to the latest stage, marking it as
+		// a modification (of flag) so that the locking phase could find it via `InspectStage` and acquire the lock.
+		latestStagingHandle := len(l.memdb.stages)
+		if latestStagingHandle > 0 && !valueAddr.AfterCheckPoint(l.memdb.stages[latestStagingHandle-1]) {
+			l.memdb.setValue(node, value)
```
cfzjywxk, line 376: Considering the worst case, if the user mistakenly enables the lazy check and executes select statements after every insert statement, would it be possible there could be 2x more memory usage?
Maybe we need add adapt this change to the previous memory tracking enhancements so it could be diagnosed.

internal/unionstore/memdb_arena.go
```diff
@@ -331,6 +337,47 @@ func (l *memdbVlog) getValue(addr memdbArenaAddr) []byte {
 	return block[valueOff:lenOff:lenOff]
 }
 
+// Get value, and possibly write a new vlog entry in the latest stage to let the locking phase be aware of it via InspectStage.
+// It should be used where the read result affect users or upper level modules. Temporary flags need to be unset in such cases.
+func (l *memdbVlog) getValue(addr memdbArenaAddr) (value []byte) {
+	if !l.memdb.enableTemporaryFlags {
+		return l.pureGetValue(addr)
+	}
+	hdrOffset := addr.off - memdbVlogHdrSize
```
cfzjywxk, line 346: Could the flag check be executed here so if it's false the function returns directly.

cfzjywxk:
> An alternative way could be recording the changed value log addresses in a separate map, and resetting them every time stage changes. I chose not to do so for better compatibility (e.g. with savepoints).

Perhaps it would make the original checkpoint-related implementation much more complex.

internal/unionstore/memdb_arena.go
ekexium: I think so, but it would be less than 2x. The memory tracker already tracks the value log. 
Did you mean separating the value log space used for such rewrite? That would be indirect since we can only track the  sizes of values. And there is no way to let users query the size, for now.

ekexium:
> BTW if the key flags are stored separately would it be possible to avoid the copy?

The key here is to let `InspectStage` be able to find the change. IIUC as long as the impl doesn't change and flags are individual from stages it won't help.

internal/unionstore/memdb_arena.go
ekexium: Sorry I don't get your point. Seems to me it just does what you say?

cfzjywxk:
> > BTW if the key flags are stored separately would it be possible to avoid the copy?
> 
> The key here is to let `InspectStage` be able to find the change. IIUC as long as the impl doesn't change and flags are individual from stages it won't help.

Maybe we could record some future enhancement tasks for the `MemDB` like separating the flags.

ekexium:
> Maybe we could record some future enhancement tasks for the `MemDB` like separating the flags.

Yes if it shows obvious increment in memory footprint. Let's see it when benchmarking.

internal/unionstore/memdb_arena.go
```diff
@@ -320,7 +321,12 @@ func (l *memdbVlog) appendValue(nodeAddr memdbArenaAddr, oldValue memdbArenaAddr
 	return addr
 }
 
-func (l *memdbVlog) getValue(addr memdbArenaAddr) []byte {
+func (addr memdbArenaAddr) AfterCheckPoint(cp MemDBCheckpoint) bool {
+	return addr.idx > uint32(cp.blocks)-1 || (addr.idx == uint32(cp.blocks)-1 && addr.off > uint32(cp.offsetInBlock))
```
cfzjywxk, line 325: Would it be possible the `cp.blocks` is zero and this function is called?

ekexium: Ah maybe. Actually there is a similar function `canModify` and I've removed `AfterCheckPoint`.

cfzjywxk: Should we remove this new function too?

ekexium: Sure

ekexium:
"Compensate locks on read" can't totally avoid the anomaly of reading results that break uniqueness constraint.
Consider there are multiple unique keys, a `select for update` statement use uk1 to read, and it locks uk1. However there are uniqueness conflict on uk2, but we are not able to know it until commit.
A more concrete case
```sql
create table t2 (id int primary key, uk int, unique key i1(uk))
insert into t2 values (1, 1)
begin pessimistic
insert into t2 values (2, 1)
select * from t2 use index(primary) for update -- (1,1) and (2,1)
commit -- fail
```

I'm considering falling back to the option of only removing the temporary flag when a key is modified. Or we can stick to the current design, but do not guarantee the integrity of read results.

sticnarf:
> I'm considering falling back to the option of only removing the temporary flag when a key is modified. Or we can stick to the current design, but do not guarantee the integrity of read results.

In my opinion, it is not so meaningful to stick to the current design if it cannot totally avoid the problem.

kv/keyflags.go
```diff
@@ -62,7 +62,13 @@ const (
 	flagAssertExist
 	flagAssertNotExist
 
-	persistentFlags = flagKeyLocked | flagKeyLockedValExist
+	// It marks the conflict check of the key is postponed to prewrite. This is only used in pessimistic transactions.
+	// When the key gets locked (and the existence is checked), the flag should be removed.
+	// It should only be applied to keys with PresumeNotExist, so that an in-place constraint check becomes
+	// (a conflict check + a constraint check) in prewrite.
+	flagNeedConstraintCheckInPrewrite
```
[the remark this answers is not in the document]
ekexium: Personally I prefer `need constraint check in prewrite`. It's clearly states our intention.

internal/unionstore/memdb.go
```diff
@@ -330,13 +331,20 @@ func (db *MemDB) set(key []byte, value []byte, ops ...kv.FlagsOp) error {
...
 	x := db.traverse(key, true)
 
-	if len(ops) != 0 {
-		flags := kv.ApplyFlagsOps(x.getKeyFlags(), ops...)
-		if flags.AndPersistent() != 0 {
-			db.dirty = true
-		}
-		x.setKeyFlags(flags)
+	// the NeedConstraintCheckInPrewrite flag is temporary,
+	// every access to the node removes it unless it's explicitly set.
```
sticnarf, line 335: Now, it should be "every update"?

sticnarf (approved):

MyonKeminta (approved):

MyonKeminta:
Since the dependency to TiDB is replaced to a fork, should we merge this PR when the corresponding PR in TiDB is ready?

PR #557, opened 2022-07-27
[WIP] Introduce Codec

AmoebaProtozoa:
This PR aims unify `CodecPDClient` and `CodecPDClientV2` by introducing a `Codec` interface.


internal/apicodec/codec.go
```diff
@@ -0,0 +1,119 @@
...
+	EncodeKey(key []byte) []byte
+	// EncodeKeys encode keys to target api version format.
+	EncodeKeys(keys [][]byte) [][]byte
+	// DecodeKey get the original key from api version encoded key.
+	// It returns error if encodedKey is not well formed.
+	DecodeKey(encodedKey []byte) ([]byte, error)
+	// EncodeRange encodes key range.
+	EncodeRange(start, end []byte) ([]byte, []byte)
+	// DecodeRange decodes key range.
+	DecodeRange(encodedStart, encodedEnd []byte) ([]byte, []byte)
```
iosmanthus, line 34: seem DecodeRange will not be used in the code.

internal/apicodec/codec.go
```diff
@@ -0,0 +1,119 @@
...
+	// DecodePairs decode KvPair's key.
+	DecodePairs(encodedPairs []*kvrpcpb.KvPair) ([]*kvrpcpb.KvPair, error)
+	// EncodeRegionKey encode region's key.
+	EncodeRegionKey(key []byte) []byte
+	// DecodeRegionKey decode region's key
+	DecodeRegionKey(encodedKey []byte) ([]byte, error)
+	// EncodeRegionRange encode region's start and end.
+	EncodeRegionRange(start, end []byte) ([]byte, []byte)
+	// DecodeRegionRange decode region's start and end.
+	DecodeRegionRange(encodedStart, encodedEnd []byte) ([]byte, []byte, error)
```
iosmanthus, line 48: For PD requests, `EncodePDQueryKey` and `DecodeRegion` should be enough.

internal/apicodec/codec_v1.go
```diff
@@ -0,0 +1,100 @@
+package apicodec
+
+import (
+	"github.com/pingcap/kvproto/pkg/kvrpcpb"
+)
+
+type codecV1 struct {
+	mode     Mode
+	memCodec memCodec
```
iosmanthus, line 9: Why do we need to store the mode into `codeV1` since we could dispatch the `memCodec` in the `NewCodecV1`

AmoebaProtozoa: It was used to implement GetMode
Was thinking that maybe some component would like to know codec mode, but it seems that's not necessary. I'll get rid of it.

internal/apicodec/codec_v2.go
```diff
@@ -0,0 +1,789 @@
...
+	return c.encodeKey(start), encodedEnd
+}
+
+// decodeRange maps encodedStart and end back to normal start and
+// end without APIv2 prefixes.
+// Note that it uses empty byte slice to mark encoded start/end
+// that lies outside the prefix's range.
+func (c *codecV2) decodeRange(encodedStart, encodedEnd []byte) ([]byte, []byte) {
+	var start, end []byte
+	if bytes.Compare(start, c.prefix) < 0 {
```
iosmanthus, line 582: This comparison is wrong, we should use `encodedStart` instead.

internal/apicodec/codec_v2.go
```diff
@@ -0,0 +1,789 @@
...
+		start = []byte{}
+	} else {
+		start = encodedStart[len(c.prefix):]
+	}
+	if len(end) == 0 || bytes.Compare(end, c.endKey) >= 0 {
```
iosmanthus, line 587: Ditto, use `encodedEnd` instead.

internal/apicodec/codec_v2.go
```diff
@@ -0,0 +1,789 @@
...
+	}
+	return encodedPairs
+}
+
+func (c *codecV2) decodePairs(encodedPairs []*kvrpcpb.KvPair) ([]*kvrpcpb.KvPair, error) {
+	var pairs []*kvrpcpb.KvPair
+	for _, encodedPair := range encodedPairs {
+		var err error
+		p := *encodedPair
+		if p.Error != nil {
```
iosmanthus, line 632: Seem this `if` could be eliminated.

internal/apicodec/codec_v2.go
```diff
@@ -0,0 +1,789 @@
...
+func NewCodecV2(mode Mode) Codec {
+	// Region keys in CodecV2 are always encoded in memory comparable form.
+	codec := &codecV2{memCodec: &memComparableCodec{}}
+	switch mode {
+	case ModeRaw:
+		codec.prefix = APIV2RawKeyPrefix
+		codec.endKey = APIV2RawEndKey
+	case ModeTxn:
+		codec.prefix = APIV2TxnKeyPrefix
+		codec.endKey = APIV2TxnEndKey
```
iosmanthus, line 44: The prefix and end key should not be hard-coded here. We could derive the end key by the prefix: `be_bytes(end) == be_bytes(be(prefix) + 1)`. We might provide a factory to construct the Codec by another prefix while we could acquire keyspace meta from PD.

iosmanthus:
There is a bug that might block this pull request: https://github.com/tikv/client-go/blob/bb026bdad8f34604cc01fdafbaf4063729583a44/txnkv/txnsnapshot/scan.go#L217. We should use to check if we need to proceed the scan:
```go
			if len(reqEndKey) == 0 || (len(loc.EndKey) > 0 && bytes.Compare(loc.EndKey, reqEndKey) < 0) {
				reqEndKey = loc.EndKey
			}
```

integration_tests/raw/api_test.go
```diff
@@ -76,7 +76,9 @@ func (s *apiTestSuite) newRawKVClient(pdCli pd.Client, addrs []string) *rawkv.Cl
 
 func (s *apiTestSuite) wrapPDClient(pdCli pd.Client, addrs []string) pd.Client {
```
ystaticy, line 77: ```addrs``` not be used?

config/config.go
```diff
@@ -180,7 +180,7 @@ func GetTxnScopeFromConfig() string {
 
 // ParsePath parses this path.
 // Path example: tikv://etcd-node1:port,etcd-node2:port?cluster=1&disableGC=false
-func ParsePath(path string) (etcdAddrs []string, disableGC bool, err error) {
+func ParsePath(path string) (etcdAddrs []string, disableGC bool, keyspace string, err error) {
```
ystaticy, line 183: All variable names and function names should clearly indicate whether they are ```keyspaceName``` or ```keyspaceID```, and do not use only ```keyspace```.
Otherwise, the code maintenance cost is very high

rawkv/rawkv.go
```diff
@@ -185,33 +197,64 @@ func NewClient(ctx context.Context, pdAddrs []string, security config.Security,
 	return NewClientWithOpts(ctx, pdAddrs, WithSecurity(security), WithPDOptions(opts...))
 }
 
+func getKeyspaceID(pdCli pd.Client, keyspaceName string) (uint32, error) {
```
ystaticy, line 200: It seems to have been implemented in txnkv/client.go ?

config/config.go
iosmanthus: Then it should be named by the query convention: `keyspace_name`?

rawkv/rawkv.go
iosmanthus: Removed

PR #561, opened 2022-08-03
PessimisticLock: Lock the specified key only if the key exists

TonsnakeLin:
**1. current process**
Currently, TiKV locks the specified keys for pessimistic lock requests whether it exists or not. TiDB sends a get request to judge whether the keys exists on the TiKV firstly, then sends a pessimistic lock request to TiKV if the key exists, It makes and additional rpc request.

**2. what changed**
We add a flag `LockIfExists` for pessimistic lock request and TiKV will check the flag to decide how to execute the lock process.

This PR is depended on the PRs bellow.

- [x] https://github.com/pingcap/kvproto/pull/963
- [x] https://github.com/tikv/tikv/issues/13210


MyonKeminta (commented):
Maybe there's better to be some unit tests, which requires the lock_if_exist logic to be implemented in mocktikv

sticnarf (approved):
LGTM

sticnarf:
Please fix the lint error

internal/mockstore/mocktikv/errors.go
```diff
@@ -155,3 +155,14 @@ func (e *ErrAssertionFailed) Error() string {
 	return fmt.Sprintf("AssertionFailed { StartTS: %v, Key: %v, Assertion: %v, ExistingStartTS: %v, ExistingCommitTS: %v }",
 		e.StartTS, hex.EncodeToString(e.Key), e.Assertion.String(), e.ExistingStartTS, e.ExistingCommitTS)
 }
+
+// ErrAssertionFailed is returned if any assertion fails on a transaction request.
+type LockIfExistsFailed struct {
```
MyonKeminta, line 160: Please fix the lint, and consider prefixing the type name with `Err` to keep naming consistent with other error types here.

internal/mockstore/mocktikv/errors.go
```diff
@@ -155,3 +155,14 @@ func (e *ErrAssertionFailed) Error() string {
...
+	StartTS uint64
+	Key     []byte
+}
+
+func (e *LockIfExistsFailed) Error() string {
+	return fmt.Sprintf("LockIfExistsFailed {Lock_only_if_exists of a pessimistic lock request is set to true, but return_value is not, start_ts: %v, key: %v}",
```
MyonKeminta, line 166: This is also inconsistent with other error types, which only list fields in the braces like a struct.

internal/mockstore/mocktikv/errors.go
MyonKeminta: Actually since there's not any specific protobuf type for this kind of error, and it's a mock implementation, Personally I think it's not very necessary to define a special error type for this case. Just returnning errors.New(...) is enough.

you06:
I'm considering this case, if a transaction called lock-if-exist requests in its first statement and all the given keys do not exist, but the lock keys requests are successful, then we set the primary key to an unlocked key, I'm afraid there will be some problems when this happened. See the following timeline:

```
begin
lock([k1, k2, k3]), TiKV resp with []
set PK as k1
```

sticnarf:
> I'm considering this case, if a transaction called lock-if-exist requests in its first statement and all the given keys do not exist, but the lock keys requests are successful, then we set the primary key to an unlocked key, I'm afraid there will be some problems when this happened. See the following timeline:
> 
> ```
> begin
> lock([k1, k2, k3]), TiKV resp with []
> set PK as k1
> ```

Besides this case, if k1 doesn't exist, but we succeed locking k2 and k3, we'll face the same problem.

If we don't want to introduce an anchor key, we can disable the feature if the primary key is not selected yet.

TonsnakeLin:
> I'm considering this case, if a transaction called lock-if-exist requests in its first statement and all the given keys do not exist, but the lock keys requests are successful, then we set the primary key to an unlocked key, I'm afraid there will be some problems when this happened. See the following timeline:
> 
> ```
> begin
> lock([k1, k2, k3]), TiKV resp with []
> set PK as k1
> ```

LockOnlyIfExists request is used only for RC isolation on TiDB, I think the scene above is reasonable,right ?
Of course, I don't limit it in client-go.

TonsnakeLin:
> > I'm considering this case, if a transaction called lock-if-exist requests in its first statement and all the given keys do not exist, but the lock keys requests are successful, then we set the primary key to an unlocked key, I'm afraid there will be some problems when this happened. See the following timeline:
> > ```
> > begin
> > lock([k1, k2, k3]), TiKV resp with []
> > set PK as k1
> > ```
> 
> Besides this case, if k1 doesn't exist, but we succeed locking k2 and k3, we'll face the same problem.
> 
> If we don't want to introduce an anchor key, we can disable the feature if the primary key is not selected yet.

TonsnakeLin:
> > I'm considering this case, if a transaction called lock-if-exist requests in its first statement and all the given keys do not exist, but the lock keys requests are successful, then we set the primary key to an unlocked key, I'm afraid there will be some problems when this happened. See the following timeline:
> > ```
> > begin
> > lock([k1, k2, k3]), TiKV resp with []
> > set PK as k1
> > ```
> 
> Besides this case, if k1 doesn't exist, but we succeed locking k2 and k3, we'll face the same problem.
> 
> If we don't want to introduce an anchor key, we can disable the feature if the primary key is not selected yet.

If lock failed with LockIfExists flag, I will unset the primary key and reset ttlManager

TonsnakeLin:
@sticnarf @MyonKeminta PTAL, thank you!!

MyonKeminta:
> If lock failed with LockIfExists flag, I will unset the primary key and reset ttlManager

It seems you meant when the pessimistic lock request succeeded but the primary lock is not acquired due to not exist, instead of failing to execute the pessimistic lock request.

Though you unset the primary and cancel ttlManager in the client side (which I think you can actually avoid starting the ttlManager instead of start it and then stop it), there still exists another problem.

Consider the case that a lock request to [k1, k2, k3] where k1 is the primary, and k2, k3 is locked but k1 is untouched due to not exist. In this case, k2 and k3 will still be locked with their primary pointing to k1. This causes other transactions cannot determine the transaction's state correctly through the lock on k2 and k3, which I'm afraid may potentially lead to other correctness issues.

cc @sticnarf @you06 @cfzjywxk

TonsnakeLin:
> > If lock failed with LockIfExists flag, I will unset the primary key and reset ttlManager
> 
> It seems you meant when the pessimistic lock request succeeded but the primary lock is not acquired due to not exist, instead of failing to execute the pessimistic lock request.
> 
> Though you unset the primary and cancel ttlManager in the client side (which I think you can actually avoid starting the ttlManager instead of start it and then stop it), there still exists another problem.
> 
> Consider the case that a lock request to [k1, k2, k3] where k1 is the primary, and k2, k3 is locked but k1 is untouched due to not exist. In this case, k2 and k3 will still be locked with their primary pointing to k1. This causes other transactions cannot determine the transaction's state correctly through the lock on k2 and k3, which I'm afraid may potentially lead to other correctness issues.
> 
> cc @sticnarf @you06 @cfzjywxk

Thanks! As @sticnarf said, I will make the LockIfExists effective only when primary key is selected

sticnarf:
> Consider the case that a lock request to [k1, k2, k3] where k1 is the primary, and k2, k3 is locked but k1 is untouched due to not exist. In this case, k2 and k3 will still be locked with their primary pointing to k1. This causes other transactions cannot determine the transaction's state correctly through the lock on k2 and k3, which I'm afraid may potentially lead to other correctness issues.

If the use case is limited to **single key** point get, the current solution may suffice, because no other key may refer to a non-existing primary lock.

txnkv/transaction/txn.go
```diff
@@ -758,14 +771,33 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
 	return nil
 }
 
+// GetLcokedCount is used for testcase
+func (txn *KVTxn) GetLcokedCount() int {
+	return txn.lockedCnt
+}
+
+// GetPrimaryKey is used for testcase
+func (txn *KVTxn) GetPrimaryKey() []byte {
```
MyonKeminta, line 791: If these methods are only used in tests, you can consider add them to `TxnProbe` instead of KVTxn itself. There already exists many useful functions. Your `GetPrimaryKey` can actually be replaced by `txnProbe.GetCommitter().GetPrimaryKey()`

MyonKeminta (commented):
Rest LGTM

txnkv/transaction/txn.go
TonsnakeLin: done

MyonKeminta (approved):

txnkv/transaction/txn.go
```diff
@@ -596,6 +596,19 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 		// it before initiating an RPC request.
 		ctx = interceptor.WithRPCInterceptor(ctx, txn.interceptor)
 	}
+
+	if lockCtx.LockOnlyIfExists {
+		if !lockCtx.ReturnValues {
+			return errors.New("LockOnlyIfExists is set for Lock Context, but ReturnValues is not set")
```
cfzjywxk, line 602: Better to wrap the error information with the transaction and key information and put it in `error/errors.go`.

TonsnakeLin: @cfzjywxk PTAL, thanks

txnkv/transaction/txn.go
```diff
@@ -599,13 +599,19 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 
 	if lockCtx.LockOnlyIfExists {
 		if !lockCtx.ReturnValues {
-			return errors.New("LockOnlyIfExists is set for Lock Context, but ReturnValues is not set")
+			return &tikverr.ErrLockOnlyIfExistsNoReturnValue{
+				ForUpdateTs: lockCtx.ForUpdateTS,
+				LockKey:     keysInput[0],
```
cfzjywxk, line 604: Need to check `len(keysInput) > 0` first?

error/error.go
```diff
@@ -238,10 +238,34 @@ type ErrAssertionFailed struct {
...
+}
+
 func (e *ErrAssertionFailed) Error() string {
 	return fmt.Sprintf("assertion failed { %s }", e.AssertionFailed.String())
 }
 
+func (e *ErrLockOnlyIfExistsNoReturnValue) Error() string {
+	return fmt.Sprintf("LockOnlyIfExists is set for Lock Context, but ReturnValues is not set, "+
+		"ForUpdateTs is {%d}, one of lock keys is {%v}.",
+		e.ForUpdateTs, string(e.LockKey))
```
cfzjywxk, line 260:
```suggestion
		e.ForUpdateTs, hex.EncodeToString(e.LockKey))
```

error/error.go
```diff
@@ -238,10 +238,34 @@ type ErrAssertionFailed struct {
...
+}
+
+func (e *ErrLockOnlyIfExistsNoPrimaryKey) Error() string {
+	return fmt.Sprintf("LockOnlyIfExists is set for Lock Context, but primary key of current transaction is not set, "+
+		"ForUpdateTs is {%d}, one of lock keys is {%s}",
+		e.ForUpdateTs, string(e.LockKey))
```
cfzjywxk, line 266:
```suggestion
		e.ForUpdateTs, hex.EncodeToString(e.LockKey))
```

error/error.go
```diff
@@ -238,10 +238,34 @@ type ErrAssertionFailed struct {
...
+	ForUpdateTs uint64
+	LockKey     []byte
+}
+
 func (e *ErrAssertionFailed) Error() string {
 	return fmt.Sprintf("assertion failed { %s }", e.AssertionFailed.String())
 }
 
+func (e *ErrLockOnlyIfExistsNoReturnValue) Error() string {
+	return fmt.Sprintf("LockOnlyIfExists is set for Lock Context, but ReturnValues is not set, "+
```
cfzjywxk, line 258: Is `StartTS` needed to be recorded so that related logs could be searched quickly?

error/error.go
TonsnakeLin: done

error/error.go
TonsnakeLin: done

txnkv/transaction/txn.go
TonsnakeLin: if len(keysInput) == 0, it prints "{}".

cfzjywxk: if len(keysInput) == 0, would `keysInput[0]` cause `index out of range`?

TonsnakeLin: done

cfzjywxk (approved):

PR #563, opened 2022-08-08
Fix that histogram is not inserted into rpcNetLatencyHistCache

sticnarf:
`latHist` should be inserted into `rpcNetLatencyHistCache` while we inserted it into `sendReqHistCache` due to a code-paste mistake.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,4 +46,6 @@ jobs:
 
       - name: Lint
         uses: golangci/golangci-lint-action@v2.5.2
+        with:
+          version: v1.47.3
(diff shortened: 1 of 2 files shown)


disksing (approved):

you06 (approved):

disksing:
there are some wired formatting errors.

sticnarf:
> there are some wired formatting errors.

The latest golangci-lint uses gofmt from Go 1.19. The formatting rules about comments are different from before. I downgrade golangci-lint to work around the issue now. We can look for solutions to the issue later.

PR #566, opened 2022-08-11
Clear intersecting regions in the cache when inserting a region

### LINKED ISSUE #565: `insertRegionToCache` should remove intersecting regions in the cache
Our region cache B-Tree uses the start key of the region as the entry key. So, `insertRegionToCache` only replaces the region with the same start key:
https://github.com/tikv/client-go/blob/516cfcdecce865d90ec05dd1cc1de060c35438dc/internal/locate/region_cache.go#L1287

Considering there's a table with lots of regions, the user removes all data from the table. So, PD will finally merge all regions into one. In our region cache implementation, only the first region (containing the table head) is updated, while all other regions still exist in the cache.

When we locate a key near the end of the table, `DescendLessOrEqual` will iterate over all regions that are left in the cache. This will waste a lot of CPU and make region search really slow, and it is also possible to cause lots of unnecessary PD `loadRegion` RPCs.

To solve the problem, when we do `insertRegionToCache`, we should iterate the B-Tree for all regions that intersect with the inserted region and remove them.

This is probably the root cause of https://github.com/tikv/client-go/issues/532

sticnarf:
Closes #565

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1316,6 +1316,11 @@ func (c *RegionCache) insertRegionToCache(cachedRegion *Region) {
 	if !ok || latest.GetVer() < newVer.GetVer() || latest.GetConfVer() < newVer.GetConfVer() {
 		c.mu.latestVersions[cachedRegion.VerID().id] = newVer
 	}
+	// The intersecting regions in the cache are probably stale, clear them.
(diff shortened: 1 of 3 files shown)


youjiali1995 (approved):

disksing:
@sticnarf please resolve conflicts

PR #574, opened 2022-08-29
*: disable PermitWithoutStream to avoid too many pings

jackysp:
Signed-off-by: Jack Yu <jackysp@gmail.com>

close: https://github.com/pingcap/tidb/issues/36861


jackysp:
TiDB may connect itself to get some info via grpc protocol, such as cluster_xx tables in information_schema database.

sticnarf:
I'm curious what was the original intention of setting `PermitWithoutStream: true`? In other words, what regression might this change bring to us?

disksing:
As I remember, keepalive is introduced to avoid the case when tikv is not working and client-go doesn't know about it, it may waste some time by sending requests to tikv.

jackysp:
I think tidb-server introduced a new way https://github.com/pingcap/tidb/pull/14145 to detect if tikv-server is alive or not.

disksing:
@jackysp maybe we can make it a configure option like `keepAlive` and `keepAliveTimeout` and make it use `true` by default so other users that don't handle it by themselves won't be affected.

disksing:
As tikv-server does not enable `keepalive_permit_without_calls`, this option has no effect and can be removed safely.

PR #575, opened 2022-08-30
Fix oob

AmoebaProtozoa:
Fix region cache out of bound problem.


internal/apicodec/codec.go
```diff
@@ -33,7 +33,7 @@ type Codec interface {
 	// EncodeRegionRange encode region's start and end.
 	EncodeRegionRange(start, end []byte) ([]byte, []byte)
 	// DecodeRegionRange decode region's start and end.
-	DecodeRegionRange(encodedStart, encodedEnd []byte) ([]byte, []byte, error)
+	DecodeRegionRange(encodedStart, encodedEnd []byte) (bool, []byte, []byte, error)
```
iosmanthus, line 36: We should return an error while the region range is invalid instead of a boolean flag.

internal/apicodec/codec_v2.go
```diff
@@ -228,7 +228,8 @@ func (c *codecV2) EncodeRequest(req *tikvrpc.Request) (*tikvrpc.Request, error)
 		newReq.Req = &r
 	case tikvrpc.CmdBatchCop:
```
iosmanthus, line 229: This request is sent to TiFlash only, we may ignore this, for now, the tiflash requests should be sent in plain.

internal/apicodec/codec_v2.go
```diff
@@ -228,7 +228,8 @@ func (c *codecV2) EncodeRequest(req *tikvrpc.Request) (*tikvrpc.Request, error)
...
 		r := *req.BatchCop()
-		// TODO(iosmanthus): handle cop Request.
+		r.Regions = c.encodeCopRegions(r.Regions)
```
iosmanthus, line 231: This request is sent to TiFlash only, we may ignore this, for now, the tiflash requests should be sent in plain.

internal/apicodec/codec_v2.go
```diff
@@ -531,7 +532,11 @@ func (c *codecV2) DecodeResponse(req *tikvrpc.Request, resp *tikvrpc.Response) (
 		}
 		r.Range = c.decodeCopRange(r.Range)
 	case tikvrpc.CmdBatchCop:
-		// TODO: handle cop response.
+		r := resp.Resp.(*coprocessor.BatchResponse)
```
iosmanthus, line 535: Ditto.

internal/locate/region_cache.go
```diff
@@ -402,9 +401,9 @@ func NewRegionCache(pdClient pd.Client) *RegionCache {
 		pdClient: pdClient,
 	}
 
-	c.apiVersion = kvrpcpb.APIVersion_V1
+	c.codec = apicodec.NewCodecV1(apicodec.ModeTxn)
```
iosmanthus, line 404: The default codec should be in raw mode.

internal/locate/region_cache.go
```diff
@@ -1743,6 +1742,17 @@ func (c *RegionCache) OnRegionEpochNotMatch(bo *retry.Backoffer, ctx *RPCContext
 	newRegions := make([]*Region, 0, len(currentRegions))
 	// If the region epoch is not ahead of TiKV's, replace region meta in region cache.
 	for _, meta := range currentRegions {
+		valid, start, end, err := c.codec.DecodeRegionRange(meta.GetStartKey(), meta.GetEndKey())
+		if err != nil {
```
iosmanthus, line 1746: We could check the cause of the `err` here.

internal/locate/region_cache.go
```diff
@@ -1743,6 +1742,17 @@ func (c *RegionCache) OnRegionEpochNotMatch(bo *retry.Backoffer, ctx *RPCContext
...
+			logutil.BgLogger().Error("failed to decode region range", zap.Error(err))
+			return false, err
+		}
+		// If cachedRegion is not valid, skip inserting it to cache.
+		if !valid {
+			continue
+		}
+		meta.StartKey = start
```
iosmanthus, line 1754: I don't think it's safe to write the meta in place since the meta slice could be shared across the goroutine.

iosmanthus:
The most straightforward implementation is to ignore the invalid region while decoding the EpochNotMatch region errors.

```go
	if errInfo := regionError.EpochNotMatch; errInfo != nil {
		for _, meta := range errInfo.CurrentRegions {
			meta.StartKey, meta.EndKey, err = c.DecodeRegionRange(meta.StartKey, meta.EndKey)
			if err != nil {
                if _, ok := errors.Cause(err).(ErrInvalidRegion); ok {
                     continue.
                }
				return nil, err
			}
		}
	}

```

internal/apicodec/codec_v2.go
```diff
@@ -755,12 +778,20 @@ func (c *codecV2) decodeRegionError(regionError *errorpb.Error) (*errorpb.Error,
 	}
 
 	if errInfo := regionError.EpochNotMatch; errInfo != nil {
+		var decodedRegions []*metapb.Region
```
iosmanthus, line 781: We might preallocate some space for this slice.

internal/apicodec/codec_v2.go
```diff
@@ -755,12 +778,20 @@ func (c *codecV2) decodeRegionError(regionError *errorpb.Error) (*errorpb.Error,
...
 		for _, meta := range errInfo.CurrentRegions {
-			meta.StartKey, meta.EndKey, err = c.DecodeRegionRange(meta.StartKey, meta.EndKey)
+			r := *meta
```
iosmanthus, line 783: No need to copy the meta?

internal/apicodec/codec_v2.go
```diff
@@ -755,12 +778,20 @@ func (c *codecV2) decodeRegionError(regionError *errorpb.Error) (*errorpb.Error,
 	}
 
 	if errInfo := regionError.EpochNotMatch; errInfo != nil {
```
iosmanthus, line 780: Should we add some tests to cover this case? We could add some TODO here, could be done in the future.

iosmanthus:
Rest LGTM.

internal/apicodec/codec_v2.go
AmoebaProtozoa: I have added some tests in codec_v2_test.go, PTAL

internal/apicodec/codec_v2.go
```diff
@@ -588,6 +596,10 @@ func (c *codecV2) DecodeRegionRange(encodedStart, encodedEnd []byte) ([]byte, []
 			return nil, nil, err
 		}
 	}
+	// If the region range does not overlap with the range of current keyspace, return ErrInvalidRegion.
```
iosmanthus, line 599: Here we may use decode key to produce the error.

internal/apicodec/codec_v2.go
```diff
@@ -639,30 +618,50 @@ func (c *codecV2) encodeRange(start, end []byte, reverse bool) ([]byte, []byte)
...
-// Note that it uses empty byte slice to mark encoded start/end
-// that lies outside the prefix's range.
-func (c *codecV2) decodeRange(encodedStart, encodedEnd []byte) ([]byte, []byte) {
+func (c *codecV2) DecodeRange(encodedStart, encodedEnd []byte) ([]byte, []byte, error) {
 	var start, end []byte
-	if bytes.Compare(encodedStart, c.prefix) < 0 {
-		start = []byte{}
-	} else {
-		start = encodedStart[len(c.prefix):]
+	var err error
```
iosmanthus, line 632: What about moving the declaration to the return value list?

like:

```go
func (c *codecV2) DecodeRange(encodedStart, encodedEnd []byte) (start []byte, end []byte, err error)
```

internal/apicodec/codec_v2.go
```diff
@@ -639,30 +618,50 @@ func (c *codecV2) encodeRange(start, end []byte, reverse bool) ([]byte, []byte)
...
 	}
-	return start, end
+	return start, end, nil
+}
+
+func (c *codecV2) EncodeKey(key []byte) []byte {
+	return append(c.prefix, key...)
+}
+
+func (c *codecV2) DecodeKey(key []byte) ([]byte, []byte, error) {
```
iosmanthus, line 648: Sorry about the interface change, the latest `api-v2` branch has changed it back to return only the user key and the error.

internal/apicodec/codec_v2.go
```diff
@@ -639,30 +618,50 @@ func (c *codecV2) encodeRange(start, end []byte, reverse bool) ([]byte, []byte)
...
+
+func (c *codecV2) decodeKey(encodedKey []byte) ([]byte, error) {
+	if len(encodedKey) < len(c.prefix) {
+		return nil, errors.Errorf("invalid encoded key prefix: %q", encodedKey)
+	}
+	// If the encodedKey does not belong to the keyspace, return error.
+	if bytes.Compare(encodedKey, c.endKey) > 0 || bytes.Compare(encodedKey, c.prefix) < 0 {
+		return nil, errKeyOutOfBound
+	}
+	return encodedKey[len(c.prefix):], nil
```
iosmanthus, line 664:
```suggestion
func (c *codecV2) decodeKey(encodedKey []byte) ([]byte, error) {
	if !bytes.HasPrefix(encodedKey, c.prefix) {
		return nil, errKeyOutOfBound
	}
	return encodedKey[len(c.prefix):], nil
}
```

iosmanthus:
LGTM

PR #577, opened 2022-09-02
update tidb dependency of integration tests & use go 1.19

ekexium:
Signed-off-by: ekexium <eke@fastmail.com>


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/tikv/client-go/v2
 
-go 1.18
+go 1.19
```
sticnarf, line 3: Can we keep it unchanged at least for the root `go.mod`? Now, the minimal supported go version for this library is still 1.18.

.github/workflows/test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.18
+        go-version: 1.19
```
sticnarf, line 31: Can we still use 1.18 for the unit tests of the library?

ekexium:
CI seems extreamely unstable even if I only changed go version for compatibility test. I don't know if it's related to go 1.19.

sticnarf (approved):

PR #579, opened 2022-09-05
*: introduce the non-retryable error FlashbackInProgress

JmPotato:

Ref https://github.com/tikv/tikv/issues/13397.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1389,6 +1389,8 @@ func regionErrorToLabel(e *errorpb.Error) string {
 		return "disk_full"
 	} else if e.GetRecoveryInProgress() != nil {
 		return "recovery_in_progress"
+	} else if e.GetFlashbackInProgress() != nil {
+		return "flashback_in_progress"
 	}
 	return "unknown"
 }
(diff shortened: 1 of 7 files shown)


internal/locate/region_request.go
```diff
@@ -1446,6 +1448,14 @@ func (s *RegionRequestSender) onRegionError(bo *retry.Backoffer, ctx *RPCContext
 		return false, nil
 	}
 
+	// Since we expect that the workload should be stopped during the flashback progress,
+	// if a request meets the FlashbackInProgress error, it should stop retrying immediately
+	// to avoid unnecessary backoff and potential unexpected data status to the user.
+	if regionErr.GetFlashbackInProgress() != nil {
+		logutil.BgLogger().Debug("tikv reports `FlashbackInProgress`", zap.Stringer("req", req), zap.Stringer("ctx", ctx))
+		return false, nil
```
sticnarf, line 1456: If you want the request to fail as soon as possible, is it better to return an error here?

If we just return `false, nil` here, I worry the caller may rebuild and request and retry.

sticnarf (commented):
What if it's in a rolling upgrade process and the old client-go ignores the new error? Is it assumed to be impossible?

JmPotato:
> What if it's in a rolling upgrade process and the old client-go ignores the new error? Is it assumed to be impossible?

Currently, a flashback SQL will be a cluster-level operation, and I think this scenario is unlikely to occur in real-world use.

JmPotato:
> What if it's in a rolling upgrade process and the old client-go ignores the new error? Is it assumed to be impossible?

For now, Flashback is a cluster-level operation, and I think this scenario is unlikely to occur in real-world use. Even if it happens, since a region in flashback progress will block all read and writes, so it's not a big deal.

sticnarf:
LGTM. The new proto doesn't seem to break old code. I think you can merge the proto and remove the replace.

PR #582, opened 2022-09-07
extend the scope of using lock_only_if_exists 

TonsnakeLin:
Signed-off-by: TonsnakeLin <lpbgytong@163.com>
At first, we enable  lock_only_if_exists mode only when the primary key of the current transaction is selected when lock keys.
We expand the use scope, if the number of  keys to be locked is one, we also enable lock_only_if_exists even though the primary key is not selected.

### DIFF
--- a/integration_tests/2pc_test.go
+++ b/integration_tests/2pc_test.go
@@ -902,11 +902,23 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
 	s.Equal(txn.GetLcokedCount(), 3)
 	s.Nil(txn.Commit(context.Background()))
 
+	// Primary key is not selected, but here is only one key.
+	txn = s.begin()
+	txn.SetPessimistic(true)
+	lockCtx = getLockOnlyIfExistsCtx(txn, 1)
+	s.Nil(txn.LockKeys(context.Background(), lockCtx, key))
+	s.Equal(txn.GetCommitter().GetPrimaryKey(), key)
+	memBuf = txn.GetMemBuffer()
+	flags, err = memBuf.GetFlags(key)
+	s.Equal(flags.HasLockedValueExists(), true)
+	s.Equal(txn.GetLcokedCount(), 1)
+	s.Nil(txn.Commit(context.Background()))
+
 	// When the primary key is not selected, it can't send a lock request with LockOnlyIfExists mode
 	txn = s.begin()
 	txn.SetPessimistic(true)
 	lockCtx = getLockOnlyIfExistsCtx(txn, 1)
-	err = txn.LockKeys(context.Background(), lockCtx, key)
+	err = txn.LockKeys(context.Background(), lockCtx, key, key2)
 	err, ok = err.(*tikverr.ErrLockOnlyIfExistsNoPrimaryKey)
 	s.Equal(ok, true)
 	s.Nil(txn.Rollback())
--- a/integration_tests/2pc_test.go
+++ b/integration_tests/2pc_test.go
@@ -801,7 +801,7 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
 	checkLockKeyResult(s, txn, lockCtx, key, key, 1, key0)
 	flags := getMembufferFlags(s, txn, key, "")
 	s.Equal(flags.HasLockedValueExists(), true)
-	s.Equal(txn.GetLcokedCount(), 2)
+	s.Equal(txn.GetLockedCount(), 2)
 	s.Nil(txn.Rollback())
 
 	// Locked "key2" unsuccessfully.
@@ -812,7 +812,7 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
 	s.Nil(txn.LockKeys(context.Background(), lockCtx, key2))
 	checkLockKeyResult(s, txn, lockCtx, key2, nil, 1, key0)
 	flags = getMembufferFlags(s, txn, key, "not exist")
-	s.Equal(txn.GetLcokedCount(), 1)
+	s.Equal(txn.GetLockedCount(), 1)
 	s.Nil(txn.Rollback())
 
 	// Lock order is key, key2, key3.
@@ -833,7 +833,7 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
 	s.Equal(err.Error(), "not exist")
 	flags, err = memBuf.GetFlags(key3)
 	s.Equal(flags.HasLockedValueExists(), true)
-	s.Equal(txn.GetLcokedCount(), 3)
+	s.Equal(txn.GetLockedCount(), 3)
 	s.Nil(txn.Rollback())
 
 	// Lock order is key2, key, key3.
@@ -854,7 +854,7 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
 	s.Equal(err.Error(), "not exist")
 	flags, err = memBuf.GetFlags(key3)
 	s.Equal(flags.HasLockedValueExists(), true)
-	s.Equal(txn.GetLcokedCount(), 3)
+	s.Equal(txn.GetLockedCount(), 3)
 	s.Nil(txn.Commit(context.Background()))
 
 	// LockKeys(key2), LockKeys(key3, key).
@@ -874,7 +874,7 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
 	s.Equal(err.Error(), "not exist")
 	flags, err = memBuf.GetFlags(key3)
 	s.Equal(flags.HasLockedValueExists(), true)
-	s.Equal(txn.GetLcokedCount(), 3)
+	s.Equal(txn.GetLockedCount(), 3)
 	s.Nil(txn.Commit(context.Background()))
 
 	// Lock order is key0, key, key3.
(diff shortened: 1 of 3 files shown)


integration_tests/2pc_test.go
```diff
@@ -902,11 +902,23 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
...
+	// Primary key is not selected, but here is only one key.
+	txn = s.begin()
+	txn.SetPessimistic(true)
+	lockCtx = getLockOnlyIfExistsCtx(txn, 1)
+	s.Nil(txn.LockKeys(context.Background(), lockCtx, key))
+	s.Equal(txn.GetCommitter().GetPrimaryKey(), key)
+	memBuf = txn.GetMemBuffer()
+	flags, err = memBuf.GetFlags(key)
+	s.Equal(flags.HasLockedValueExists(), true)
+	s.Equal(txn.GetLcokedCount(), 1)
```
sticnarf, line 914: The function name has a typo, can you help fix it?

integration_tests/2pc_test.go
```diff
@@ -902,11 +902,23 @@ func (s *testCommitterSuite) TestPessimisticLockIfExists() {
 	s.Equal(txn.GetLcokedCount(), 3)
 	s.Nil(txn.Commit(context.Background()))
 
+	// Primary key is not selected, but here is only one key.
```
sticnarf, line 905: I think with this condition, there should be two test cases depending on whether the key exists.

When the key does not exist, we should ensure the primary remains `nil` after `LockKeys`.

txnkv/transaction/test_probe.go
```diff
@@ -92,8 +92,8 @@ func (txn TxnProbe) GetStartTime() time.Time {
 	return txn.startTime
 }
 
-// GetLcokedCount is used for testcase
-func (txn TxnProbe) GetLcokedCount() int {
+// GetLockedCount is used for testcase
```
MyonKeminta, line 95:
```suggestion
// GetLockedCount returns the count of locks acquired by the transaction.
```

txnkv/transaction/txn.go
```diff
@@ -763,6 +763,13 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 			checkedExistence = true
 		}
 	}
+	if assignedPrimaryKey && lockCtx.LockOnlyIfExists && len(keys) == 1 {
+		if !txn.unsetPrimaryKeyIfNeed(lockCtx) {
+			memBuf.UpdateFlags(keys[0], tikv.SetKeyLocked, tikv.DelNeedCheckExists, tikv.SetKeyLockedValueExists)
+			txn.lockedCnt++
+		}
+		return nil
```
MyonKeminta, line 771: Why not only unset the primary, and then do nothing but continue to the result handling and membuf updating logic below?
It seems when we are going to introduce other handling logic to these part (like adding some more flags), we will need to maintain both (L767-771 and L774-795). I'm afraid the new part will be very easily forgotten.

TonsnakeLin: In actually, I thought about doing that before.  I'll modify it.

txnkv/transaction/txn.go
```diff
@@ -789,6 +792,23 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 	return nil
 }
 
+// unsetPrimaryKeyIfNeed is used to unset pk of current transaction.
+// When lock only one key with LockOnlyIfExists flag, this key is selected as pk, if the key doesn't
+// exist on TiKV, it doesn't generate a lock for this key, we should unset this pk for this scene.
+// The caller should ensure bellow conditions.
+// (1) only one key to be locked (2) pk is not selected before (2) with LockOnlyIfExists
+// retun true means the pk has been unseted
```
MyonKeminta, line 800:
```suggestion
// unsetPrimaryKeyIfNeed is used to unset pk of the transaction after performing LockOnlyIfExists.
// When locking only one key with LockOnlyIfExists flag, the key will be selected as primary if
// it's the first lock of the transaction. If the key doesn't exist on TiKV, the key won't be
// locked, in which case we should unset the primary of the transaction.
// The caller must ensure bellow conditions:
// (1) only one key to be locked (2) primary is not selected before (2) with LockOnlyIfExists
// Returns true if the primary has been unset.
```

&ast;Trying to make the English expression more clear...

* Mention the requirement of LockOnlyIfExists in the first sentence to avoid potential misusing
* Prefer not using `pk` to stand for the primary of a transaction in documentation (I think)
* Some grammar adjustments...

TonsnakeLin:
@sticnarf @MyonKeminta  Please review again, thanks!

txnkv/transaction/txn.go
```diff
@@ -789,6 +792,24 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
...
 }
 
+// unsetPrimaryKeyIfNeed is used to unset pk of the transaction after performing LockOnlyIfExists.
+// When locking only one key with LockOnlyIfExists flag, the key will be selected as primary if
+// it's the first lock of the transaction. If the key doesn't exist on TiKV, the key won't be
+// locked, in which case we should unset the primary of the transaction.
+// The caller must ensure bellow conditions:
+// (1) only one key to be locked (2) primary is not selected before (2) with LockOnlyIfExists
+// Returns true if the primary has been unset.
+func (txn *KVTxn) unsetPrimaryKeyIfNeed(lockCtx *tikv.LockCtx) bool {
```
sticnarf, line 802: unsetPrimaryKeyIfNeed**ed** is more grammarly correct.

txnkv/transaction/txn.go
```diff
@@ -789,6 +792,24 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 	return nil
 }
 
+// unsetPrimaryKeyIfNeed is used to unset pk of the transaction after performing LockOnlyIfExists.
+// When locking only one key with LockOnlyIfExists flag, the key will be selected as primary if
+// it's the first lock of the transaction. If the key doesn't exist on TiKV, the key won't be
+// locked, in which case we should unset the primary of the transaction.
+// The caller must ensure bellow conditions:
```
sticnarf, line 799: Change to "The caller must ensure the conditions below:"

txnkv/transaction/txn.go
sticnarf: What is the return value used for?

txnkv/transaction/txn.go
```diff
@@ -763,6 +763,9 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 			checkedExistence = true
 		}
 	}
+	if assignedPrimaryKey && lockCtx.LockOnlyIfExists && len(keys) == 1 {
```
sticnarf, line 766: Here I prefer to make the `&& len(keys) == 1` part an assertion. For example, add
```go
if len(keys) != 1 {
	panic("LockOnlyIfExists only assigns the primary key when locking only one key")
}
```
Then, if `LockOnlyIfExists` extends its usage in the future, it's more possible for us to find we need to change here too.

sticnarf (commented):
LGTM except for some small suggestions.

txnkv/transaction/txn.go
TonsnakeLin: good idea

TonsnakeLin:
@MyonKeminta @sticnarf PTAL, thanks!

txnkv/transaction/txn.go
```diff
@@ -792,22 +796,20 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 	return nil
 }
 
-// unsetPrimaryKeyIfNeed is used to unset pk of the transaction after performing LockOnlyIfExists.
+// unsetPrimaryKeyIfNeed is used to unset primary key of the transaction after performing LockOnlyIfExists.
 // When locking only one key with LockOnlyIfExists flag, the key will be selected as primary if
 // it's the first lock of the transaction. If the key doesn't exist on TiKV, the key won't be
 // locked, in which case we should unset the primary of the transaction.
-// The caller must ensure bellow conditions:
+// The caller must ensure the conditions bellow:
```
sticnarf, line 803: bellow -> below ...

TonsnakeLin: done
PTAL @sticnarf @MyonKeminta  thanks

txnkv/transaction/txn.go
```diff
@@ -789,6 +796,22 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 	return nil
 }
 
+// unsetPrimaryKeyIfNeed is used to unset primary key of the transaction after performing LockOnlyIfExists.
+// When locking only one key with LockOnlyIfExists flag, the key will be selected as primary if
+// it's the first lock of the transaction. If the key doesn't exist on TiKV, the key won't be
+// locked, in which case we should unset the primary of the transaction.
+// The caller must ensure the conditions below:
+// (1) only one key to be locked (2) primary is not selected before (2) with LockOnlyIfExists
+// Returns true if the primary has been unset.
```
MyonKeminta, line 805:
```suggestion
```

The return value is removed.

txnkv/transaction/txn.go
```diff
@@ -763,6 +763,13 @@ func (txn *KVTxn) LockKeys(ctx context.Context, lockCtx *tikv.LockCtx, keysInput
 			checkedExistence = true
 		}
 	}
+	if assignedPrimaryKey && lockCtx.LockOnlyIfExists {
+		if len(keys) != 1 {
+			panic("When using the LockOnlyIfExists mode to lock keys, if one of the keys is selected as " +
+				"the primary key, the number of input keys must be one.")
```
MyonKeminta, line 769: Use brief  English sentence or incomplete sentence to describe the reason of panicking. I think you can learn the expression style in our existing logs and error messages. You can adopt @sticnarf 's suggestion directly.

MyonKeminta (commented):
Rest LGTM

MyonKeminta (approved):

PR #584, opened 2022-09-09
Don't let pessimistic txn prewrite in the optimistic way

ekexium:
If a pessimistic transaction has its for_update_ts = 0, we set it to its start_ts so that it is prewritten in the pessimistic way in TiKV, making it seems more consistent.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -1398,6 +1398,17 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		c.hasTriedOnePC = true
 	}
 
+	// if lazy uniqueness check is enabled in TiDB (@@constraint_check_in_place_pessimistic=0), for_update_ts might be
+	// zero for a pessimistic transaction. We set it to the start_ts to force the PrewritePessimistic path in TiKV.
+	if c.forUpdateTS == 0 {
+		for i := 0; i < c.mutations.Len(); i++ {
+			if c.mutations.NeedConstraintCheckInPrewrite(i) {
+				c.forUpdateTS = c.startTS
+				break
+			}
+		}
+	}
+
 	// TODO(youjiali1995): It's better to use different maxSleep for different operations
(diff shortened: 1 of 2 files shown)


ekexium:
cc @sticnarf @cfzjywxk

ekexium:
Wait, there's something wrong..

txnkv/transaction/2pc.go
```diff
@@ -1398,6 +1398,17 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		c.hasTriedOnePC = true
 	}
 
+	// if lazy uniqueness check is enabled in TiDB (@@constraint_check_in_place_pessimistic=0), for_update_ts might be
+	// zero for a pessimistic transaction. We set it to the start_ts to force the PrewritePessimistic path in TiKV.
+	if c.forUpdateTS == 0 {
+		for i := 0; i < c.mutations.Len(); i++ {
+			if c.mutations.NeedConstraintCheckInPrewrite(i) {
+				c.forUpdateTS = c.startTS
```
cfzjywxk, line 1406: Or could we just set 
```
if c.forUpdateTS == 0 {
  if c.is_pessimistic_txn
    c.forUpdateTS = c.startTS
}
```

ekexium: Lazy uniqueness check is the only situation that we know this could happen. If we set it for whatever pessimistic transaction, we may miss unknown bugs. The chance is extremely low though. I'm fine with either option. What do you think?

cfzjywxk: To reduce the risk it's ok to use the current change. Maybe we could more investigations letter, we could record an issue about it in the tidb repo.

cfzjywxk (approved):
LGTM

txnkv/transaction/2pc.go
ekexium: Let's just put a TODO here.

PR #589, opened 2022-09-22
ignore encode for tiflash requests

iosmanthus:


internal/client/client.go
```diff
@@ -535,7 +535,7 @@ func (c *RPCClient) sendRequest(ctx context.Context, addr string, req *tikvrpc.R
 
 // SendRequest sends a Request to server and receives Response.
 func (c *RPCClient) SendRequest(ctx context.Context, addr string, req *tikvrpc.Request, timeout time.Duration) (*tikvrpc.Response, error) {
-	if c.option == nil || c.option.codec == nil {
+	if c.option == nil || c.option.codec == nil || req.StoreTp == tikvrpc.TiFlash {
```
disksing, line 538: should the change only affect when using api v2?

iosmanthus: If in API V1, we should not encode the request either.

PR #591, opened 2022-09-28
fix codecV1 encode region key

AmoebaProtozoa:
Signed-off-by: David <8039876+AmoebaProtozoa@users.noreply.github.com>

Remove the special treatment of nil key encode when using apiv1

This should result in consistent behavior of codecV1 with the original.


internal/apicodec/codec_v1.go
```diff
@@ -146,7 +143,10 @@ func (c *codecV1) DecodeRegionKey(encodedKey []byte) ([]byte, error) {
 }
 
 func (c *codecV1) EncodeRegionRange(start, end []byte) ([]byte, []byte) {
-	return c.EncodeRegionKey(start), c.EncodeRegionKey(end)
+	if len(end) > 0 {
+		return c.EncodeRegionKey(start), c.EncodeRegionKey(end)
+	}
+	return c.EncodeRegionKey(start), end
```
iosmanthus, line 149: What if the start key is empty since the regions we see in `pd-ctl` have the empty start key which means it's not encoded?

PR #596, opened 2022-10-11
Fix the false positive assertion error caused by loss of pessimistic locks 

ekexium:
To fix https://github.com/pingcap/tidb/issues/38365
Stash the assertion error found in client-go; report it after prewrite succeeds

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -181,7 +181,7 @@ type twoPhaseCommitter struct {
 	prewriteTotalReqNum int
 
 	// assertion error happened when initializing mutations, could be false positive if pessimistic lock is lost
-	assertionErr error
+	stashedAssertionError error
 }
 
 type memBufferMutations struct {
@@ -633,7 +633,7 @@ func (c *twoPhaseCommitter) initKeysAndMutations(ctx context.Context) error {
 				// from the prewrite phase.
 				if err1 != nil && assertionError == nil {
 					assertionError = errors.WithStack(err1)
-					c.assertionErr = assertionError
+					c.stashedAssertionError = assertionError
 					c.txn.enableAsyncCommit = false
 					c.txn.enable1PC = false
 				}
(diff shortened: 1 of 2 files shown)


txnkv/transaction/2pc.go
```diff
@@ -622,8 +625,17 @@ func (c *twoPhaseCommitter) initKeysAndMutations(ctx context.Context) error {
 				// Do not exit immediately here. To rollback the pessimistic locks (if any), we need to finish
 				// collecting all the keys.
 				// Keep only the first assertion error.
+
+				// assertion errors is treated differently from other errors. If there is an assertion error,
+				// it's probably cause by loss of pessimistic locks, so we can't directly return the assertion error.
+				// Instead, we stash the error, forbid async commit and 1PC, then let the prewrite continue.
```
sticnarf, line 631: Let me confirm. Forbidding async commit and 1PC is to give us the chance to abort the transaction by the client. And the main goal of this PR is to change the assertion error in this case to become a constraint check failure or write failure. Do I understand it correctly?

ekexium: Yes exactly

ekexium:
@MyonKeminta PTAL

txnkv/transaction/2pc.go
```diff
@@ -179,6 +179,9 @@ type twoPhaseCommitter struct {
 
 	// The total number of kv request after batch split.
 	prewriteTotalReqNum int
+
+	// assertion error happened when initializing mutations, could be false positive if pessimistic lock is lost
```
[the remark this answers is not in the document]
ekexium: IMO it's a false positive of the type assertion failure, considering the error type directly indicates a risk of data corruption.

sticnarf (approved):

txnkv/transaction/2pc.go
```diff
@@ -1507,6 +1515,12 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		return err
 	}
 
+	// return assertion error found in TiDB after prewrite succeeds to prevent false positive. Note this is only visible
+	// when async commit or 1PC is disabled.
+	if c.assertionErr != nil {
+		return c.assertionErr
```
MyonKeminta, line 1521: May it be better if it panics when either 1pc or async commit is enabled?

txnkv/transaction/2pc.go
```diff
@@ -179,6 +179,9 @@ type twoPhaseCommitter struct {
 
 	// The total number of kv request after batch split.
 	prewriteTotalReqNum int
+
+	// assertion error happened when initializing mutations, could be false positive if pessimistic lock is lost
+	assertionErr error
```
MyonKeminta, line 184: It seems that the `twoPhaseCommitter.execute` handles assertion errors from everywhere. How about renaming this to something like `stashedAssertionError`?

disksing:
@MyonKeminta Please take another look.

MyonKeminta (approved):

PR #598, opened 2022-10-11
region_cache: fix bug when reload tiflash_compute store cache

guo-shaoge:
1. fix bug in `reloadTiFlashComputeStores`, should only reload store whose state is up.
2. change all tiflash_mpp to tiflash_compute

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -878,8 +878,8 @@ func (s *RegionRequestSender) getRPCContext(
 		return s.regionCache.GetTiFlashRPCContext(bo, regionID, true)
 	case tikvrpc.TiDB:
 		return &RPCContext{Addr: s.storeAddr}, nil
-	case tikvrpc.TiFlashMPP:
-		rpcCtxs, err := s.regionCache.GetTiFlashMPPRPCContextByConsistentHash(bo, []RegionVerID{regionID})
+	case tikvrpc.TiFlashCompute:
+		rpcCtxs, err := s.regionCache.GetTiFlashComputeRPCContextByConsistentHash(bo, []RegionVerID{regionID})
 		if err != nil {
 			return nil, err
 		}
@@ -1308,8 +1308,8 @@ func (s *RegionRequestSender) onSendFail(bo *retry.Backoffer, ctx *RPCContext, e
 		}
 	}
 
-	if ctx.Store != nil && ctx.Store.storeType == tikvrpc.TiFlashMPP {
-		s.regionCache.InvalidateTiFlashMPPStoresIfGRPCError(err)
+	if ctx.Store != nil && ctx.Store.storeType == tikvrpc.TiFlashCompute {
+		s.regionCache.InvalidateTiFlashComputeStoresIfGRPCError(err)
 	} else if ctx.Meta != nil {
 		if s.replicaSelector != nil {
 			s.replicaSelector.onSendFailure(bo, err)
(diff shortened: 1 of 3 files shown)


disksing:
@guo-shaoge Please fix linter issue. You can see the detail in CI log.

guo-shaoge:
Actually I only changed 3 files, other 10 files are changed by gofmt. Don't know why other 10 files need change.

sticnarf:
> Actually I only changed 3 files, other 10 files are changed by gofmt. Don't know why other 10 files need change. @disksing PTAL~

I guess you're formatting the whole project with Go 1.19.

Go 1.19 has different fmt rules from 1.18. It enforces the format of comments. Let's keep them unchanged for now.

guo-shaoge:
> > Actually I only changed 3 files, other 10 files are changed by gofmt. Don't know why other 10 files need change. @disksing PTAL~
> 
> I guess you're formatting the whole project with Go 1.19.
> 
> Go 1.19 has different fmt rules from 1.18. It enforces the format of comments. Let's keep them unchanged for now.



> > Actually I only changed 3 files, other 10 files are changed by gofmt. Don't know why other 10 files need change. @disksing PTAL~
> 
> I guess you're formatting the whole project with Go 1.19.
> 
> Go 1.19 has different fmt rules from 1.18. It enforces the format of comments. Let's keep them unchanged for now.

You mean I should not change the other 10 files? But the ci will break.

sticnarf:
> You mean I should not change the other 10 files? But the ci will break.

Emm, that's strange. I don't see similar problems in ther PRs.

internal/locate/region_request.go
```diff
@@ -264,8 +264,9 @@ type replicaSelector struct {
 // selectorState is the interface of states of the replicaSelector.
 // Here is the main state transition diagram:
 //
-//                                    exceeding maxReplicaAttempt
-//           +-------------------+   || RPC failure && unreachable && no forwarding
+//	                         exceeding maxReplicaAttempt
+//	+-------------------+   || RPC failure && unreachable && no forwarding
```
sticnarf, line 268: You can see the new formatter breaks the diagrams in the comments. At least this should be reverted.

guo-shaoge: Looks like I only need to format my 3 changed files.

tikvrpc/endpoint.go
```diff
@@ -81,8 +81,8 @@ func GetStoreTypeByMeta(store *metapb.Store) EndpointType {
 		if label.Key == EngineLabelKey && label.Value == EngineLabelTiFlash {
 			return TiFlash
 		}
-		if label.Key == EngineLabelKey && label.Value == EngineLabelTiFlashMPP {
-			return TiFlashMPP
+		if label.Key == EngineLabelKey && label.Value == EngineLabelTiFlashCompute {
+			return TiFlashCompute
```
sticnarf, line 85: Is it backward compatible? What if such a new client connects to an old cluster with `tiflash_mpp` as its engine label?

guo-shaoge: Older version of tiflash's lable is fiexed at compilation time(can only be 'tiflash').
And in new versoin of tiflash(disaggregated tiflash mode), we ensure tiflash lable can only be 'tiflash' or 'tiflash_compute'.

guo-shaoge: e.g. 'tiflash_mpp' is never used.

sticnarf: OK.

sticnarf (approved):

PR #606, opened 2022-10-14
Introduce the FlashbackNotPrepared error

JmPotato:


Defined2014 (approved):

disksing:
It looks there is somethong wrong with integration test. Could you check why it is failing?

Connor1996:
> Signed-off-by: JmPotato [ghzpotato@gmail.com](mailto:ghzpotato@gmail.com)
> 
> Since the flashback state check is before other checks like EpochNotMatch, so we need to invalidate the region cache also.

Why not move the order of the check after `EpochNotMatch`?

JmPotato:
> It looks there is somethong wrong with integration test. Could you check why it is failing?

I think it's the same problem as https://github.com/tikv/client-go/issues/449.

PR #610, opened 2022-10-26
implement the external timestamp client

YangKeao:
Signed-off-by: YangKeao <yangkeao@chunibyo.icu>

This PR also upgrades the integration test to 1.19, as TiDB cannot compile under 1.18.


go.mod
```diff
@@ -58,3 +58,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 	stathat.com/c/consistent v1.0.0 // indirect
 )
+
+replace github.com/tikv/pd/client v0.0.0-20220725055910-7187a7ab72db => github.com/lhy1024/pd/client v0.0.0-20221026120600-7cf30bf5ce7d
```
lhy1024, line 62: plz update kvproto

internal/mockstore/mocktikv/pd.go
```diff
@@ -116,6 +120,32 @@ func (c *pdClient) GetLocalTSAsync(ctx context.Context, dcLocation string) pd.TS
...
+		return err
+	}
+
+	currentTSO := oracle.ComposeTS(p, l)
+	if newTimestamp > currentTSO {
+		return errors.New("external timestamp is greater than global tso")
+	}
+	for {
+		externalTimestamp := c.externalTimestamp.Load()
+		if externalTimestamp > newTimestamp {
```
lhy1024, line 135: we also may skip equal case, which is unnecessary to update.

oracle/oracles/local_external_timestamp.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return err
+	}
+
+	if newTimestamp > currentTSO {
+		return errors.New("external timestamp is greater than global tso")
+	}
+	for {
+		externalTimestamp := l.externalTimestamp.Load()
+		if externalTimestamp > newTimestamp {
```
lhy1024, line 40: ditto

internal/mockstore/mocktikv/pd.go
YangKeao: Fixed.

oracle/oracles/local.go
```diff
@@ -51,6 +51,8 @@ type localOracle struct {
 	hook            *struct {
 		currentTime time.Time
 	}
+
+	localExternalTimestamp
```
lhy1024, line 55: Do we need to set local external ts?

YangKeao: The empty value should be valid :thinking: , it only contains an atomic 0.

lhy1024 (approved):

integration_tests/go.mod
```diff
@@ -69,21 +76,26 @@ require (
...
+	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
 replace github.com/tikv/client-go/v2 => ../
+
+replace (
+	github.com/pingcap/tidb v1.1.0-beta.0.20220902042024-0482b2e83ed2 => github.com/YangKeao/tidb v1.1.0-beta.0.20221027084838-c574366d0e5a
+	github.com/pingcap/tidb/parser v0.0.0-20220724090709-5484002f1963 => github.com/YangKeao/tidb/parser v0.0.0-20221027084838-c574366d0e5a
```
YangKeao, line 100: They will be removed after merging the https://github.com/pingcap/tidb/pull/38689

sticnarf:
The version policy of client-go is supporting the oldest version under official maintanence. I'm afraid we should stick to 1.18 at present.

The uber atomic can be an alternative right now.

YangKeao:
> The version policy of client-go is supporting the oldest version under official maintanence. I'm afraid we should stick to 1.18 at present.
> 
> The uber atomic can be an alternative right now.

Thanks for the information. I could use uber atomic for implementation and unit test. For integration test, I have to upgrade it to 1.19, as TiDB cannot compile under 1.18 now :thinking: .

YangKeao:
> The version policy of client-go is supporting the oldest version under official maintanence. I'm afraid we should stick to 1.18 at present.
> 
> The uber atomic can be an alternative right now.

Fixed. PTAL @sticnarf

sticnarf:
> Thanks for the information. I could use uber atomic for implementation and unit test. For integration test, I have to upgrade it to 1.19, as TiDB cannot compile under 1.18 now  .

Yep, it's okay. Only the main lib needs to compile on 1.18.

PR #615, opened 2022-11-03
tikv: skip 0 when call `getMinSafeTSByStores`

Defined2014:
fix https://github.com/tikv/tikv/issues/13675
TiKV will return `0` for `CmdStoreSafeTS` RPC when there is no region leader on a store. We should ignore this store when caculate the minSafeTS.


sticnarf (approved):

sticnarf:
Do you want to take another look? @JmPotato

JmPotato (approved):

PR #616, opened 2022-11-07
util: update GCTimeFormat to be the same as the value stored in PD

Defined2014:
The GC-related time stored in TiKV is inconsistent with the time stored in PD. The value in TiKV, the time format is "20060102-15:04:05 -0700", and the precision is only seconds. In PD, it stores a uint64 with millisecond precision.

The if we use the value got from TiKV to do some processes, it will got a error like this,

```
mysql> flashback cluster to timestamp '2022-11-07 10:34:00';
ERROR 9006 (HY000): GC life time is shorter than transaction duration, transaction starts at 2022-11-07 10:34:00 +0800 CST, GC safe point is 2022-11-07 10:34:00.366 +0800 CST
```


Defined2014:
@sticnarf PTAL

sticnarf (approved):
LGTM

PR #618, opened 2022-11-17
txn: make `txn_source` be u64 type

xiongjiwei:
we use `u64` to reserve more space for future use. For now, the upper application is limited to setting this value under `0x80`, so there will no more cost to change it to `u64`

Signed-off-by: xiongjiwei <xiongjiwei1996@outlook.com>


sticnarf:
Maybe change the TiKV lock encoding part first?

xiongjiwei:
> Maybe change the TiKV lock encoding part first?

yes, I will do

MyonKeminta (approved):
LGTM. Please merge after merging the corresponding PR in TiKV.

PR #621, opened 2022-11-23
fix: avoid data race when setting memdb footprint hook

ekexium:


sticnarf:
How does it avoid the race? Will the read at `db.allocator.memChangeHook == nil` race with `db.allocator.memChangeHook = innerHook`?

ekexium:
> Will the read at db.allocator.memChangeHook == nil race with db.allocator.memChangeHook = innerHook?

Well I have no idea how it works :(. But the race test passes. Maybe use CAS instead?

internal/unionstore/memdb.go
```diff
@@ -871,8 +872,8 @@ func (db *MemDB) SetMemoryFootprintChangeHook(hook func(uint64)) {
 	innerHook := func() {
 		hook(db.allocator.capacity + db.vlog.capacity)
 	}
-	db.allocator.memChangeHook = innerHook
-	db.vlog.memChangeHook = innerHook
+	atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&db.allocator.memChangeHook)), nil, unsafe.Pointer(&innerHook))
+	atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&db.vlog.memChangeHook)), nil, unsafe.Pointer(&innerHook))
```
sticnarf, line 876: Should we skip the second CAS if the first fails? Then, we won't assign different hooks for allocator and vlog.

ekexium: Sure we can but I think it unnecessary. By design they are always set together.

sticnarf (approved):

PR #622, opened 2022-11-23
precompute min safe ts

hihihuhu:
compute the min safe ts on the query path is expensive because it has to allocate thousands of small objects with lock, when there are hundreds of tikv node. this makes stale read using session variable `tidb_read_staleness` noticeable slower than normal read.

the safe ts from tikv are only updated every 2 seconds, thus the min safe ts could be precomputed on the update path to avoid the expensive computation on query path.

result on our benchmark 
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-74d134e2-7fff-1006-4985-b6a1f5d8d040"><div dir="ltr" style="margin-left:0pt;" align="left">

  | p99 | p999 | Cpu usage
-- | -- | -- | --
Leader read + tidb_low_resolution_tso | 4.0ms | 14.1ms | 413%
Stale read | 10.1ms | 40.3ms | 486%
Stale read with precomputed min safe ts | 5.5ms | 13.6ms | 416%

</div></b>

normal read:
Screen Shot 2022-11-23 at 10 22 26 AM

stale read before the change:
Screen Shot 2022-11-23 at 10 22 45 AM
zoom in:
Screen Shot 2022-11-23 at 10 23 00 AM

stale read after the change:
Screen Shot 2022-11-23 at 10 30 14 AM
zoom in:
Screen Shot 2022-11-23 at 10 31 05 AM

to further reduce the cpu usage, could remove the warning logging from https://github.com/pingcap/tidb/blob/v6.5.0-alpha/expression/builtin_time.go#L6576

### DIFF
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -564,18 +564,10 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 
 	txnScopeMap := make(map[string][]uint64)
 	for _, store := range stores {
-		storeIDs, ok := txnScopeMap[oracle.GlobalTxnScope]
-		if !ok {
-			storeIDs = make([]uint64, 0)
-		}
-		txnScopeMap[oracle.GlobalTxnScope] = append(storeIDs, store.StoreID())
+		txnScopeMap[oracle.GlobalTxnScope] = append(txnScopeMap[oracle.GlobalTxnScope], store.StoreID())
 
 		if label, ok := store.GetLabelValue(DCLabelKey); ok {
-			storeIDs, ok = txnScopeMap[label]
-			if !ok {
-				storeIDs = make([]uint64, 0)
-			}
-			txnScopeMap[label] = append(storeIDs, store.StoreID())
+			txnScopeMap[label] = append(txnScopeMap[label], store.StoreID())
 		}
 	}
 	for txnScope, storeIDs := range txnScopeMap {
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -47,7 +47,6 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/pingcap/kvproto/pkg/kvrpcpb"
-	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pkg/errors"
 	"github.com/tikv/client-go/v2/config"
 	tikverr "github.com/tikv/client-go/v2/error"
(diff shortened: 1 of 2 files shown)


tikv/kv.go
```diff
@@ -571,6 +556,26 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
 		}(ctx, wg, storeID, storeAddr)
 	}
+
+	txnScopeMap := make(map[string][]uint64)
+	for _, store := range stores {
+		storeIDs, ok := txnScopeMap[oracle.GlobalTxnScope]
+		if !ok {
+			storeIDs = make([]uint64, 0)
+		}
+		txnScopeMap[oracle.GlobalTxnScope] = append(storeIDs, store.StoreID())
```
Connor1996, line 566:
```suggestion
		txnScopeMap[oracle.GlobalTxnScope] = append(txnScopeMap[oracle.GlobalTxnScope], store.StoreID())
```

tikv/kv.go
```diff
@@ -571,6 +556,26 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
+
+		if label, ok := store.GetLabelValue(DCLabelKey); ok {
+			storeIDs, ok = txnScopeMap[label]
+			if !ok {
```
Connor1996, line 570: ditto

Connor1996:
PTAL @Yisaer

tikv/kv.go
```diff
@@ -571,6 +556,18 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 			metrics.TiKVMinSafeTSGapSeconds.WithLabelValues(storeIDStr).Set(time.Since(safeTSTime).Seconds())
 		}(ctx, wg, storeID, storeAddr)
 	}
+
+	txnScopeMap := make(map[string][]uint64)
+	for _, store := range stores {
```
Connor1996, line 561: why not keep the original way

hihihuhu: sorry, what do you mean by `original` here?

Connor1996: `if store.IsLabelsMatch([]*metapb.StoreLabel{`

hihihuhu: `IsLabelsMatch` requires a specific value for the label, where here tries to precompute the min safe ts for all the labelled value

Connor1996 (approved):
LGTM

sticnarf (approved):
LGTM

hihihuhu:
the failed test seems related to https://github.com/tikv/tikv/issues/13866, @Connor1996 do you know how often the tikv nightly image is updated?

sticnarf:
Thanks for your contribution!

hihihuhu:
hi @sticnarf, qq: is there a chance that we could port this to upcoming v6.5? if possible, do you know how is the process? thanks!

Connor1996:
Update tikv-client in TiDB. And v6.5 branch has already been check-out, you need to cherry-pick it to v6.5 branch. But not sure if the cherry-pick would be approved as usually only bugfix is allowed to be merged after code freeze。

hihihuhu:
i c, thanks for the info

PR #628, opened 2022-11-30
export region and peer for batch cop

you06:
Signed-off-by: you06 <you1474600@gmail.com>

ref pingcap/tidb#39361

This PR export the region and peer structs for batch cop usage.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -806,10 +806,11 @@ func (c *RegionCache) GetTiFlashComputeRPCContextByConsistentHash(bo *retry.Back
 
 // KeyLocation is the region and range that a key is located.
 type KeyLocation struct {
-	Region   RegionVerID
-	StartKey []byte
-	EndKey   []byte
-	Buckets  *metapb.Buckets
+	Region    RegionVerID
+	StartKey  []byte
+	EndKey    []byte
+	Buckets   *metapb.Buckets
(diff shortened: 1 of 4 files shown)


internal/locate/region_cache.go
```diff
@@ -806,10 +806,11 @@ func (c *RegionCache) GetTiFlashComputeRPCContextByConsistentHash(bo *retry.Back
...
 type KeyLocation struct {
-	Region   RegionVerID
-	StartKey []byte
-	EndKey   []byte
-	Buckets  *metapb.Buckets
+	Region    RegionVerID
+	StartKey  []byte
+	EndKey    []byte
+	Buckets   *metapb.Buckets
+	RawRegion *Region
```
sticnarf, line 813: Now this struct looks very strange. The other fields can be get from `RawRegion`...

you06: That's right, but I don't want to break the compatibility...

sticnarf: Then, it adds something strange that cannot be deleted forever in the future for compatibility...

Now, I think we can just use `GetTiKVRPCContext` for the information we want. We can polish the way in the future.

cfzjywxk: `GetTiKVRPCContext` +1, clock is ticking, better to keep it easy to main

PR #631, opened 2022-12-01
*: fix data race in the metadb

hawkingrei:
ref https://github.com/pingcap/tidb/issues/39558

### DIFF
--- a/internal/unionstore/memdb.go
+++ b/internal/unionstore/memdb.go
@@ -43,6 +43,7 @@ import (
 
 	tikverr "github.com/tikv/client-go/v2/error"
 	"github.com/tikv/client-go/v2/kv"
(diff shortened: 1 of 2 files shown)


disksing (approved):

sticnarf:
The `UpdateFlags` function is protected by the `RWMutex` in `MemDB`. So, I think the race should be fixed by locking in `MemDB.Dirty()`.

I worry that using an atomic here just hides potential logical error.

hawkingrei:
> The `UpdateFlags` function is protected by the `RWMutex` in `MemDB`. So, I think the race should be fixed by locking in `MemDB.Dirty()`.
> 
> I worry that using an atomic here just hides potential logical error.

I agree with you.

PR #632, opened 2022-12-05
fix: do not call onMemChange in alloc

ekexium:
ref: pingcap/tidb#39611
We shall not call a.onMemChange() in `alloc`, since it will make the latest block empty if the callback finally decides to panic, which breaks a precondition  for some operations (e.g. revertToCheckpoint)


sticnarf (approved):
LGTM

ekexium:
@disksing PTAL

PR #636, opened 2022-12-07
don't call onMemChange before all work is done in memdb

ekexium:
ref https://github.com/pingcap/tidb/issues/39611

This is a workaround.
The real problem is that we broke the assumption that some functions should not panic. This is related to how the memory tracking system works, which should be discussed later.


ekexium:
PTAL @sticnarf @disksing

PR #637, opened 2022-12-08
use goroutine pool to avoid the `runtime.newstack` call

### LINKED ISSUE #630: There are room for performance improvement here
image

[profiling_2022-12-01_13-58-16.zip](https://github.com/tikv/client-go/files/10133278/profiling_2022-12-01_13-58-16.zip)

Here can be optimized:
https://github.com/tikv/client-go/blob/5dc09b15e7506f2170ba9031b13cd558ba60321f/internal/client/client_batch.go#L766

```
type batchCommandsEntry struct {
    ...
    res chan *tikvpb.BatchCommandsResponse_Response
}

<-entry.res
```

can be replaced by 

```
type batchCommandsEntry struct {
    ...
    wg sync.WaitGroup
    res *tikvpb.BatchCommandsResponse_Response
}
entry.wg.Wait()
```

The difference is that the later way avoid the channel allocation cost.

A more important optimization is here:

https://github.com/tikv/client-go/blob/5dc09b15e7506f2170ba9031b13cd558ba60321f/txnkv/transaction/2pc.go#L2062-L2079

The stack of the goroutine grows too large, cause `newstack` `copystack` in the flamegraph, which is the dominant cost.
It can be avoided by make a goroutine pool and reuse the goroutine (which has grow already and won't `newstack` `copystack` any more)

tiancaiamao:
close #630

Benchmark program 

```
package main

import (
	"context"
	"database/sql"
	"flag"
	"fmt"
	"log"
	"time"
	"sync/atomic"

	"github.com/jamiealquiza/tachymeter"
	_ "github.com/go-sql-driver/mysql"
)

var (
	host     = flag.String("h", "127.0.0.1", "host")
	port     = flag.Int("P", 4000, "port")
	user     = flag.String("u", "root", "username")
	password = flag.String("p", "", "password")
)

func openDB() (*sql.DB, error) {
	dbDSN := fmt.Sprintf("%s:%s@tcp(%s:%d)/test", *user, *password, *host, *port)
	db, err := sql.Open("mysql", dbDSN)
	if err != nil {
		return nil, err
	}
	return db, nil
}

func mustExec(db *sql.DB, query string, args ...interface{}) {
	_, err := db.Exec(query, args...)
	if err != nil {
		log.Fatal(err)
	}
}

func assert(err error) {
	if err != nil {
		log.Fatal(err)
	}
}	

func main() {
	flag.Parse()
	db, err := openDB()
	if err != nil {
		log.Fatal(err)
	}

	mustExec(db, `use test`)
	mustExec(db, `drop table if exists test;`)
	mustExec(db, `create table test (id int unsigned key nonclustered auto_increment) shard_row_id_bits=4 auto_id_cache 1;`)

	t := tachymeter.New(&tachymeter.Config{Size: 300})
	var count int64

	now := time.Now()
	fmt.Println("benchmark start ...")

	for i := 0; i < 64; i++ {
		go func() {
			conn, err := db.Conn(context.Background())
			assert(err)

			_, err = conn.ExecContext(context.Background(), "use test")
			assert(err)

			for {
				start := time.Now()
				_, err := conn.ExecContext(context.Background(), "insert into test values ()")
				assert(err)
				atomic.AddInt64(&count, 1)
				t.AddTime(time.Since(start))
			}
		}()
	}


	tick := time.NewTicker(5 * time.Second)
	for range tick.C {
		total := atomic.LoadInt64(&count)
		fmt.Println("qps ==", total/5)
		atomic.StoreInt64(&count, 0)

		fmt.Println(t.Calc())
		fmt.Println()
	}


	fmt.Println("benchmark end ...", time.Since(now))
}
```

After this change, the `runtime.newstack` -> `runtime.copystack` disappear.

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/stathat/consistent v1.0.0
 	github.com/stretchr/testify v1.7.0
-	github.com/tiancaiamao/gp v0.0.0-20221208073724-7c59693b66d9
+	github.com/tiancaiamao/gp v0.0.0-20221214071713-abacb15f16f1
 	github.com/tikv/pd/client v0.0.0-20221031025758-80f0d8ca4d07
 	github.com/twmb/murmur3 v1.1.3
 	go.etcd.io/etcd/api/v3 v3.5.2
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/stathat/consistent v1.0.0
 	github.com/stretchr/testify v1.7.0
+	github.com/tiancaiamao/gp v0.0.0-20221208073724-7c59693b66d9
 	github.com/tikv/pd/client v0.0.0-20221031025758-80f0d8ca4d07
 	github.com/twmb/murmur3 v1.1.3
 	go.etcd.io/etcd/api/v3 v3.5.2
@@ -47,7 +48,6 @@ require (
 	github.com/prometheus/common v0.26.0 // indirect
 	github.com/prometheus/procfs v0.6.0 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20200410134404-eec4a21b6bb0 // indirect
-	github.com/tiancaiamao/gp v0.0.0-20221208073724-7c59693b66d9 // indirect
 	go.etcd.io/etcd/client/pkg/v3 v3.5.2 // indirect
 	go.uber.org/multierr v1.7.0 // indirect
 	golang.org/x/net v0.0.0-20210428140749-89ef3d95e781 // indirect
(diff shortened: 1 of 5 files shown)


tiancaiamao:
PTAL @disksing

sticnarf:
@tiancaiamao Would you please add an Apache-compatible license to your project [tiancaiamao/gp](https://github.com/tiancaiamao/gp)?

sticnarf:
You pointed out the problem in #630 that the goroutine started in `startWorker` enlarged the stack. But why using goroutine in `doActionOnGroupMutations` solves the problem in this PR?

tiancaiamao:
> @tiancaiamao Would you please add an Apache-compatible license to your project [tiancaiamao/gp](https://github.com/tiancaiamao/gp)?

Done

tiancaiamao:
> You pointed out the problem in #630 that the goroutine started in `startWorker` enlarged the stack. But why using goroutine in `doActionOnGroupMutations` solves the problem in this PR?

It's `startWorker.func1` which is actually this closure function (closure function is named func1 func2 etc)  @sticnarf

sticnarf (approved):

PR #641, opened 2022-12-14
RuntimeStat: Clone `SnapshotRuntimeStats` Completely

TonsnakeLin:
Signed-off-by: TonsnakeLin <lpbgytong@163.com>
TiDB writes `SnapshotRuntimeStats` to slow log, but we found that the TimeDetail and 'ScanDetail' is lost. It is because that Cloning SnapshotRuntimeStats omits some information.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/stathat/consistent v1.0.0
 	github.com/stretchr/testify v1.7.0
+	github.com/tiancaiamao/gp v0.0.0-20221214071713-abacb15f16f1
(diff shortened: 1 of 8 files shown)


TonsnakeLin:
@sticnarf @ekexium @MyonKeminta @cfzjywxk . thanks for your review !

sticnarf (approved):
LGTM

sticnarf:
Maybe we need to verify the performance impact of cloning the stats.

cfzjywxk:
I think just copying the pointer is enough as the underlying data would not change, the price is it's not real clone.

TonsnakeLin:
@cfzjywxk PTAL

cfzjywxk (approved):

PR #644, opened 2022-12-17
region_cache: refine GetTiFlashComputeRPCContextByConsistentHash

guo-shaoge:
Signed-off-by: guo-shaoge <shaoge1994@163.com>

Move `GetTiFlashComputeStores()` outside of `GetTiFlashComputeRPCContextByConsistentHash()` to help filter alive tiflash_compute nodes.

Related tidb pr: https://github.com/pingcap/tidb/pull/40008

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -879,7 +879,11 @@ func (s *RegionRequestSender) getRPCContext(
 	case tikvrpc.TiDB:
 		return &RPCContext{Addr: s.storeAddr}, nil
 	case tikvrpc.TiFlashCompute:
-		rpcCtxs, err := s.regionCache.GetTiFlashComputeRPCContextByConsistentHash(bo, []RegionVerID{regionID})
+		stores, err := s.regionCache.GetTiFlashComputeStores(bo)
+		if err != nil {
+			return nil, err
+		}
+		rpcCtxs, err := s.regionCache.GetTiFlashComputeRPCContextByConsistentHash(bo, []RegionVerID{regionID}, stores)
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -879,7 +879,11 @@ func (s *RegionRequestSender) getRPCContext(
 	case tikvrpc.TiDB:
 		return &RPCContext{Addr: s.storeAddr}, nil
 	case tikvrpc.TiFlashCompute:
-		rpcCtxs, err := s.regionCache.GetTiFlashComputeRPCContextByConsistentHash(bo, []RegionVerID{regionID})
+		stores, err := s.regionCache.GetTiFlashComputeStores(bo)
+		if err != nil {
+			return nil, err
+		}
+		rpcCtxs, err := s.regionCache.GetTiFlashComputeRPCContextByConsistentHash(bo, []RegionVerID{regionID}, stores)
```
cosven, line 886: How about getting tiflash stores inside the `GetTiFlashComputeRPCContextByConsistentHash` method? In this way, I think we need not to check the store label in `GetTiFlashComputeRPCContextByConsistentHash` anymore :thinking:

cosven (commented):
Does it need to add unit test?

internal/locate/region_request.go
guo-shaoge: We need to filter alive store, then call GetTiFlashComputeRPCContextByConsistentHash(), so, put it outside.

gengliqi (approved):
LGTM

PR #645, opened 2022-12-21
bind the goroutine pool with store and close on Close

tiancaiamao:
Fix tidb CI 

image

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -49,7 +49,6 @@ import (
 	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/tiancaiamao/gp"
 	"github.com/tikv/client-go/v2/config"
 	tikverr "github.com/tikv/client-go/v2/error"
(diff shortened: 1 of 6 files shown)


tiancaiamao:
PTAL @sticnarf @hehechen

txnkv/transaction/2pc.go
```diff
@@ -1014,7 +1011,12 @@ func (c *twoPhaseCommitter) doActionOnGroupMutations(bo *retry.Backoffer, action
 					zap.Error(e))
 				metrics.SecondaryLockCleanupFailureCounterCommit.Inc()
 			}
-		})
+		}
+		if gP, ok := c.store.(interface{ Go(func()) }); ok {
+			gP.Go(f)
+		} else {
+			go f()
+		}
```
sticnarf, line 1019: I can imagine the goroutine pool could be used elsewhere. Then, this snippet may be repeated.

So, do you think it acceptable to make `Go(func())` a function inside of `kvstore`? Indeed, this seems a bit strange considering the interface name is `kvstore`. But `kvstore` is a private interface, so it will have less worse influence.

tiancaiamao: Done

sticnarf (approved):

PR #651, opened 2022-12-29
Fix incorrect insertion behavior in aggressive locking mode and add tests

MyonKeminta:
This PR fixes a bug that when inserting on a key that's locked during the previous aggressive locking stage, it may not check the key's existence correctly. Also added more test cases to cover more cornercases of aggressive locking.

Ref: https://github.com/tikv/tikv/issues/13298


cfzjywxk (approved):
LGTM

MyonKeminta:
@sticnarf @ekexium PTAL when you have time, thank you!

sticnarf (approved):
LGTM

PR #655, opened 2023-01-05
fix data race in the LockKeys

hawkingrei:

ref https://github.com/pingcap/tidb/pull/40350


sticnarf (approved):
LGTM

Issue #659, opened 2023-01-10
client keep print error log after tikv node force scale in


Smityz:
TiKV version: v5.3.0

After using tiup force scale-in one tikv node, client will keep print logs like
```
[2023/01/10 17:41:55.984 +08:00] [INFO] [region_cache.go:2487] ["[health check] check health error"] [store=10.169.48.76:20000] [error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 10.169.48.76:20000: connect: connection refused\""]
[2023/01/10 17:41:56.983 +08:00] [INFO] [region_cache.go:2487] ["[health check] check health error"] [store=10.169.48.76:20000] [error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 10.169.48.76:20000: connect: connection refused\""]
[2023/01/10 17:41:57.983 +08:00] [INFO] [region_cache.go:2487] ["[health check] check health error"] [store=10.169.48.76:20000] [error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 10.169.48.76:20000: connect: connection refused\""]
```

PR #660, opened 2023-01-10
add tombstone check

Smityz:
Signed-off-by: Smityz <smityz@qq.com>

fix https://github.com/tikv/client-go/issues/659

The root cause is `reResolve()` there will return true, even if the store's state is tombstone.
https://github.com/tikv/client-go/blob/master/internal/locate/region_cache.go#L2442

The existing unit test assumes if pd client gets a tombstone store, it will return nil, but in the force scale-in scenario, it won't.
https://github.com/tikv/client-go/blob/master/internal/mockstore/mocktikv/pd.go#L202-L208

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -2231,7 +2231,7 @@ func (s *Store) initResolve(bo *retry.Backoffer, c *RegionCache) (addr string, e
 			continue
 		}
 		// The store is a tombstone.
-		if store == nil {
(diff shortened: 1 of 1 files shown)


disksing:
hi @Smityz thanks for the fix, but our community requires DCO for commits. (for more details about DCO  https://www.secondstate.io/articles/dco/)

Smityz:
> hi @Smityz thanks for the fix, but our community requires DCO for commits. (for more details about DCO https://www.secondstate.io/articles/dco/)

done for it

Issue #532, opened 2022-06-22
performance of searchCachedRegion could be enhanced


chrysan:
When many regions are cached in memory, searchCachedRegion becomes slower and holds RegionCache global read lock for longer time, and then makes other queries who load new regions wait for write lock. When QPS grows, the mutex contention becomes even worse and query latency grows.

![image](https://user-images.githubusercontent.com/5151781/175061823-b029747f-a51c-4fef-95a1-791997943c09.png)

findRegionByKey waits for write lock:
```
goroutine 8144205746 [semacquire]:goroutine 8144205746 [semacquire]:sync.runtime_SemacquireMutex(0xc0003ae014, 0x0, 0x1) /usr/local/go/src/runtime/sema.go:71 +0x47sync.(*Mutex).lockSlow(0xc0003ae010) /usr/local/go/src/sync/mutex.go:138 +0xfcsync.(*Mutex).Lock(...) /usr/local/go/src/sync/mutex.go:81sync.(*RWMutex).Lock(0xc0003ae010) /usr/local/go/src/sync/rwmutex.go:98 +0x97github.com/pingcap/tidb/store/tikv.(*RegionCache).findRegionByKey(0xc0003ae000, 0xc4040d51a8, 0xc0eb43fce0, 0x13, 0x13, 0xc031612e00, 0x7fcaad8be1f0, 0x0, 0x40) /home/jenkins/agent/workspace/optimization-build-tidb-linux-amd/go/src/github.com/pingcap/tidb/store/tikv/region_cache.go:582 +0x6c2
```
searchCachedRegion holds read lock:
```
[...]

PR #664, opened 2023-01-11
Add a background region cache GC goroutine

sticnarf:
Refers to https://github.com/pingcap/tidb/issues/40461.

https://github.com/tikv/client-go/pull/566 can solve high CPU cost problems like https://github.com/tikv/client-go/issues/532. But in some cases, there will still be memory leaks when some of the stale regions are never touched again.

This PR adds a background goroutine to iterate over the cache to clean up stale region info. We don't want this goroutine to affect performance, so it runs at low frequency and each round is very fast.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1912,8 +1912,8 @@ func (c *RegionCache) UpdateBucketsIfNeeded(regionID RegionVerID, latestBucketsV
 	}
 }
 
-const cleanCacheInterval = 100 * time.Millisecond
-const cleanRegionNumPerRound = 10
+const cleanCacheInterval = time.Second
+const cleanRegionNumPerRound = 50
 
 // This function is expected to run in a background goroutine.
 // It keeps iterating over the whole region cache, searching for stale region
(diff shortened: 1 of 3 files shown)


internal/locate/region_cache.go
```diff
@@ -1906,6 +1912,56 @@ func (c *RegionCache) UpdateBucketsIfNeeded(regionID RegionVerID, latestBucketsV
 	}
 }
 
+const cleanCacheInterval = 100 * time.Millisecond
```
cfzjywxk, line 1915: This seems to be a bit frequent for recycling memory, how about several seconds with a larger `cleanRegionNumPerRound`?

ekexium (approved):
LGTM. Have we met such cases?

internal/locate/region_cache.go
sticnarf: It requires a lock to access the region cache. I hope the time that it holds the lock is short enough. So, I am very cautious about increasing `cleanRegionNumPerRound`.

I just tried `cleanRegionNumPerRound = 50`, the time of each round can be controlled under 20us. (But this does not involve removal). Maybe it's acceptable to change the value to 50 regions every 1 second.

sticnarf:
> LGTM. Have we met such cases?

Probably. Some customer finds memory leak but the version they use does not include #566 yet.

In their case, they drop `day -14` to `day -7` tables every week. The regions of the removed tables may be merged into the `day -6` table. However, it's possible that the `day -6` table will not be touched at all.

cfzjywxk (approved):
LGTM

Issue #522, opened 2022-06-08
Rawkv can not recovery after one of the tikv-servers restart


pingyu:
### Version
master, https://github.com/tikv/client-go/commit/2807409d4968a3dd7b49ac073eadeaf3e6bffc8f

##### RawKV cluster deploy by TiUP, v6.1.0-alpha
```
tikv-server --version
TiKV
Release Version:   6.1.0-alpha
Edition:           Community
Git Commit Hash:   c26134bbf04235f61484d4f5ed58c47b520e8015
Git Commit Branch: heads/refs/tags/v6.1.0
UTC Build Time:    2022-06-07 11:55:22
Rust Version:      rustc 1.60.0-nightly (1e12aef3f 2022-02-13)
Enable Features:   jemalloc mem-profiling portable sse test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure
Profile:           dist_release
```

##### Cluster setup
4 x TiKV + 3 x PD (deployed together with TiKV), 4 x 32C64GB + 500GB cloud SSD. Kingsoft cloud.

##### Cluster config
```
  tikv:
    log.level: "info"
    log.file.max-size: 1024
    log.file.max-backups: 30
    storage.api-version: 1
    storage.enable-ttl: false
```

### Steps to reproduce
1. Run the test code: [rawkv_ha.go](https://github.com/pingyu/client-go/blob/rawkv-ha/examples/rawkv/rawkv_ha.go)
2. Kill one of the tikv-server (10.2.103.99) by kill -9. Then it will be restarted by systemctl.

##### 10.2.103.99 is recovered in less than 30 seconds.
<img width="593" alt="image" src="https://user-images.githubusercontent.com/1907938/172564595-71c2da7c-db44-414c-bf54-7172988a19ee.png">

##### But no request is sent to it for minutes.
<img width="1181" alt="image" src="https://user-images.githubusercontent.com/1907938/172564928-289b57fa-b7c4-47c8-a3ef-2821693ca675.png">

##### Even there are external retires ([rawkv_ha.go](https://github.com/pingyu/client-go/blob/7b22bfe7ae2cca54970bdc77d6072bb761eac48f/examples/rawkv/rawkv_ha.go#L83))
```go
func workload(ctx context.Context, cli *rawkv.Client, idx int64) (err error) {
	...
	withRetry := func(f func() error) (err error) {
	Loop:
		for i := 0; i < *flagRetryCnt; i++ {
[...]

Issue #653, opened 2023-01-03
Improve monitoring for gRPC connections in client


Smityz:
We would like to be able to use monitoring metrics to determine which network connections between endpoints are experiencing issues, in order to better troubleshoot. For example, we would like to be able to use monitoring metrics to identify if certain gRPC connections are more prone to dropping at certain times, or if there are recurring network failures between certain endpoints.

PR #665, opened 2023-01-11
Add gRPC connection monitor and fix related bug

Smityz:
Signed-off-by: Smityz [smityz@qq.com](mailto:smityz@qq.com)

fix: https://github.com/tikv/client-go/issues/522 https://github.com/tikv/client-go/issues/653

After a tikv server restart, the grpc connection state on the client becomes IDLE and needs a request to be sent through this connection or a manual connect to change it to the READY state. In the current implementation, I believe trying to`Connect()` when the state is IDLE is a relatively small modification.

And add a monitor of grpc connection state can help us to find related bugs

Screenshot 2023-01-11 at 18 11 38


disksing:
@hicqu @MyonKeminta Please take a look.

internal/client/client.go
```diff
@@ -114,30 +114,119 @@ type connArray struct {
...
+			go c.start()
+		},
+	)
+}
+
+func (c *connMonitor) Stop() {
+	c.stopOnce.Do(
+		func() {
+			if c.stop != nil {
+				close(c.stop)
```
hicqu, line 175: Should set stop to nil here?

Smityz: `stop` is a chan here, why should set it to nil?

Smityz: `stopOnce` can  promise

disksing:
@hicqu any more comments?

hicqu (approved):

PR #671, opened 2023-01-17
ReadMode: introduce AutoFollowerRead mechanism by adding a new `ReadMode == PreferLeader`

### LINKED ISSUE #670: ReadMode: support filtering out slow nodes to improve availabilities on Read workloads
## Development description
This feature is used to improve the availability on Read workloads when there exists slow nodes in the TiKV cluster.
In current stage, in a single AZ, the Read availability will be damaged when there exists slow TiKV nodes whose processing progress is delayed or hung by CPU / Disk IO, even though we have optional "ReplicaReadMode == Leader || Follower || LeaderAndFollower" to redirect Read flows to other nodes. And we do not have an accessible way to recognize slow nodes, making Read flows not to be redirected to these nodes.

So, I try to introduce a simple way / algorithm to detect slow nodes in TiKV cluster(just for TiKV nodes, currently not compatible to TiFlash nodes). And when there exists slow nodes, Client can perceive these slow nodes and do not send Read flows to these nodes to try it best to keep the high availabilities on Read requests. I just call it `AutoFollowerRead`.
Meanwhile, given that there already exists on Read mode -- `ReplicaReadMode == LeaderAndFollower` with Round-Robin algorithm to balance Read flows, I just wanna introduce `AutoFollowerRead` into a new read mode `PreferLeader` for convenient. That is, `AutoFollowerRead` is introduced as an enhancement on `ReplicaReadMode == PreferLeader`.

LykxSassinator:
## Description

Introduce AutoFollowerRead mechanism to improve the Read availability on Read workloads when setting with `ReadMode == PreferLeader`. Here, a new read mode `PreferLeader` has been introduced for enabling this mechanism.

Close #670.

Signed-off-by: lucasliang <nkcs_lykx@hotmail.com>


LykxSassinator:
@tonyxuqqi PTAL,thx

disksing:
I wonder If the load of different stores is different (for example, some stores are full of small requests, while others are scanned extensively), then the method of determining store slow may not be accurate.

LykxSassinator:
> I wonder If the load of different stores is different (for example, some stores are all point get requests, while others are full table scans), then the method of determining store slow may not be accurate.

Thx for your comments. 
What u've pointed about the influence of different workloads has been considered in the design of the detection algorithm.
### Design in brief introduction
One store in TiKV clusters may be slow because of network delays or IO delays on disk. Currently, we cannot have a very concise method to aggregate all informations into the Client and make the determination on which store is slow.
So, we use a measurement, so-called **Slow Score**, to represent the severity of slow on each store. And only when the **Slow Score** exceed the threshold(80 in current implementation), will the corresponding store be marked as **SLOW**. Then, the read flows will be redirected to other nodes.

And about the calculation of **Slow Score**, we introduce two dimensions to determine it on each store:
* **Requests** in one timing tick. This dimension can be regarded as QPS on each store.
* **Average timecost** on each request in one timing tick. This dimension is used to detect whether the relative store is busy on processing requests.

And as we known, if one store is slow, its **Requests** will keep decreasing gradually, but **Average timecost** will keep ascending. 
image

So, we just introduce an appropriate algorithm(just like Sliding window) to combine these two factors and convert them to **Slow Score**, for representing the business of each node. That is:
* **Slow Score** focus the severity of slowness on each store, and each store has its own counting statistics for calculating **Slow Score**.
* **Slow Score** of one store will gradually ascend if this store is gradually slow on processing requests. And this store will be marked as **SLOW** if **Slow Score** exceed the threshold.

Of course, about the point on:
> store slow may not be accurate.

**Slow Score** can be more accurate if we make the statistics divided by the type of commands, which can be reviewed in later generation of this algorithm. We just wanna introduce an appropriate and simple way, for less costs of resources, to optimize the `replica_read_mode == leader_and_follower` mode.

disksing:
@LykxSassinator Thank you for the clarification.

internal/locate/region_cache.go
```diff
@@ -233,8 +234,8 @@ func (r *regionStore) kvPeer(seed uint32, op *storeSelectorOp) AccessIndex {
 
 func (r *regionStore) filterStoreCandidate(aidx AccessIndex, op *storeSelectorOp) bool {
 	_, s := r.accessStore(tiKVOnly, aidx)
-	// filter label unmatched store
-	return s.IsLabelsMatch(op.labels)
+	// filter label unmatched store and slow stores
+	return s.IsLabelsMatch(op.labels) && (!op.preferLeader || (op.preferLeader && aidx == r.workTiKVIdx && !s.isSlow()))
```
tonyxuqqi, line 238:
```suggestion
	return s.IsLabelsMatch(op.labels) && (!op.preferLeader || (aidx == r.workTiKVIdx && !s.isSlow()))
```

internal/locate/region_cache.go
```diff
@@ -2629,6 +2649,195 @@ func invokeKVStatusAPI(addr string, timeout time.Duration) (l livenessState) {
 	return
 }
 
+const (
+	slowScoreInitVal     = 1
+	slowScoreThreshold   = 80
+	slowScoreMax         = 100
+	slowScoreInitTimeout = 500000   // unit: us
+	slowScoreMaxTimeout  = 30000000 // max timeout of one txn, unit: us
```
tonyxuqqi, line 2657:
```suggestion
        slowScoreInitTimeoutInUs = 500000 // unit: us
	slowScoreMaxTimeoutInUs  = 30000000 // max timeout of one txn, unit: us
```
Or use duration instead.

internal/locate/region_cache.go
```diff
@@ -2629,6 +2649,195 @@ func invokeKVStatusAPI(addr string, timeout time.Duration) (l livenessState) {
...
+	slidingWindowSize    = 10       // default size of sliding window
+)
+
+// CntSlidingWindow represents the statistics on a bunch of sliding windows.
+type CntSlidingWindow struct {
```
tonyxuqqi, line 2662: i think slow score related code can be a separate file.

internal/locate/region_cache.go
```diff
@@ -2629,6 +2649,195 @@ func invokeKVStatusAPI(addr string, timeout time.Duration) (l livenessState) {
...
+	tsCntSlidingWindow  CntSlidingWindow // sliding window on timecost
+	updCntSlidingWindow CntSlidingWindow // sliding window on update count
+}
+
+func (ss *SlowScoreStat) getSlowScore() uint64 {
+	return atomic.LoadUint64(&ss.avgScore)
+}
+
+// Update the statistics on SlowScore periodically.
+func (ss *SlowScoreStat) updateSlowScore() bool {
```
tonyxuqqi, line 2711: please add the comment for the algorithm

internal/locate/region_cache.go
tonyxuqqi: Is this for all read queries? For large scan query,  30s may be the legitimate latency.

internal/locate/region_cache.go
```diff
@@ -2629,6 +2649,195 @@ func invokeKVStatusAPI(addr string, timeout time.Duration) (l livenessState) {
...
+	atomic.CompareAndSwapUint64(&ss.avgTimecost, avgTimecost, ss.tsCntSlidingWindow.Avg())
+
+	// Resets the counter of inteval timecost
+	atomic.StoreUint64(&ss.intervalTimecost, 0)
+	atomic.StoreUint64(&ss.intervalUpdCount, 0)
+	return true
+}
+
+// recordSlowScoreStat records the timecost of each request.
+func (ss *SlowScoreStat) recordSlowScoreStat(timecost time.Duration) bool {
```
tonyxuqqi, line 2753: add the comment for return value's meaning.

internal/locate/region_cache.go
```diff
@@ -2629,6 +2649,195 @@ func invokeKVStatusAPI(addr string, timeout time.Duration) (l livenessState) {
...
+func (s *Store) recordSlowScoreStat(timecost time.Duration) bool {
+	return s.slowScore.recordSlowScoreStat(timecost)
+}
+
+func (s *Store) markAlreadySlow() {
+	s.slowScore.markAlreadySlow()
+}
+
+// asyncUpdateStoreSlowScore updates the slow score of each store periodically.
+func (c *RegionCache) asyncUpdateStoreSlowScore(interval time.Duration) {
```
tonyxuqqi, line 2806: Ideally this asyncUpdateStoreSlowScore should not be called when the mode is not preferLeader. Or at least the expensive part of this function should not run.

internal/locate/region_request.go
```diff
@@ -592,7 +594,10 @@ func (state *accessFollower) isCandidate(idx AccessIndex, replica *replica) bool
 		// The request can only be sent to the leader.
 		((state.option.leaderOnly && idx == state.leaderIdx) ||
 			// Choose a replica with matched labels.
-			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && replica.store.IsLabelsMatch(state.option.labels) && (!state.learnerOnly || replica.peer.Role == metapb.PeerRole_Learner)))
+			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && replica.store.IsLabelsMatch(state.option.labels) && (!state.learnerOnly || replica.peer.Role == metapb.PeerRole_Learner)) &&
+				// And If the leader store is abnormal to be accessed under `ReplicaReadPreferLeader` mode, we should choose other valid followers
+				// as candidates to serve the Read request.
+				(!state.option.preferLeader || !replica.store.isSlow()))
```
tonyxuqqi, line 600: it's better to to make the `(!state.option.preferLeader || !replica.store.isSlow()))` close to other condition check. Add the comment below is fine.

internal/locate/region_request.go
```diff
@@ -644,8 +650,15 @@ func newReplicaSelector(regionCache *RegionCache, regionID RegionVerID, req *tik
 		for _, op := range opts {
 			op(&option)
 		}
+		if req.ReplicaReadType == kv.ReplicaReadPreferLeader {
+			WithPerferLeader()(&option)
+		}
+		tryLeader := false
+		if req.ReplicaReadType == kv.ReplicaReadMixed || req.ReplicaReadType == kv.ReplicaReadPreferLeader {
+			tryLeader = true
```
tonyxuqqi, line 658:
```suggestion
			tryLeader := req.ReplicaReadType == kv.ReplicaReadMixed || req.ReplicaReadType == kv.ReplicaReadPreferLeader
```

internal/locate/region_request.go
```diff
@@ -1202,6 +1216,10 @@ func (s *RegionRequestSender) sendReqToRegion(bo *retry.Backoffer, rpcCtx *RPCCo
 	if !injectFailOnSend {
 		start := time.Now()
 		resp, err = s.client.SendRequest(ctx, sendToAddr, req, timeout)
+		// Record timecost of this request into the related Store.
+		if !util.IsInternalRequest(req.RequestSource) {
+			rpcCtx.Store.recordSlowScoreStat(time.Since(start))
```
tonyxuqqi, line 1221: is this called when preferleader mode is off?

internal/locate/region_cache.go
LykxSassinator: > legitimate

Yes, 30s is the largest and tolerable TIMEOUT setting used in Client by default.

internal/locate/region_cache.go
LykxSassinator: No recommend. 
`preferLeader` is a ReplicaReadMode for request-level or session-level setting. If and only if we have the global statistics of slowScores on all TiKV nodes by `asyncUpdateStoreSlowScore`, can we make the judgement whether a leader replica is slow or not and do following redirecting operations.
That is, we need the background statistics by the co-routine `asyncUpdateStoreSlowScore`.

internal/locate/region_request.go
LykxSassinator: Yes, this condition is just filtering out unnecessary internal requests sent to TiKV node by ticking, making Client focus on the timecost of external requests, triggered by users.

internal/locate/region_request.go
LykxSassinator: Let's not, it's recommended to mark this judgement in a newline here and just with the above comments for clarifying this condition will make sense when `preferLeader` is on.

tonyxuqqi (approved):

LykxSassinator:
cc @disksing PTAL, thx.

PR #675, opened 2023-01-20
Support load-based replica read

sticnarf:
Refers to https://github.com/tikv/rfcs/pull/105.

Corresponding TiDB PR: https://github.com/pingcap/tidb/pull/40742


txnkv/txnsnapshot/scan.go
```diff
@@ -233,6 +233,7 @@ func (s *Scanner) getData(bo *retry.Backoffer) error {
 				ResourceGroupTag:  s.snapshot.mu.resourceGroupTag,
 				RequestSource:     s.snapshot.GetRequestSource(),
 				ResourceGroupName: s.snapshot.mu.resourceGroupName,
+				BusyThresholdMs:   uint32(s.snapshot.mu.busyThreshold.Milliseconds()),
```
ekexium, line 236: I don't understand why it is not protected by the RWLock?

internal/locate/region_request.go
```diff
@@ -259,13 +260,17 @@ type replicaSelector struct {
...
 }
 
 // selectorState is the interface of states of the replicaSelector.
 // Here is the main state transition diagram:
 //
-//                                    exceeding maxReplicaAttempt
-//           +-------------------+   || RPC failure && unreachable && no forwarding
+//	                         exceeding maxReplicaAttempt
+//	+-------------------+   || RPC failure && unreachable && no forwarding
+//
```
ekexium, line 273: Unexpected formats?

txnkv/txnsnapshot/scan.go
sticnarf: Sorry, I think the code before was actually incorrect. `Context` should not be set here. It should be set below in L243 instead.

internal/locate/region_request.go
```diff
@@ -1158,6 +1265,7 @@ func (s *RegionRequestSender) sendReqToRegion(bo *retry.Backoffer, rpcCtx *RPCCo
 	if e := tikvrpc.SetContext(req, rpcCtx.Meta, rpcCtx.Peer); e != nil {
 		return nil, false, err
 	}
+	rpcCtx.overrides.patchContext(&req.Context)
```
ekexium, line 1268: Just my personal preference:
If this is written as something like `rpcCtx.contextPatcher.applyTo(&req.Context)` then it would be easier to understand without checking out what the function does.

internal/locate/region_request.go
sticnarf: I don't find how to let `go fmt` ignore it. I add a bar at the line beginning to work around it.

internal/locate/region_request.go
sticnarf: OK, I change the type name.

internal/locate/region_request.go
you06: You can leave an empty line after the graph and before the following code.

internal/locate/region_request.go
```diff
@@ -1548,13 +1655,10 @@ func (s *RegionRequestSender) onRegionError(bo *retry.Backoffer, ctx *RPCContext
 		return retry, err
 	}
 
-	if regionErr.GetServerIsBusy() != nil {
-		// Mark the server is busy (the next incoming READs could be redirect
-		// to expected followers. )
-		if ctx != nil && ctx.Store != nil {
-			ctx.Store.markAlreadySlow()
+	if serverIsBusy := regionErr.GetServerIsBusy(); serverIsBusy != nil {
+		if s.replicaSelector != nil {
```
cfzjywxk, line 1659: If the `replicaSelector` is nil would the `markAlreadySlow` be skipped unexpectedly?

sticnarf: Actually, this is impossible for TiKV requests. `RegionRequestSender` always creates a new `replicaSelector` if not exists.

cfzjywxk (approved):
LGTM

you06 (approved):

you06:
A question about `ServerIsBusy.AppliedIndex`, is the enhancement of sending it to the follower replicas the further plan in your design?

sticnarf:
> A question about `ServerIsBusy.AppliedIndex`, is the enhancement of sending it to the follower replicas the further plan in your design?

Yes, it's a future plan.

It's not implemented it in TiKV now because I haven't found a good way to get it efficiently. The gRPC threads don't have access to the engine `RaftKv` while the threads with `RaftKv` may be too busy at the time.

Maybe it's possible to move local reader to gRPC threads, but there do have a lot to consider.

PR #683, opened 2023-02-01
update go version to 1.19 and config support both path and names  

HuSharp:
Signed-off-by: husharp <jinhao.hu@pingcap.com>
Part of https://github.com/tikv/pd/pull/5886. We need to update client-go first to update TiDB. Should be merged after https://github.com/tikv/pd/pull/5886
And for integration test, need to change TiDB temporarily
After https://github.com/pingcap/tidb/pull/40955 merged will update TiDB version which is in integration_test.


HuSharp:
Need to update go-version to 1.19
66beeecb-da53-4110-be2b-8b80895ac850

disksing (approved):

PR #684, opened 2023-02-02
Consider the number of replicas when calculating RU

JmPotato:

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1183,6 +1183,9 @@ func (s *RegionRequestSender) sendReqToRegion(bo *retry.Backoffer, rpcCtx *RPCCo
 		sendToAddr = rpcCtx.ProxyAddr
 	}
 
+	// TODO: include TiFlash replica(s) in the future.
+	req.ReplicaNumber = int64(rpcCtx.TiKVNum)
+
 	var sessionID uint64
 	if v := bo.GetCtx().Value(util.SessionID); v != nil {
 		sessionID = v.(uint64)
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1183,7 +1183,6 @@ func (s *RegionRequestSender) sendReqToRegion(bo *retry.Backoffer, rpcCtx *RPCCo
 		sendToAddr = rpcCtx.ProxyAddr
 	}
 
-	// TODO: include TiFlash replica(s) in the future.
 	req.ReplicaNumber = int64(rpcCtx.TiKVNum)
 
 	var sessionID uint64
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1190,7 +1190,7 @@ func (s *RegionRequestSender) sendReqToRegion(bo *retry.Backoffer, rpcCtx *RPCCo
 		for _, peer := range rpcCtx.Meta.GetPeers() {
 			role := peer.GetRole()
 			if role == metapb.PeerRole_Voter || role == metapb.PeerRole_Learner {
-				req.ReplicaNumber += 1
+				req.ReplicaNumber++
 			}
 		}
 	}
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1183,7 +1183,17 @@ func (s *RegionRequestSender) sendReqToRegion(bo *retry.Backoffer, rpcCtx *RPCCo
 		sendToAddr = rpcCtx.ProxyAddr
 	}
 
-	req.ReplicaNumber = int64(rpcCtx.TiKVNum)
+	// Count the replica number as the RU cost factor.
+	req.ReplicaNumber = 1
+	if rpcCtx.Meta != nil && len(rpcCtx.Meta.GetPeers()) > 0 {
+		req.ReplicaNumber = 0
+		for _, peer := range rpcCtx.Meta.GetPeers() {
+			role := peer.GetRole()
+			if role == metapb.PeerRole_Voter || role == metapb.PeerRole_Learner {
+				req.ReplicaNumber += 1
+			}
+		}
+	}
 
 	var sessionID uint64
 	if v := bo.GetCtx().Value(util.SessionID); v != nil {
--- a/internal/resourcecontrol/resource_control.go
+++ b/internal/resourcecontrol/resource_control.go
@@ -55,7 +55,7 @@ func MakeRequestInfo(req *tikvrpc.Request) *RequestInfo {
 		}
 	}
 
-	return &RequestInfo{writeBytes: writeBytes}
+	return &RequestInfo{writeBytes: writeBytes * req.ReplicaNumber}
 }
 
 // IsWrite returns whether the request is a write request.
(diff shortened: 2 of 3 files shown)


sticnarf:
Could you give some context about this?

Multiplying it by replica num simplifies the problem too much. As you commented in TODO, TiFlash nodes are not considered. And, when there is witness, the write cost in witness and non-witness nodes are different.

Why not just do resource control according to front end flow bytes?

nolouch:
> 
> Why not just do resource control according to front end flow bytes?

ptal @SunRunAway

internal/resourcecontrol/resource_control.go
```diff
@@ -48,7 +48,7 @@ func MakeRequestInfo(req *tikvrpc.Request) *RequestInfo {
 		writeBytes += int64(unsafe.Sizeof(r.Keys))
 	}
 
-	return &RequestInfo{writeBytes: writeBytes}
+	return &RequestInfo{writeBytes: writeBytes * req.ReplicaNumber}
```
BornChanger, line 51: It makes things complicated. How about the deployment of cluster with witness node?

BornChanger: On the other hand, customer already pay for the multiple storage replicas.

SunRunAway:
Please use the following logic to calculate Write Unit for serverless metering:
```
ReplicaNumber:=3
if table has tiflash {
    ReplicaNumber = 4
}
```
In the Serverless metering scenario, we currently fix the number of tikv copies to tell the user is 3, so as long as replica should be fixed to 3 regardless of whether the deployment structure has witness or not.

If a table has tiflash copies, replica should be considered to be 4, also a fixed number.

@sticnarf @BornChanger @JmPotato @nolouch

JmPotato:
> Please use the following logic to calculate Write Unit for serverless metering:

I think the point here is that Resource Control doesn't really need to care about the replica number because it's used for throttling. If we really need to count the replica number for Serverless metering, why not count it in the billing since it's a fixed number for now?

Also, hard-coding this number here doesn't sound like a good idea to me.

SunRunAway:
> why not count it in the billing since it's fixed for now
I don't care where the logic is.

JmPotato:
> I don't care where the logic is in. As long as the metering logic is implemented

Well, then it makes sense to me. We can implement the replica counting somewhere else later, maybe inside the Serverless platform. WDYT? @sticnarf @nolouch

nolouch:
I prefer the ru convert in one place.  and we should make sure how we report to the billing, the replicas is only for write part.

nolouch:
and IMO,  billing need understand the cluster config is unreasonable, the tiflash may will be a elastic pool in the serverless.

nolouch:
I think there is no need to distinguish witness, tiflash and others, they all need to perform log replication. The total `replica num` is equivalent to the write traffic amplification factor.

sticnarf:
> I think there is no need to distinguish witness, tiflash and others, they all need to perform log replication. The total `replica num` is equivalent to the write traffic amplification factor.

The CPU/IO cost of RocksDB compaction is greater than replication itself by _orders of magnitude_. That's why I have the question of not distinguishing them.

sticnarf:
The ideal solution IMO is to let TiKV return the estimated RU cost because the server tends to have more information about how many resources a request costs. If this has difficulty, I am happy with a simple way like this now.

nolouch:
> > I think there is no need to distinguish witness, tiflash and others, they all need to perform log replication. The total `replica num` is equivalent to the write traffic amplification factor.
> 
> The CPU/IO cost of RocksDB compaction is greater than replication itself by _orders of magnitude_. That's why I have the question of not distinguishing them.

Yes. the compaction can be seen as the background flow, it's not easy to bind the resource usage to a request, we may have other ways to control it(there are already some), here we mainly consider the front write flow, and replication can be seen as the front.  moreover, in the cloud, we can offload the compaction.

JmPotato:
> The ideal solution IMO is to let TiKV estimate RU cost because the server tends to have more information about how many resources a request costs. If this has difficulty, I am happy with a simple way like this now.

Let's use this simple way for now. I updated the way to count the replica, PTAL @nolouch @sticnarf

sticnarf (commented):
It seems to record write bytes for failed requests, should we do that? (I think it's not reasonable to record this at least on region errors, which is unlikely to be triggered by the user.) And please fix fint.

JmPotato:
> It seems to record write bytes for failed requests, should we do that? (I think it's not reasonable to record this at least on region errors, which is unlikely to be triggered by the user.) And please fix fint.

Nice catch. I think we need to pay back this part cost when it fails. I will file another PR to do this.

sticnarf (approved):

sticnarf:
cc @BornChanger

nolouch (approved):

BornChanger (approved):

internal/locate/region_request.go
```diff
@@ -1183,6 +1183,18 @@ func (s *RegionRequestSender) sendReqToRegion(bo *retry.Backoffer, rpcCtx *RPCCo
 		sendToAddr = rpcCtx.ProxyAddr
 	}
 
+	// Count the replica number as the RU cost factor.
+	req.ReplicaNumber = 1
```
BornChanger, line 1187: I think we need to optimize here. The write node number should be calculated for once instead of calculating every time (Though some node role might change). The overhead is too much...

BornChanger (approved):

internal/locate/region_request.go
sticnarf: The loop is simple enough, so it seems not big overhead to me. A benchmark should tell.

JmPotato: Before:

```shell
BenchmarkSendReqToRegion/replicaNum_3-8         	  440764	      2661 ns/op	    2075 B/op	      49 allocs/op
BenchmarkSendReqToRegion/replicaNum_5-8         	  416978	      2887 ns/op	    2396 B/op	      54 allocs/op
BenchmarkSendReqToRegion/replicaNum_50-8        	  180320	      6511 ns/op	    8211 B/op	     147 allocs/op
BenchmarkSendReqToRegion/replicaNum_100-8       	  108415	     10799 ns/op	   14886 B/op	     248 allocs/op
BenchmarkSendReqToRegion/replicaNum_1000-8      	   15418	     78474 ns/op	  139241 B/op	    2053 allocs/op
```

After:

```shell
BenchmarkSendReqToRegion/replicaNum_3-8         	  438793	      2678 ns/op	    2072 B/op	      49 allocs/op
BenchmarkSendReqToRegion/replicaNum_5-8         	  416216	      2836 ns/op	    2396 B/op	      54 allocs/op
BenchmarkSendReqToRegion/replicaNum_50-8        	  175693	      6606 ns/op	    8219 B/op	     147 allocs/op
BenchmarkSendReqToRegion/replicaNum_100-8       	   96657	     10785 ns/op	   14921 B/op	     248 allocs/op
BenchmarkSendReqToRegion/replicaNum_1000-8      	   15133	     79340 ns/op	  139246 B/op	    2053 allocs/op
```

BornChanger: I am fine with it now.

PR #687, opened 2023-02-03
Add metrics and statistics about aggressive locking

MyonKeminta:
Ref: https://github.com/tikv/tikv/issues/13298


metrics/metrics.go
```diff
@@ -97,6 +97,7 @@ var (
 	TiKVUnsafeDestroyRangeFailuresCounterVec *prometheus.CounterVec
 	TiKVPrewriteAssertionUsageCounter        *prometheus.CounterVec
 	TiKVGrpcConnectionState                  *prometheus.GaugeVec
+	TiKVAggressiveLockedKeysCounter          *prometheus.CounterVec
```
cfzjywxk, line 100: Do we also need the metric that we could calculate:
- The `aggressive locking keys /  The total locking keys`, so we know how many keys are affected by the locking mode.
- The `aggressive locked transactions/ The total pessimistic transactions`, so we know the percentage affected from transaction perspective.

txnkv/transaction/txn.go
```diff
@@ -1124,13 +1132,32 @@ func (txn *KVTxn) lockKeys(ctx context.Context, lockCtx *tikv.LockCtx, fn func()
...
 				continue
 			}
 			memBuf.UpdateFlags(key, tikv.SetKeyLocked, tikv.DelNeedCheckExists, setValExists)
 		}
 	}
-	txn.lockedCnt += len(keys) - skipedLockKeys
+
+	// Update statistics information.
+	txn.lockedCnt += len(keys) - skippedLockKeys
+	if txn.aggressiveLockingContext != nil && lockCtx.Stats != nil {
```
cfzjywxk, line 1144: How about wrapping a statistic or telemetry function.

txnkv/transaction/txn.go
```diff
@@ -988,24 +994,26 @@ func (txn *KVTxn) lockKeys(ctx context.Context, lockCtx *tikv.LockCtx, fn func()
 				return err
 			}
 
+			filteredAggressiveLockedKeysCount = len(allKeys) - len(keys)
+			metrics.AggressiveLockedKeysDerived.Add(float64(filteredAggressiveLockedKeysCount))
+			metrics.AggressiveLockedKeysNew.Add(float64(len(keys)))
+
 			if len(keys) == 0 {
+				if lockCtx.Stats != nil {
+					lockCtx.Stats.AggressiveLockDerivedCount += filteredAggressiveLockedKeysCount
```
sticnarf, line 1003: What about moving this out of `if len(keys) == 0`? Then, we make all metrics update and stats update together.

metrics/metrics.go
cfzjywxk: Oh I see them in the tidb pr.

sticnarf (commented):
Only a small optional suggestion.

txnkv/transaction/txn.go
MyonKeminta: Does it matter if `Stats` counts failed ones?

sticnarf: Ah, seems better not. I see TiDB merges the stats no matter whether there's an error.

metrics/shortcuts.go
```diff
@@ -136,6 +136,11 @@ var (
 	PrewriteAssertionUsageCounterExist    prometheus.Counter
 	PrewriteAssertionUsageCounterNotExist prometheus.Counter
 	PrewriteAssertionUsageCounterUnknown  prometheus.Counter
+
+	AggressiveLockedKeysNew                prometheus.Counter
```
ekexium, line 140: Shall we add brief comments explaining them? For example "derived" is vague if not diving into code.

ekexium (approved):

cfzjywxk (approved):

PR #708, opened 2023-02-20
internal: treat default resource group normally

BornChanger:
In order to address the issue https://github.com/pingcap/tidb/issues/40875, we should take the `default` resource group as a normal one.


nolouch:
Does we should support init `default` in resource manager first?

disksing (approved):

glorv (commented):
LGTM

PR #709, opened 2023-02-21
ResourceControl: check resp RegionErr

CabinfeverB:
ref https://github.com/tikv/client-go/pull/684#pullrequestreview-1288681385 and https://github.com/tikv/pd/pull/5941


CabinfeverB:
IMO, RPC error requires pay back, but which situations in RegionError require pay back? 
@sticnarf @glorv @JmPotato @nolouch @BornChanger

JmPotato:
> IMO, RPC error requires pay back, but which situations in RegionError require pay back? 
> 
> @sticnarf @glorv @JmPotato @nolouch @BornChanger

AFAIK, basically all region errors are returned before applying any data to the storage, so once a region error occurs, the pay back should be performed.

nolouch:
> IMO, RPC error requires pay back, but which situations in RegionError require pay back? @sticnarf @glorv @JmPotato @nolouch @BornChanger

I think may need to make sure the error will not retry again, which means the request failed about that sql.

PR #710, opened 2023-02-22
upgrade go version

nolouch:
ref https://github.com/tikv/pd/issues/6029 https://github.com/pingcap/tidb/issues/40969

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.19
+        go-version: "1.20.x"
 
     - name: Test
       run: go test ./...
@@ -28,7 +28,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.19
+        go-version: "1.20.x"
 
     - name: Test with race
       run: go test -race ./...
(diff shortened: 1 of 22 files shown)


nolouch:
ptal @disksing @sticnarf @JmPotato

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.19
+        go-version: "1.20.x"
```
sticnarf, line 18: Can we keep this 1.19? We expect client-go itself to be compilable on the oldest officially supported Golang version.

nolouch: https://github.com/tikv/client-go/actions/runs/4240854947/jobs/7370351699

if we upgrade the dependency of pd or tidb, will report the error:
```
go: go.mod file indicates go 1.20, but maximum version supported by tidy is 1.19
```

nolouch: oh, it uses the latest TiDB. so the compatibility test will be failed always.

sticnarf: Ah, perhaps it's the PD client that specifies a too new toolchain...

sticnarf: > oh, it uses the latest TiDB. so the compatibility test will be failed always. maybe we can only change the version for compatibility test.

The compatibility tests and integration tests can use 1.20.

nolouch: updated in https://github.com/tikv/client-go/pull/711.

PR #718, opened 2023-02-27
txn: set group name in twoPhaseCommitter

glorv:
We need to also support track PessimisticLock request by resource group. The correspending changes at tidb side is https://github.com/glorv/tidb/commit/99b36924e7a8d4b836052de1648be02e0b374054.


glorv:
@Connor1996 @disksing PTAL

disksing (approved):

PR #721, opened 2023-02-28
metric: seperate source for txn cmd

cfzjywxk:
ref: https://github.com/pingcap/tidb/issues/41203

Add source scope for the `TiKVTxnCmdHistogram` metric.

Before
image

After
image


ekexium:
Most of the time we only care about user queries. Is it better to split it into 2 panels or hide internals by default?

cfzjywxk:
> Most of the time we only care about user queries. Is it better to split it into 2 panels or hide internals by default? Especially for "KV Transaction Duration", it looks a bit messy to me

Yes, in the future `tidb.json` change the `internal` types would be hidden by default.

zyguan (approved):

PR #723, opened 2023-03-02
metrics: seperate metrics with source scope for txn command

cfzjywxk:
ref: https://github.com/pingcap/tidb/issues/41203

Add source scope for several metrics.

The grafana change result would be posted latter, related PR would be opened in the tidb repo.


metrics/shortcuts.go
```diff
@@ -74,14 +74,26 @@ var (
...
+	TxnRegionsNumHistogramPrewrite                     prometheus.Observer
+	TxnRegionsNumHistogramCommitInternal               prometheus.Observer
+	TxnRegionsNumHistogramCommit                       prometheus.Observer
+	TxnRegionsNumHistogramCleanupInternal              prometheus.Observer
+	TxnRegionsNumHistogramCleanup                      prometheus.Observer
+	TxnRegionsNumHistogramPessimisticLockInternal      prometheus.Observer
+	TxnRegionsNumHistogramPessimisticLock              prometheus.Observer
+	TxnRegionsNumHistogramPessimisticRollbackInternal  prometheus.Observer
+	TxnRegionsNumHistogramPessimisticRollback          prometheus.Observer
+	TxnRegionsNumHistogramWithCoprocessor              prometheus.Observer
```
you06, line 89: `TxnRegionsNumHistogramWithCoprocessorInternal` is missing

metrics/shortcuts.go
```diff
@@ -185,14 +197,22 @@ func initShortcuts() {
...
+	TxnRegionsNumHistogramCleanupInternal = TiKVTxnRegionsNumHistogram.WithLabelValues("2pc_cleanup", LblInternal)
+	TxnRegionsNumHistogramCleanup = TiKVTxnRegionsNumHistogram.WithLabelValues("2pc_cleanup", LblGeneral)
+	TxnRegionsNumHistogramPessimisticLockInternal = TiKVTxnRegionsNumHistogram.WithLabelValues("2pc_pessimistic_lock", LblInternal)
+	TxnRegionsNumHistogramPessimisticLock = TiKVTxnRegionsNumHistogram.WithLabelValues("2pc_pessimistic_lock", LblGeneral)
+	TxnRegionsNumHistogramPessimisticRollbackInternal = TiKVTxnRegionsNumHistogram.WithLabelValues("2pc_pessimistic_rollback", LblInternal)
+	TxnRegionsNumHistogramPessimisticRollback = TiKVTxnRegionsNumHistogram.WithLabelValues("2pc_pessimistic_rollback", LblGeneral)
+	TxnRegionsNumHistogramWithBatchCoprocessorInternal = TiKVTxnRegionsNumHistogram.WithLabelValues("coprocessor", LblInternal)
+	TxnRegionsNumHistogramWithBatchCoprocessor = TiKVTxnRegionsNumHistogram.WithLabelValues("batch_coprocessor", LblGeneral)
+	TxnWriteKVCountHistogramInternal = TiKVTxnWriteKVCountHistogram.WithLabelValues(LblInternal)
+	TxnWriteKVCountHistogramGeneral = TiKVTxnWriteKVCountHistogram.WithLabelValues(LblGeneral)
```
you06, line 215: `TxnWriteKVCountHistogramInternal` and `TxnWriteKVCountHistogramGeneral` are not initialized.

Unluckily there is no need codegen by macro for such works :(

you06 (commented):
And may change the implementation of [`KVTxn.isInternal`](https://github.com/tikv/client-go/blob/3f7860f109599a7149665e842eab9974478353f5/txnkv/transaction/txn.go#L601-L603) and [`KVSnapshot.IsInternal`](https://github.com/tikv/client-go/blob/3f7860f109599a7149665e842eab9974478353f5/txnkv/txnsnapshot/snapshot.go#L193-L196) to inherit IsInternal from `RequestSource`.

```go
func (r *RequestSource) IsInternal() bool {
	if r == nil {
		return false
	}
	return r.RequestSourceInternal
}
```

metrics/shortcuts.go
cfzjywxk: Missing `TxnWriteSizeHistogramInternal` and `TxnWriteSizeHistogramGeneral` are added.

cfzjywxk:
> And may change the implementation of [`KVTxn.isInternal`](https://github.com/tikv/client-go/blob/3f7860f109599a7149665e842eab9974478353f5/txnkv/transaction/txn.go#L601-L603) and [`KVSnapshot.IsInternal`](https://github.com/tikv/client-go/blob/3f7860f109599a7149665e842eab9974478353f5/txnkv/txnsnapshot/snapshot.go#L193-L196) to inherit IsInternal from `RequestSource`.
> 
> ```go
> func (r *RequestSource) IsInternal() bool {
> 	if r == nil {
> 		return false
> 	}
> 	return r.RequestSourceInternal
> }
> ```

@you06 
There's a method wrapped in https://github.com/tikv/client-go/blob/master/util/request_source.go#L64, I think it's better to check   nil first and void using nil references.

you06:
@

> There's a method wrapped in https://github.com/tikv/client-go/blob/master/util/request_source.go#L64, I think it's better to check nil first and void using nil references.

Yes the nil check is required, what I mean is to avoid making string and check it, we can check `RequestSource.RequestSourceInternal` directly if possible. The function `IsInternalRequest` is used when we have request context, the request source is formatted as a string in it.

cfzjywxk:
@you06 
Got it, I've changed the `util.IsRequestSourceInternal` implementation to use the boolean `RequestSourceInternal` directly.

zyguan (approved):

you06 (approved):

PR #727, opened 2023-03-06
txn: support force locking and lock only if exists be used together

MyonKeminta:
Ref: https://github.com/tikv/tikv/issues/14293
This PR makes client-go allow force-locking be used when `lockOnlyIfExists` is set. Also adjusted mocktikv's behavior in this case, according to https://github.com/tikv/tikv/issues/14293 .

Requires merging https://github.com/tikv/tikv/pull/14330 first.


MyonKeminta:
The failing test `TestRequestPriority` doesn't seem to be caused by this PR.

cfzjywxk (approved):
LGTM

ekexium (approved):

PR #732, opened 2023-03-08
Introduce the RURuntimeStats

JmPotato:
Ref https://github.com/pingcap/tidb/issues/41972. Should merge after ~~https://github.com/pingcap/kvproto/pull/1073 and~~ https://github.com/tikv/pd/pull/6125.

Before `EXPLAIN ANALYZE` is executed, it will set `RURuntimeStats` with its `start_ts`, then the `OnRequest` and `OnResponse` could get the corresponding `RURuntimeStats` and update its RU info.


tikv/kv.go
```diff
@@ -127,6 +127,9 @@ type KVStore struct {
 
 	replicaReadSeed uint32 // this is used to load balance followers / learners when replica read is enabled
 
+	// StartTS -> RURuntimeStats, stores the RU runtime stats for certain transaction.
+	ruRuntimeStatsMap sync.Map
```
glorv, line 131: Seems the keys in this map is never deleted or clean, will this cause memory usage problem?

JmPotato: I added a goroutine to do the clean-up work, PTAL again.

txnkv/txnsnapshot/snapshot.go
```diff
@@ -391,6 +391,7 @@ func (s *KVSnapshot) batchGetSingleRegion(bo *retry.Backoffer, batch batchKeys,
 			Keys:    pending,
 			Version: s.version,
 		}, s.mu.replicaRead, &s.replicaReadSeed, kvrpcpb.Context{
+			StartTs:           s.version,
```
Connor1996, line 394: all the places that you set `StartTs` is just a copy of start version in each type of command. How about getting the start ts with a match of the request type instead of adding a new field in context

internal/client/client_interceptor.go
```diff
@@ -34,17 +36,17 @@ var _ Client = interceptedClient{}
...
-	return interceptedClient{client}
+func NewInterceptedClient(client Client, ruRuntimeStatsMap *sync.Map) Client {
+	return interceptedClient{client, ruRuntimeStatsMap}
 }
 
 func (r interceptedClient) SendRequest(ctx context.Context, addr string, req *tikvrpc.Request, timeout time.Duration) (*tikvrpc.Response, error) {
 	// Build the resource control interceptor.
-	resourceGroupName := req.GetResourceGroupName()
-	var finalInterceptor interceptor.RPCInterceptor = buildResourceControlInterceptor(ctx, req, resourceGroupName)
+	var finalInterceptor interceptor.RPCInterceptor = buildResourceControlInterceptor(ctx, req, r.getRURuntimeStats(req.GetStartTs()))
```
nolouch, line 49: I wonder `point get` use same ts, if run point get workload and run a `explain analazy` may get wrong result?

JmPotato: I checked the TiDB code and found that the point get inside an `EXPLAIN ANALYZE` would not use max uint64 due to the following code:

https://github.com/pingcap/tidb/blob/cc56b21242a2ad82a2f079d7f518606f8705fec4/sessiontxn/isolation/optimistic.go#L132

It will not be regarded as a pure point get since it's an `EXPLAIN` SQL after all. So its `start_ts` will remain unique.

nolouch (approved):
lgtm. ptal @Connor1996

glorv (approved):
LGTM

Issue #735, opened 2023-03-09
race test failed


Smityz:
```
go test -race ./...
```
race test may failed

error info:
```
WARNING: DATA RACE
Read at 0x00c000406610 by goroutine 410:
  github.com/tikv/client-go/v2/internal/locate.(*Store).requestLiveness()
      /Users/yanzhao.tang/Code/client-go/internal/locate/region_cache.go:2594 +0x1bb
  github.com/tikv/client-go/v2/internal/locate.(*Store).checkUntilHealth()
      /Users/yanzhao.tang/Code/client-go/internal/locate/region_cache.go:2571 +0x59a
  github.com/tikv/client-go/v2/internal/locate.(*Store).startHealthCheckLoopIfNeeded.func1()
      /Users/yanzhao.tang/Code/client-go/internal/locate/region_cache.go:2542 +0x47

Previous write at 0x00c000406610 by goroutine 398:
  github.com/tikv/client-go/v2/internal/locate.(*testRegionRequestToThreeStoresSuite).TestReplicaSelector()
      /Users/yanzhao.tang/Code/client-go/internal/locate/region_request3_test.go:498 +0x396b
  runtime.call16()
      /usr/local/Cellar/go/1.19.5/libexec/src/runtime/asm_amd64.s:724 +0x48
  reflect.Value.Call()
      /usr/local/Cellar/go/1.19.5/libexec/src/reflect/value.go:368 +0xc7
  github.com/stretchr/testify/suite.Run.func1()
      /Users/yanzhao.tang/go/pkg/mod/github.com/stretchr/testify@v1.8.1/suite/suite.go:175 +0x6e6
  testing.tRunner()
      /usr/local/Cellar/go/1.19.5/libexec/src/testing/testing.go:1446 +0x216
  testing.(*T).Run.func1()
      /usr/local/Cellar/go/1.19.5/libexec/src/testing/testing.go:1493 +0x47
```

PR #736, opened 2023-03-09
fix data race

Smityz:
close https://github.com/tikv/client-go/issues/735


Smityz:
@disksing Hi master, could you please help me to review this fix? It is only related to unit testing.

PR #741, opened 2023-03-16
resource_control: consider default as a normal group

nolouch:
Currently we support to config the default resource group, relative https://github.com/pingcap/tidb/pull/41606. after remove resource group, we can see the metrics:
image


nolouch:
PTAL @JmPotato @glorv @Connor1996

glorv:
duplicated with #708

PR #748, opened 2023-03-28
txnkv: add new API for lock->put optimization

zyguan:
This is a part of hotfix for https://github.com/pingcap/tidb/issues/28011 . It allow us do lock->put optimization without touching membuf.

### DIFF
--- a/integration_tests/2pc_test.go
+++ b/integration_tests/2pc_test.go
@@ -42,6 +42,7 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -1731,7 +1732,7 @@ func (s *testCommitterSuite) TestFlagsInMemBufferMutations() {
 
 	forEachCase(func(op kvrpcpb.Op, key []byte, value []byte, i int, isPessimisticLock, assertExist, assertNotExist bool) {
 		handle := db.IterWithFlags(key, nil).Handle()
-		mutations.Push(op, isPessimisticLock, assertExist, assertNotExist, handle)
+		mutations.Push(op, isPessimisticLock, assertExist, assertNotExist, handle, nil)
 	})
 
 	forEachCase(func(op kvrpcpb.Op, key []byte, value []byte, i int, isPessimisticLock, assertExist, assertNotExist bool) {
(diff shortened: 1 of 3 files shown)


txnkv/transaction/2pc.go
```diff
@@ -1758,7 +1784,7 @@ func (c *twoPhaseCommitter) tryAmendTxn(ctx context.Context, startInfoSchema Sch
 				return false, err
 			}
 			handle := c.txn.GetMemBuffer().IterWithFlags(key, nil).Handle()
-			c.mutations.Push(op, addMutations.IsPessimisticLock(i), addMutations.IsAssertExists(i), addMutations.IsAssertNotExist(i), handle)
+			c.mutations.Push(op, addMutations.IsPessimisticLock(i), addMutations.IsAssertExists(i), addMutations.IsAssertNotExist(i), handle, nil)
```
cfzjywxk, line 1787: Some comments are left [here](https://github.com/zyguan/client-go/pull/1/files), I think avoid using `nil` checking in the `push` function is safer.

txnkv/transaction/2pc.go
```diff
@@ -581,8 +606,8 @@ func (c *twoPhaseCommitter) initKeysAndMutations(ctx context.Context) error {
 		if c.txn.schemaAmender != nil || c.txn.assertionLevel == kvrpcpb.AssertionLevel_Off {
 			mustExist, mustNotExist, hasAssertUnknown = false, false, false
 		}
-		c.mutations.Push(op, isPessimistic, mustExist, mustNotExist, it.Handle())
-		size += len(key) + len(value)
+		c.mutations.Push(op, isPessimistic, mustExist, mustNotExist, it.Handle(), cachedValue)
```
cfzjywxk, line 609: Some comments are left [here](https://github.com/zyguan/client-go/pull/1/files), maybe avoiding using `nil` checking in the `push` function is safer.

txnkv/transaction/2pc.go
```diff
@@ -252,6 +261,12 @@ func (m *memBufferMutations) Push(op kvrpcpb.Op, isPessimisticLock, assertExist,
 	}
 	handle.UserData = aux
 	m.handles = append(m.handles, handle)
+	if value != nil {
+		if m.overlay == nil {
+			m.overlay = make(map[unionstore.MemKeyHandle][]byte)
+		}
+		m.overlay[handle] = value
```
cfzjywxk, line 268: Maybe we could assert `isPessimisticLock == true` here?

txnkv/transaction/2pc.go
zyguan: `Push` needs to take a parameter of type `[]byte` as the lock->put value, since we do not overwrite membuf anymore (even on `initKeysAndMutations`).

txnkv/transaction/2pc.go
zyguan: Do you mean panic on assertion failed?

integration_tests/2pc_test.go
```diff
@@ -1742,3 +1743,90 @@ func (s *testCommitterSuite) TestFlagsInMemBufferMutations() {
...
+func (s *testCommitterSuite) TestSetLockedKeyValue() {
+	ctx := context.Background()
+	k1 := []byte("k1")
+	v1 := []byte("v1")
+	v2 := []byte("v2")
+
+	mustLockKey := func(txn transaction.TxnProbe, key []byte) {
+		s.Require().NoError(txn.LockKeys(ctx, &kv.LockCtx{ForUpdateTS: txn.StartTS(), WaitStartTime: time.Now()}, key))
+	}
+	mutOpVals := func(opVals ...interface{}) func(m transaction.CommitterMutations) {
```
ekexium, line 1756: I just have a question: what does `mut` mean here?

ekexium (commented):
LGTM

integration_tests/2pc_test.go
zyguan: It's the abbr of 'mutation'. The function means to create a check function from mutation op-val pairs. I just want to make the name shorter, but yes, it's little hard to understand.

integration_tests/2pc_test.go
```diff
@@ -1742,3 +1743,90 @@ func (s *testCommitterSuite) TestFlagsInMemBufferMutations() {
...
+			mutOpVals(kvrpcpb.Op_Put, v2),
+		},
+		{
+			"LockAndDelete",
+			[]func(txn transaction.TxnProbe){
+				func(txn transaction.TxnProbe) { txn.SetLockedKeyValue(k1, v1) },
+				func(txn transaction.TxnProbe) { mustLockKey(txn, k1) },
+				func(txn transaction.TxnProbe) { s.Require().NoError(txn.Delete(k1)) },
+			},
+			mutOpVals(kvrpcpb.Op_Del, []byte{}),
```
MyonKeminta, line 1803: Is it necessary to test in different order, like calling `txn.SetLockedKeyValue` after calling `LockKeys` or any mutations?

MyonKeminta (approved):

integration_tests/2pc_test.go
zyguan: In tidb, `SetLockedKeyValue` is always called after `LockKeys` in point-get/batch-point-get (like the following). But I think it's ok to test it in arbitrary orders.
```
LockKeys(...)
for each unique keys {
  if uk not in membuf {
    SetLockedKeyValue(uk, val)
  }
}
```

txnkv/transaction/2pc.go
cfzjywxk: Yes, as it may prevent some data inconsistency risks.

cfzjywxk (approved):

PR #752, opened 2023-03-28
txnkv: convert LOCK on untouched kv into PUT

zyguan:
#748 only works for select-for-update, lock->put optimization should also works for [update-non-index-column](https://github.com/pingcap/tidb/issues/25659).

### DIFF
--- a/integration_tests/2pc_test.go
+++ b/integration_tests/2pc_test.go
@@ -1747,6 +1747,7 @@ func (s *testCommitterSuite) TestFlagsInMemBufferMutations() {
 func (s *testCommitterSuite) TestSetLockedKeyValue() {
 	ctx := context.Background()
 	k1 := []byte("k1")
+	k2 := []byte("t00000001_i000000001")
 	v1 := []byte("v1")
 	v2 := []byte("v2")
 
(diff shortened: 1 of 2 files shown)


integration_tests/2pc_test.go
```diff
@@ -1811,12 +1832,26 @@ func (s *testCommitterSuite) TestSetLockedKeyValue() {
...
+			[]func(txn transaction.TxnProbe){
+				func(txn transaction.TxnProbe) { txn.SetLockedKeyValue(k1, v1) },
+				func(txn transaction.TxnProbe) { mustLockKey(txn, k1) },
+				func(txn transaction.TxnProbe) {
+					s.Require().NoError(txn.GetMemBuffer().DeleteWithFlags(k1, kv.SetNewlyInserted))
+				},
+			},
+			checkByOpVals(kvrpcpb.Op_Lock, []byte{}),
+			checkByOpVals(kvrpcpb.Op_Lock, []byte{}),
+		},
```
cfzjywxk, line 1846: We could add cases where `PUT` and `Delete` are executed first then the `SetLockedKeyValue`.

ekexium (approved):

PR #756, opened 2023-03-30
Support specifying checking forUpdateTS in prewrite

MyonKeminta:
Ref: https://github.com/tikv/tikv/issues/14311
Requires: 
- https://github.com/pingcap/kvproto/pull/1096
- https://github.com/tikv/tikv/pull/14492
- https://github.com/pingcap/tidb/pull/42843


txnkv/transaction/txn.go
```diff
@@ -1185,10 +1199,15 @@ func (txn *KVTxn) lockKeys(ctx context.Context, lockCtx *tikv.LockCtx, fn func()
 		}
 
 		if txn.IsInAggressiveLockingMode() {
+			actualForUpdateTS := lockCtx.ForUpdateTS
+			if val.LockedWithConflictTS > actualForUpdateTS {
```
cfzjywxk, line 1203: Should we add check or assert `if val.LockedWithConflictTS > 0 then val.LockedWithConflictTS > lockCtx.ForUpdateTS` to ensure no unexpected cases would happen.

cfzjywxk (approved):
LGTM

ekexium (approved):

txnkv/transaction/txn.go
MyonKeminta: Considering that the content of the response depends on TiKV, I think returning error is better than panicking; then considering that the loop updates keys' states to membuffer, I think it's better not to break it and return on the halfway. So I checked the error after this loop. PTAL again.

cfzjywxk (approved):
LGTM

PR #757, opened 2023-03-30
fix api v2 decode of region bucket's keys

iosmanthus:
This pull request decodes the region bucket key by mapping them into the keyspace range.
If the bucket key is in the keyspace, then decode it as usual, otherwise map it as `nil`.

### DIFF
--- a/internal/apicodec/codec_v2.go
+++ b/internal/apicodec/codec_v2.go
@@ -953,3 +953,16 @@ func (c *codecV2) decodeLockInfos(locks []*kvrpcpb.LockInfo) ([]*kvrpcpb.LockInf
 	}
 	return locks, nil
 }
+
+func (c *codecV2) DecodeBucketKey(encodedKey []byte) ([]byte, error) {
+	key, err := c.memCodec.decodeKey(encodedKey)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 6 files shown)


internal/apicodec/codec_v2.go
```diff
@@ -953,3 +953,16 @@ func (c *codecV2) decodeLockInfos(locks []*kvrpcpb.LockInfo) ([]*kvrpcpb.LockInf
...
+
+func (c *codecV2) DecodeBucketKey(encodedKey []byte) ([]byte, error) {
+	key, err := c.memCodec.decodeKey(encodedKey)
+	if err != nil {
+		return nil, err
+	}
+
+	if bytes.Compare(key, c.endKey) >= 0 || bytes.Compare(key, c.prefix) < 0 {
+		return []byte{}, nil
+	}
```
AmoebaProtozoa, line 965: should we return the corresponding prefix or endkey of the codec instead of return empty slice?

iosmanthus: This function aims for decoding the key, for example, if there is a region ranged in `[x0001, x0003)`, and the keyspace 002 tries to decode the bucket key "x0001test", then the key for keyspace 002 doesn't exist, we might decode them as "".

AmoebaProtozoa: got it

PR #758, opened 2023-03-31
For KeyIsLocked error reported for timeout, if a lock is recently updated, don't try to resolve it.

ekexium:
ref 
- https://github.com/pingcap/tidb/issues/42717


ekexium:
PTAL @MyonKeminta @cfzjywxk . Performance test reveals that the PR introduces slight performance regression. But I haven't figure it out.

txnkv/transaction/pessimistic.go
```diff
@@ -253,17 +267,32 @@ func (action actionPessimisticLock) handleKeyError(c *twoPhaseCommitter, keyErrs
 			return nil, true, errors.WithStack(&tikverr.ErrDeadlock{Deadlock: deadlock})
 		}
 
+		// Do not resolve the lock if the lock was recently updated which indicates the txn holding the lock is
+		// much likely alive.
+		// This should only happen for wait timeout.
+		if lockInfo := keyErr.GetLocked(); lockInfo != nil &&
+			lockInfo.DurationToLastUpdateMs > 0 &&
+			lockInfo.DurationToLastUpdateMs < skipResolveThresholdMs {
```
MyonKeminta, line 275: The default wait time in TiKV is 1s and the threshold is 1.2s, it seems that it will never resolve lock if the lock state have ever updated, and the threshold doesn't take any effect.
I think a proper threshold can depend on the actual wait timeout in TiKV, such as something like 0.3*waitTimeout.

cfzjywxk: Maybe `300ms` is fine as the `1 second` would not be changed in most cases.
Is there any situation that this setting that would impact performance a lot? I don't come up with one yet.

ekexium: I've verified 300ms threshold in the hot-update workload. It works.

cfzjywxk (approved):

cfzjywxk:
@MyonKeminta PTAL again

MyonKeminta (approved):

PR #764, opened 2023-04-07
add grpc metrics label

ystaticy:


internal/client/client.go
```diff
@@ -672,13 +672,106 @@ func (c *RPCClient) SendRequest(ctx context.Context, addr string, req *tikvrpc.R
...
 }
 
+func (c *RPCClient) setMetrics(req *tikvrpc.Request, rpcDuration time.Duration) {
+
+	ctx := &req.Context
+	keyspace_id := fmt.Sprintf("%d", ctx.KeyspaceId)
+	val := float64(rpcDuration.Milliseconds())
+	switch req.Type {
+	// Transaction Request Types.
+	case tikvrpc.CmdGet:
```
iosmanthus, line 693: How about creating a hashmap to map the CmdType to string?

disksing (approved):

PR #765, opened 2023-04-12
avoid dataIsNotReady error while retrying stale read on the leader

Tema:
When [Stale Read](https://docs.pingcap.com/tidb/stable/stale-read) fails with `DataIsNotReady` it [reties on the leader.](https://github.com/Tema/client-go/blob/b6dc236b463520ea26720a96dc7916d06d1ec708/internal/locate/region_request.go#L560-L565)  as a Stale Read again and still can fail with `DataIsNotReady` if `safe_ts` is not advanced on leader either. We can easily avoid it by sending the read against leader without stale flag. This way it always succeeds if there is no active conflict. This should considerably help in cases when safe_ts is not advancing due to long commit transactions.


internal/locate/region_request.go
```diff
@@ -1194,6 +1194,12 @@ func (s *RegionRequestSender) SendReqCtx(
 			}
 		}
 
+		if retryTimes > 0 && s.replicaSelector != nil && s.replicaSelector.regionStore != nil &&
+			s.replicaSelector.targetIdx == s.replicaSelector.regionStore.workTiKVIdx {
+			// retry on the leader should not use stale read to avoid possible DataIsNotReady error as it always can serve any read
+			req.StaleRead = false
```
you06, line 1200: The region state may not up-to-date in the client, may set `req.ReplicaRead` to `false` to avoid it served by a follower replica.

you06: Sorry `req.ReplicaRead` is set to `false` when stale read, the comment is not correct, my idea is to avoid follower read when retrying, and I'm inclined to read from leader after `DataIsNotReady` is returned once.
The request may got multi `DataIsNotReady` before it choosing the leader replica and switch off stale read in your implementation.
If the resolved ts is blocked by write locks(maybe large transactions), there will be `DataIsNotReady` in another replica node with high possibiliy.
Maybe you can checkout this [commit](https://github.com/you06/client-go/commit/83819131ab705950b837cf12645bdfb14d1d1c1a) to figure out my suggestion.

zhangjinpeng1987:
If stale read encounter a `DataIsNotReady` error, it will backoff a duration and retry here https://github.com/tikv/client-go/blob/master/internal/locate/region_request.go#L1863-L1877, in your situation, does the backoff is necessary since the request will transfer to read leader directly?

internal/locate/region_request.go
you06: This implementation is correct, but I can find some region miss with it, from the metrics, 1/3 of the DataIsNotReady suffers from region miss.

image

I've carefully debugged with it and find the problem may happens here, when the stale read is served by leader in the first time, it may receive a DataIsNotReady error, the `leader.attempts` is increased, then the check `leader.isExhausted(1)` will fail here. Response with region error will retry later, but introduce unexpected latency.

https://github.com/tikv/client-go/blob/92db9f7b151f568637c590eb76331d97b61cf046/internal/locate/region_request.go#L595-L599

I think we have 2 options:

- Reset `leader.attempts` in the path with `state.option.leaderOnly`.
- Disable stale read if choosing leader replica to serve the read in the first try.

cfzjywxk: The `leaderOnly` flag in the previous implementation of `accessFollower` does not function as intended, which can be considered a bug. 
To implement the "retry leader peer without stale read flag" strategy, an alternative approach is to modify `accessFollower.next` to return the corresponding leader peer's `RPCContext`. Then, in `SendReqCtx`, we can remove the `StaleRead` flag from the request based on information provided by the returned `RPCContext`. This ensures that peer selection strategy remains unified and centralized within the `selector`.

you06 (approved):
Great improvement by removing backoffer!

cfzjywxk (approved):
LGTM. Greate improvement~

PR #773, opened 2023-04-19
Add integration test for the issue that stale pessimistic lock request with force-locking enabled may cause data inconsistency after resolving lock

MyonKeminta:
Ref: https://github.com/tikv/tikv/issues/14551

This PR adds a integration test case to cover the issue stated in https://github.com/tikv/tikv/issues/14551

It's confirmed that the test will fail when running on older TiKV versions without [the fix](https://github.com/tikv/tikv/pull/14557)

```
    lock_test.go:1483:
                Error Trace:    /home/tidb/misono/sync/client-go/integration_tests/lock_test.go:1483
                                                        /home/tidb/misono/sync/client-go/integration_tests/lock_test.go:1510
                                                        /home/tidb/misono/sync/client-go/integration_tests/lock_test.go:1513
                Error:          Received unexpected error:
                                not exist
                                github.com/tikv/client-go/v2/error.init
                                        /home/tidb/misono/sync/client-go/error/error.go:53
                                runtime.doInit
                                        /usr/local/go/src/runtime/proc.go:6506
                                runtime.doInit
                                        /usr/local/go/src/runtime/proc.go:6483
                                runtime.doInit
                                        /usr/local/go/src/runtime/proc.go:6483
                                runtime.doInit
                                        /usr/local/go/src/runtime/proc.go:6483
                                runtime.doInit
                                        /usr/local/go/src/runtime/proc.go:6483
                                runtime.main
                                        /usr/local/go/src/runtime/proc.go:233
                                runtime.goexit
                                        /usr/local/go/src/runtime/asm_amd64.s:1598
                Test:           TestLockWithTiKV/TestStaleForcePessimisticLockNotCommitted
    lock_test.go:1484:
                Error Trace:    /home/tidb/misono/sync/client-go/integration_tests/lock_test.go:1484
                                                        /home/tidb/misono/sync/client-go/integration_tests/lock_test.go:1510
                                                        /home/tidb/misono/sync/client-go/integration_tests/lock_test.go:1513
                Error:          Not equal:
                                expected: []byte{0x76, 0x32}
                                actual  : []byte(nil)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,4 +1,2 @@
                                -([]uint8) (len=2) {
                                - 00000000  76 32                                             |v2|
                                -}
                                +([]uint8) <nil>

                Test:           TestLockWithTiKV/TestStaleForcePessimisticLockNotCommitted
```

### DIFF
--- a/integration_tests/lock_test.go
+++ b/integration_tests/lock_test.go
@@ -1354,6 +1354,7 @@ storeIter:
 		}
 	}
 
+	// Prevent goleak from complaining about its internal connections.
 	http.DefaultClient.CloseIdleConnections()
 
 	if len(recoverConfigs) > 0 {
@@ -1368,6 +1369,9 @@ storeIter:
 				t.Logf("failed to recover config for store at %s: %v", item.url, err)
 			}
 		}
+
+		// Prevent goleak from complaining about its internal connections.
+		http.DefaultClient.CloseIdleConnections()
 	}
 }
 
--- a/integration_tests/lock_test.go
+++ b/integration_tests/lock_test.go
@@ -1248,3 +1248,131 @@ func (s *testLockWithTiKVSuite) TestPrewriteCheckForUpdateTS() {
 	test(true, true, false)
 	test(true, false, true)
 }
+
+func (s *testLockWithTiKVSuite) TestStaleForcePessimisticLockNotCommitted() {
+	// See https://github.com/tikv/tikv/issues/14551 for detail about the issue this test is supposed to cover.
+	if !*withTiKV {
+		// The issue only exists in TiKV's storage model. Unnecessary to run with mocks.
+		return
+	}
+
+	ctx := context.WithValue(context.Background(), util.SessionID, uint64(1))
+
+	var success bool
+	for iteration := 0; iteration < 20; iteration++ {
+		success = func() bool {
+			// Test several times repeatedly for greater possibility to reproduce the issue if it exists.
+			s.T().Logf("TestStaleForcePessimisticLockNotCommitted starts iteration %d", iteration)
+
+			// Add prefix to the key to make that more likely to locate in the same region.
+			pk := []byte(fmt.Sprintf("TestStaleForcePessimisticLockNotCommitted-k%d-p", iteration))
+			key1 := []byte(fmt.Sprintf("TestStaleForcePessimisticLockNotCommitted-k%d-1", iteration))
+			key2 := []byte(fmt.Sprintf("TestStaleForcePessimisticLockNotCommitted-k%d-1", iteration))
+
+			v1 := []byte("v1")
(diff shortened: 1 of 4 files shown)


integration_tests/lock_test.go
```diff
@@ -1248,3 +1254,269 @@ func (s *testLockWithTiKVSuite) TestPrewriteCheckForUpdateTS() {
...
+			lockCtx = kv.NewLockCtx(t1.StartTS(), 200, lockTime)
+			s.NoError(t1.LockKeys(ctx, lockCtx, key1))
+			t1.CancelAggressiveLocking(ctx)
+			// Wait the pessimistic lock of t1 on key1 to be rolled back.
+			s.checkIsKeyLocked(key1, false)
+
+			// Try to make transaction t1 and t2 commit at the same commitTS.
+			// First, block at finishing getting minCommitTS from PD and before prewriting
+			s.NoError(failpoint.Enable("tikvclient/beforePrewrite", "pause"))
+			resCh := make(chan error)
```
cfzjywxk, line 1443: Is it better to use `resCh := make(chan error, 2)` so they two results would not block each other on the result channel?

integration_tests/lock_test.go
```diff
@@ -1248,3 +1254,269 @@ func (s *testLockWithTiKVSuite) TestPrewriteCheckForUpdateTS() {
...
+				mutations.Push(kvrpcpb.Op_PessimisticLock, key1, nil, true, false, false, false)
+				err = simulatedCommiter.PessimisticLockMutations(ctx, lockCtx, kvrpcpb.PessimisticLockWakeUpMode_WakeUpModeForceLock, &mutations)
+				s.NoError(err)
+
+				s.Equal(t2.GetCommitTS(), lockCtx.Values[string(key1)].LockedWithConflictTS)
+			}
+
+			// Trigger resolving locks.
+			// For normal resolve lock procedures, it always uses special path for resolving pessimistic locks, which
+			// avoids this problem. However, `BatchResolveLock` which is used by GC doesn't have that handling (and
```
cfzjywxk, line 1504: As discussed maybe we need to fix `BatchResolveLock` first and make the resolving behaviour and interface consistent

MyonKeminta: Yes. I'll try to find some other way to do this test later.

ekexium (approved):

MyonKeminta:
This case actually won't happen, since when we meets pessimistic lock, we always call `resolvePessimisticLock` which only pessimistic-rollback the lock instead of performing normal resolve-lock procedure, and when we meets optimistic lock we always know the real primary.

But perhaps we also need to consider the case that a optimistic lock points to the real primary, but the primary is locked by a stale pessimistic lock pointing to another key which is the invalidated primary...

PR #776, opened 2023-04-23
add metrics for stale-read traffic

you06:
Add the traffic metrics for stale-read, this tells the hit ratio and traffics produced by stale-read miss.

The metrics example:

image

Before 15:15, the stale-read hit ratio is high, and after 15:15, there are over 80% stale-read misses.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1155,6 +1155,14 @@ func (s *RegionRequestSender) SendReqCtx(
 			metrics.TiKVRequestRetryTimesHistogram.Observe(float64(retryTimes))
 		}
 	}()
+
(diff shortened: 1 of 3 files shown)


zyguan (approved):

zyguan:
BTW, do we need/have metrics for retry traffic?

you06:
> retry traffic

You mean the retry traffic should be counted multi times?

zyguan:
> > retry traffic
> 
> You mean the retry traffic should be counted multi times?

No, I think it's ok for only tracking the final req/resp (as the current implementation does). I just wonder that is it necessary for observing retry traffic as well (maybe by another metric).

disksing:
Merged! If we decide to add metrics later, please submit a new PR. Thanks!

PR #777, opened 2023-04-23
lock_resolver: support verifying primary for check_txn_status

MyonKeminta:
Ref: https://github.com/pingcap/tidb/issues/42937
Requires: https://github.com/pingcap/kvproto/pull/1105
Requires: https://github.com/tikv/tikv/pull/14637


MyonKeminta:
This PR is now reviewable. The integration test failed because the TiKV image is not updated yet and doesn't include the latest change. It should pass tomorrow.

txnkv/txnlock/lock_resolver.go
```diff
@@ -448,7 +448,11 @@ func (lr *LockResolver) resolveLocks(bo *retry.Backoffer, opts ResolveLocksOptio
 	var resolve func(*Lock, bool) (TxnStatus, error)
 	resolve = func(l *Lock, forceSyncCommit bool) (TxnStatus, error) {
 		status, err := lr.getTxnStatusFromLock(bo, l, callerStartTS, forceSyncCommit, detail)
-		if err != nil {
+
+		if _, ok := errors.Cause(err).(primaryMismatch); ok && l.LockType == kvrpcpb.Op_PessimisticLock {
```
cfzjywxk, line 452: Maybe we could print an `ERROR` or `FATAL` log for the `l.LockType != kvrpcpb.Op_PessimisticLock` path since it's unexpected?

cfzjywxk (approved):

txnkv/txnlock/lock_resolver.go
MyonKeminta: I'll add an error here. Actually the error will be returned to TiDB, and TiDB will print it as an unknown error anyway.

MyonKeminta: I noticed that there might be other problem about the primary when fair-locking is used...

cfzjywxk:
@ekexium @zyguan 
PTAL

ekexium (approved):

txnkv/txnlock/lock_resolver.go
cfzjywxk: @MyonKeminta 
The check issue for fair lock is recorded in https://github.com/pingcap/tidb/issues/43540, could we merge this first?

MyonKeminta: > @MyonKeminta The check issue for fair lock is recorded in [pingcap/tidb#43540](https://github.com/pingcap/tidb/issues/43540), could we merge this first?

yes. lets merge it.

PR #779, opened 2023-04-24
lock_resolver: handle pessimistic locks in BatchResolveLocks

MyonKeminta:
Ref: https://github.com/pingcap/tidb/issues/43243

### DIFF
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -238,6 +238,8 @@ func (lr *LockResolver) BatchResolveLocks(bo *retry.Backoffer, locks []*Lock, lo
 	txnInfos := make(map[uint64]uint64)
 	startTime := time.Now()
 	for _, l := range expiredLocks {
+		logutil.Logger(bo.GetCtx()).Info("BatchResolveLocks handling lock", zap.Stringer("lock", l))
+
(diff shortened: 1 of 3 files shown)


txnkv/txnlock/lock_resolver.go
```diff
@@ -238,11 +238,27 @@ func (lr *LockResolver) BatchResolveLocks(bo *retry.Backoffer, locks []*Lock, lo
 	txnInfos := make(map[uint64]uint64)
 	startTime := time.Now()
 	for _, l := range expiredLocks {
+		logutil.Logger(bo.GetCtx()).Info("BatchResolveLocks handling lock", zap.Stringer("lock", l))
```
cfzjywxk, line 241: Should we use `logutil.Debug(...)`?

cfzjywxk (approved):

txnkv/txnlock/lock_resolver.go
MyonKeminta: Ah sorry. I added this when debugging and forgot to remove it. Keep it as a debug-level log might be a good idea.

MyonKeminta:
Oops something went wrong about the tests... I'll figure it out when I have time..

MyonKeminta:
fixed in #794

PR #781, opened 2023-04-24
add const label for metrics

zeminzhou:
image

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -127,537 +127,598 @@ const (
 	LblGeneral         = "general"
 )
 
-func initMetrics(namespace, subsystem string) {
+func initMetrics(namespace, subsystem string, constLabels prometheus.Labels) {
 	TiKVTxnCmdHistogram = prometheus.NewHistogramVec(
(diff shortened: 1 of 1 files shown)


metrics/metrics.go
```diff
@@ -127,537 +127,603 @@ const (
...
 }
 
 // InitMetrics initializes metrics variables with given namespace and subsystem name.
 func InitMetrics(namespace, subsystem string) {
-	initMetrics(namespace, subsystem)
+	initMetrics(namespace, subsystem, nil)
+}
+
+// InitMetricsWithConstLabels initializes metrics variables with given namespace, subsystem name and const labels.
+func InitMetricsWithConstLabels(namespace, subsystem string, constLabels prometheus.Labels) {
```
iosmanthus, line 725: When will we call this function to init metrics?

zeminzhou: TiDB will call this function with const labels. If TiDB configures a keyspace when it starts, the client-go's metrics will have keyspace label.

disksing (approved):

PR #791, opened 2023-05-09
txn: disable follower read for max ts point get read

cfzjywxk:
Ref: https://github.com/pingcap/tidb/issues/43583

What's changed:
When autocommit point get is used, the linearizability could be broken.Disable follower-read path for autocommit point get read.

Side effects:
There would be read hotspot issue if follower read is used and most of the reads are autocommit point get read. Should it be a configurable option?

### DIFF
--- a/txnkv/txnsnapshot/snapshot.go
+++ b/txnkv/txnsnapshot/snapshot.go
@@ -635,12 +635,18 @@ func (s *KVSnapshot) get(ctx context.Context, bo *retry.Backoffer, k []byte) ([]
 	if len(matchStoreLabels) > 0 {
 		ops = append(ops, locate.WithMatchLabels(matchStoreLabels))
 	}
-	if req.ReplicaReadType.IsFollowerRead() && replicaAdjuster != nil {
-		op, readType := replicaAdjuster(1)
-		if op != nil {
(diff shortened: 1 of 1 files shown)


txnkv/txnsnapshot/snapshot.go
```diff
@@ -635,12 +635,18 @@ func (s *KVSnapshot) get(ctx context.Context, bo *retry.Backoffer, k []byte) ([]
 	if len(matchStoreLabels) > 0 {
 		ops = append(ops, locate.WithMatchLabels(matchStoreLabels))
 	}
-	if req.ReplicaReadType.IsFollowerRead() && replicaAdjuster != nil {
-		op, readType := replicaAdjuster(1)
-		if op != nil {
-			ops = append(ops, op)
+	if req.ReplicaReadType.IsFollowerRead() {
```
you06, line 638: Also need to set `BusyThresholdMs` to 0 when `s.version == math.MaxUint64 && !req.ReplicaReadType.IsFollowerRead()`

cfzjywxk:
Try to use another solution described in https://github.com/pingcap/tidb/issues/43583#issuecomment-1539580909.

PR #792, opened 2023-05-09
Remove stale-read flag after resolving lock

you06:
Ref pingcap/tidb#43659

Remove the stale-read flag after resolving lock.

Before this PR:

before

After this PR:

after

Because the request struct is reused when retrying in get, it'll suffer not leader error, because the stale read field is  already [reset from the context patcher](https://github.com/tikv/client-go/blob/d3c53d30ff5dd0e480a31bc8385fb027d2444f04/internal/locate/region_cache.go#L633). But the request is regenerated in batch get, so it'll meet data is not ready instead. For both cases, this PR reduces unnecessary RPC.


tikvrpc/tikvrpc.go
```diff
@@ -275,6 +275,12 @@ func (req *Request) EnableStaleRead() {
 	req.ReplicaRead = false
 }
 
+// DisableStaleRead is called when stale-read fallbacks to leader read after meeting key-is-locked error.
+func (req *Request) DisableStaleRead() {
```
cfzjywxk, line 279: How about `DisableStaleReadAfterFallback`

cfzjywxk (approved):

tikvrpc/tikvrpc.go
you06: Rename it with `DisableStaleReadMeetLock` because it should only take effect when key is locked, other rollback won't affect this flag

zyguan (approved):

PR #793, opened 2023-05-10
pd http: support api to get store min resolved ts 

### LINKED ISSUE #795: minResolvedTS: use pd API to fetch the store resolved_ts firstly
## Development Task

TiDB is currently unable to advance the min_resolved_ts when any TiKV node is down, as reported in https://github.com/pingcap/tidb/issues/43386. However, PD can handle this situation by utilizing more detailed information such as leader count to exclude problematic TIKV nodes. To ensure consistent behavior across the system, we provide an API to fetch store resolved_ts, which will replace the current implementation in client-go.

HuSharp:
Close #795
after https://github.com/tikv/pd/pull/6413 merged, We can replace resolved-ts's current implementation in client-go
- support api mechanism to pd
- support api to get store min resolved ts

### DIFF
--- a/integration_tests/pd_api_test.go
+++ b/integration_tests/pd_api_test.go
@@ -66,11 +66,10 @@ type apiTestSuite struct {
 
 func (s *apiTestSuite) SetupTest() {
 	addrs := strings.Split(*pdAddrs, ",")
-	// Set PD HTTP client.
 	pdClient, err := pd.NewClient(addrs, pd.SecurityOption{})
 	s.Require().Nil(err)
-
 	rpcClient := tikv.NewRPCClient()
+	// Set PD HTTP client.
 	store, err := tikv.NewTestTiKVStore(rpcClient, pdClient, nil, nil, 0, tikv.WithPDHTTPClient(nil, addrs))
 	s.store = store
 	storeID := uint64(1)
--- a/integration_tests/pd_api_test.go
+++ b/integration_tests/pd_api_test.go
@@ -52,9 +52,9 @@ import (
 	pd "github.com/tikv/pd/client"
 )
 
-func TestPDApi(t *testing.T) {
+func TestPDAPI(t *testing.T) {
 	if !*withTiKV {
-		t.Skip("skipping TestPDApi because with-tikv is not enabled")
+		t.Skip("skipping TestPDAPI because with-tikv is not enabled")
 	}
 	suite.Run(t, new(apiTestSuite))
 }
@@ -106,7 +106,8 @@ func (c *storeSafeTsMockClient) CloseAddr(addr string) error {
 func (s *apiTestSuite) TestGetStoreMinResolvedTS() {
 	util.EnableFailpoints()
 	// Try to get the minimum resolved timestamp of the store from PD.
-	s.Require().Nil(failpoint.Enable("tikvclient/InjectMinResolvedTS", `return(100)`))
+	require := s.Require()
+	require.Nil(failpoint.Enable("tikvclient/InjectMinResolvedTS", `return(100)`))
 	mockClient := storeSafeTsMockClient{
 		Client: s.store.GetTiKVClient(),
 	}
@@ -119,12 +120,12 @@ func (s *apiTestSuite) TestGetStoreMinResolvedTS() {
 		}
 		retryCount++
 	}
-	s.Require().Equal(atomic.LoadInt32(&mockClient.requestCount), int32(0))
-	s.Require().Equal(uint64(100), s.store.GetMinSafeTS(oracle.GlobalTxnScope))
-	s.Require().Nil(failpoint.Disable("tikvclient/InjectMinResolvedTS"))
+	require.Equal(atomic.LoadInt32(&mockClient.requestCount), int32(0))
+	require.Equal(uint64(100), s.store.GetMinSafeTS(oracle.GlobalTxnScope))
+	require.Nil(failpoint.Disable("tikvclient/InjectMinResolvedTS"))
 
 	// Try to get the minimum resolved timestamp of the store from TiKV.
-	s.Require().Nil(failpoint.Enable("tikvclient/InjectMinResolvedTS", `return(0)`))
+	require.Nil(failpoint.Enable("tikvclient/InjectMinResolvedTS", `return(0)`))
 	defer func() {
 		s.Require().Nil(failpoint.Disable("tikvclient/InjectMinResolvedTS"))
 	}()
@@ -136,8 +137,8 @@ func (s *apiTestSuite) TestGetStoreMinResolvedTS() {
 		}
 		retryCount++
 	}
-	s.Require().GreaterOrEqual(atomic.LoadInt32(&mockClient.requestCount), int32(1))
-	s.Require().Equal(uint64(150), s.store.GetMinSafeTS(oracle.GlobalTxnScope))
+	require.GreaterOrEqual(atomic.LoadInt32(&mockClient.requestCount), int32(1))
+	require.Equal(uint64(150), s.store.GetMinSafeTS(oracle.GlobalTxnScope))
 }
 
 func (s *apiTestSuite) TearDownTest() {
--- a/integration_tests/pd_api_test.go
+++ b/integration_tests/pd_api_test.go
@@ -1,0 +1,148 @@
+// Copyright 2023 TiKV Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NOTE: The code in this file is based on code from the
+// TiDB project, licensed under the Apache License v 2.0
+//
+// https://github.com/pingcap/tidb/tree/cc5e161ac06827589c4966674597c137cc9e809c/store/tikv/tests/prewrite_test.go
+//
+
+// Copyright 2023 PingCAP, Inc.
+//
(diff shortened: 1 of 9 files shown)


tikv/kv.go
```diff
@@ -193,6 +194,13 @@ func WithPool(gp Pool) Option {
 	}
 }
 
+// WithPDAddrsAndTls set the pd addrs and tls for pdHttpClient.
+func WithPDAddrsAndTls(tlsConf *tls.Config, pdaddrs []string) Option {
```
JmPotato, line 198: Suggest separating into two option functions.

```suggestion
func WithPDAddrs(pdaddrs []string) Option
func WithTLSConfig(tlsConf *tls.Config) Option
```

tikv/kv.go
```diff
@@ -581,25 +589,30 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
-							StartKey: []byte(""),
-							EndKey:   []byte(""),
-						},
-					}, kvrpcpb.Context{
-						RequestSource: util.RequestSourceFromCtx(ctx),
-					},
-				), client.ReadTimeoutShort,
-			)
+
+			safeTS, err := s.pdHttpClient.GetStoreMinResolvedTS(ctx, storeID)
```
JmPotato, line 593: What about checking whether `s.pdHttpClient` is nil first?

util/pd.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+)
+
+// pd request retry time when connection fail
+const (
+	pdRequestRetryTime = 10
+
```
JmPotato, line 58:
```suggestion
const (
	// pd request retry time when connection fail
	pdRequestRetryTime = 10
```

tikv/kv.go
```diff
@@ -115,6 +115,7 @@ type KVStore struct {
 		client Client
 	}
 	pdClient     pd.Client
+	pdHttpClient util.PdController
```
JmPotato, line 118:
```suggestion
	pdHttpClient *util.PdController
```

util/pd.go
```diff
@@ -0,0 +1,204 @@
...
+	storeMinResolvedTSPrefix = "pd/api/v1/min-resolved-ts"
+)
+
+// PdController manage get/update config from pd.
+type PdController struct {
```
JmPotato, line 63: Suggest using a more direct naming like `PDHTTPClient`.

util/pd.go
```diff
@@ -0,0 +1,204 @@
...
+	var err error
+	for _, addr := range p.addrs {
+		query := fmt.Sprintf("%s/store_id=%d", storeMinResolvedTSPrefix, storeID)
+		v, e := pdRequest(ctx, addr, query, p.cli, http.MethodGet, nil)
+		if e != nil {
+			log.Warn("failed to get min resolved ts", zap.String("addr", addr), zap.Error(e))
+			err = e
+			continue
+		}
+		log.Info("store min resolved ts", zap.String("resp", string(v)))
```
JmPotato, line 100:
```suggestion
		log.Debug("store min resolved ts", zap.String("resp", string(v)))
```

util/pd.go
```diff
@@ -0,0 +1,204 @@
...
+			return 0, errors.Trace(err)
+		}
+		if !d.IsRealTime {
+			message := "store min resolved ts not enabled"
+			log.Error(message, zap.String("addr", addr))
+			return 0, errors.Trace(errors.New(message))
+		}
+		return d.MinResolvedTS, nil
+	}
+	return 0, errors.Trace(err)
```
JmPotato, line 116: Check if the error is nil first.

util/pd.go
```diff
@@ -0,0 +1,204 @@
...
+		d := struct {
+			IsRealTime    bool   `json:"is_real_time,omitempty"`
+			MinResolvedTS uint64 `json:"min_resolved_ts"`
+		}{}
+		err = json.Unmarshal(v, &d)
+		if err != nil {
+			return 0, errors.Trace(err)
+		}
+		if !d.IsRealTime {
+			message := "store min resolved ts not enabled"
```
JmPotato, line 110: What about using `fmt.Errorf`?

util/pd.go
```diff
@@ -0,0 +1,204 @@
...
+// pdRequest is a func to send an HTTP to pd and return the result bytes.
+func pdRequest(
+	ctx context.Context,
+	addr string, prefix string,
+	cli *http.Client, method string, body io.Reader) ([]byte, error) {
+	_, respBody, err := pdRequestWithCode(ctx, addr, prefix, cli, method, body)
+	return respBody, err
+}
+
+func pdRequestWithCode(
```
JmPotato, line 128: It seems that this function is unnecessary since the returned code has never been used.

util/pd_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/pingcap/failpoint"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPDRequestRetry(t *testing.T) {
```
JmPotato, line 48: Can we add an integration test to connect a real PD also?

HuSharp: Done, Add at `api_test`

tikv/kv.go
HuSharp: Thx, good catch!

util/pd.go
HuSharp: `errors.Trace` will check whether err is nil

tikv/kv.go
HuSharp: I changed it to `WithPDHttpClient` instead of supporting 2 options, do you think this is appropriate?

integration_tests/raw/api_test.go
```diff
@@ -95,6 +102,20 @@ func (s *apiTestSuite) SetupTest() {
 	client := s.newRawKVClient(pdClient, addrs)
 	s.client = client
 
+	// Set PD HTTP client.
+	var pdAddresses []string
+	for _, addr := range addrs {
+		pdAddresses = append(pdAddresses, addr)
+	}
+	s.pdHttpClient = util.NewPDHTTPClient(nil, pdAddresses)
```
JmPotato, line 110: Why not just use `addrs` directly?

integration_tests/raw/api_test.go
```diff
@@ -515,14 +536,84 @@ func (s *apiTestSuite) TestEmptyValue() {
...
+
+func (c *storeSafeTsMockClient) Close() error {
+	return c.Client.Close()
+}
+
+func (c *storeSafeTsMockClient) CloseAddr(addr string) error {
+	return c.Client.CloseAddr(addr)
+}
+
+func (s *apiTestSuite) TestGetStoreMinResolvedTS() {
```
JmPotato, line 566: I don't think it's proper to put `resolved_ts` test inside `integration_tests/raw/api_test.go` since it's for the RawKV API testing.

tikv/kv.go
```diff
@@ -193,6 +194,13 @@ func WithPool(gp Pool) Option {
 	}
 }
 
+// WithPDHttpClient set the pd addrs and tls for pdHttpClient.
+func WithPDHttpClient(tlsConf *tls.Config, pdaddrs []string) Option {
```
JmPotato, line 198:
```suggestion
// WithPDHTTPClient set the PD HTTP client with the given address and TLS config.
func WithPDHTTPClient(tlsConf *tls.Config, addrs []string) Option {
```

tikv/kv.go
```diff
@@ -581,25 +589,36 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
-							EndKey:   []byte(""),
-						},
-					}, kvrpcpb.Context{
-						RequestSource: util.RequestSourceFromCtx(ctx),
-					},
-				), client.ReadTimeoutShort,
-			)
+
+			var safeTS uint64
+			var err error
```
JmPotato, line 594:
```suggestion
			var (
				safeTS uint64
				err error
			)
```

integration_tests/raw/api_test.go
HuSharp: Sorry my bad...

integration_tests/raw/api_test.go
HuSharp: move to `pd_api_test` file

HuSharp:
@Defined2014 PTAL, thx!

integration_tests/pd_api_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (c *storeSafeTsMockClient) CloseAddr(addr string) error {
+	return c.Client.CloseAddr(addr)
+}
+
+func (s *apiTestSuite) TestGetStoreMinResolvedTS() {
+	util.EnableFailpoints()
+	// Try to get the minimum resolved timestamp of the store from PD.
+	s.Require().Nil(failpoint.Enable("tikvclient/InjectMinResolvedTS", `return(100)`))
```
JmPotato, line 109: What about using `re := s.Require()`?

Defined2014 (approved):
LGTM

integration_tests/pd_api_test.go
HuSharp: Good idea!

tikv/kv.go
```diff
@@ -581,25 +592,38 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
+				err    error
 			)
 			storeIDStr := strconv.Itoa(int(storeID))
-			if err != nil {
-				metrics.TiKVSafeTSUpdateCounter.WithLabelValues("fail", storeIDStr).Inc()
-				logutil.BgLogger().Debug("update safeTS failed", zap.Error(err), zap.Uint64("store-id", storeID))
-				return
+			// Try to get the minimum resolved timestamp of the store from PD.
+			if s.pdHttpClient != nil {
+				safeTS, err = s.pdHttpClient.GetStoreMinResolvedTS(ctx, storeID)
```
JmPotato, line 603: We could add a debug log to output `safeTS` and `err` when it's not nil.

JmPotato (commented):
The rest LGTM.

JmPotato (approved):

PR #796, opened 2023-05-11
Fallback to leader read when replica read meets flashback-in-progress error

you06:
Ref pingcap/tidb#43673

Currently the flashback doesn't wait followers before responsing and continue workload, this PR handles such failure requests and redirect them to leader.


Defined2014 (approved):

you06:
@disksing PTAL

zyguan (approved):

PR #806, opened 2023-05-17
Use a new way to create client for v2

rleungx:


disksing (approved):

tikv/kv.go
```diff
@@ -260,6 +260,34 @@ func NewPDClient(pdAddrs []string) (pd.Client, error) {
 	return pdCli, nil
 }
 
+// NewPDClientWithAPIContext returns an unwrapped pd client with API context.
+func NewPDClientWithAPIContext(pdAddrs []string, apiContext pd.APIContext) (pd.Client, error) {
```
iosmanthus, line 264: How about implementing `NewPDClient` with `NewPDClientWithAPIContext` by passing the APIV1 context?

iosmanthus (commented):
Rest LGTM

iosmanthus (approved):

PR #808, opened 2023-05-19
support remote coprocessor

coocood:


internal/locate/region_request.go
```diff
@@ -1344,6 +1345,9 @@ func (s *RegionRequestSender) sendReqToRegion(
 		req.ForwardedHost = rpcCtx.Addr
 		sendToAddr = rpcCtx.ProxyAddr
 	}
+	if req.StoreTp == tikvrpc.TiKVRemoteCoprocessor {
```
iosmanthus, line 1348: Is there only one remote cop worker address for each request?

coocood: Yes.

PR #828, opened 2023-06-07
add gc options

weedge:
add gc options WithConcurrency


disksing (approved):

txnkv/transaction/txn.go
```diff
@@ -206,7 +206,7 @@ func (txn *KVTxn) GetVars() *tikv.Variables {
 func (txn *KVTxn) Get(ctx context.Context, k []byte) ([]byte, error) {
 	ret, err := txn.us.Get(ctx, k)
 	if tikverr.IsErrNotFound(err) {
-		return nil, err
+		return nil, nil
```
disksing, line 209: em, why change here?

weedge: in rawkv.go  https://github.com/tikv/client-go/blob/master/rawkv/rawkv.go#L287 get  not found key return nil, nil;
so want txnkv return nil, nil  when get not found key

weedge: change back https://github.com/tikv/client-go/pull/828/commits/85c65998e04e2ff54bcc17aef1bc20b1f4ed45ba

disksing (approved):

PR #832, opened 2023-06-09
kvrpc: deduplicate chained intercept by name

glorv:
deduplicate chained intercept by name so we can replace SetInterceptor with AddInterceptor


glorv:
@disksing @nolouch PTAL

txnkv/txnsnapshot/snapshot.go
```diff
@@ -903,6 +903,7 @@ func (s *KVSnapshot) SetRPCInterceptor(it interceptor.RPCInterceptor) {
 }
 
 // AddRPCInterceptor adds an interceptor, the order of addition is the order of execution.
+// the chained interceptors will be dedupcated by its name.
```
nolouch, line 906: Why need this?

glorv: Currently, the same interceptor is set in different places in tidb, I'm not sure if they are set several because the current behavior is replace. But I need to change the behavior to add

nolouch (approved):

tikvrpc/interceptor/interceptor.go
```diff
@@ -122,29 +153,49 @@ func NewRPCInterceptorChain() *RPCInterceptorChain {
 // Link is used to link the next RPCInterceptor.
 // Multiple interceptors will be executed in the order of link time.
```
Connor1996, line 154: add a comment about the behavior of linking duplicated interceptor

Connor1996 (approved):
rest LGTM

tikvrpc/interceptor/interceptor.go
glorv: done

PR #833, opened 2023-06-09
batch_client: auto re-connect idle connection when wait connection ready

crazycs520:


zyguan (approved):

internal/client/client_batch.go
```diff
@@ -555,6 +555,8 @@ func (c *batchCommandsClient) waitConnReady() (err error) {
 			cancel()
 			break
 		}
+		// Trigger idle connection to reconnection
+		c.conn.Connect()
```
cfzjywxk, line 559: This seems to be an experimental API which may not be stable enough?

cfzjywxk: Besides, it seems each time the API is triggered a new goroutine would be created which could be a risk.

crazycs520: Nice catch. For now, I've only found `Connect ` API to work, I try to use the TiKV RPC interface such as heartbeat or health check RPC API, but I didn't find such an interface.

Anyway, I just added a test about this.

zyguan: Actually `GetState` and `WaitForStateChange` are both experimental, I think it's ok to use `Connect` here. How about to trigger `Connect` only once (maybe just before the for loop) and only when the state is idle?

crazycs520: `Connect ` API internal already checks the state, and only do connect when state is idle.

cfzjywxk: `WaitForStateChange` has been used for some time and the risk is relatively low.
Maybe we'd better check the gRPC client source code in details about the state management and how could the `Connect` impact the underlying gRPC logic.

cfzjywxk (approved):
It seems that there is no standard operation for this reconnection. The transient failure state will automatically be converted to connecting for retry and backoff internally. I have approved it for now, but we need to continue monitoring whether there are any abnormalities in terms of availability testing, stress testing, and disaster recovery testing. In addition, we also need to further organize and improve the code of batch client/region cache for maintainability.

PR #834, opened 2023-06-12
client_batch: tiny optimize

crazycs520:
tiny optimize for #833

### DIFF
--- a/internal/client/client_batch.go
+++ b/internal/client/client_batch.go
@@ -541,7 +541,8 @@ func (c *batchCommandsClient) failPendingRequests(err error) {
 }
 
 func (c *batchCommandsClient) waitConnReady() (err error) {
-	if c.conn.GetState() == connectivity.Ready {
+	state := c.conn.GetState()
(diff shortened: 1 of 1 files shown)


cfzjywxk (approved):

internal/client/client_batch.go
```diff
@@ -549,14 +549,15 @@ func (c *batchCommandsClient) waitConnReady() (err error) {
 		metrics.TiKVBatchClientWaitEstablish.Observe(time.Since(start).Seconds())
 	}()
 	dialCtx, cancel := context.WithTimeout(context.Background(), c.dialTimeout)
+	// Trigger idle connection to reconnection
+	// Put it outside loop to avoid unnecessary reconnecting.
+	c.conn.Connect()
```
zyguan, line 554: How about to check the state here? We only need to call `Connect` when the state is IDLE.

crazycs520: Ok, done.

zyguan (approved):

PR #837, opened 2023-06-13
client-go: label the background request source

nolouch:
ref https://github.com/pingcap/tidb/issues/44517


nolouch:
ptal @glorv @HuSharp

util/request_source.go
```diff
@@ -52,6 +55,12 @@ func (r *RequestSource) SetRequestSourceType(tp string) {
 	r.RequestSourceType = tp
 }
 
+// IsBackground checks whether the request is a background job.
+func (r *RequestSource) IsBackground() bool {
+	// TODO: supports set as background by user in the future.
+	return r.RequestSourceType == InternalTxnTools
```
glorv, line 61: I wander what scenario it is to expected pass this `tools` source-type, as most ecosystem tools are not running in txn mode. BTW, tidb-lightning's tidb backend is running with normal txn, but since it's executed via normal sql statements, it's request source is unlikely to be tools either.

util/request_source.go
```diff
@@ -52,6 +55,12 @@ func (r *RequestSource) SetRequestSourceType(tp string) {
 	r.RequestSourceType = tp
 }
 
+// IsBackground checks whether the request is a background job.
+func (r *RequestSource) IsBackground() bool {
```
glorv, line 59: Question: Why not let the caller(tidb in most case) directly pass a `is_background` field?

nolouch: I want to directly define `background`  by job type. but let the caller define it is also ok to me.

util/request_source.go
HuSharp: @glorv  Do you think adding a type in these 2 places is a comprehensive way? https://github.com/pingcap/tidb/pull/44628/files

glorv: I'd like to add a `IsBackground` field the same way as the `ResourceGroupName` field. Determine the value of background implicitly by `RequestSourceType` is a bit weird to me.

HuSharp: make sense to me.

util/request_source.go
```diff
@@ -91,6 +112,15 @@ func RequestSourceFromCtx(ctx context.Context) string {
 	return SourceUnknown
 }
 
+// IsBackgroundFromCtx checks whether the request is a background job.
+func IsBackgroundFromCtx(ctx context.Context) bool {
```
HuSharp, line 116: When reviewing the code, I noticed `WithResouceGroupName` missed a `r`, can u fix it BTW? :)

util/request_source.go
```diff
@@ -52,12 +56,29 @@ func (r *RequestSource) SetRequestSourceType(tp string) {
...
+// SetRunInBackground sets the request to run in background.
+func (r *RequestSource) SetRunInBackground(background bool) {
+	r.RunInBackground = background
+}
+
+// RequestSourceOption is used to set the request source.
+type RequestSourceOption func(*RequestSource)
+
+// RequestRunInBackground sets the request to run in background.
+var RequestRunInBackgrounOption = func(rs *RequestSource) {
```
HuSharp, line 68:
```suggestion
var RequestRunInBackgroundOption = func(rs *RequestSource) {
```
When reviewing the code, I noticed WithResouceGroupName missed a r, can u fix it BTW? :)

PR #839, opened 2023-06-15
mock: impl resource manager get default resource group

CabinfeverB:
ref https://github.com/pingcap/tidb/pull/44339, it creates resource manager client in domain, so it use mock pd client from client-go.


Connor1996 (approved):
LGTM

PR #843, opened 2023-06-16
reload region cache when store is resolved from invalid status

### LINKED ISSUE #841: need to reload region cache when invalid store become health
Follower selector will avoid invalid stores when selecting replica. After the invalid store is up, the status is still invalid in client-go, thus the load is inbalance.

### Follower read

image

### Stale read

image

you06:
close #841

Before this PR:
image

image

With this PR:
image

image


internal/locate/region_request.go
```diff
@@ -573,18 +573,22 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	if state.option.preferLeader {
 		state.lastIdx = state.leaderIdx
 	}
+	offset := rand.Intn(replicaSize)
```
LykxSassinator, line 576: Under `PreferLeader` mode, offset should be set to `0` to make the selection starting from `leaderIdx`.

you06: Yeah, I changed the code and not to use offset in the first selection.

internal/locate/region_request.go
```diff
@@ -573,18 +573,40 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
...
+				if state.option.preferLeader && (i+offset)%replicaSize == 0 {
+					// skip leader because prefer leader already tried it.
+					offset++
+				}
+				idx = AccessIndex((int(state.lastIdx) + i + offset) % replicaSize)
+			}
+		} else {
+			// If the given store is abnormal to be accessed under `ReplicaReadMixed` mode, we should choose other followers or leader
+			// as candidates to serve the Read request. Meanwhile, we should make the choice of next() meet Uniform Distribution.
+			idx = AccessIndex((int(state.lastIdx) + i + offset) % replicaSize)
```
LykxSassinator, line 594: It's a bit more complicated. Recommend just directly use the previous commit.

And here, `PreferLeader` is a Read mode which select the leader replica as priority, and only if is leader slow or abnormal, the followers will be chosen.

```suggestion
		// If the given store is abnormal to be accessed under `ReplicaReadMixed` mode, we should choose other followers or leader
		// as candidates to serve the Read request. Meanwhile, we should make the choice of next() meet Uniform Distribution.
		idx := AccessIndex((int(state.lastIdx) + i + offset) % replicaSize)
```

internal/locate/region_request.go
```diff
@@ -573,18 +573,40 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	if state.option.preferLeader {
 		state.lastIdx = state.leaderIdx
```
LykxSassinator, line 574:
```suggestion
		state.lastIdx = state.leaderIdx
		offset = 0
```

internal/locate/region_request.go
LykxSassinator: Recommend move `offset` before the judgement of `if state.option.preferLeader {`.

internal/locate/region_request.go
```diff
@@ -573,18 +573,40 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
...
 	}
+	offset := rand.Intn(replicaSize)
+	reloadRegion := false
 	for i := 0; i < replicaSize && !state.option.leaderOnly; i++ {
-		idx := AccessIndex((int(state.lastIdx) + i) % replicaSize)
-		// If the given store is abnormal to be accessed under `ReplicaReadMixed` mode, we should choose other followers or leader
-		// as candidates to serve the Read request. Meanwhile, we should make the choice of next() meet Uniform Distribution.
-		for cnt := 0; cnt < replicaSize && !state.isCandidate(idx, selector.replicas[idx]); cnt++ {
-			idx = AccessIndex((int(idx) + rand.Intn(replicaSize)) % replicaSize)
+		var idx AccessIndex
```
LykxSassinator, line 579: Still recommend to revert to the original one, current implementation is bit more complex and hard to read.

internal/locate/region_cache.go
```diff
@@ -1226,6 +1227,36 @@ func (c *RegionCache) LocateRegionByID(bo *retry.Backoffer, regionID uint64) (*K
 	}, nil
 }
 
+func (c *RegionCache) asyncReloadRegion(region *Region) {
+	if region == nil || !region.asyncReload.CompareAndSwap(false, true) {
+		// async reload triggered by other thread.
+		return
+	}
+	go func() {
```
cfzjywxk, line 1235: It can be risky to spawn a reload task for each region, especially if many regions become invalid due to the related store being marked as such. To mitigate this risk, we should implement a refreshing strategy in the region cache-related workers and consider incorporating the store's "health checker."

@crazycs520 is now considering refactoring the region cache component, he could give some advice about it.

internal/locate/region_request.go
```diff
@@ -573,18 +573,36 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	if state.option.preferLeader {
 		state.lastIdx = state.leaderIdx
 	}
+	var offset int
```
cfzjywxk, line 576: Peer selection can become quite complex when the `accessFollower` is used for all follower read and stale read processing, with various configurations. To prevent any potential issues, it may be beneficial to separate them into different categories.

`ReplicaSelector`  itself is quite complex too..

internal/locate/region_cache.go
cfzjywxk: For the v6.5.x release version, a quick fix may still be needed.

internal/locate/region_request.go
cfzjywxk: After discussing with @you06, we could do some refactoring on the master branch like
```
match access_type {
     leade_only =>
     follower_read_random =>
     follower_read_replica_only =>
     follower_read_closet => 
     stale_read_random => 
     stale_read_closest =>
}
```
while keeping the change on release-6.5 as simple as possible to reduce the risk.

you06: Maybe the refactor can be in another PR?

cfzjywxk: Ok let's merge this first

internal/locate/region_cache.go
```diff
@@ -457,6 +458,12 @@ type RegionCache struct {
 		// requestLiveness always returns unreachable.
 		mockRequestLiveness atomic.Pointer[livenessFunc]
 	}
+
+	regionsNeedReload struct {
```
cfzjywxk, line 462: Comments on the tidb-6.5 branch need to be ported here.

internal/locate/region_cache.go
```diff
@@ -1399,8 +1455,11 @@ func (mu *regionIndexMu) insertRegionToCache(cachedRegion *Region) {
 		if InvalidReason(atomic.LoadInt32((*int32)(&oldRegion.invalidReason))) == NoLeader {
 			store.workTiKVIdx = (oldRegionStore.workTiKVIdx + 1) % AccessIndex(store.accessStoreNum(tiKVOnly))
 		}
-		// Invalidate the old region in case it's not invalidated and some requests try with the stale region information.
-		oldRegion.invalidate(Other)
+		// If the region info is async reloaded, the old region is still valid.
+		if !oldRegion.asyncReload.Load() {
```
cfzjywxk, line 1459: Ditto.

cfzjywxk (approved):

crazycs520 (approved):

PR #845, opened 2023-06-19
internal: fix rpc interceptor data race

glorv:
ref: pingcap/tidb#44758

This PR changed `ChainRPCInterceptors` to always build a new interceptor to avoid data race will the original interceptor is used by other requests.


glorv:
@Connor1996 @nolouch PTAL, thanks

Connor1996 (approved):
LGTM

nolouch (approved):

PR #846, opened 2023-06-20
 reload region cache when store is resolved from invalid status (#843)

you06:
Cherry-pick #843 to tidb-6.5.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -2435,8 +2435,7 @@ func (s *Store) reResolve(c *RegionCache) (bool, error) {
 
 func (s *Store) getResolveState() resolveState {
 	if s == nil {
-		var state resolveState
-		return state
+		return unresolved
 	}
 	return resolveState(atomic.LoadUint64(&s.state))
 }
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -364,6 +364,8 @@ func (r *Region) isValid() bool {
 	return r != nil && !r.checkNeedReload() && r.checkRegionCacheTTL(time.Now().Unix())
 }
 
+type livenessFunc func(s *Store, bo *retry.Backoffer) livenessState
+
 // RegionCache caches Regions loaded from PD.
 // All public methods of this struct should be thread-safe, unless explicitly pointed out or the method is for testing
 // purposes only.
@@ -396,7 +398,7 @@ type RegionCache struct {
 	testingKnobs struct {
 		// Replace the requestLiveness function for test purpose. Note that in unit tests, if this is not set,
 		// requestLiveness always returns unreachable.
-		mockRequestLiveness func(s *Store, bo *retry.Backoffer) livenessState
+		mockRequestLiveness atomic.Pointer[livenessFunc]
 	}
 
 	regionsNeedReload struct {
@@ -2605,8 +2607,12 @@ func (s *Store) requestLiveness(bo *retry.Backoffer, c *RegionCache) (l liveness
 			return unknown
 		}
 	}
-	if c != nil && c.testingKnobs.mockRequestLiveness != nil {
-		return c.testingKnobs.mockRequestLiveness(s, bo)
+
+	if c != nil {
+		livenessFunc := c.testingKnobs.mockRequestLiveness.Load()
+		if livenessFunc != nil {
+			return (*livenessFunc)(s, bo)
+		}
 	}
 
 	if storeLivenessTimeout == 0 {
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -126,7 +126,7 @@ type Region struct {
 	syncFlag      int32          // region need be sync in next turn
 	lastAccess    int64          // last region access time, see checkRegionCacheTTL
 	invalidReason InvalidReason  // the reason why the region is invalidated
-	asyncReload   atomic.Bool    // the region need to be reloaded in async mode
+	asyncReload   int32          // the region need to be reloaded in async mode
 }
 
 // AccessIndex represent the index for accessIndex array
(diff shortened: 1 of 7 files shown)


disksing:
cannot build.

you06:
Hold this PR because it may spawns too many goroutines.

cfzjywxk:
See the discussions in https://github.com/tikv/client-go/pull/843.
TODOS for the release-v6.5 branch:
- Make the change about `acccessFollower.next` simple to reduce risk, verify the correctness.
- Verify there's no extra or hidden cache reloading logic in the `RegionCache`, the background async reload worker is the only entry to resolve the region actively.

internal/locate/region_cache.go
```diff
@@ -395,7 +398,13 @@ type RegionCache struct {
 	testingKnobs struct {
 		// Replace the requestLiveness function for test purpose. Note that in unit tests, if this is not set,
 		// requestLiveness always returns unreachable.
-		mockRequestLiveness func(s *Store, bo *retry.Backoffer) livenessState
+		mockRequestLiveness atomic.Value
+	}
+
+	regionsNeedReload struct {
```
cfzjywxk, line 404: How about using a channel so `Mutex` could be saved and operations on the `RegionCache` are already synchronized?

internal/locate/region_cache.go
```diff
@@ -2371,8 +2434,8 @@ func (s *Store) reResolve(c *RegionCache) (bool, error) {
 }
 
 func (s *Store) getResolveState() resolveState {
-	var state resolveState
 	if s == nil {
+		var state resolveState
 		return state
```
crazycs520, line 2439:
```suggestion
		return unresolved
```

internal/locate/region_cache.go
```diff
@@ -1315,8 +1375,11 @@ func (c *RegionCache) insertRegionToCache(cachedRegion *Region) {
 		if InvalidReason(atomic.LoadInt32((*int32)(&oldRegion.invalidReason))) == NoLeader {
 			store.workTiKVIdx = (oldRegionStore.workTiKVIdx + 1) % AccessIndex(store.accessStoreNum(tiKVOnly))
 		}
-		// Invalidate the old region in case it's not invalidated and some requests try with the stale region information.
-		oldRegion.invalidate(Other)
+		// If the region info is async reloaded, the old region is still valid.
+		if atomic.LoadInt32(&oldRegion.asyncReload) == 0 {
```
crazycs520, line 1379: This logic is a bit strange, why need this judgment?

you06: If the region is async reloaded, the old region info is still valid(just , so requests from other threads using the old one should be able to continue instead of failed by invalid region cache.

internal/locate/region_cache.go
```diff
@@ -466,6 +480,22 @@ func (c *RegionCache) asyncCheckAndResolveLoop(interval time.Duration) {
 				// there's a deleted store in the stores map which guaranteed by reReslve().
 				return state != unresolved && state != tombstone && state != deleted
 			})
+
+		case <-reloadRegionTicker.C:
+			for regionID := range c.regionsNeedReload.toReload {
+				c.reloadRegion(regionID)
+				delete(c.regionsNeedReload.toReload, regionID)
+			}
```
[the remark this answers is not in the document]
you06: Delay the reload to next loop which avoids some errors and backoffs, see the comment of Line491 to Line494

internal/locate/region_cache.go
cfzjywxk: It seems we can't judge if the `region` is async reloaded or not only with the condition `atomic.LoadInt32(&oldRegion.asyncReload) == 0`? The zero value could be either no-op or async reload has been finished.

If our purpose is to skip invalidation if the `insertRegionToCache` is triggered by an `async reload`, how about passing the triggering reason as a function input parameter instead of an atomic variable `asyncReload`?

you06: > how about passing the triggering reason as a function input parameter

What do you mean by passing the reason, it can be an enum or a bool value indicates whether invalidate the old one, I'm inclined to make it easy like a paramater `invalidateOldRegion bool` is enough.

cfzjywxk: @you06 
A parameter `invalidateOldRegion` bool is fine by now.

internal/locate/region_cache.go
```diff
@@ -1300,7 +1359,7 @@ func (c *RegionCache) removeVersionFromCache(oldVer RegionVerID, regionID uint64
 
 // insertRegionToCache tries to insert the Region to cache.
 // It should be protected by c.mu.Lock().
-func (c *RegionCache) insertRegionToCache(cachedRegion *Region) {
+func (c *RegionCache) insertRegionToCache(cachedRegion *Region, invalidateOldRegion bool) {
```
cfzjywxk, line 1362: Need to comment about the new parameter.

cfzjywxk (approved):

you06:
What's wrong with golangci-lint 🤔️

```text
run golangci-lint
  Running [/home/runner/golangci-lint-1.47.3-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  panic: load embedded ruleguard rules: rules/rules.go:13: can't load fmt
  
  goroutine 1 [running]:
  github.com/go-critic/go-critic/checkers.init.22()
  	github.com/go-critic/go-critic@v0.6.3/checkers/embedded_rules.go:47 +0x4b4
  
  Error: golangci-lint exit with code 2
  Ran golangci-lint in 12405ms
```

cfzjywxk:
@crazycs520 @ekexium @zyguan 
PTAL

crazycs520 (approved):
LGTM

PR #847, opened 2023-06-20
lock_resolver: let resolve lock timeout properly

zyguan:
fix https://github.com/pingcap/tidb/issues/44822 , after this PR tidb can pass the test.

### DIFF
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -693,6 +693,7 @@ func (lr *LockResolver) getTxnStatusFromLock(bo *retry.Backoffer, l *Lock, calle
 		// success before the primary region.
 		if err := bo.Backoff(retry.BoTxnNotFound, err); err != nil {
 			logutil.Logger(bo.GetCtx()).Warn("getTxnStatusFromLock backoff fail", zap.Error(err))
+			return TxnStatus{}, err
 		}
 	}
 }


disksing (approved):

txnkv/txnlock/lock_resolver.go
```diff
@@ -693,6 +693,7 @@ func (lr *LockResolver) getTxnStatusFromLock(bo *retry.Backoffer, l *Lock, calle
 		// success before the primary region.
 		if err := bo.Backoff(retry.BoTxnNotFound, err); err != nil {
 			logutil.Logger(bo.GetCtx()).Warn("getTxnStatusFromLock backoff fail", zap.Error(err))
+			return TxnStatus{}, err
```
cfzjywxk, line 696: I vaguely remember that this was intentionally designed in 4.0 to solve some problems, but I can't recall the details.
@MyonKeminta Do you remember?

cfzjywxk:
I think it's better to merge PRs affecting critical path with 2-3 approvals.

txnkv/txnlock/lock_resolver.go
cfzjywxk: It seems to address some issues with resolving v4.0 optimistic mode, but it has no impact on the current default pessimistic mode. Maybe we still need to look into historical problems.

PR #848, opened 2023-06-20
make ResourceControlInterceptor atomic

glorv:
ref https://github.com/pingcap/tidb/issues/44818

https://github.com/pingcap/tidb/pull/44819 is another attempt to fix the race but the test still failed https://do.pingcap.net/jenkins/job/pingcap/job/tidb/job/ghpr_unit_test/26946/display/redirect.

while in theory, we can init the `ResourceControlInterceptor` once when the server is bootstrapped. But due to some implementation restriction, it's hard to do so in tidb and the current code can case data race though the data race is harmless. 

So this PR changes the type to atomic ptr for simplicity.

### DIFF
--- a/internal/client/client_interceptor.go
+++ b/internal/client/client_interceptor.go
@@ -29,7 +29,6 @@ import (
 
 func init() {
 	ResourceControlSwitch.Store(false)
-	ResourceControlInterceptor = nil
 }
 
 var _ Client = interceptedClient{}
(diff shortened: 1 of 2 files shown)


nolouch (approved):
lgtm

hawkingrei:
@disksing PTAL

Connor1996 (approved):
LGTM

Connor1996:
test failed

PR #852, opened 2023-06-29
region_requst: add req info in flashback region error

Defined2014:
add more log for flashback when it meets error, ref https://github.com/pingcap/tidb/issues/43673

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1737,6 +1737,7 @@ func (s *RegionRequestSender) onRegionError(
 			zap.Stringer("req", req),
 			zap.Stringer("ctx", ctx),
 		)
+		var reqExtraInfo, ctxExtraInfo string
 		if req != nil {
 			// if the failure is caused by replica read, we can retry it with leader safely.
 			if ctx.contextPatcher.replicaRead != nil && *ctx.contextPatcher.replicaRead {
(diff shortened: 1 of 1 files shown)


internal/locate/region_request.go
```diff
@@ -1752,6 +1752,9 @@ func (s *RegionRequestSender) onRegionError(
 				return true, nil
 			}
 		}
+		logutil.BgLogger().Error("Region is in flashback progress",
+			zap.Stringer("req", req),
```
you06, line 1756: I'm inclined to contain these info inside the returned error, also consider the req can be nil.

Also, `req.String()` return weird things that no meaningful information is inside it. Maybe we should format some key fields of it.

you06:
@Defined2014 I suggest to attach the addtional info like this:
https://github.com/tikv/client-go/compare/master...you06:client-go:add-log?expand=1

Defined2014:
> @Defined2014 I suggest to attach the addtional info like this: https://github.com/tikv/client-go/compare/master...you06:client-go:add-log?expand=1

ok. I found that we already have debug log in this branch, maybe we could try open debug log firstly.

And not sure return such internal information in Error Msg is good or not. It's useless for user.

you06:
> ok. I found that we already have debug log in for this part, maybe we could try open debug log firstly.

Debug mode print too many logs...

internal/locate/region_request.go
Defined2014: Updated.

PR #858, opened 2023-06-29
fix stale read unbalance read

### LINKED ISSUE #857: Stale-Read load is unbalanced
## TiDB Cluster Topology

```yaml
tidb_servers:
  - host: 192.168.78.2
      labels: { zone: "z1", host: "tidb0" }
tikv_servers:
  - host: 192.168.80.2
    config:
      server.labels: { zone: "z1", host: "tikv0" }
  - host: 192.168.80.3
    config:
      server.labels: { zone: "z1", host: "tikv1" }
  - host: 192.168.80.4
    config:
      server.labels: { zone: "z2", host: "tikv2" }
```

Then send stale-read point-select SQL request to `tidb0`, then the stale-read request in TiKV should be balanced in `tikv0` and `tikv1`, and doesn't have stale-read request in `tikv2`. But in TiDB v6.5.3, the stale-read request in TiKV is unbalanced. Following is the TiKV metrics:

image

As you can see, the stale-read get request 

The stale-read get request on tikv0 is twice as high as on tikv1, which is not as expected, and there is no stale-read get request on tikv2, which is expected.

crazycs520:
close #857

image

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -524,21 +524,25 @@ type accessFollower struct {
 }
 
 func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector) (*RPCContext, error) {
+	matchReplicasIdx := make([]int, 0, len(selector.replicas))
+	if len(state.option.labels) == 0 {
+		for i := 0; i < len(selector.replicas); i++ {
+			if !state.tryLeader && AccessIndex(i) == state.leaderIdx {
+				continue
+			}
+			matchReplicasIdx = append(matchReplicasIdx, i)
+		}
+	} else {
+		for i, replica := range selector.replicas {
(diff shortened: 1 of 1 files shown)


internal/locate/region_request.go
```diff
@@ -524,21 +524,25 @@ type accessFollower struct {
...
-		} else {
-			if len(selector.replicas) <= 1 {
-				state.lastIdx = state.leaderIdx
-			} else {
-				// Randomly select a non-leader peer
-				state.lastIdx = AccessIndex(rand.Intn(len(selector.replicas) - 1))
-				if state.lastIdx >= state.leaderIdx {
-					state.lastIdx++
-				}
+	matchReplicasIdx := make([]int, 0, len(selector.replicas))
```
crazycs520, line 527: todo: Initialize this variable when initialize accessFollower.

PR #865, opened 2023-07-03
make cop request timeout a config paramter

SpadeA-Tang:


cfzjywxk:
There's a [proposal](https://github.com/pingcap/tidb/pull/45093) about configurable timeout that could be referenced, it's not needed to add a seperate configuration in `kv-client`.
For temporary test it's ok to add configurations as it's simple.

SpadeA-Tang:
> There's a [proposal](https://github.com/pingcap/tidb/pull/45093) about configurable timeout that could be referenced, it's not needed to add a seperate configuration in `kv-client`. For temporary test it's ok to add configurations as it's simple.

It's okay.

cfzjywxk:
Use configuration change on the master branch as the configurable timeout proposal is still under discussion.

cfzjywxk (approved):
It's needed to comment that it could be removed in the future.

SpadeA-Tang:
> It's needed to comment that it could be removed in the future.

Fine.

config/client.go
```diff
@@ -150,6 +153,7 @@ func DefaultTiKVClient() TiKVClient {
 			AdmissionMaxResultMB:  10,
 			AdmissionMinProcessMs: 5,
 		},
+		CoprReqTimeout: 60 * time.Second,
```
overvenus, line 156:
```suggestion
		CoprReqTimeout: tikv.ReadTimeoutMedium,
```

overvenus (approved):

PR #866, opened 2023-07-03
Fix the issue that primary pessimistic lock may be left not cleared after GC

MyonKeminta:
Ref https://github.com/pingcap/tidb/issues/45134

The problem is that `BatchResolveLock` calls `resolvePessimsticLock` without calling `getTxnStatus`, but `resolvePessimsticLock` assumes `getTxnStatus` (in which `CheckTxnStatus` RPC is called) is done so it doesn't do anything to primary pessimistic lock. This may result in residue pessimistic locks in some rare cases.

This PR moves `resolvePessimsticLock` calling to after `getTxnStatus` to avoid the problem.

### DIFF
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -556,6 +556,7 @@ func (txn *KVTxn) Rollback() error {
 		txn.CancelAggressiveLocking(context.Background())
 	}
 
+	// `skipPessimisticRollback` may be true only when set by failpoint in tests.
 	skipPessimisticRollback := false
 	if val, err := util.EvalFailpoint("onRollback"); err == nil {
 		if s, ok := val.(string); ok {
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -556,10 +556,26 @@ func (txn *KVTxn) Rollback() error {
 		txn.CancelAggressiveLocking(context.Background())
 	}
 
+	skipPessimisticRollback := false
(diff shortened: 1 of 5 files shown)


txnkv/transaction/txn.go
```diff
@@ -556,10 +556,26 @@ func (txn *KVTxn) Rollback() error {
 		txn.CancelAggressiveLocking(context.Background())
 	}
 
+	skipPessimisticRollback := false
```
cfzjywxk, line 559: Better to add comments if it's used just in tests.

txnkv/txnlock/lock_resolver.go
```diff
@@ -245,6 +245,12 @@ func (lr *LockResolver) BatchResolveLocks(bo *retry.Backoffer, locks []*Lock, lo
 		}
 		metrics.LockResolverCountWithExpired.Inc()
 
+		// Use currentTS = math.MaxUint64 means rollback the txn, no matter the lock is expired or not!
+		status, err := lr.getTxnStatus(bo, l.TxnID, l.Primary, 0, math.MaxUint64, true, false, l)
+		if err != nil {
+			return false, err
+		}
+
 		if l.LockType == kvrpcpb.Op_PessimisticLock {
```
cfzjywxk, line 254: Should we add some comments on `resolvePessimisticLock` that it should be called only after the `getTxnStatus` is called?

ekexium (approved):

PR #868, opened 2023-07-04
add explicit request source type to label the external request like lightning/br

nolouch:
if use
```
set @tidb_request_source_type='lightling'
select count(*) from t;
```
the request source field will like:
```
external_Select_lightning
```

### DIFF
--- a/util/request_source.go
+++ b/util/request_source.go
@@ -29,9 +29,9 @@ const (
 
 const (
 	// InternalRequest is the scope of internal queries
-	InternalRequest = "internal_"
+	InternalRequest = "internal"
 	// ExternalRequest is the scope of external queries
-	ExternalRequest = "external_"
+	ExternalRequest = "external"
 	// SourceUnknown keeps same with the default value(empty string)
 	SourceUnknown = "unknown"
 )
@@ -40,6 +40,9 @@ const (
 type RequestSource struct {
 	RequestSourceInternal bool
 	RequestSourceType     string
+	// ExplicitRequestSoureType is set from the session variable, it may specified by the client or users. `explicit_request_source_type`. it's a complement of RequestSourceType.
+	// The value maybe "lightning", "br", "dumpling" etc.
+	ExplicitRequestSoureType string
 }
 
 // SetRequestSourceInternal sets the scope of the request source.
@@ -50,6 +53,11 @@ func (r *RequestSource) SetRequestSourceInternal(internal bool) {
 // SetRequestSourceType sets the type of the request source.
 func (r *RequestSource) SetRequestSourceType(tp string) {
 	r.RequestSourceType = tp
+}
+
+// SetExplicitRequestSourceType sets the type of the request source.
+func (r *RequestSource) SetExplicitRequestSourceType(tp string) {
+	r.ExplicitRequestSoureType = tp
 }
 
 // WithInternalSourceType create context with internal source.
@@ -76,10 +84,24 @@ func (r *RequestSource) GetRequestSource() string {
 	if r == nil || r.RequestSourceType == "" {
 		return SourceUnknown
 	}
-	if r.RequestSourceInternal {
-		return InternalRequest + r.RequestSourceType
+	appendType := func(list []string) []string {
+		if len(r.ExplicitRequestSoureType) > 0 {
+			list = append(list, r.ExplicitRequestSoureType)
+			return list
+		}
+		return list
 	}
-	return ExternalRequest + r.RequestSourceType
+	if r.RequestSourceInternal {
+		labels := []string{InternalRequest, r.RequestSourceType}
+		labels = appendType(labels)
+		if len(r.ExplicitRequestSoureType) > 0 {
+			labels = append(labels, r.ExplicitRequestSoureType)
+		}
+		return strings.Join(labels, "_")
+	}
+	labels := []string{ExternalRequest, r.RequestSourceType}
+	labels = appendType(labels)
+	return strings.Join(labels, "_")
 }
 
 // RequestSourceFromCtx extract source from passed context.
--- a/util/request_source.go
+++ b/util/request_source.go
@@ -27,6 +27,18 @@ const (
 	InternalTxnMeta = InternalTxnOthers
 )
 
+// explict source types.
+const (
+	ExplicitTypeDefault    = ""
+	ExplicitTypeLightning  = "lightning"
+	ExplicitTypeBR         = "br"
+	ExplicitTypeDumpling   = "dumpling"
+	ExplicitTypeBackground = "background"
+)
+
+// ExplictTypeList is the list of all explict source types.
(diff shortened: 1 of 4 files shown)


util/request_source.go
```diff
@@ -76,10 +79,24 @@ func (r *RequestSource) GetRequestSource() string {
 	if r == nil || r.RequestSourceType == "" {
```
HuSharp, line 79:
```suggestion
	if r == nil || r.RequestSourceType == "" && r. ExplicitRequestSoureType {
```

util/request_source.go
```diff
@@ -76,10 +79,24 @@ func (r *RequestSource) GetRequestSource() string {
...
-		return InternalRequest + r.RequestSourceType
+		internalLabels := []string{InternalRequest, r.RequestSourceType}
+		internalLabels = appendType(internalLabels)
+		if len(r.ExplicitRequestSoureType) > 0 {
+			internalLabels = append(internalLabels, r.ExplicitRequestSoureType)
+		}
+		return strings.Join(internalLabels, "_")
 	}
-	return ExternalRequest + r.RequestSourceType
+	internalLabels := []string{ExternalRequest, r.RequestSourceType}
```
HuSharp, line 97:
```suggestion
	externalLabels := []string{ExternalRequest, r.RequestSourceType}
```

util/request_source.go
```diff
@@ -73,13 +81,27 @@ func IsRequestSourceInternal(reqSrc *RequestSource) bool {
...
-		return InternalRequest + r.RequestSourceType
+		labels := []string{InternalRequest, r.RequestSourceType}
+		labels = appendType(labels)
+		if len(r.ExplicitRequestSoureType) > 0 {
+			labels = append(labels, r.ExplicitRequestSoureType)
+		}
+		return strings.Join(labels, "_")
 	}
-	return ExternalRequest + r.RequestSourceType
+	labels := []string{ExternalRequest, r.RequestSourceType}
```
HuSharp, line 102: if RequestSourceType is "" && ExplicitRequestSoureType is not "", 
 It will join a string like `external__XXX` which with double `_`, 
 does it make sense?

util/request_source.go
```diff
@@ -27,11 +27,23 @@ const (
...
 )
 
+// explicit source types.
+const (
+	ExplicitTypeDefault    = ""
+	ExplicitTypeLightning  = "lightning"
+	ExplicitTypeBR         = "br"
+	ExplicitTypeDumpling   = "dumpling"
+	ExplicitTypeBackground = "background"
+)
```
HuSharp, line 37: I think they are both a type it's more convenient
```suggestion
type ExplicitType string

const (
	ExplicitTypeDefault    ExplicitType = ""
	ExplicitTypeLightning  ExplicitType = "lightning"
	ExplicitTypeBR         ExplicitType = "br"
	ExplicitTypeDumpling   ExplicitType = "dumpling"
	ExplicitTypeBackground ExplicitType = "background"
)
```

And then we can point which is background type like:
```
func IsBackgroundRequest(sourceType ExplicitType) bool {
	return sourceType != ExplicitTypeDefault
}
```
What do you think?

nolouch: It should judge by the resource group setting I think.

util/request_source_test.go
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NOTE: The code in this file is based on code from the
+// TiDB project, licensed under the Apache License v 2.0
+//
+// https://github.com/pingcap/tidb/tree/cc5e161ac06827589c4966674597c137cc9e809c/store/tikv/util/execdetails.go
```
HuSharp, line 18: I am not sure, do we need to remove the redundant License?

nolouch:
ptal @HuSharp  @glorv

util/request_source.go
```diff
@@ -73,13 +93,28 @@ func IsRequestSourceInternal(reqSrc *RequestSource) bool {
...
+	}
 	if r.RequestSourceInternal {
-		return InternalRequest + r.RequestSourceType
+		labels := []string{InternalRequest, r.RequestSourceType, r.ExplicitRequestSourceType}
+		labels = clearEmpty(labels)
+
+		return strings.Join(labels, "_")
 	}
-	return ExternalRequest + r.RequestSourceType
+	labels := []string{ExternalRequest, r.RequestSourceType, r.ExplicitRequestSourceType}
```
glorv, line 115: Currently, tikv uses the last segments of `resquest_source.split("_")` as the explicit_request_source(task_names) value, so if it is empty, the logic won't be correct. Because there are some request_source_types such as `ddl_refill`, we can't depend on the number of segments either. So I suggest always make ExplicitRequestSourceType not empty.

nolouch: changed,ptal @glorv

util/request_source.go
```diff
@@ -73,13 +94,32 @@ func IsRequestSourceInternal(reqSrc *RequestSource) bool {
...
+			if len(v) != 0 {
+				list[i] = v
+				i++
+			}
+		}
+		return list[:i]
+	}
+	explicitSourceType := r.ExplicitRequestSourceType
+	if len(explicitSourceType) == 0 {
+		explicitSourceType = ExplicitTypeDB
```
glorv, line 112: NIT: not sure if `db` is a good name for the default value. Maybe `default` is more conventional?

glorv (approved):
LGTM

HuSharp (approved):
LGTM!

util/request_source.go
```diff
@@ -40,6 +53,9 @@ const (
 type RequestSource struct {
 	RequestSourceInternal bool
 	RequestSourceType     string
+	// ExplicitRequestSourceType is set from the session variable, it may specified by the client or users. `explicit_request_source_type`. it's a complement of RequestSourceType.
```
HuSharp, line 56:
```suggestion
	// ExplicitRequestSourceType is set from the session variable, it may specified by the client or users with `tidb_request_source_type `. 
	// It's a complement of RequestSourceType.
```

util/request_source.go
```diff
@@ -73,13 +94,32 @@ func IsRequestSourceInternal(reqSrc *RequestSource) bool {
 func (r *RequestSource) GetRequestSource() string {
 	// if r.RequestSourceType is not set, it's mostly possible that r.RequestSourceInternal is not set
 	// to avoid internal requests be marked as external(default value), return unknown source here.
```
HuSharp, line 96: Maybe need to change comment

overvenus (approved):

overvenus:
Some tests fails.

HuSharp:
@overvenus Can you leave a merge label? thx!

PR #877, opened 2023-07-10
Replace stale-read traffic with local-remote traffic

you06:
By recording stale-read-miss traffic, we can know the traffic of remote reading.

But in some cases, the replica selector just skip the unavailable local replica and do a **remote stale read**. Even such stale-read is successful, billed traffic may be produced and need to be specified.

This PR will record the traffic of stale-read requests by local and remote types when there is labels. In the example metrics, the local store is in restart progress, but the stale-read success rate is still high, and we can find the remote traffic with this PR.

image

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -37,18 +37,12 @@ package locate
 import (
 	"context"
 	"fmt"
-	"github.com/tikv/client-go/v2/oracle"
-	"github.com/tikv/client-go/v2/tikv"
 	"math/rand"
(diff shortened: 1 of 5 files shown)


cfzjywxk:
@crazycs520 PTAL

cfzjywxk:
@you06 Is this included in https://github.com/tikv/client-go/pull/878?

you06:
> @you06 Is this included in #878?

Yeah, this PR can be closed.

PR #878, opened 2023-07-11
*: fix stale read ops metric

crazycs520:
The stale read ops metric is not right for some time. I found this issue when test stale read with reload all tikv node.

This PR is base on https://github.com/tikv/client-go/pull/877

image

after fixed in this PR, the `stale read ops`, `stale read req ops` and `stale read traffic` metrics is like following when reload all tikv:

image

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1002,6 +1002,13 @@ func (s *RegionRequestSender) SendReqCtx(
 	var staleReadCollector *staleReadMetricsCollector
 	if req.StaleRead {
 		staleReadCollector = &staleReadMetricsCollector{}
+		defer func() {
+			if tryTimes == 0 {
+				metrics.StaleReadHitCounter.Add(1)
+			} else {
+				metrics.StaleReadMissCounter.Add(1)
+			}
+		}()
 	}
 
 	for {
@@ -1096,7 +1103,7 @@ func (s *RegionRequestSender) SendReqCtx(
 			}
 		}
 		if staleReadCollector != nil {
-			staleReadCollector.onResp(resp, isLocalTraffic)
+			staleReadCollector.onResp(req.Type, resp, isLocalTraffic)
 		}
 		return resp, rpcCtx, nil
 	}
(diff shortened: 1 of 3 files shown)


metrics/shortcuts.go
```diff
@@ -241,8 +244,11 @@ func initShortcuts() {
 	PrewriteAssertionUsageCounterNotExist = TiKVPrewriteAssertionUsageCounter.WithLabelValues("not-exist")
 	PrewriteAssertionUsageCounterUnknown = TiKVPrewriteAssertionUsageCounter.WithLabelValues("unknown")
 
-	StaleReadHitInTraffic = TiKVStaleReadSizeSummary.WithLabelValues("hit", "in")
-	StaleReadHitOutTraffic = TiKVStaleReadSizeSummary.WithLabelValues("hit", "out")
-	StaleReadMissInTraffic = TiKVStaleReadSizeSummary.WithLabelValues("miss", "in")
-	StaleReadMissOutTraffic = TiKVStaleReadSizeSummary.WithLabelValues("miss", "out")
+	StaleReadHitCounter = TiKVStaleReadCounter.WithLabelValues("hit")
+	StaleReadMissCounter = TiKVStaleReadCounter.WithLabelValues("miss")
```
crazycs520, line 248: I prefer following change.
```suggestion
	StaleReadLocalCounter = TiKVStaleReadCounter.WithLabelValues("local")
	StaleReadRemoteCounter = TiKVStaleReadCounter.WithLabelValues("remote")
```

internal/locate/region_request.go
```diff
@@ -669,6 +670,7 @@ func newReplicaSelector(regionCache *RegionCache, regionID RegionVerID, req *tik
 		cachedRegion,
 		regionStore,
 		replicas,
+		option,
```
you06, line 673: Here we make the copy of `option`, and currently we do not use `option.leaderOnly` so it's safe, but I'm worring about the risk, because `replicaSelector.option.leaderOnly` and `accessFollower.option.leaderOnly` may be different.

For me, it's better to make `replicaSelector.option` a reference.

internal/locate/region_request.go
```diff
@@ -1666,35 +1678,36 @@ func (s *RegionRequestSender) onRegionError(bo *retry.Backoffer, ctx *RPCContext
...
-		size += req.Cop().Size()
+		size = req.Cop().Size()
 	default:
 		// ignore non-read requests
 		return
 	}
-	s.tp = req.Type
-	size += req.Context.Size()
-	s.out = size
+	size += +req.Context.Size()
```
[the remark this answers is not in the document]
crazycs520: A mistake, fixed, PTAL.

internal/locate/region_request.go
crazycs520: Great, done.

you06 (approved):

internal/locate/region_request.go
```diff
@@ -1040,6 +1044,14 @@ func (s *RegionRequestSender) SendReqCtx(
 			return resp, nil, err
 		}
 
+		var isLocalTraffic bool
+		if staleReadCollector != nil && s.replicaSelector != nil {
+			if target := s.replicaSelector.targetReplica(); target != nil {
+				isLocalTraffic = target.store.IsLabelsMatch(s.replicaSelector.labels)
```
cfzjywxk, line 1050: If there's no label set, is the logic working as expected?

internal/locate/region_request.go
```diff
@@ -1040,6 +1044,14 @@ func (s *RegionRequestSender) SendReqCtx(
...
+				staleReadCollector.onReq(req, isLocalTraffic)
```
cfzjywxk, line 1051: Should we move the statistic code after `s.sendReqToRegion` finishes with no rpc error?

internal/locate/region_request.go
crazycs520: If no label set, `isLocalTraffic` will always be true. All traffic and qps will be local. I think this is expected.

internal/locate/region_request.go
crazycs520: No, since function `staleReadCollector .onResp` will be called after rpc finished.

crazycs520: And always need to record request qps and traffic no matter request success.

cfzjywxk (approved):

PR #885, opened 2023-07-12
handle `mismatch peer id` region error

you06:
As pingcap/tidb#45297 reports, TiKV may report `mismatch peer id` error, currently the client-go doesn't handle such error, this PR checks this error and invalidate the region cache when receiving this error.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1938,6 +1938,7 @@ func (s *RegionRequestSender) onRegionError(
 	if s.replicaSelector != nil {
 		if errLabel == "mismatch_peer_id" {
 			s.replicaSelector.invalidateRegion()
+			return false, nil
 		}
 		// Try the next replica.
 		return true, nil
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1645,6 +1645,10 @@ func regionErrorToLabel(e *errorpb.Error) string {
 		return "flashback_not_prepared"
 	} else if e.GetIsWitness() != nil {
 		return "peer_is_witness"
+	} else if strings.HasPrefix(e.Message, "mismatch peer id") {
+		// the `mismatch peer id` error does not has a specific error type, so we have to match the error message.
+		// TODO: add a specific error type for `mismatch peer id`.
+		return "mismatch_peer_id"
 	}
 	return "unknown"
 }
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -1931,6 +1931,11 @@ func (s *RegionRequestSender) onRegionError(
 	)
 
 	if s.replicaSelector != nil {
+		// the `mismatch peer id` error does not has a specific error type, so we have to match the error message.
```
cfzjywxk, line 1934: The above code line `metrics.TiKVRegionErrorCounter.WithLabelValues(regionErrorToLabel(regionErr)).Inc()` should be changed to so this error is not `unknown`.

internal/locate/region_request.go
```diff
@@ -1931,6 +1931,11 @@ func (s *RegionRequestSender) onRegionError(
...
+		// TODO: add a specific error type for `mismatch peer id`.
+		if strings.HasPrefix(regionErr.Message, "mismatch peer id") {
+			s.replicaSelector.region.invalidate(NoLeader)
```
cfzjywxk, line 1937: Should we use `s.replicaSelector.invalidateRegion()` to invalide the cache?

internal/locate/region_request.go
you06: Changed.

zyguan (approved):

internal/locate/region_request.go
```diff
@@ -1931,6 +1936,9 @@ func (s *RegionRequestSender) onRegionError(
 	)
 
 	if s.replicaSelector != nil {
+		if errLabel == "mismatch_peer_id" {
+			s.replicaSelector.invalidateRegion()
```
cfzjywxk, line 1940: In this case it seems we should return `false, nil` to let the region reloading instead of retry directly.

internal/locate/region_request3_test.go
```diff
@@ -1052,3 +1052,48 @@ func (s *testRegionRequestToThreeStoresSuite) TestReplicaReadWithFlashbackInProg
...
+	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
+	bo := retry.NewBackoffer(ctx, -1)
+	s.Nil(err)
+	resp, retry, err := s.regionRequestSender.SendReq(bo, req, regionLoc.Region, time.Second)
+	s.Nil(err)
+	s.GreaterOrEqual(retry, 1)
+	regionErr, err := resp.GetRegionError()
+	s.Nil(err)
+	// return non-epoch-not-match region error and the upper layer can auto retry.
+	s.Nil(regionErr.GetEpochNotMatch())
```
cfzjywxk, line 1096: Better to add `regionErr` message assertion here.

cfzjywxk (approved):

PR #893, opened 2023-07-17
add more log

crazycs520:


internal/locate/region_request.go
```diff
@@ -557,11 +557,35 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
...
+			}
+			for _, label := range state.option.labels{
+				labels = append(labels, label.Key + ": " + label.Value)
+			}
+			for _, replica := range selector.replicas {
+				replicaStatus = append(replicaStatus, fmt.Sprintf("peer: %v, store: %v, isEpochStale: %v, IsLabelsMatch: %v, attempts: %v",
+					replica.peer.GetId(),
+					replica.store.storeID,
+					replica.isEpochStale(),
+					replica.store.IsLabelsMatch(state.option.labels),
```
cfzjywxk, line 577: Maybe the `isExhausted(1)` result could be printed for each peer too, so we know if the peer has been tried in this `replicaSelector` round. Another choice to reduce log printing is to print the detailed peer status when `epochNotMatchError` is returned, which means we print peer details just one time for one `replicaSelector` round.

Besides, it's better if we could print the reason for the failure retry on each peer, the related information would be collected at the `OnRegionError` call.

internal/locate/region_request.go
```diff
@@ -1046,6 +1071,7 @@ func (s *RegionRequestSender) SendReqCtx(
 			return nil, nil, err
 		}
 		if regionErr != nil {
+			lastRegionErr = regionErrorToLabel(regionErr)
 			retry, err = s.onRegionError(bo, rpcCtx, req, regionErr)
```
cfzjywxk, line 1075: We may need to collect information about retry reason of region error details for each `replicaSelectorRound`.

internal/locate/region_request.go
```diff
@@ -1046,6 +1085,12 @@ func (s *RegionRequestSender) SendReqCtx(
 			return nil, nil, err
 		}
 		if regionErr != nil {
+			errStr := getRegionErrorString(regionErr)
+			if len(totalErrors) == 0 || totalErrors[len(totalErrors)-1].error != errStr {
```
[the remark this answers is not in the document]

Issue #879, opened 2023-07-11
unexpected continuous cross AZ traffic after tikv shutdown


zyguan:
Let the cluster has 3 AZs and each AZ has 2 tikvs, then shutdown 1 tikv in AZ1. Since there is still one alive tikv in AZ1, it will hold all regions finally, thus replica reads with closest-replica from AZ1 can be performed on that tikv. However, the tidb from AZ1 still continuously access replicas from other AZs.

![image](https://github.com/tikv/client-go/assets/6850317/1c8f066f-1d10-4b7a-a8be-a4220f93e2a5)
![image](https://github.com/tikv/client-go/assets/6850317/0209a902-a3b2-4879-92c0-37ab97f9e6c0)

zyguan:
fixed by #1122

PR #901, opened 2023-07-24
let access-follower reload region on no candidate

zyguan:
Fix https://github.com/tikv/client-go/issues/879.

https://github.com/tikv/client-go/pull/843 only reload region when a failed store becomes reachable, thus can not handle issues like https://github.com/tikv/client-go/issues/879. This PR solve the issue by scheduling region reload when there is no candidate. To avoid reload region too frequently, it also limited the max number of pending reload regions.

With this PR, cross AZ traffic shall be reduced during the down-peer scheduling:

```
2023-07-24T05:31:31+00:00 graceful shutdown tikv@node-3:20161
2023-07-24T05:57:48+00:00 restart tikv@node-3:20161
```

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1178,8 +1178,13 @@ func (c *RegionCache) scheduleReloadRegion(region *Region) {
 	}
 	regionID := region.GetID()
 	if regionID > 0 {
+		const maxPendingReloadRegions = 512
 		c.regionsNeedReload.Lock()
-		c.regionsNeedReload.regions = append(c.regionsNeedReload.regions, regionID)
+		if len(c.regionsNeedReload.regions) < maxPendingReloadRegions {
+			c.regionsNeedReload.regions = append(c.regionsNeedReload.regions, regionID)
+		} else {
+			c.regionsNeedReload.regions[rand.Intn(maxPendingReloadRegions)] = regionID
+		}
 		c.regionsNeedReload.Unlock()
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/locate/region_cache.go
```diff
@@ -1178,8 +1178,13 @@ func (c *RegionCache) scheduleReloadRegion(region *Region) {
 	}
 	regionID := region.GetID()
 	if regionID > 0 {
+		const maxPendingReloadRegions = 512
 		c.regionsNeedReload.Lock()
-		c.regionsNeedReload.regions = append(c.regionsNeedReload.regions, regionID)
+		if len(c.regionsNeedReload.regions) < maxPendingReloadRegions {
+			c.regionsNeedReload.regions = append(c.regionsNeedReload.regions, regionID)
+		} else {
+			c.regionsNeedReload.regions[rand.Intn(maxPendingReloadRegions)] = regionID
```
cfzjywxk, line 1186: For this strategy, there seem to be starvation possibilities if `scheduleReloadRegion` keeps happening and no one can make progress.

you06: And the region whose reload is canceled should be set to non-async status(set `region.asyncReload` to 0), unless it'll not be reloaded in the future.

crazycs520: Why not just ignore in this situation? Random indicates unpredictable...

zyguan:
#1122 is the better solution.

Issue #902, opened 2023-07-24
The try times are not limited when stale read is used


cfzjywxk:
After https://github.com/tikv/client-go/pull/765/files#diff-0c36d07fbd1fc71c850ff57d323fe12bdf7f2c053ef71b7fec5bb44358f1070dR598, the retry time limit is removed when stale read is used and `leaderOnly` flag is set.

As mentioned in https://github.com/pingcap/tidb/issues/45297, if region errors are encountered and the leader peer is not available, the retry may continue to happen and cause many unnecessary retries.

cfzjywxk:
Closed by https://github.com/tikv/client-go/pull/903

PR #903, opened 2023-07-24
Resume the max retry check for stale read processing

cfzjywxk:
Ref: https://github.com/tikv/client-go/issues/902

Allow only another extra leader retry when processing the `DataIsNotReady` error.

### DIFF
--- a/internal/locate/region_request3_test.go
+++ b/internal/locate/region_request3_test.go
@@ -1032,8 +1032,13 @@ func (s *testRegionRequestToThreeStoresSuite) TestStaleReadFallback() {
 	s.Nil(err)
 	s.NotNil(regionLoc)
 	value := []byte("value")
-	isFirstReq := true
 
+	type testState struct {
+		tryTimes uint8
+		succ     bool
+	}
+
+	state := &testState{}
 	s.regionRequestSender.client = &fnClient{fn: func(ctx context.Context, addr string, req *tikvrpc.Request, timeout time.Duration) (response *tikvrpc.Response, err error) {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


crazycs520 (approved):
LGTM

internal/locate/region_request3_test.go
```diff
@@ -1018,3 +1019,58 @@ func (s *testRegionRequestToThreeStoresSuite) TestAccessFollowerAfter1TiKVDown()
 		s.Equal(0, bo.GetTotalBackoffTimes())
 	}
 }
+
+func (s *testRegionRequestToThreeStoresSuite) TestStaleReadFallback() {
```
crazycs520, line 1023: This test was also passed before this PR.

cfzjywxk: It would fail after removing the `!state.option.leaderOnly` check which is the early version.

crazycs520 (approved):

internal/locate/region_request3_test.go
crazycs520: ok

internal/locate/region_request3_test.go
```diff
@@ -1018,3 +1019,58 @@ func (s *testRegionRequestToThreeStoresSuite) TestAccessFollowerAfter1TiKVDown()
...
+		default:
+		}
+		// Return `DataIsNotReady` for the first time on leader.
+		if isFirstReq {
+			isFirstReq = false
+			return &tikvrpc.Response{Resp: &kvrpcpb.GetResponse{RegionError: &errorpb.Error{
+				DataIsNotReady: &errorpb.DataIsNotReady{},
+			}}}, nil
+		}
+		return &tikvrpc.Response{Resp: &kvrpcpb.GetResponse{Value: value}}, nil
```
you06, line 1050: It tests the max retry works, may this mock client looks like:
- 1st time returns data-is-not-ready error
- 2nd time returns other errors like unavailable store
- 3rd time fail the test because there should be leader retry limitation

cfzjywxk: Fail path test are added.

PR #905, opened 2023-07-24
ci: update actions

disksing:


you06:
By updating the actions, it's hard to guarantee tidb-6.5 branch can be built with go 1.18.

https://github.com/tikv/client-go/blob/59adec2a979819d12508d7aa103697a333c1cf16/go.mod#L3

disksing:
seems that 6.5 tidb cannot pass new linter. so cancel the PR.

PR #907, opened 2023-07-25
Only fallback to leader read with data-is-not-ready error

you06:
Inject server-is-busy error for stale read and leader read, so direct fallback to leader won't success.

fail

This patch will continue trying other replicas(the duration is high because of backoff).

success

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -517,11 +517,12 @@ func (state *tryNewProxy) onNoLeader(selector *replicaSelector) {
 type accessFollower struct {
 	stateBase
 	// If tryLeader is true, the request can also be sent to the leader.
-	tryLeader   bool
-	isStaleRead bool
-	option      storeSelectorOp
-	leaderIdx   AccessIndex
-	lastIdx     AccessIndex
+	tryLeader      bool
+	isStaleRead    bool
+	resetStaleRead bool
+	option         storeSelectorOp
+	leaderIdx      AccessIndex
+	lastIdx        AccessIndex
 }
 
 func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector) (*RPCContext, error) {
@@ -542,7 +543,7 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 		}
 	} else {
 		// Stale Read request will retry the leader only by using the WithLeaderOnly option,
-		if state.isStaleRead {
+		if state.isStaleRead && state.resetStaleRead {
 			WithLeaderOnly()(&state.option)
 			// retry on the leader should not use stale read flag to avoid possible DataIsNotReady error as it always can serve any read.
 			resetStaleRead = true
@@ -617,7 +618,7 @@ func (state *accessFollower) isCandidate(idx AccessIndex, replica *replica) bool
 		// The request can only be sent to the leader.
 		((state.option.leaderOnly && idx == state.leaderIdx) ||
 			// Choose a replica with matched labels.
-			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && replica.store.IsLabelsMatch(state.option.labels))) &&
+			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && (state.lastIdx >= 0 || replica.store.IsLabelsMatch(state.option.labels)))) &&
 		// Make sure the replica is not unreachable.
 		replica.store.getLivenessState() != unreachable
 }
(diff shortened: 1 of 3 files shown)


cfzjywxk:
@you06 
Please add descriptions and test results.

crazycs520 (approved):

crazycs520 (approved):
LGTM

internal/locate/region_request.go
```diff
@@ -1064,6 +1065,24 @@ func (s *RegionRequestSender) SendReqCtx(
 			}
 		}
 
+		//resp = nil
```
cfzjywxk, line 1068: Is this temporary?

internal/locate/region_request3_test.go
```diff
@@ -36,6 +36,7 @@ package locate
 
 import (
 	"context"
+	"github.com/tikv/client-go/v2/metrics"
```
cfzjywxk, line 39: Need to sort.

you06: It's updated in [my repo](https://github.com/you06/client-go/tree/reset-stale-read-6.5), but the display in PR is inconsistent.

internal/locate/region_request.go
you06: It's updated in [my repo](https://github.com/you06/client-go/tree/reset-stale-read-6.5), but the display in PR is inconsistent.

zyguan (approved):

internal/locate/region_request.go
```diff
@@ -1656,14 +1657,17 @@ func (s *RegionRequestSender) onRegionError(bo *retry.Backoffer, ctx *RPCContext
...
+			if af, ok := s.replicaSelector.state.(*accessFollower); ok && !af.resetStaleRead {
+				af.resetStaleRead = true
+				// always retry data-is-not-ready error.
+				return true, nil
 			}
 		}
-		return true, nil
+		// backoff other cases to avoid infinite retries.
+		err = bo.Backoff(retry.BoMaxDataNotReady, errors.New("data is not ready"))
+		return err == nil, err
```
cfzjywxk, line 1670:
```suggestion
		err = bo.Backoff(retry.BoMaxDataNotReady, errors.New("data is not ready"))
		if err != nil {
		    return false, err
		}
		return true, nil
```

It's better to return `nil` value directly and follow the error processing convention of Golang, as the `nil`, `error`, and interface types in Golang have some design drawbacks.

Consider this:
```
func testError() *TestError {
   return nil
}
func test() error {
   return testError()
}

func main() {
    err := test()
    fmt.Println(err == nil) // The result is false.
}
```

cfzjywxk:
@you06 Please resolve the conflicts.

internal/locate/region_request.go
```diff
@@ -604,7 +605,7 @@ func (state *accessFollower) isCandidate(idx AccessIndex, replica *replica) bool
 		// The request can only be sent to the leader.
 		((state.option.leaderOnly && idx == state.leaderIdx) ||
 			// Choose a replica with matched labels.
-			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && replica.store.IsLabelsMatch(state.option.labels))) &&
+			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && (state.lastIdx >= 0 || replica.store.IsLabelsMatch(state.option.labels)))) &&
 		// Make sure the replica is not unreachable.
 		replica.store.getLivenessState() != unreachable
```
cfzjywxk, line 610: It's better to split this check into different checks like the master branch here like:
```
func (state *accessFollower) isCandidate(idx AccessIndex, replica *replica) bool {
   if condition1 not match 
        return false
   if condition2 not match 
        return false
   if condition3 matches
        return true
   return false    
}
``` 
, it's quite difficult to maintain.

internal/locate/region_request.go
```diff
@@ -604,7 +605,7 @@ func (state *accessFollower) isCandidate(idx AccessIndex, replica *replica) bool
 		// The request can only be sent to the leader.
 		((state.option.leaderOnly && idx == state.leaderIdx) ||
 			// Choose a replica with matched labels.
-			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && replica.store.IsLabelsMatch(state.option.labels))) &&
+			(!state.option.leaderOnly && (state.tryLeader || idx != state.leaderIdx) && (state.lastIdx >= 0 || replica.store.IsLabelsMatch(state.option.labels)))) &&
```
cfzjywxk, line 608: Why adding `(state.lastIdx >= 0 ||` here?

you06: I change this condition to `leaderExhaust`.

The try order is local follower -> leader -> remote follower.

When `leaderExhaust` is true, we need to ignore the labels to select the remote follower.

internal/locate/region_request.go
you06: Done, and I did this for master branch before.

internal/locate/region_request.go
cfzjywxk: Is is working as expected when the `local follower` is the leader peer picked first and it's unavailable for both stale and follower read requests?
It's better to comment about the details here.

internal/locate/region_request.go
```diff
@@ -551,10 +552,11 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	}
 
 	reloadRegion := false
+	leaderExhaust := state.IsLeaderExhausted(selector.replicas[state.leaderIdx])
 	for i := 0; i < len(selector.replicas) && !state.option.leaderOnly; i++ {
 		idx := AccessIndex((int(state.lastIdx) + i) % len(selector.replicas))
 		selectReplica := selector.replicas[idx]
-		if state.isCandidate(idx, selectReplica) {
+		if state.isCandidate(idx, selectReplica, leaderExhaust) {
```
cfzjywxk, line 559: The `leaderExhaust` which is the result of `state.IsLeaderExhausted` should not be used here in the `isCandidate` function as the  `leaderExhaust` checks `isExhausted` with **2** and the `isCandidate` checks only **1**. The `leaderExhaust` is designed to be only used in the L574 `if selector.targetIdx < 0 {` branch to allow an extra retry on the leader peer if no peers are available.

cfzjywxk:
@you06 Is this to be closed?

you06:
> @you06 Is this to be closed?

Yes, with the retry policy of #910 and #916, we do not care about the type of retriable error, and remote stale-read does not make sense, so reset it anyway.

PR #910, opened 2023-07-26
add region cache state test & fix some issues of replica selector

you06:
This PR add the selector replica test for stale read, cover the most possible errors, and fix the bug the test finds:

Fixes:
- when fallback to followers from leader, stale-read flag should be removed.
- in some cases, the fallback to follower not work.
- when using stale read, allow async reload region when leader is exhausted, because leader may just be unavailable for a while, and the late update of region info won't affect stale read processing.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -969,7 +969,8 @@ func (s *replicaSelector) canFallback2Follower() bool {
 	if !state.isStaleRead {
 		return false
 	}
-	return state.lastIdx == state.leaderIdx
+	// can fallback to follower only when the leader is exhausted.
+	return state.lastIdx == state.leaderIdx && state.IsLeaderExhausted(s.replicas[state.leaderIdx])
 }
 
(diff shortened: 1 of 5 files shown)


internal/locate/region_request.go
```diff
@@ -609,14 +615,30 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	// If there is no candidate, fallback to the leader.
 	if selector.targetIdx < 0 {
 		leader := selector.replicas[state.leaderIdx]
-		leaderInvalid := leader.isEpochStale() || state.IsLeaderExhausted(leader)
+		leaderEpochStale := leader.isEpochStale()
+		leaderInvalid := leaderEpochStale || state.IsLeaderExhausted(leader)
```
cfzjywxk, line 619: Maybe we could refactor the implememtation of `state.IsLeaderExhausted`, cheking if the leader is exausted by states instead of `leader.isExhausted(2)`.

cfzjywxk (approved):

PR #916, opened 2023-07-26
fallback to follower when leader is busy

you06:
Fallback to follower when leader is busy.

Inject data-is-not-ready for stale read and server-is-busy for leader, so fallback to leader will be stucked.

fail

With this patch, server-is-busy on leader will try followers.

success

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -572,12 +572,16 @@ func (c *RPCContext) String() string {
 }
 
 type contextPatcher struct {
-	staleRead *bool
+	staleRead   *bool
+	replicaRead *bool
 }
 
 func (patcher *contextPatcher) applyTo(pbCtx *kvrpcpb.Context) {
 	if patcher.staleRead != nil {
 		pbCtx.StaleRead = *patcher.staleRead
+	}
+	if patcher.replicaRead != nil {
+		pbCtx.ReplicaRead = *patcher.replicaRead
 	}
 }
 
@@ -1191,9 +1195,11 @@ func (c *RegionCache) reloadRegion(regionID uint64) {
 		// ignore error and use old region info.
 		logutil.Logger(bo.GetCtx()).Error("load region failure",
 			zap.Uint64("regionID", regionID), zap.Error(err))
+		c.mu.RLock()
 		if oldRegion := c.getRegionByIDFromCache(regionID); oldRegion != nil {
-			atomic.StoreInt32(&oldRegion.asyncReload, 0)
+			atomic.CompareAndSwapInt32(&oldRegion.asyncReload, 1, 0)
 		}
+		c.mu.RUnlock()
 		return
 	}
 	c.mu.Lock()
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -371,6 +371,7 @@ func (state *accessKnownLeader) onNoLeader(selector *replicaSelector) {
 // the leader will be updated to replicas[0] and give it another chance.
 type tryFollower struct {
 	stateBase
+	// if the leader is unavailable, but it still holds the leadership, fallbackFromLeader is true and replica read is enabled.
 	fallbackFromLeader bool
 	leaderIdx          AccessIndex
 	lastIdx            AccessIndex
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -379,7 +379,6 @@ type tryFollower struct {
 }
 
 func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (*RPCContext, error) {
-	var targetReplica *replica
 	filterReplicas := func(fn func(*replica) bool) (AccessIndex, *replica) {
 		for i := 0; i < len(selector.replicas); i++ {
 			idx := AccessIndex((int(state.lastIdx) + i) % len(selector.replicas))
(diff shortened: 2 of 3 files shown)


internal/locate/region_cache.go
```diff
@@ -1191,9 +1195,11 @@ func (c *RegionCache) reloadRegion(regionID uint64) {
 		// ignore error and use old region info.
 		logutil.Logger(bo.GetCtx()).Error("load region failure",
 			zap.Uint64("regionID", regionID), zap.Error(err))
+		c.mu.RLock()
```
cfzjywxk, line 1198: Is this an extra fix?

internal/locate/region_request.go
```diff
@@ -371,8 +371,9 @@ func (state *accessKnownLeader) onNoLeader(selector *replicaSelector) {
 // the leader will be updated to replicas[0] and give it another chance.
 type tryFollower struct {
 	stateBase
-	leaderIdx AccessIndex
-	lastIdx   AccessIndex
+	fallbackFromLeader bool
```
cfzjywxk, line 374: Need to comment about its usage.

internal/locate/region_request.go
```diff
@@ -397,12 +398,25 @@ func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (
...
+	}
+	if state.fallbackFromLeader {
+		replicaRead := true
+		rpcCtx.contextPatcher.replicaRead = &replicaRead
+	}
+	return rpcCtx, err
 }
 
 func (state *tryFollower) onSendSuccess(selector *replicaSelector) {
-	if !selector.region.switchWorkLeaderToPeer(selector.targetReplica().peer) {
```
cfzjywxk, line 404: The former naming and meaning of the `switchWorkLeaderToPeer` function is quite confusing, I don't understand what's the purpose of it..

internal/locate/region_request.go
```diff
@@ -888,6 +902,22 @@ func (s *replicaSelector) updateLeader(leader *metapb.Peer) {
 	s.region.invalidate(StoreNotFound)
 }
 
+// For some reason, the leader is unreachable by now, try followers instead.
+func (s *replicaSelector) fallback2Follower(ctx *RPCContext) bool {
```
cfzjywxk, line 906: By now is the only situation that would be used `the stale read fallback -> leader -> fallback replicas`?

internal/locate/region_cache.go
you06: Yes, a possible data race.

internal/locate/region_request.go
you06: The former usage of `tryFollower` is after the failure of `accessKnownLeader`, in this case, if one of the follower can serve the leader-read request, it's the new leader, so switch the leader to this peer.

crazycs520 (approved):
LGTM

internal/locate/region_request.go
```diff
@@ -397,12 +399,25 @@ func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (
...
-	return selector.buildRPCContext(bo)
+	rpcCtx, err := selector.buildRPCContext(bo)
+	if err != nil || rpcCtx == nil {
+		return rpcCtx, err
+	}
+	if state.fallbackFromLeader {
+		replicaRead := true
+		rpcCtx.contextPatcher.replicaRead = &replicaRead
+	}
+	return rpcCtx, err
```
cfzjywxk, line 410:
```suggestion
	return rpcCtx, nil
```

internal/locate/region_request.go
you06: Yes, when fallbacking to replica from leader, it's a follower read request, not stale read.

ekexium (commented):
Is it necessary to give it a hint or constraint on which follower to try (first)? For example we may want it to do the follower read in its local zone as much as possible?

you06:
> Is it necessary to give it a hint or constraint on which follower to try (first)? For example we may want it to do the follower read in its local zone as much as possible?

Implemented this strategy, PTAL.

internal/locate/region_request.go
```diff
@@ -371,38 +371,78 @@ func (state *accessKnownLeader) onNoLeader(selector *replicaSelector) {
...
+			selector.targetIdx = idx
+			targetReplica = selectReplica
 		}
-		targetReplica = selector.replicas[idx]
-		// Each follower is only tried once
-		if !targetReplica.isExhausted(1) && targetReplica.store.getLivenessState() != unreachable {
+		// labels only take effect for first try.
+		state.labels = nil
+	}
+	if targetReplica == nil {
```
cfzjywxk, line 409: Is it possible to do the checks with the `selector.targetIdx` so we could avoid using two states?

internal/locate/region_request.go
```diff
@@ -371,38 +371,75 @@ func (state *accessKnownLeader) onNoLeader(selector *replicaSelector) {
...
 		}
-		targetReplica = selector.replicas[idx]
-		// Each follower is only tried once
-		if !targetReplica.isExhausted(1) && targetReplica.store.getLivenessState() != unreachable {
+		return -1, nil
+	}
+
+	if len(state.labels) > 0 {
+		idx, selectReplica := filterReplicas(func(selectReplica *replica) bool {
+			return selectReplica.store.IsLabelsMatch(state.labels)
```
cfzjywxk, line 398: Is the `selectReplica.isExhausted(1)` check missing here? How about putting it into the default `filterReplicas` and pass a `nil` checker function if there's no labels?

you06: The replica may be exhausted by data-is-not-ready, which does not affect follower read.

cfzjywxk (approved):

PR #921, opened 2023-08-02
tikv/safeTS: obtain min-resolved-ts by store list

HuSharp:
### Background
Nowadays we have 2 api interface for obtaining min resolved ts

There are 2 repair step in this pr:
1. We can use the `approach I` to get cluster's min resolved ts when @@txn_scope is global.
2. For client-go's updateSafeTS, we call `approach II` for each store, which is not necessary.
We can extend the `approach I` params to obtain specified store via a list to reduce the cost per call after https://github.com/tikv/pd/pull/6880 merged

#### step 1
judge @@txn_scope value which from the config:

- when find is global, we can get cluster-level resolved ts to avoid using kv grpc.
- when find is not global, we need to use kv grpc.

#### step 2
use `approach I`'s extension to get all stores.


HuSharp:
@JmPotato @lhy1024 PTAL as well, thx!

integration_tests/go.mod
```diff
@@ -112,7 +112,5 @@ require (
 
 replace (
 	github.com/go-ldap/ldap/v3 => github.com/YangKeao/ldap/v3 v3.4.5-0.20230421065457-369a3bab1117
-	github.com/pingcap/tidb => github.com/HuSharp/tidb v1.1.0-beta.0.20230726045237-a2b0085ad7c5
-	github.com/pingcap/tidb/parser => github.com/HuSharp/tidb/parser v0.0.0-20230726045237-a2b0085ad7c5
```
HuSharp, line 116: FIY, here update is independent of this pr, just for the other merged pr which wants to pass ci...

tikv/kv.go
```diff
@@ -584,6 +584,23 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 	tikvClient := s.GetTiKVClient()
 	wg := &sync.WaitGroup{}
 	wg.Add(len(stores))
+	// Try to get the minimum resolved timestamp of the store from PD.
+	var (
+		err error
+	)
+	var storeIDs []string
```
JmPotato, line 591:
```suggestion
	var (
		storeIDs []string
		err           error
	)
```

tikv/kv.go
```diff
@@ -595,16 +612,11 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 
 			var (
 				safeTS uint64
-				err    error
 			)
```
JmPotato, line 615: Please use `var safeTS uint64` if there is only one variable to init.

tikv/kv.go
```diff
@@ -595,16 +612,11 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
-			storeIDStr := strconv.Itoa(int(storeID))
-			// Try to get the minimum resolved timestamp of the store from PD.
 			if s.pdHttpClient != nil {
```
JmPotato, line 616: No need to do this check again.

tikv/kv.go
```diff
@@ -595,16 +612,11 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
-				safeTS, err = s.pdHttpClient.GetStoreMinResolvedTS(ctx, storeID)
-				if err != nil {
-					logutil.BgLogger().Debug("get resolved TS from PD failed", zap.Error(err), zap.Uint64("store-id", storeID))
-				}
+				safeTS = StoreMinResolvedTSs[storeID]
 			}
+			storeIDStr := strconv.FormatUint(storeID, 10)
```
JmPotato, line 619: What about moving this into the `if` block below?

HuSharp: - metrics need string type to report... so need to at if block outer
- StoreMinResolvedTSs need unit type, so we cannot put it in func's params :(

.github/workflows/integration.yml
```diff
@@ -69,7 +69,10 @@ jobs:
           mv ../${{steps.extract-tikv.outputs.destination}}/tikv-server .
           ./pd-server > pd.log 2>&1 &
           sleep 5
-          ./tikv-server -C tikv.toml > tikv.log 2>&1 &
+          TIKV_COUNT=3
+          for i in $(seq $TIKV_COUNT); do
+            ./tikv-server -C tikv.toml > "tikv${i}.log" 2>&1 &
```
JmPotato, line 74: Will this cause the 3 TiKV nodes to use the same configuration? The listen address, data path, etc, may be conflicted.

tikv/kv.go
```diff
@@ -584,6 +584,23 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
 	wg.Add(len(stores))
+	// Try to get the minimum resolved timestamp of the store from PD.
+	var (
+		storeIDs []string
+		err      error
+	)
+	for _, store := range stores {
+		storeIDs = append(storeIDs, strconv.FormatUint(store.StoreID(), 10))
+	}
+	StoreMinResolvedTSs := make(map[uint64]uint64)
```
JmPotato, line 595: Why make it exported?

HuSharp: for `for` block will use it.

.github/workflows/integration.yml
HuSharp: Oh, good catch! changed it for diff binding, PTAL again, thx!

tikv/kv.go
JmPotato: The exportable attribute is only meaningful when referenced across packages. The definition and use of `StoreMinResolvedTSs` are in the same function, so it is meaningless to make it exportable here.

HuSharp: Sorry... I misunderstand your mean...

HuSharp:
@nolouch PTAL, thx!

lhy1024 (approved):

HuSharp:
friendly ping :) @JmPotato @nolouch

tikv/kv.go
```diff
@@ -584,6 +584,21 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 	tikvClient := s.GetTiKVClient()
 	wg := &sync.WaitGroup{}
 	wg.Add(len(stores))
+	// Try to get the minimum resolved timestamp of the store from PD.
+	var err error
+	var storeMinResolvedTSs map[uint64]uint64
```
JmPotato, line 589:
```suggestion
	var (
		err error
		storeMinResolvedTSs map[uint64]uint64
	)
```

tikv/kv.go
```diff
@@ -593,20 +608,10 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
-			storeIDStr := strconv.Itoa(int(storeID))
-			// Try to get the minimum resolved timestamp of the store from PD.
-			if s.pdHttpClient != nil {
-				safeTS, err = s.pdHttpClient.GetStoreMinResolvedTS(ctx, storeID)
-				if err != nil {
-					logutil.BgLogger().Debug("get resolved TS from PD failed", zap.Error(err), zap.Uint64("store-id", storeID))
-				}
-			}
+			var safeTS uint64
+			storeIDStr := strconv.FormatUint(storeID, 10)
```
JmPotato, line 612: Could we reuse the results from lines 591~594?

util/pd.go
```diff
@@ -86,45 +86,65 @@ func NewPDHTTPClient(
...
+						// Should be val.(uint64) but failpoint doesn't support that.
+						if tmp, ok := val.(int); ok {
+							d.StoresMinResolvedTS[storeID] = uint64(tmp)
+							logutil.BgLogger().Info("inject min resolved ts", zap.Uint64("storeID", storeID), zap.Uint64("ts", uint64(tmp)))
+						}
+					}
+				}
+			} else {
 				// Should be val.(uint64) but failpoint doesn't support that.
 				if tmp, ok := val.(int); ok {
```
JmPotato, line 133:
```suggestion
			} else if tmp, ok := val.(int); ok {
				// Should be val.(uint64) but failpoint doesn't support that.
```

tikv/kv.go
HuSharp: good catch, reused by `for range index`

integration_tests/pd_api_test.go
```diff
@@ -103,42 +105,107 @@ func (c *storeSafeTsMockClient) CloseAddr(addr string) error {
 	return c.Client.CloseAddr(addr)
 }
 
-func (s *apiTestSuite) TestGetStoreMinResolvedTS() {
+func (s *apiTestSuite) TestGetStoresMinResolvedTS() {
 	util.EnableFailpoints()
-	// Try to get the minimum resolved timestamp of the store from PD.
 	require := s.Require()
+	require.Nil(failpoint.Enable("tikvclient/mockFastSafeTSUpdater", `return()`))
```
JmPotato, line 111:
```suggestion
	require.NoError(failpoint.Enable("tikvclient/mockFastSafeTSUpdater", `return()`))
```

integration_tests/pd_api_test.go
```diff
@@ -103,42 +105,107 @@ func (c *storeSafeTsMockClient) CloseAddr(addr string) error {
...
+	defer func() {
+		require.Nil(failpoint.Disable("tikvclient/mockFastSafeTSUpdater"))
```
JmPotato, line 113: Ditto.

tikv/kv.go
```diff
@@ -655,6 +672,40 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
 }
 
+var (
+	skipSafeTSUpdateCounter    = metrics.TiKVSafeTSUpdateCounter.WithLabelValues("skip", "cluster")
+	successSafeTSUpdateCounter = metrics.TiKVSafeTSUpdateCounter.WithLabelValues("success", "cluster")
+	clusterMinSafeTSGap        = metrics.TiKVMinSafeTSGapSeconds.WithLabelValues("cluster")
+)
+
+// setClusterMinSafeTSByPD check whether it is needed to get cluster-level's min resolved ts from PD.
+func (s *KVStore) setClusterMinSafeTSByPD(ctx context.Context) bool {
```
JmPotato, line 682: This function's name sounds like a method that will set something, which is not, however.

integration_tests/pd_api_test.go
HuSharp: seems that this usage in `client-go` needs to be unified

tikv/kv.go
HuSharp: updated, what do u think :)

JmPotato (approved):

PR #928, opened 2023-08-07
prefix safepoint kv with keyspace name

AmoebaProtozoa:
Add a optional namespace prefix when creating spkv, this allows tidb to create prefixed spkv to be used in gcV2 when each keyspace have their own safepoint.

### DIFF
--- a/tikv/safepoint.go
+++ b/tikv/safepoint.go
@@ -46,6 +46,7 @@ import (
 	"github.com/tikv/client-go/v2/internal/logutil"
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/client/v3/namespace"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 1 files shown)


txnkv/client.go
```diff
@@ -93,6 +99,8 @@ func NewClient(pdAddrs []string, opts ...ClientOpt) (*Client, error) {
 		if err != nil {
 			return nil, err
 		}
+		// Set etcd namespace prefix if we are using api v2.
+		namespacePrefix = fmt.Sprintf(tidbKeyspaceEtcdPathPrefix+"%d", codecCli.GetCodec().GetKeyspaceID())
```
iosmanthus, line 103: How about making the namespace prefix an option for `NewClient`, this hard-coded path might couple the logic with TiDB.

AmoebaProtozoa: done, PTAL

tikv/safepoint.go
```diff
@@ -123,11 +124,17 @@ type EtcdSafePointKV struct {
 }
 
 // NewEtcdSafePointKV creates an instance of EtcdSafePointKV
-func NewEtcdSafePointKV(addrs []string, tlsConfig *tls.Config) (*EtcdSafePointKV, error) {
+func NewEtcdSafePointKV(addrs []string, tlsConfig *tls.Config, namespacePrefix string) (*EtcdSafePointKV, error) {
```
iosmanthus, line 127: How about making this an option too?

iosmanthus (approved):

PR #929, opened 2023-08-07
log: group replica selector logging and split not leader errors

cfzjywxk:
1. Refactor and group the logging-related logic into `RegionRequestSender` and `replicaSelector`.
2. Split the `totalErros` into details for `NotLeader` error responses, 
from
```
[total-region-errors="not_leader-nil:3"]
```
```
[Prefix: target peer id]-not_leader-[Suffix: Leader information in the returned error and nil if no leader]
[total-region-errors="6-not_leader-nil:1, 4-not_leader-nil:1, 5-not_leader-nil:1"]
```

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -112,6 +112,10 @@ type RegionRequestSender struct {
 	RegionRequestRuntimeStats
 }
 
+func (s *RegionRequestSender) String() string {
+	return fmt.Sprintf("{replicaSelector: %v}", s.replicaSelector.String())
+}
+
 // RegionRequestRuntimeStats records the runtime stats of send region requests.
 type RegionRequestRuntimeStats struct {
 	Stats map[tikvrpc.CmdType]*RPCRuntimeStats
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -271,6 +275,64 @@ type replicaSelector struct {
...
+			}
+		}
+		for _, replica := range s.replicas {
+			replicaStatus = append(replicaStatus, fmt.Sprintf("peer: %v, store: %v, isEpochStale: %v, "+
+				"attempts: %v, replica-epoch: %v, store-epoch: %v, store-state: %v, store-liveness-state: %v",
+				replica.peer.GetId(),
+				replica.store.storeID,
+				replica.isEpochStale(),
+				replica.attempts,
+				replica.epoch,
```
ekexium, line 325: Do we need to use atomic load on this variable?

cfzjywxk: Thanks for the reminder. I understand the `replica` should not be used concurrently as its lifetime is the same as the `replicaSelector`, but I forget why atomic operations are used for the epoch updating.

ekexium (approved):

internal/locate/region_request3_test.go
```diff
@@ -1284,3 +1284,31 @@ func (s *testRegionRequestToThreeStoresSuite) TestSendReqFirstTimeout() {
 		}
 	}
 }
+
+func (s *testRegionRequestToThreeStoresSuite) TestLogging() {
```
you06, line 1288: Seems a manual test?

you06 (approved):
rest LGTM

internal/locate/region_request3_test.go
cfzjywxk: Yes.

PR #937, opened 2023-08-11
Fix the check of scatter status

rleungx:
Ref https://github.com/pingcap/tidb/issues/46001


rleungx:
Hold it since there are some problems.

Issue #924, opened 2023-08-03
code annotation misleading 


jiyfhust:
https://github.com/tikv/client-go/blob/cbfd64018c2194af29694547bfcad1972cb24512/internal/locate/region_cache.go#L993

```
// LoadRegionsInKeyRange lists regions in [start_key,end_key].
func (c *RegionCache) LoadRegionsInKeyRange(bo *retry.Backoffer, startKey, endKey []byte) (regions []*Region, err error) {
	var batchRegions []*Region
	for {
		batchRegions, err = c.Bat
....
....
```
Actually it is regions in [start_key,end_key), not [start_key,end_key], right?
It is Puzzling.

Maybe somewhere else same in file "internal/locate/region_cache.go".

PR #941, opened 2023-08-12
fix misleading annotation of LoadRegionsInKeyRange

jiyfhust:
The annotation of LoadRegionsInKeyRange seems to be misleading.
I think it is better to fix it. 
It once wasts time of me when i am troubleshooting a tidb bug.  

As issue #924 describe.


bufferflies (approved):

bufferflies:
plz fix the DCO check.

disksing:
low risk, merging.

PR #942, opened 2023-08-14
add region cache state test & fix some issues of replica selector (#910) 

### LINKED ISSUE #906: When leader is selected as target peer at first, no other replicas could be retied for stale read
When the leader peer is first picked as the target with labels and it's unavailable because of some region errors, the replica selector would keep trying the leader peer but not recover though replica peers may be able to process this request successfully.

For example:
1. consider there are 3AZs with 1 tidb and 1 tikv server in each AZ, and there is a region whose leader peer is in AZ a. 
2. Then a stale read request is triggered by the tidb-server in AZ a, the tikv-server in AZ a is selected as the target peer because of the label configurations. 
3. Meanwhile, network partition happens between the tidb and tikv nodes in AZ a, and when errors are encountered processing stale read requests the leader peer is unconditionally retried. 
4. The leader is unavailable and tidb kept raising pseudo epoch not match errors and retrying those peers.

you06:
Cherry pick #910 to master branch.
Close #906


HuSharp:
seems like `TestRegionCacheStaleRead` is a flaky test :(
Can you help to check it?
https://github.com/tikv/client-go/actions/runs/5865148918/job/15901509130?pr=944#step:4:521
https://github.com/tikv/client-go/actions/runs/5873468298/job/15926773544?pr=947#step:4:522

you06:
> seems like `TestRegionCacheStaleRead` is a flaky test :(
> Can you help to check it?

Will take a look.

cfzjywxk (approved):

zyguan (approved):

PR #943, opened 2023-08-15
rawkv: add TTL argument for client method CompareAndSwap

tutububug:


tutububug:
to fix issue #371 .

disksing:
Hi, @tutububug , can you make it an option so that it won't break the API?

tutububug:
Actually, in our use scenario, we use the CAS method to implement the distribute lock function. 
The key names 'lock', and the value names 'locker'. Users can lock or unlock a key only when its value equals the value the user passes. In normal cases there is no problem, but when the locker client crashes, there should be a semantic 'lock timeout', so the TTL here needs to be added to the CAS parameters.

Smityz:
@disksing how about adding a new function `CompareAndSwapTTL`

PR #944, opened 2023-08-15
min-resolved-ts: check dc label

HuSharp:
### Background
1. api interface
	Nowadays we have 2 api interfaces for obtaining min resolved ts

	For client-go's updateSafeTS, we call `approach II` for each store, which is not necessary to call API so many times 
We can use the `approach I` to get cluster's min resolved ts when `@@txn_scope` is `global`

2. avoid using kv request reason:
if TiDB failed to connect a TiKV store to fetch the latest resolved_ts, it will remain the out-of-dated value until the TiKV comes back, though PD will mark the store as unavailable and keep updating the minimal resolved_ts with other TiKV stores. We may need to keep these two behaviors the same or unify them.

### Summary pr
judge `@@txn_scope` value which from the config:
- when find is `global`, we can get cluster-level resolved ts to avoid using kv grpc.
- when find is not `global`, we need to use kv grpc.

### DIFF
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -542,33 +542,13 @@ func (s *KVStore) safeTSUpdater() {
 	}
 }
 
-var (
-	skipSafeTSUpdateCounter    = metrics.TiKVSafeTSUpdateCounter.WithLabelValues("skip", "cluster")
-	successSafeTSUpdateCounter = metrics.TiKVSafeTSUpdateCounter.WithLabelValues("success", "cluster")
-	clusterMinSafeTSGap        = metrics.TiKVMinSafeTSGapSeconds.WithLabelValues("cluster")
-)
-
 func (s *KVStore) updateSafeTS(ctx context.Context) {
-	stores := s.regionCache.GetStoresByType(tikvrpc.TiKV)
-	// Try getting the cluster-level minimum resolved timestamp from PD first.
-	clusterMinSafeTS, txnScopeMap, err := s.buildTxnScopeMap(ctx, stores)
-	if clusterMinSafeTS != 0 && err == nil {
-		preClusterMinSafeTS := s.GetMinSafeTS(oracle.GlobalTxnScope)
-		if preClusterMinSafeTS > clusterMinSafeTS {
-			skipSafeTSUpdateCounter.Inc()
-			preSafeTSTime := oracle.GetTimeFromTS(preClusterMinSafeTS)
-			clusterMinSafeTSGap.Set(time.Since(preSafeTSTime).Seconds())
-			return
-		} else {
-			s.minSafeTS.Store(oracle.GlobalTxnScope, clusterMinSafeTS)
-			successSafeTSUpdateCounter.Inc()
-			safeTSTime := oracle.GetTimeFromTS(clusterMinSafeTS)
-			clusterMinSafeTSGap.Set(time.Since(safeTSTime).Seconds())
-		}
-
+	// Try to get the cluster-level minimum resolved timestamp from PD first.
+	if s.setClusterMinSafeTSByPD(ctx) {
 		return
 	}
 
+	stores := s.regionCache.GetStoresByType(tikvrpc.TiKV)
 	tikvClient := s.GetTiKVClient()
 	wg := &sync.WaitGroup{}
 	wg.Add(len(stores))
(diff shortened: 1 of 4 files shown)


HuSharp:
@nolouch @JmPotato @lhy1024 PTAL, thx!

lhy1024 (commented):
ci failed

HuSharp:
> ci failed

seems like is a non-related failed

we can check another pr meeting as well. https://github.com/tikv/client-go/actions/runs/5873468298/job/15926773544?pr=947

lhy1024 (approved):

tikv/kv.go
```diff
@@ -601,18 +592,40 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 		}(ctx, wg, storeID, storeAddr)
 	}
 
-	txnScopeMap := make(map[string][]uint64)
+	if clusterTS != 0 && err == nil {
```
nolouch, line 595: Its better improve the readable like:
```
clusterMinSafeTS, txnScopeMap, err := s.buildTxnScopeMap(ctx, stores)
// fast path get cluster scope MinSafeTs
if clusterMinSafeTS!= 0 && err == nil {
     s.minSafeTS.Store(oracle.GlobalTxnScope, clusterMinSafeTS)
     return
}

// Get SafeTS from TiKV by Store
for _, store := range stores {
...
}
```

HuSharp: good idea! fixed.

you06:
I check out the flacky test, can you help change L255 of the test into `atomic.StoreInt64(&reloadRegionInterval, originReloadRegionInterval)` in this PR?
https://github.com/tikv/client-go/blob/6f2442951b2bf717081ff383e3164f84675bf3d9/internal/locate/region_request_state_test.go#L255

HuSharp:
> I check out the flacky test, can you help change L255 of the test into `atomic.StoreInt64(&reloadRegionInterval, originReloadRegionInterval)` in this PR?
> 
> https://github.com/tikv/client-go/blob/6f2442951b2bf717081ff383e3164f84675bf3d9/internal/locate/region_request_state_test.go#L255
> 
> I will update other branches.

fixed

tikv/kv.go
```diff
@@ -601,18 +614,36 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
+		s.updateMinSafeTS(txnScope, storeIDs)
+	}
+	wg.Wait()
+}
+
+// build txnScopeMap and judge whether it is needed to get safeTS from PD.
+// - if stores label are global, return get cluster min resolved ts from pd.
+// - if contains dc label store, return try to get it from TiKV.
+func (s *KVStore) buildTxnScopeMap(ctx context.Context, stores []*Store) (safeTS uint64, txnScopeMap map[string][]uint64, err error) {
+	isGlobal := true
```
nolouch, line 627: We do not need judge from the tikv labels, I think should judge from the config like the usage in tidb:
https://github.com/pingcap/tidb/blob/88225787f3c3da18323415c72208bce20876009a/expression/builtin_time.go#L6617-L6632

HuSharp: go it

tikv/kv.go
```diff
@@ -615,6 +608,40 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
+)
+
+// setClusterMinSafeTSByPD try to get cluster-level's min resolved timestamp from PD when @@txn_scope is `global`.
+func (s *KVStore) setClusterMinSafeTSByPD(ctx context.Context) bool {
+	isGlobal := config.GetTxnScopeFromConfig() == oracle.GlobalTxnScope
+	// Try to get the minimum resolved timestamp of the cluster from PD.
+	if s.pdHttpClient != nil && isGlobal {
+		clusterMinSafeTS, err := s.pdHttpClient.GetClusterMinResolvedTS(ctx)
+		if err != nil {
+			logutil.BgLogger().Debug("get cluster-level min resolved timestamp from PD failed", zap.Error(err))
```
nolouch, line 624: I think we can warn it, it is called every 2 secs.

nolouch (commented):
rest lgtm

tikv/kv.go
HuSharp: When not enable `min-resolved-ts` I think it's a really noisy info... What do you think?
image

HuSharp: For clusters upgraded from v6.0.0~v6.2.0 will meet this error, does it make sense?
 https://docs.pingcap.com/tidb/stable/pd-configuration-file#min-resolved-ts-persistence-interval-new-in-v600

iosmanthus (approved):

PR #945, opened 2023-08-15
gc: add resolve locks interface for tidb gc_worker

3pointer:
This PR is partial of https://github.com/pingcap/tidb/pull/45904.

It tried to abstract a new inteface of gc lock resolver for GCWorker and log backup.


tikv/gc.go
```diff
@@ -81,8 +81,9 @@ func WithConcurrency(concurrency int) GCOpt {
 }
 
 func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrency int) error {
+	lockResolver := &BaseLockResolver{store: s}
 	handler := func(ctx context.Context, r kv.KeyRange) (rangetask.TaskStat, error) {
-		return s.resolveLocksForRange(ctx, safePoint, r.StartKey, r.EndKey)
+		return ResolveLocksForRange(ctx, "GC", lockResolver, safePoint, r.StartKey, r.EndKey)
```
MyonKeminta, line 86: If the second parameter is named `uuid`, it doesn't make sense to pass `"GC"` to it. As there isn't a GCWorker here, consider generating one.
Actually I think we don't need to keep `uuid` anyway. For example, we can call it `identifier`, and pass it `"gc-client-go-api"` here, and pass `"gcworker-{uuid}"` from TiDB.

tikv/gc.go
```diff
@@ -94,72 +95,115 @@ func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrenc
...
+	bo    *Backoffer
+}
+
+func NewBaseLockResolver(store Storage) *BaseLockResolver {
+	return &BaseLockResolver{
+		store: store,
+	}
+}
+
+func (l *BaseLockResolver) ResolveLocks(ctx context.Context, locks []*txnlock.Lock, loc *locate.KeyLocation) (*locate.KeyLocation, error) {
```
MyonKeminta, line 109: * The name `BaseLockResolver` looks confused with those stuff in `lock_resovler.go`. Consider distinguish them by naming this `BaseGCLockResolver`. (and btw what does `Base` mean here?)
* The function `ResolveLocks` looks also confused with those methods in `lock_resolver.go`, which is used by transactions that meet stale locks. `BatchResolveLocks`, which is used for GC, has some differences with `ResolveLocks`. Here I suggest you to consider still using the name `BatchResolveLocks`, or more clearly, using somethig like `ResolveLocksForGC`.

tikv/gc.go
```diff
@@ -94,72 +95,115 @@ func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrenc
...
+	// otherwise create it to prevent panic.
+	if l.bo == nil {
+		l.bo = NewGcResolveLockMaxBackoffer(ctx)
+	}
+	return batchResolveLocksInARegion(l.bo, l.GetStore(), locks, loc)
+}
+
+func (l *BaseLockResolver) ScanLocks(ctx context.Context, key []byte, maxVersion uint64) ([]*txnlock.Lock, *locate.KeyLocation, error) {
+	// every time ScanLocks need a new backoffer
+	l.bo = NewGcResolveLockMaxBackoffer(ctx)
```
MyonKeminta, line 120: How about providing the backoff from the caller?
I'm afraid it might easily cause misusing to making it a state of the lock resolver.

tikv/gc.go
```diff
@@ -94,72 +95,115 @@ func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrenc
...
+		resolvedLocation, err := resolver.ResolveLocks(ctx, locks, loc)
+		if err != nil {
+			return stat, err
 		}
 		// resolve locks failed since the locks are not in one region anymore, need retry.
 		if resolvedLocation == nil {
 			continue
 		}
-		if len(locks) < gcScanLockLimit {
+		if len(locks) < GCScanLockLimit {
```
MyonKeminta, line 172: If this function is going to be used by TiDB, please also see if TiDB has its own settings of the scan limit. If so, consider making it customizable.

3pointer: TiDB has a `mergeLockScanner` which use `gcScanLockLimit` to scan locks.

tikv/gc.go
3pointer: the origin logic is `ScanLocks` and `ResolveLocks` **must use backoff and must share one backoff**, and gc worker test using failpoint to control the backoff. 
So I think backoff is an internal logic for GCWorker and the test doesn't need it.

If we public it, then it should be caller's responsibility to call `SetBackoffer` and `GetBackoffer` when using `ScanLocks` and `ResolveLocks` for LockResolver? and caller may not know they need **share** the same backoff. Do you think it's still ok?

tikv/gc.go
3pointer: Got it

tikv/gc.go
3pointer: For TiDB GC Worker. it use BaseLockResolver to scan and resolve locks.
For TiDB Advacner(log backup). it use BaseLockResolver to scan locks and **override** `ResolveLocks` to handle locks with a small region range and check txn status one by one.(which is not BatchResolveLocks, force delete locks before safepoint)

So, BaseGCLockResolver is ok for me. and `BatchResolveLocks` seems not clear.

tikv/gc.go
3pointer: They both use the same setting 128. I think we can unified here

tikv/gc.go
```diff
@@ -94,72 +95,124 @@ func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrenc
...
 		select {
 		case <-ctx.Done():
 			return stat, errors.New("[gc worker] gc job canceled")
 		default:
 		}
-
-		locks, loc, err := s.scanLocksInRegionWithStartKey(bo, key, safePoint, gcScanLockLimit)
+		// create new backoffer for every scan and resolve locks
+		bo := createBackoffFn(ctx)
+		locks, loc, err := resolver.ScanLocks(bo, key, maxVersion, scanLimit)
```
MyonKeminta, line 168: TiDB has this test code: 
```go
		if w.testingKnobs.scanLocks != nil {
			locks = append(locks, w.testingKnobs.scanLocks(key, loc.Region.GetID(), tryResolveLocksTS)...)
		}
```
This is used to inject some simulated locks for tests. Did you notice it, and do you have any idea how to support that kind of tests after the change?
This

3pointer: Yes，In TiDB test a new test struct will override `ScanLock` method to generate expected locks.

tikv/gc.go
```diff
@@ -94,72 +95,124 @@ func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrenc
...
+		store:      store,
+	}
+}
+
+func (l *BaseLockResolver) Identifier() string {
+	return l.identifier
+}
+
+func (l *BaseLockResolver) ResolveLocks(bo *Backoffer, locks []*txnlock.Lock, loc *locate.KeyLocation) (*locate.KeyLocation, error) {
+	return batchResolveLocksInARegion(bo, l.GetStore(), locks, loc)
```
MyonKeminta, line 115: The function `batchResolveLocksInARegion` only resolves locks in the single region specified by `loc`. If error happens, it may reload the region info, but still only handles the first region containing the key `locks[0]`. If `locks` locates in more than one region, the remaining will be ignored. This works in the GC logic (or the scanlocks-resolvelocks for range logic) because it guarantees `locks` passed to this function is sorted, and the returned `KeyLocation` indicates the next key to scan the next batch, so no lock will be missed. However if you public this function to be accessible everywhere, it's very weird that it doesn't promise to completely handle the given locks and returns a `KeyLocation`, and is very very very likely to cause misusing. Please consider adjust the implementation to suite the interface's semantics or any other way to avoid the problem.

tikv/gc.go
```diff
@@ -94,72 +98,129 @@ func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrenc
...
+
+	// ResolveLocksInOneRegion tries to resolve expired locks for one region.
+	// 1. For GCWorker it will scan locks before *safepoint*,
+	// and force remove these locks. rollback the txn, no matter the lock is expired of not.
+	// 2. For log backup advancer, it will scan all locks for a small range.
+	// and it will check status of the txn. resolve the locks if txn is expired, Or do nothing.
+	//
+	// regionLocation should return if resolve locks succeed. if regionLocation return nil,
+	// which means not all locks are resolved in someway. the caller should retry scan locks.
+	ResolveLocksInOneRegion(bo *Backoffer, locks []*txnlock.Lock, regionLocation *locate.KeyLocation) (*locate.KeyLocation, error)
```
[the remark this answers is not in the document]
3pointer: updated. PTAL

tikv/gc.go
```diff
@@ -94,72 +98,129 @@ func (s *KVStore) resolveLocks(ctx context.Context, safePoint uint64, concurrenc
...
-		if err1 != nil {
-			return stat, err1
+		resolvedLocation, err := resolver.ResolveLocksInOneRegion(bo, locks, loc)
+		if err != nil {
+			return stat, err
 		}
 		// resolve locks failed since the locks are not in one region anymore, need retry.
 		if resolvedLocation == nil {
 			continue
 		}
```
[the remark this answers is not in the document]
3pointer: fixed. good catch.

zyguan (approved):
LGTM

PR #946, opened 2023-08-15
Refactor region error handler

you06:


cfzjywxk:
@you06 @crazycs520 
Is this to be continued/reused or closed?

zyguan:
I've tried to make some abstractions [here](https://gist.github.com/zyguan/31e78683b8a5f7ab901ebd889429c73c#file-send_request-go-L781-L823), which is similar to this PR.

PR #919, opened 2023-07-28
use tidb_kv_read_timeout as first kv request timeout

crazycs520:
issue: https://github.com/pingcap/tidb/issues/45380
related TiDB PR: https://github.com/pingcap/tidb/pull/45601

PR #948, opened 2023-08-17
use tidb_kv_read_timeout as first kv request timeout (#919)

crazycs520:
cherry-pick #919

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -240,6 +240,8 @@ type replica struct {
 	peer     *metapb.Peer
 	epoch    uint32
 	attempts int
+	// deadlineErrUsingConfTimeout indicates the replica is already tried, but the received deadline exceeded error.
+	deadlineErrUsingConfTimeout bool
 }
 
 func (r *replica) isEpochStale() bool {
@@ -337,7 +339,7 @@ func (state *accessKnownLeader) next(bo *retry.Backoffer, selector *replicaSelec
 	// a request. So, before the new leader is elected, we should not send requests
 	// to the unreachable old leader to avoid unnecessary timeout.
 	if liveness != reachable || leader.isExhausted(maxReplicaAttempt) {
-		selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx}
+		selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx, fromAccessKnownLeader: true}
 		return nil, stateChanged{}
 	}
 	selector.targetIdx = state.leaderIdx
@@ -352,7 +354,7 @@ func (state *accessKnownLeader) onSendFailure(bo *retry.Backoffer, selector *rep
 		return
 	}
 	if liveness != reachable || selector.targetReplica().isExhausted(maxReplicaAttempt) {
-		selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx}
+		selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx, fromAccessKnownLeader: true}
 	}
 	if liveness != reachable {
 		selector.invalidateReplicaStore(selector.targetReplica(), cause)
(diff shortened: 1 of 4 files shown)


internal/locate/region_request.go
```diff
@@ -436,7 +446,7 @@ func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (
 }
 
 func (state *tryFollower) onSendSuccess(selector *replicaSelector) {
-	if !state.fallbackFromLeader {
+	if state.fromOnNotLeader {
```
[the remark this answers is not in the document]
you06: The flag is different between 6.5 and master, I think here is to unify them.

internal/locate/region_request.go
```diff
@@ -377,17 +379,21 @@ type tryFollower struct {
 	leaderIdx          AccessIndex
 	lastIdx            AccessIndex
 	labels             []*metapb.StoreLabel
+	// fromOnNotLeader indicates whether the state is changed from onNotLeader.
+	fromOnNotLeader bool
```
you06, line 383: By adding fromOnNotLeader, it's better to remove `state.fallbackFromLeader` flag

internal/locate/region_request.go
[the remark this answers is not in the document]
cfzjywxk: But L429 they are still used seperately, it's confusing.

internal/locate/region_request.go
you06: And the flag should also set in these places.

https://github.com/tikv/client-go/blob/6f2442951b2bf717081ff383e3164f84675bf3d9/internal/locate/region_request.go#L340

https://github.com/tikv/client-go/blob/6f2442951b2bf717081ff383e3164f84675bf3d9/internal/locate/region_request.go#L355

internal/locate/region_request3_test.go
```diff
@@ -1190,3 +1194,84 @@ func (s *testRegionRequestToThreeStoresSuite) TestStaleReadFallback2Follower() {
...
+			s.Equal(1, len(s.regionRequestSender.Stats))
+			if staleRead {
+				rpcNum := s.regionRequestSender.Stats[tikvrpc.CmdGet].Count
+				s.True(rpcNum == 1 || rpcNum == 2) // 1 rpc or 2 rpc
+			} else {
+				s.Equal(int64(3), s.regionRequestSender.Stats[tikvrpc.CmdGet].Count) // 3 rpc
+				s.Equal(3, len(reqTargetAddrs))                                      // each rpc to a different store.
+			}
+			s.Equal(0, bo.GetTotalBackoffTimes()) // no backoff since fast retry.
+			// warn: must rest MaxExecutionDurationMs before retry.
```
you06, line 1257:
```suggestion
			// warn: must reset MaxExecutionDurationMs before retry.
```

I guess it's "reset".

internal/locate/region_request.go
```diff
@@ -436,7 +446,7 @@ func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (
 }
 
 func (state *tryFollower) onSendSuccess(selector *replicaSelector) {
-	if !state.fallbackFromLeader {
+	if !state.fallbackFromLeader || state.fromOnNotLeader {
```
cfzjywxk, line 449: Should it be `!state.fallbackFromLeader && state.fromOnNotLeader`?

cfzjywxk:
@you06 PTAL

internal/locate/region_request.go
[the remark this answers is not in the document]
crazycs520: Now, this logic is keep consistent with the master branch after https://github.com/tikv/client-go/pull/952.

internal/locate/region_request.go
```diff
@@ -426,17 +434,17 @@ func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (
...
 	}
-	if state.fallbackFromLeader {
-		staleRead := false
-		rpcCtx.contextPatcher.staleRead = &staleRead
+	if !state.fromAccessKnownLeader {
 		replicaRead := true
 		rpcCtx.contextPatcher.replicaRead = &replicaRead
 	}
+	staleRead := false
+	rpcCtx.contextPatcher.staleRead = &staleRead
```
cfzjywxk, line 442:
```suggestion
	if !state.fromAccessKnownLeader {
		staleRead := false
	        rpcCtx.contextPatcher.staleRead = &staleRead
	}
	replicaRead := true
	rpcCtx.contextPatcher.replicaRead = &replicaRead
```

Should it be like this?

cfzjywxk:
@crazycs520 
It seems the DCO fails, commits need to be signed with "-s" in git.

internal/locate/region_request.go
you06: No, if it's from `accessKnownLeader`, replica read should not be set.

crazycs520: The logic is consistent with master branch. https://github.com/tikv/client-go/blob/master/internal/locate/region_request.go#L461

bufferflies (approved):

PR #949, opened 2023-08-17
do not try leader if it's unreachable

zyguan:
When isolate two AZs, tidb may still try unreachable leader replicas, which leads to long-tail queries.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -616,7 +616,8 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	if selector.targetIdx < 0 {
 		leader := selector.replicas[state.leaderIdx]
 		leaderEpochStale := leader.isEpochStale()
-		leaderInvalid := leaderEpochStale || state.IsLeaderExhausted(leader)
+		leaderUnreachable := leader.store.getLivenessState() != reachable
+		leaderInvalid := leaderEpochStale || leaderUnreachable || state.IsLeaderExhausted(leader)
 		if len(state.option.labels) > 0 {
 			logutil.BgLogger().Warn("unable to find stores with given labels",
 				zap.Uint64("region", selector.region.GetID()),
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -616,7 +616,8 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	if selector.targetIdx < 0 {
 		leader := selector.replicas[state.leaderIdx]
 		leaderEpochStale := leader.isEpochStale()
-		leaderInvalid := leaderEpochStale || state.IsLeaderExhausted(leader)
+		leaderUnreachable := leader.store.getLivenessState() != reachable
+		leaderInvalid := leaderEpochStale || leaderUnreachable || state.IsLeaderExhausted(leader)
```
cfzjywxk, line 620: It's better to print the states `leaderEpochStale`,  `leaderUnreachable` and `state.IsLeaderExhausted(leader)` in the logs at L622.

cfzjywxk (approved):

internal/locate/region_request.go
```diff
@@ -616,7 +616,8 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 	if selector.targetIdx < 0 {
 		leader := selector.replicas[state.leaderIdx]
 		leaderEpochStale := leader.isEpochStale()
-		leaderInvalid := leaderEpochStale || state.IsLeaderExhausted(leader)
+		leaderUnreachable := leader.store.getLivenessState() != reachable
```
ekexium, line 619: If liveness state is unknown, should it be treated as invalid?

zyguan: I'm not sure actually, I just see `accessKnownLeader` will fallback to `tryFollower` when `liveness != reachable`.

cfzjywxk (approved):

you06 (approved):

PR #952, opened 2023-08-21
use leader read when `tryFollower` is fallback from `accessKnownLeader`

you06:
Recently, we made some changes to the selector states, the `tryFollower` is not only used as a fallback state of `accessKnownLeader`, this PR keeps the compatibility of `tryFollower`'s requests.

Before the changes we made:

`accessKnownLeader` -> `tryFollower` -> switch leader when success, and the flag `KVContext.ReplicaRead` is always false here.

Now there is another path lead to `tryFollower` state:

`accessFollower` -> `tryFollower`, in this case, `KVContext.ReplicaRead` is set to true in `tryFollower` state.

To make the behavior compatible with earlier versions, we need to set a flag for the `accessKnownLeader` -> `tryFollower` path, and that's `fromKnownLeader`. It's renamed from `fromOnNotLeader` for better semantic.

cc @crazycs520

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -347,7 +347,7 @@ func (state *accessKnownLeader) next(bo *retry.Backoffer, selector *replicaSelec
 	// a request. So, before the new leader is elected, we should not send requests
 	// to the unreachable old leader to avoid unnecessary timeout.
 	if liveness != reachable || leader.isExhausted(maxReplicaAttempt) {
-		selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx}
+		selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx, fromKnownLeader: true}
 		return nil, stateChanged{}
 	}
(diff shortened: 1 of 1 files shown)


crazycs520 (approved):

crazycs520:
@cfzjywxk @zyguan @ekexium PTAL

internal/locate/region_request.go
```diff
@@ -393,8 +393,8 @@ type tryFollower struct {
 	stateBase
 	leaderIdx AccessIndex
 	lastIdx   AccessIndex
-	// fromOnNotLeader indicates whether the state is changed from onNotLeader.
-	fromOnNotLeader bool
+	// fromKnownLeader indicates whether the state is changed from onNotLeader.
+	fromKnownLeader bool
```
cfzjywxk, line 397: If it's to represent a fallback from `accessKnownLeader` to `tryFollower`, naming it `fromAccessknownLeader` is more accurate and understandable.

zyguan (approved):

cfzjywxk (approved):

PR #953, opened 2023-08-24
add retry info to request source

you06:
image

As the image shows, this PR will change the request srouce when there is retry.

The first request's request source will be set to `{request source}-{read type}_read`.

The retried request's request soruce will be set to:

- `{request source}-{read type}_retry_read` if the read type doesn't change.
- `{request source}-{first read type}_to{second read type}_read` if the read type is changed.

e.g.

- `external_Select-stale_read` stands for first request with stale read.
- `external_Select-stale_to_leader_read` stands for retried request with leader read, and it's fallback from a stale read request.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1375,6 +1375,12 @@ func (s *RegionRequestSender) SendReqCtx(
 		}()
 	}
 
+	requestSource := req.Context.RequestSource
+	defer func() {
+		// reset the request source because it may be retried from upper layer.
+		req.Context.RequestSource = requestSource
+	}()
+	readType := getReadType(req)
 	totalErrors := make(map[string]int)
 	for {
 		if retryTimes > 0 {
@@ -1430,6 +1436,12 @@ func (s *RegionRequestSender) SendReqCtx(
 				h(req)
 			}
 		}
+
+		if e := tikvrpc.SetContext(req, rpcCtx.Meta, rpcCtx.Peer); e != nil {
+			return nil, nil, retryTimes, err
+		}
+		rpcCtx.contextPatcher.applyTo(&req.Context)
+		patchRequestSource(req, requestSource, readType, retryTimes)
 
 		var retry bool
 		resp, retry, err = s.sendReqToRegion(bo, rpcCtx, req, timeout)
@@ -1582,10 +1594,6 @@ func fetchRespInfo(resp *tikvrpc.Response) string {
 func (s *RegionRequestSender) sendReqToRegion(
 	bo *retry.Backoffer, rpcCtx *RPCContext, req *tikvrpc.Request, timeout time.Duration,
 ) (resp *tikvrpc.Response, retry bool, err error) {
-	if e := tikvrpc.SetContext(req, rpcCtx.Meta, rpcCtx.Peer); e != nil {
-		return nil, false, err
-	}
-	rpcCtx.contextPatcher.applyTo(&req.Context)
 	// judge the store limit switch.
 	if limit := kv.StoreLimit.Load(); limit > 0 {
 		if err := s.getStoreToken(rpcCtx.Store, limit); err != nil {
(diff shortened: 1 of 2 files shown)


cfzjywxk:
@crazycs520 PTAL

internal/locate/region_request.go
```diff
@@ -1431,6 +1437,12 @@ func (s *RegionRequestSender) SendReqCtx(
 			}
 		}
 
+		if e := tikvrpc.SetContext(req, rpcCtx.Meta, rpcCtx.Peer); e != nil {
+			return nil, nil, retryTimes, err
+		}
+		rpcCtx.contextPatcher.applyTo(&req.Context)
+		patchRequestSource(req, requestSource, readType, retryTimes)
```
cfzjywxk, line 1444: If finally "nil, nil" is returned, the next round `RegionRequestSender.SendReqCtx` retry, stale flag would not be considered again, in this case it seems the `readType` may not represent the original read type.

internal/locate/region_request.go
```diff
@@ -2302,3 +2310,34 @@ func (s *staleReadMetricsCollector) onResp(tp tikvrpc.CmdType, resp *tikvrpc.Res
...
+	if req.ReplicaRead {
+		return "replica"
+	}
+	return "leader"
+}
+
+func patchRequestSource(req *tikvrpc.Request, source, readType string, retryTimes int) {
+	var sb strings.Builder
+	sb.WriteString(source)
+	sb.WriteByte('-')
```
crazycs520, line 2327: Both `-` and `_` exist here, how about using only one for unification?

internal/locate/region_request.go
you06: I change the implemention and it now handles the retry from upper layer.

internal/locate/region_request.go
you06: `-` devides request source and read type, `_` is the inner separator of each one.

internal/locate/region_request.go
```diff
@@ -2302,3 +2306,34 @@ func (s *staleReadMetricsCollector) onResp(tp tikvrpc.CmdType, resp *tikvrpc.Res
 		metrics.StaleReadRemoteInBytes.Add(float64(size))
 	}
 }
+
+func GetReadType(req *tikvrpc.Request) string {
```
ekexium, line 2310: I suppose staleness and replica are two independent dimensions? For example shall we distinguish stale-follower-read and stale-leader-read?

you06: Do you mean there is 3 dimenstions:

- is retry
- staleness
- the selected replica is follower or leader(it does not depend on the `ReplicaReadType`)

ekexium: Yes, this approach seems clearer to me. The only concern is its verbosity, especially when we need to record the type of the last attempt.

tikvrpc/tikvrpc.go
```diff
@@ -234,6 +234,10 @@ type Request struct {
 	ForwardedHost string
 	// ReplicaNumber is the number of current replicas, which is used to calculate the RU cost.
 	ReplicaNumber int64
+	// The initial read type.
+	ReadType string
+	// InputRequestSource is the input source of the request, if it's not empty, the final RequestSource sent to store will be attached with the retry info.
+	InputRequestSource string
```
cfzjywxk, line 240: Would the `InputRequestSource`  of coprocessor related requests be set accordingly in tidb?

you06: Coprocessor code also need to be updated.

internal/locate/region_request.go
you06: So it's look like:

- `leader` means request sent to leader first time(it can be either leader read or replica read)
- `follower` means request sent to follower first time(it should be replica read)
- `stale_leader` means stale read on leader replica first time.
- `stale_follower` means stale read on follower replica first time.
- `retry_stale_leader_leader` means fallback to leader read on leader replica from stale read on leader replica.
- `retry_stale_follower_leader` means fallback to leader read on leader replica from stale read on follower replica.
- `retry_stale_leader_follower` means fallback to replica read on follower replica from stale read on leader replica(the leader is exhausted).
- `retry_stale_follower_follower` means fallback to replica read on follower replica from stale read on follower replica(the leader is exhausted).

tikvrpc/tikvrpc.go
cfzjywxk: Ok, let's merge the client-go part first

zyguan (approved):

cfzjywxk:
The retry flag should also be handled when key error is returned to the `snapshot or cop send request` layer, setting `req.IsRetryRequest` is not enough for such cases.

cfzjywxk:
@you06 
If there's no "stale" word in the description, does it mean this is not a stale request but it's retried for some reason?

you06:
> If there's no "stale" word in the description, does it mean this is not a stale request but it's retried for some reason?

Yes, if there is no "stale" word, it's not a stale read request.

There are two figures in the description, the first one is tested with stale read and the second one is not.

nolouch:
`{request source}-{retry}_{stale}_{replica type}`, breaks the dependency for the background control: 
https://github.com/tikv/pd/blob/41eae1a0a306d763c505f2cadaaec40795c120bc/client/resource_group/controller/controller.go#L490-L511

can we adjust to like : `{retry}_{stale}_{replica type}_{request source}`?

internal/locate/region_request.go
```diff
@@ -1431,8 +1430,17 @@ func (s *RegionRequestSender) SendReqCtx(
 			}
 		}
 
+		if e := tikvrpc.SetContext(req, rpcCtx.Meta, rpcCtx.Peer); e != nil {
+			return nil, nil, retryTimes, err
```
ekexium, line 1434: Should it return `e` rather than `err`?

PR #956, opened 2023-08-30
min-resolved-ts: check dc label 

HuSharp:
### Background
1. api interface
	Nowadays we have 2 api interfaces for obtaining min resolved ts

	For client-go's updateSafeTS, we call `approach II` for each store, which is not necessary to call API so many times 
We can use the `approach I` to get cluster's min resolved ts when `@@txn_scope` is `global`

2. avoid using kv request reason:
if TiDB failed to connect a TiKV store to fetch the latest resolved_ts, it will remain the out-of-dated value until the TiKV comes back, though PD will mark the store as unavailable and keep updating the minimal resolved_ts with other TiKV stores. We may need to keep these two behaviors the same or unify them.

### Summary pr
judge `@@txn_scope` value which from the config:
- when find is `global`, we can get cluster-level resolved ts to avoid using kv grpc.
- when find is not `global`, we need to use kv grpc.

### DIFF
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -580,10 +580,17 @@ func (s *KVStore) safeTSUpdater() {
 }
 
 func (s *KVStore) updateSafeTS(ctx context.Context) {
+	// Try to get the cluster-level minimum resolved timestamp from PD first.
+	if s.setClusterMinSafeTSByPD(ctx) {
+		return
+	}
+
 	stores := s.regionCache.GetAllStores()
 	tikvClient := s.GetTiKVClient()
 	wg := &sync.WaitGroup{}
 	wg.Add(len(stores))
+	// If getting the cluster-level minimum resolved timestamp from PD failed or returned 0,
+	// try to get it from TiKV.
 	for _, store := range stores {
 		storeID := store.StoreID()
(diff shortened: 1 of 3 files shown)


nolouch (approved):

tikv/kv.go
```diff
@@ -593,39 +600,25 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
...
+				ctx, storeAddr, tikvrpc.NewRequest(
+					tikvrpc.CmdStoreSafeTS, &kvrpcpb.StoreSafeTSRequest{
+						KeyRange: &kvrpcpb.KeyRange{
+							StartKey: []byte(""),
+							EndKey:   []byte(""),
+						},
+					}, kvrpcpb.Context{
+						RequestSource: util.RequestSourceFromCtx(ctx),
+					},
+				), client.ReadTimeoutShort,
```
bufferflies, line 613: 30s maybe too long.

bufferflies (approved):

tikv/kv.go
HuSharp: maybe need to handle it in another pr if concerned about it.

PR #958, opened 2023-09-04
log: Enhanced log key output

HuSharp:
Signed-off-by: husharp <jinhao.hu@pingcap.com>


internal/locate/region_cache.go
```diff
@@ -1376,6 +1377,8 @@ func (c *RegionCache) BatchLoadRegionsWithKeyRange(bo *retry.Backoffer, startKey
 		return
 	}
 	if len(regions) == 0 {
+		startKey = c.codec.EncodeRegionKey(startKey)
```
nolouch, line 1380: because it change the behavior, I think it's better to add a new field.

disksing (approved):

PR #965, opened 2023-09-11
golangci-lint: fix lint error on tidb-6.5 branch

3pointer:


zyguan (approved):

.github/workflows/test.yml
```diff
@@ -47,5 +47,5 @@ jobs:
       - name: Lint
         uses: golangci/golangci-lint-action@v2.5.2
         with:
-          version: v1.47.3
+          version: v1.52.2
```
HuSharp, line 50: I noticed master is `version: v1.51.2`, do we need to update the master as well?
https://github.com/tikv/client-go/blob/master/.github/workflows/test.yml#L50

disksing: I think we can .

PR #978, opened 2023-09-15
*: configurable kv_read_timeout should not affect write request

### LINKED ISSUE #972: configurable KV Timeout should't affect write request
https://github.com/tikv/client-go/pull/948/files#diff-0c36d07fbd1fc71c850ff57d323fe12bdf7f2c053ef71b7fec5bb44358f1070dR1550  

This check maybe affect write request, since `prewrite`, `commit` use `MaxWriteExecutionTime` as `MaxExecutionDurationMs`, and `MaxWriteExecutionTime = ReadTimeoutShort - 10*time.Second`.

crazycs520:
close #972

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1087,13 +1087,25 @@ func (s *replicaSelector) onSendFailure(bo *retry.Backoffer, err error) {
 	s.state.onSendFailure(bo, s, err)
 }
 
-func (s *replicaSelector) onDeadlineExceeded() {
-	if target := s.targetReplica(); target != nil {
-		target.deadlineErrUsingConfTimeout = true
+func (s *replicaSelector) onReadReqConfigurableTimeout(req *tikvrpc.Request) bool {
+	if req.MaxExecutionDurationMs >= uint64(client.ReadTimeoutShort.Milliseconds()) {
+		// Configurable timeout should less than `ReadTimeoutShort`.
+		return false
 	}
-	if accessLeader, ok := s.state.(*accessKnownLeader); ok {
-		// If leader return deadline exceeded error, we should try to access follower next time.
-		s.state = &tryFollower{leaderIdx: accessLeader.leaderIdx, lastIdx: accessLeader.leaderIdx}
+	switch req.Type {
+	case tikvrpc.CmdGet, tikvrpc.CmdBatchGet, tikvrpc.CmdScan,
+		tikvrpc.CmdCop, tikvrpc.CmdBatchCop, tikvrpc.CmdCopStream:
+		if target := s.targetReplica(); target != nil {
+			target.deadlineErrUsingConfTimeout = true
+		}
+		if accessLeader, ok := s.state.(*accessKnownLeader); ok {
+			// If leader return deadline exceeded error, we should try to access follower next time.
+			s.state = &tryFollower{leaderIdx: accessLeader.leaderIdx, lastIdx: accessLeader.leaderIdx}
+		}
+		return true
+	default:
+		// Only work for read requests, return false for non-read requests.
+		return false
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -2360,3 +2360,13 @@ func (s *replicaSelector) patchRequestSource(req *tikvrpc.Request, rpcCtx *RPCCo
 	}
 	sb.WriteString(req.ReadType)
 }
+
+func isReadReq(tp tikvrpc.CmdType) bool {
```
cfzjywxk, line 2364: How about the read requests triggered by write statements?

crazycs520: Nice catch, already refined the logic, PTAL.

internal/locate/region_request.go
```diff
@@ -1777,9 +1789,8 @@ func (s *RegionRequestSender) onSendFail(bo *retry.Backoffer, ctx *RPCContext, r
 		return errors.WithStack(err)
 	} else if LoadShuttingDown() > 0 {
 		return errors.WithStack(tikverr.ErrTiDBShuttingDown)
-	} else if errors.Cause(err) == context.DeadlineExceeded && req.MaxExecutionDurationMs < uint64(client.ReadTimeoutShort.Milliseconds()) {
-		if s.replicaSelector != nil {
-			s.replicaSelector.onDeadlineExceeded()
+	} else if errors.Cause(err) == context.DeadlineExceeded {
+		if s.replicaSelector != nil && s.replicaSelector.onReadReqConfigurableTimeout(req) {
```
ekexium, line 1793: Are all DeadlineExceeded errors caused by configurable timeout? This seems a bit confusing to me

internal/locate/region_request.go
```diff
@@ -1087,13 +1087,25 @@ func (s *replicaSelector) onSendFailure(bo *retry.Backoffer, err error) {
 	s.state.onSendFailure(bo, s, err)
 }
 
-func (s *replicaSelector) onDeadlineExceeded() {
-	if target := s.targetReplica(); target != nil {
-		target.deadlineErrUsingConfTimeout = true
+func (s *replicaSelector) onReadReqConfigurableTimeout(req *tikvrpc.Request) bool {
+	if req.MaxExecutionDurationMs >= uint64(client.ReadTimeoutShort.Milliseconds()) {
```
cfzjywxk, line 1091: It seems we don't check like if `rea.stmtType == Read`, if the session variable of `kv timeout` is set to a small value like `1s`, the read requests from write statements would still take effect? For example, the batch get requests coming from the `INSERT` statements.

crazycs520: Nice catch. But currently, the write statement doesn't set `kv_read_timeout` in TiDB.

Or just in case, we should add `stmtType` field and check `req.stmtType == Read`?

crazycs520: But check `stmtType` in `client-go` is a little bit strange. If the write statement should not use `kv_read_timeout`, then the write statement should not set it.

cfzjywxk: Could the `request source` be used to check the statement source? For example, check if it's a `select` source type.
It's more robust to do all the necessary checks to avoid possible mis-usages though users may have some assumptions on the application side.

crazycs520: Check if it's `select` is not work well. Such as `explain analyze select ...`, the `request source` will be `ExplainAnalyzeSQL`.

ekexium (commented):
Looks OK as a fix for now. But I do think we will have to distinguish different timeouts, instead of mixing the use a ServerIsError and using strings to match

cfzjywxk:
> Looks OK as a fix for now. But I do think we will have to distinguish different timeouts, instead of mixing the use a ServerIsError and using strings to match

+1

cfzjywxk (approved):
As a quick fix for standalone branch is fine, we need to defensive add checks for write statements in the future.

ekexium (approved):

PR #990, opened 2023-09-22
retry short read reqs immediately on server busy

zyguan:
If a read-only workload running with a short max-execution-time (eg. 500ms), injecting IO delay faults will cause the QPS amost drop to zero, because the queries will be interrupted during server-busy backoff and have no chance to try other available peers.

To address the issue, let the sender retry those requests immediately and do backoff lazily. Here is the test results of injecting IO delay (1s) on one tikv for 3 minutes.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1384,6 +1384,48 @@ func (s *RegionRequestSender) SendReqCtx(
 		}()
 	}
 
+	type backoffArgs struct {
+		cfg *retry.Config
+		err error
+	}
+	pendingBackoffs := make(map[string]*backoffArgs)
+	delayBoTiKVServerBusy := func(addr string, reason string) {
+		if _, ok := pendingBackoffs[addr]; !ok {
+			pendingBackoffs[addr] = &backoffArgs{
+				cfg: retry.BoTiKVServerBusy,
+				err: errors.Errorf("server is busy, reason: %s", reason),
+			}
+		}
+	}
+	backoffOnRetry := func(addr string) error {
+		if args, ok := pendingBackoffs[addr]; ok {
+			delete(pendingBackoffs, addr)
+			logutil.Logger(bo.GetCtx()).Warn(
+				"apply pending backoff on retry",
+				zap.String("target", addr),
+				zap.String("bo", args.cfg.String()),
+				zap.String("err", args.err.Error()))
+			return bo.Backoff(args.cfg, args.err)
+		}
+		return nil
+	}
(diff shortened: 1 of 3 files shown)


ekexium:
The change makes backoff more complex. It only takes effect when max execution time is smaller than read timeout, which may itself be considered a misuse, I suppose? I'm not sure if it's worth to cover this case

internal/locate/region_request.go
```diff
@@ -1388,6 +1402,48 @@ func (s *RegionRequestSender) SendReqCtx(
...
+		}
+		if args != nil {
+			logutil.Logger(bo.GetCtx()).Warn(
+				"apply pending backoff on pseudo region error",
+				zap.String("bo", args.cfg.String()),
+				zap.String("err", args.err.Error()))
+			return bo.Backoff(args.cfg, args.err)
+		}
+		return nil
+	}
```
MyonKeminta, line 1446: Could you explain the purpose of these code? I didn't understand it 🤔

zyguan: It's used to maintain delayed backoffs.
-  "`pendingBackoffs[addr]` is not nil" means there is a delayed backoff which should be applied on retrying this store.
- `delayBoTiKVServerBusy` is the callback for recording delayed backoffs.
- `backoffOnRetry` is used to apply pending backoffs on retry.
- when there is no candidate, the kv client just return a fake error and let the caller do retry. to avoid potential frequent rpcs, we call `backoffOnFakeErr` before returning a fake error, which chooses a pending backoff with the largest base duration and applies it.

MyonKeminta: Can this be done by adding "busy" flag to the stores, instead of maintaining maps here?

zyguan: If you mean the [store](https://github.com/tikv/client-go/blob/dbea404a846b5404206f1455da4e94639da9895f/internal/locate/region_cache.go#L2535) struct in region cache, probably no I think. They have different scope/lifetime, that is, each `SendReqCtx` call should has its own `pendingBackoffs`, and has no effect to each others. Maybe we can add a field to the [replica](https://github.com/tikv/client-go/blob/master/internal/locate/region_request.go#L247) struct, however, this PR is just a quick-and-dirty fix to the customer issue, I just try to make the least change.

MyonKeminta (approved):
The logic looks fine to me. I wonder if the code can be better structured...

ekexium:
Is this PR still active?

zyguan:
It's included https://github.com/tikv/client-go/pull/1184

PR #994, opened 2023-09-27
min-safe-ts: fix MinSafeTS might be set to MaxUint64 permanently

### LINKED ISSUE #991: MinSafeTS might be set to MaxUint64 permanently
If the cluster just bootstraps and the `min-resolved-ts` hasn't been initialized yet, `MinSafeTS` might be set to `MaxUint64`. And since the `MinSafeTS` is not allowed to fall back, it will remain being this value permanently.

https://github.com/tikv/client-go/blob/c8832b846685d303f5afe29e1f8da50c7aff9a2e/tikv/kv.go#L546-L563

HuSharp:
close #991 

The core question is the getter 
- we introduce `PD API` to not execute `go func which for KV request`, resulting in not updating `safeTSMap`. 		 
    - `updateMinSafeTS` relies on  `safeTSMap` which makes sense(because actually, we can call `updateMinSafeTS` to `kvReuqestUpdater`[to indicate func base]).
- And we need to update `minsafeTS` to make sure when API fails we can fall back to the original way which is by kv request.
- But the core problem is:  **`updateMinSafeTS` will return `maxUnit64` when the first kv request returns 0 and then although `PD API` returns correctly[maybe kv is not initialized], TS can not change `maxUnit64`.**
   - to resolve this question, we need to regard `maxUnit64` as 0 which means there is an initial state.

### DIFF
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -500,6 +500,9 @@ func (s *KVStore) GetTiKVClient() (client Client) {
 // GetMinSafeTS return the minimal safeTS of the storage with given txnScope.
 func (s *KVStore) GetMinSafeTS(txnScope string) uint64 {
 	if val, ok := s.minSafeTS.Load(txnScope); ok {
+		if val.(uint64) == uint64(math.MaxUint64) {
+			return 0
+		}
 		return val.(uint64)
 	}
 	return 0
@@ -850,3 +853,15 @@ type SchemaVer = transaction.SchemaVer
 // MaxTxnTimeUse is the max time a Txn may use (in ms) from its begin to commit.
 // We use it to abort the transaction to guarantee GC worker will not influence it.
 const MaxTxnTimeUse = transaction.MaxTxnTimeUse
+
+// SetMinSafeTS set the minimal safeTS of the storage with given txnScope.
+// Note: it is only used for test.
+func (s *KVStore) SetMinSafeTS(txnScope string, safeTS uint64) {
+	s.minSafeTS.Store(txnScope, safeTS)
+}
+
+// SetStoreSafeTS set the safeTS of the store with given storeID.
+// Note: it is only used for test.
+func (s *KVStore) SetStoreSafeTS(storeID, safeTS uint64) {
+	s.safeTSMap.Store(storeID, safeTS)
+}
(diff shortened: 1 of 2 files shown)


disksing (approved):

HuSharp:
@JmPotato PTAL, thx!

tikv/kv.go
```diff
@@ -500,6 +500,9 @@ func (s *KVStore) GetTiKVClient() (client Client) {
 // GetMinSafeTS return the minimal safeTS of the storage with given txnScope.
 func (s *KVStore) GetMinSafeTS(txnScope string) uint64 {
 	if val, ok := s.minSafeTS.Load(txnScope); ok {
+		if val.(uint64) == uint64(math.MaxUint64) {
+			return 0
+		}
```
JmPotato, line 505: The key point is to prevent the `MinSafeTS` from not being updated permanently because of `MaxUint64`. We should handle it from the updater rather than just the getter.

HuSharp: The core question is the getter 
- we introduce `PD API` to not execute `go func which for KV request`, resulting in not updating `safeTSMap`. 		 
    - `updateMinSafeTS` relies on  `safeTSMap` which makes sense(because actually, we can call `updateMinSafeTS` to `kvReuqestUpdater`[to indicate func base]).
- And we need to update `minsafeTS` to make sure when API fails we can fall back to the original way which is by kv request.
- But the core question is:  **`updateMinSafeTS` will return `maxUnit64` when there is no kv request.**
   - to resolve this question, we need to regard `maxUnit64` as 0 which means there is an initial state.

HuSharp:
@nolouch PTAL as well, thx!

nolouch (approved):
lgtm

tikv/kv.go
```diff
@@ -690,6 +690,11 @@ func (s *KVStore) updateGlobalTxnScopeTSFromPD(ctx context.Context) bool {
 		} else if clusterMinSafeTS != 0 {
 			// Update ts and metrics.
 			preClusterMinSafeTS := s.GetMinSafeTS(oracle.GlobalTxnScope)
+			// If preClusterMinSafeTS is maxUint64, it means that the min safe ts has not been initialized.
+			// related to https://github.com/tikv/client-go/issues/991
+			if preClusterMinSafeTS == math.MaxUint64 {
+				preClusterMinSafeTS = 0
+			}
 			if preClusterMinSafeTS > clusterMinSafeTS {
```
JmPotato, line 698: What about just:

```suggestion
			if preClusterMinSafeTS != math.MaxUint64 && preClusterMinSafeTS > clusterMinSafeTS {
```

HuSharp: good idea. Fixed.

JmPotato (approved):

Issue #1001, opened 2023-10-11
Backoff retries for "server is busy" type errors may experience infinite retries in some extreme scenarios


cfzjywxk:
From https://github.com/tikv/client-go/pull/271 the backoff caused by `ServerIsBusy` error type would not be counted in the total backoff time calculation, this may lead to infinite retries in some extreme scenarios, like:

```
# Query_time: 14189.777571764
# Parse_time: 0
# Compile_time: 0.000188387
# Rewrite_time: 0.000056979
# Optimize_time: 0.000031731
# Wait_TS: 0.00120544
[...]

PR #1002, opened 2023-10-11
add retry limit for excluded backoff type to avoid infinite retry

cfzjywxk:
Ref: https://github.com/tikv/client-go/issues/1001

Try to add retry limit for excluded backoff type to avoid infinite retry.

### DIFF
--- a/internal/retry/backoff.go
+++ b/internal/retry/backoff.go
@@ -143,7 +143,13 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 	if b.noop {
 		return err
 	}
-	if b.maxSleep > 0 && (b.totalSleep-b.excludedSleep) >= b.maxSleep {
+	maxBackoffTimeExceeded := (b.totalSleep - b.excludedSleep) >= b.maxSleep
+	maxExcludedTimeExceeded := false
+	if maxLimit, ok := isSleepExcluded[cfg.name]; ok {
+		maxExcludedTimeExceeded = b.excludedSleep >= maxLimit && b.excludedSleep >= b.maxSleep
+	}
+	maxTimeExceeded := maxBackoffTimeExceeded || maxExcludedTimeExceeded
+	if b.maxSleep > 0 && maxTimeExceeded {
 		longestSleepCfg, longestSleepTime := b.longestSleepCfg()
 		errMsg := fmt.Sprintf("%s backoffer.maxSleep %dms is exceeded, errors:", cfg.String(), b.maxSleep)
 		for i, err := range b.errors {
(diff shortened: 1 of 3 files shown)


zyguan (approved):

internal/retry/backoff.go
```diff
@@ -143,7 +143,13 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 	if b.noop {
 		return err
 	}
-	if b.maxSleep > 0 && (b.totalSleep-b.excludedSleep) >= b.maxSleep {
+	maxBackoffTimeExceeded := (b.totalSleep - b.excludedSleep) >= b.maxSleep
+	maxExcludedTimeExceeded := false
+	if maxLimit, ok := isSleepExcluded[cfg.name]; ok {
+		maxExcludedTimeExceeded = b.excludedSleep >= maxLimit && b.excludedSleep >= b.maxSleep
```
ekexium, line 149: From the name `maxLimit`, I would suppose the latter part is not necessary. In practice it won't make big difference though.

ekexium (approved):

internal/retry/backoff.go
```diff
@@ -163,7 +169,8 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 			backoffDetail.WriteString(":")
 			backoffDetail.WriteString(strconv.Itoa(times))
 		}
-		errMsg += fmt.Sprintf("\ntotal-backoff-times: %v, backoff-detail: %v", totalTimes, backoffDetail.String())
+		errMsg += fmt.Sprintf("\ntotal-backoff-times: %v, backoff-detail: %v, maxBackoffTimeExceeded: %v, maxBackoffTimeExceeded: %v",
```
crazycs520, line 172:
```suggestion
		errMsg += fmt.Sprintf("\ntotal-backoff-times: %v, backoff-detail: %v, maxBackoffTimeExceeded: %v, maxExcludedTimeExceeded: %v",
```

internal/retry/backoff.go
cfzjywxk: In normal circumstances, this check at the end is not necessary. It's mainly to guard against some erroneous inputs or configurations.

internal/retry/config.go
```diff
@@ -130,11 +130,18 @@ var (
 	BoTxnLockFast = NewConfig(txnLockFastName, &metrics.BackoffHistogramLockFast, NewBackoffFnCfg(2, 3000, EqualJitter), tikverr.ErrResolveLockTimeout)
 )
 
-var isSleepExcluded = map[string]struct{}{
-	BoTiKVServerBusy.name: {},
+var isSleepExcluded = map[string]int{
+	BoTiKVServerBusy.name: 600000, // The max excluded limit is 10min.
```
jackysp, line 134: It's a bit too far to add a new limit of it, I'm afraid that someday it will be required to be configurable.

cfzjywxk: This could only be used in some extreme cases like the service is unavailable but the leader is still valid.

PR #1004, opened 2023-10-11
internal/locate: change some retriable error logs to debug level

jackysp:
If the requests are retriable, high log level will make the log files too noisy.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -2125,7 +2125,7 @@ func (s *RegionRequestSender) onRegionError(
 		if s.replicaSelector != nil {
 			return s.replicaSelector.onServerIsBusy(bo, ctx, req, serverIsBusy)
 		}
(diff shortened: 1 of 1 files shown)


cfzjywxk (approved):

zyguan (approved):

internal/locate/region_request.go
```diff
@@ -2232,7 +2232,7 @@ func (s *RegionRequestSender) onRegionError(
 	// This error is specific to stale read and the target replica is randomly selected. If the request is sent
 	// to the leader, the data must be ready, so we don't backoff here.
 	if regionErr.GetDataIsNotReady() != nil {
-		logutil.Logger(bo.GetCtx()).Warn(
```
ekexium, line 2235: We used to use this for diagnosis. A better but complex idea might be sampling logs.

PR #1032, opened 2023-10-25
collecting the RU information by pasing point through context.Value

zeminzhou:
To collect all query RU consumption into statement log, ref https://github.com/tidbcloud/tidb-cse/pull/678


tikv/kv.go
```diff
@@ -695,44 +687,6 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 	wg.Wait()
 }
 
-func (s *KVStore) ruRuntimeStatsMapCleaner() {
```
ystaticy, line 698: Why remove those code?

PR #1043, opened 2023-10-26
client: add GRPC client metrics

jackysp:
Related to: https://github.com/pingcap/tidb/issues/45242


jackysp:
It seems it's not very useful.

PR #1044, opened 2023-10-28
*: upgrade grpc and etcd

hawkingrei:
ref https://github.com/pingcap/tidb/issues/48060

It is tested by https://github.com/pingcap/tidb/pull/48057 in the tidb.


hawkingrei:
@disksing PTAL

PR #1049, opened 2023-11-03
Support to integrate with the PD HTTP client

JmPotato:
Ref https://github.com/tikv/pd/issues/7300, https://github.com/tikv/pd/pull/7304


lhy1024 (approved):

HuSharp:
Can you pass the integration test in your local host?

tikv/kv.go
```diff
@@ -195,9 +198,9 @@ func WithPool(gp Pool) Option {
 }
 
 // WithPDHTTPClient set the PD HTTP client with the given address and TLS config.
-func WithPDHTTPClient(tlsConf *tls.Config, pdaddrs []string) Option {
+func WithPDHTTPClient(pdAddrs []string, tlsConf *tls.Config) Option {
```
HuSharp, line 201: Does it make a compatibility error to TiDB?

JmPotato: When upgrading client-go, we will naturally perceive this modification before and after compilation. I don't think it will cause compatibility problems.

tikv/kv.go
```diff
@@ -640,8 +651,10 @@ func (s *KVStore) updateSafeTS(ctx context.Context) {
 					return
 				}
 				safeTS = resp.Resp.(*kvrpcpb.StoreSafeTSResponse).GetSafeTs()
+				logutil.BgLogger().Info("get min resolved ts from tikv", zap.Uint64("store-id", storeID), zap.Uint64("safe-ts", safeTS))
```
HuSharp, line 654: Do we need this info?

JmPotato: Removed.

HuSharp (approved):
LGTM!

Issue #1051, opened 2023-11-08
Region cache stuck once try to load region until +inf


YuJuncen:
I have noticed this call never return:
```go
	rs, err := c.Cache.LoadRegionsInKeyRange(bo, startKey, []byte(""))
```
According to `dlv`, this call generated a infinite loop in `github.com/tikv/client-go/v2/internal/locate.(*RegionCache).LoadRegionsInKeyRange()`, which tests whether the last region contains the end key is by a buggy function:
```go
// ContainsByEnd check the region contains the greatest key that is less than key.
// for the maximum region endKey is empty.
// startKey < key <= endKey.
func (r *Region) ContainsByEnd(key []byte) bool {
	return bytes.Compare(r.meta.GetStartKey(), key) < 0 &&
		(bytes.Compare(key, r.meta.GetEndKey()) <= 0 || len(r.meta.GetEndKey()) == 0)
}
```
The left hand side of `&&` simply always evaluates to `false` when the `key` is empty (which means inf). Which returns a false `false` when the input region has range `(*, +inf]`.

disksing:
fixed.

PR #1052, opened 2023-11-08
make `ContainsByEnd` return `true` when the key is empty and region's right bound is inf

YuJuncen:
Signed-off-by: hillium <yujuncen@pingcap.com>

Ref: https://github.com/tikv/client-go/issues/1051

I think the semantic of `ContainsByEnd` should be: "Please check whether this key is in the range, if the input is empty, treat it as infinity.". But I have noticed there are already some test cases that believes "" shouldn't be in ("", ""]. I'm not sure whether it is by design.

Anyway this will bring unexpected behavior sometimes (say, make infinity loops), I think perhaps we can make it better.


disksing:
I think you're right, but I need to check all usages of this function to prevent this modification from accidentally breaking something.

internal/locate/region_cache.go
```diff
@@ -2468,8 +2468,9 @@ func (r *Region) Contains(key []byte) bool {
 // for the maximum region endKey is empty.
 // startKey < key <= endKey.
 func (r *Region) ContainsByEnd(key []byte) bool {
-	return bytes.Compare(r.meta.GetStartKey(), key) < 0 &&
-		(bytes.Compare(key, r.meta.GetEndKey()) <= 0 || len(r.meta.GetEndKey()) == 0)
+	return len(key) == 0 && len(r.EndKey()) == 0 ||
+		bytes.Compare(r.meta.GetStartKey(), key) < 0 &&
+			(bytes.Compare(key, r.meta.GetEndKey()) <= 0 || len(r.meta.GetEndKey()) == 0)
```
disksing, line 2473: maybe this will make it easier to read:
```
if len(key) == 0 {
    return len(r.EndKey()) == 0
}
return bytes.Compare(r.meta.GetStartKey(), key) < 0 &&
			(bytes.Compare(key, r.meta.GetEndKey()) <= 0 || len(r.meta.GetEndKey()) == 0)
```

disksing:
> I think you're right, but I need to check all usages of this function to prevent this modification from accidentally breaking something.

checked, it should be fine.

disksing (approved):

PR #1054, opened 2023-11-09
tso: merge lastTS and lastArrivalTS into an atomic pointer

### LINKED ISSUE #1053: `GetStaleTimestamp` may return future TSO
This test may return future TSO when calling `GetStaleTimestamp`, which is a correctness risk.

```go
func TestNonFutureStaleTSO(t *testing.T) {
	o := oracles.NewEmptyPDOracle()
	oracles.SetEmptyPDOracleLastTs(o, oracle.GoTimeToTS(time.Now()))
	for i := 0; i < 100; i++ {
		time.Sleep(10 * time.Millisecond)
		now := time.Now()
		upperBound := now.Add(5 * time.Millisecond) // allow 5ms time drift

		closeCh := make(chan struct{})
		go func() {
			time.Sleep(100 * time.Microsecond)
			oracles.SetEmptyPDOracleLastTs(o, oracle.GoTimeToTS(now))
			close(closeCh)
		}()
	CHECK:
		for {
			select {
			case <-closeCh:
				break CHECK
			default:
				ts, err := o.GetStaleTimestamp(context.Background(), oracle.GlobalTxnScope, 0)
				assert.Nil(t, err)
				staleTime := oracle.GetTimeFromTS(ts)
				if staleTime.After(upperBound) && time.Since(now) < time.Millisecond /* only check staleTime within 1ms */ {
					assert.Less(t, staleTime, upperBound, i)
					t.FailNow()
				}
			}
		}
	}
}
```

The test should pass because stale timestamp should not be a future ts. But we got:

```
Error:      	"2023-11-09 15:44:50.757 +0900 JST" is not less than "2023-11-09 15:44:50.751395 +0900 JST m=+0.030711501"
```

you06:
close #1053

The update of `lastTS` and `lastArrivalTS` should be atomic, unless we may get new `lateTS` with old `lastArrivalTS` when calling `GetStaleTimestamp` which may lead to a future TSO.

This PR merge `lastTS` and `lastArrivalTS` into an atomic pointer and fix the future stale TSO bug.


disksing:
@JmPotato Please take a look

JmPotato (approved):

oracle/oracles/pd.go
```diff
@@ -165,61 +169,49 @@ func (o *pdOracle) setLastTS(ts uint64, txnScope string) {
...
-	lastTSPointer := lastTSInterface.(*uint64)
-	for {
-		lastTS := atomic.LoadUint64(lastTSPointer)
-		if ts <= lastTS {
-			return
+		if current == nil {
+			current = &lastTSO{
+				tso:     ts,
+				arrival: o.getArrivalTimestamp(),
+			}
```
cfzjywxk, line 185: How about defining `current` outside the for loop as it would not change.

oracle/oracles/pd.go
```diff
@@ -165,61 +169,49 @@ func (o *pdOracle) setLastTS(ts uint64, txnScope string) {
...
+		lastTSInterface, _ = o.lastTSMap.LoadOrStore(txnScope, &atomic.Pointer[lastTSO]{})
 	}
-	lastTSPointer := lastTSInterface.(*uint64)
+	lastTSPointer := lastTSInterface.(*atomic.Pointer[lastTSO])
+	var current *lastTSO
 	for {
-		lastTS := atomic.LoadUint64(lastTSPointer)
-		if ts <= lastTS {
+		last := lastTSPointer.Load()
+		if last != nil && ts <= last.tso {
```
cfzjywxk, line 178: Is it ok to put a nil value into the `lastTSPointer`? How about using a constant value instead of nil?

you06: Made it a non-nil pointer.

cfzjywxk (approved):
LGTM

PR #1055, opened 2023-11-10
config: enable user configuration for retry strategy.

Smityz:
We have observed that in some cases, the configuration of backoff has a significant impact on the delay. Therefore, we would like to provide the option for manual adjustment of the retry strategy on the client-side.
This PR moved folder `internal/retry` to `config/retry`


disksing (approved):

disksing:
Thanks @Smityz

ekexium:
After this PR, two duplicate files of backoff.go exist: config/retry/backoff.go and internal/client/retry/backoff.go. What is the purpose of maintaining two copies? Is it possible to eliminate one of them?

ekexium:
I'm just wondering why this PR both
1. does a rename
image

2. and create new files 
image

If this is by mistake, we can eliminate one of them

Smityz:
> I'm just wondering why this PR both
> 
> 1. does a rename
> 
> image
> 2. and creates new files
> 
> image
> If this is by mistake, we can eliminate one of them

Yes, it's a mistake. I will raise another PR to fix it. Thanks for your reminding

PR #1057, opened 2023-11-13
Fix comment of `SuspendTime`

ekexium:
Based on current TiKV implementation, modify the comments to make them more precise.


util/execdetails.go
```diff
@@ -609,12 +609,13 @@ type TimeDetail struct {
...
 	// this field is called wall time instead of CPU time.
 	ProcessTime time.Duration
-	// Cpu wall time elapsed that task is waiting in queue.
+	// Time elapsed when a coprocessor task yields itself.
 	SuspendTime time.Duration
 	// Off-cpu wall time elapsed in TiKV side. Usually this includes queue waiting time and
 	// other kind of waits in series.
 	WaitTime time.Duration
-	// KvReadWallTime is the time used in KV Scan/Get.
+	// KvReadWallTime is the time used in KV Scan/Get. For get/batch_get,
```
glorv, line 617:
```suggestion
	// KvReadWallTime is the time used in KV Scan/Get. For coprocessor requests,
```
Coprocessor request queue time is more notable due to reschedule

ekexium: My point was that for get/batch_get, the field covers more procedures

glorv: oo, you are right.

glorv (approved):
LGTM

disksing (approved):

PR #1059, opened 2023-11-15
increase large transaction preSplitSizeThreashold

coocood:
cloud-storage-engine has larger region size 512MB, we don't want to split regions into 32MB size on large transaction write.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -786,8 +786,8 @@ func txnLockTTL(startTime time.Time, txnSize int) uint64 {
 	return lockTTL + uint64(elapsed)
 }
 
-var preSplitDetectThreshold uint32 = 100000
-var preSplitSizeThreshold uint32 = 32 << 20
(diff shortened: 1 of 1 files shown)


zeminzhou (approved):

txnkv/transaction/2pc.go
```diff
@@ -786,8 +786,8 @@ func txnLockTTL(startTime time.Time, txnSize int) uint64 {
 	return lockTTL + uint64(elapsed)
 }
 
-var preSplitDetectThreshold uint32 = 100000
-var preSplitSizeThreshold uint32 = 32 << 20
+var preSplitDetectThreshold uint32 = 1000000
```
iosmanthus, line 789: How about making these optional and changing the default value in the tidb-cse?

coocood: That would be much more work to do.

iosmanthus (approved):

PR #1060, opened 2023-11-16
resource_control: set override_priority with resource group priority by default

glorv:
ref tikv/tikv#15994. Pass the priority from client side to avoid performance regression in high cpu usage scenario.


glorv:
@bufferflies @nolouch PTAL

nolouch (approved):

bufferflies (approved):

PR #1077, opened 2023-12-13
region_cache: no need to metrics when inserting for scan regions

CabinfeverB:
close https://github.com/pingcap/tidb/issues/49254


CabinfeverB:
cc @you06 @disksing

you06 (approved):
LGTM

PR #1092, opened 2023-12-27
*: update pd client

glorv:
https://github.com/tikv/pd/pull/7617 introduced some interface change, so need to update pd to newest version.

NOTE: due to the interface change, need to use a patched tidb to pass integration change, we can remove the patch by updated to newest tidb after https://github.com/pingcap/tidb/pull/49833 is merged.


HuSharp:
Maybe now it needs to be replaced with a specific version of TiDB

nolouch (approved):

bufferflies (approved):

PR #1100, opened 2023-12-28
avoid too frequent logs

### LINKED ISSUE #1050: always print  "unable to find stores with given labels" which need to be limited frequency
1、run mussel workload with stale read and closest-replicas
2、enable ttl
3、disk capacity of tikv is near full

always print  "unable to find stores with given labels"

zyguan:
fix #1050


cfzjywxk (approved):
LGTM

cfzjywxk:
@crazycs520 PTAL

PR #1102, opened 2024-01-03
fix: check kill signal against 0

ekexium:
Previously the kill signal was checked against `1`, however TiDB uses non-zero values to indicate a kill signal.

### DIFF
--- a/config/retry/backoff.go
+++ b/config/retry/backoff.go
@@ -218,7 +218,12 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 	}
 
 	if b.vars != nil && b.vars.Killed != nil {
-		if atomic.LoadUint32(b.vars.Killed) == 1 {
+		killed := atomic.LoadUint32(b.vars.Killed)
+		if killed != 0 {
+			logutil.BgLogger().Info(
+				"backoff stops because a killed signal is received",
+				zap.Uint32("signal", killed),
+			)
 			return errors.WithStack(tikverr.ErrQueryInterrupted)
 		}
 	}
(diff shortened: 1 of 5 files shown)


ekexium:
Why are there two identical `backoff.go` ? I just noticed this😓 @disksing any idea?

config/retry/backoff.go
```diff
@@ -218,7 +218,12 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 	}
 
 	if b.vars != nil && b.vars.Killed != nil {
-		if atomic.LoadUint32(b.vars.Killed) == 1 {
+		killed := atomic.LoadUint32(b.vars.Killed)
+		if killed != 0 {
+			logutil.BgLogger().Info(
+				"backoff stops because a killed signal is received",
+				zap.Uint32("signal", killed),
+			)
```
cfzjywxk, line 226: How about wrapping the kill signal check into a function inside the `Backoffer` type.
Besides, we need to make sure this method is used by all backoff types.

ekexium: All backoff methods go into `BackoffWithCfgAndMaxSleep` which checks the kill signal. Is this what you meant?

cfzjywxk: Yes

internal/locate/region_request.go
```diff
@@ -1478,8 +1478,15 @@ func (s *RegionRequestSender) SendReqCtx(
 
 		// recheck whether the session/query is killed during the Next()
 		boVars := bo.GetVars()
-		if boVars != nil && boVars.Killed != nil && atomic.LoadUint32(boVars.Killed) == 1 {
-			return nil, nil, retryTimes, errors.WithStack(tikverr.ErrQueryInterrupted)
+		if boVars != nil && boVars.Killed != nil {
+			killed := atomic.LoadUint32(boVars.Killed)
```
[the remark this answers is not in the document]
ekexium: Yes

zyguan:
> Why are there two identical `backoff.go` ? I just noticed this😓 @disksing any idea?

I guess the `internal/client/retry` is redundant and can be removed.

cfzjywxk (approved):

cfzjywxk:
@ekexium Please update the branch

Issue #1104, opened 2024-01-04
RegionCache Enhancement


zyguan:
Region cache is the core component of kv client. It maintains the information of regions and stores, which is required to send requests to the right TiKV nodes.

Currently, there are some stability issues related to region cache:
- follower read may access a slow store, which leads to remarkable performance degradation.
- region cache can not reload region properly when there is down peer, which leads to issues like https://github.com/tikv/client-go/issues/879 .
- region cache may invoke GetRegion too frequently, which leads to heavy load on PD.

In order to resolve these issues and build an unified health controller, here are some preparatory tasks related to region cache:

- Store Management
  - [ ] Simplify store life cycle (unify state and liveness).
  - [ ] Extract store index from region cache.
  - [ ] Maintain health feedback from each store (shall be the input of replica selection).
- Region Management
  - [ ] Simplify region state (deprecate fields like sync_flag).
  - [ ] Let region cache be aware of down peers.
  - [ ] Reload region properly according to its state and peers' state.
  - [ ] Investigate the root cause of too frequent GetRegion problem.

PR #1106, opened 2024-01-05
region_cache: restrict access to store related fields

zyguan:
This is the first stage of extracting store management from region cache. (ref https://github.com/tikv/client-go/issues/1104)


cfzjywxk:
@crazycs520 PTAL

crazycs520 (approved):

PR #1109, opened 2024-01-10
make experimental improvements to replica selector

zyguan:
- pick #1100
- [stale read] support increasing timeout when retry leader
- [stale read] support disabling retry follower when leader is unavailable

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -80,6 +80,34 @@ func LoadShuttingDown() uint32 {
 	return atomic.LoadUint32(&shuttingDown)
 }
 
+// ReplicaSelectorExperimentalOptions defines experimental options of replica selector.
+type ReplicaSelectorExperimentalOptions struct {
+	StaleRead struct {
+		PreventRetryFollower     bool
+		RetryLeaderTimeoutFactor float64
+	}
+}
+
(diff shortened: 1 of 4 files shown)


cfzjywxk:
@zyguan @crazycs520 @MyonKeminta @you06 
We need to revisit this PR as the read index confirm of follower read is quite slow and it does not help increase the request success rate but cause read pool tasks to accumulate.

internal/locate/region_request.go
```diff
@@ -80,6 +80,34 @@ func LoadShuttingDown() uint32 {
...
+type ReplicaSelectorExperimentalOptions struct {
+	StaleRead struct {
+		PreventRetryFollower     bool
+		RetryLeaderTimeoutFactor float64
+	}
+}
+
+var (
+	selectorExpOpts     ReplicaSelectorExperimentalOptions
+	selectorExpOptsLock sync.RWMutex
```
crazycs520, line 93: How about use `atomic.Value` to avoid use global mutex.

crazycs520 (approved):

crazycs520:
@MyonKeminta @cfzjywxk PTAL

cfzjywxk (approved):

PR #1114, opened 2024-01-12
txn: introduce pipelined memdb, Flush and BufferBatchGet for pipelined DML

you06:
Introduce pipelined memdb which support flushing mutations into storage engine during execution.

The pipelined memdb is the basic support of pipelined DML, which avoids linear memory requirement when the DML's size increasing.

The main changes:
* A memdb which can be dynamically flushed into stores, when the flushing is slow, the memdb may block writing to avoid OOM.
* Add metrics of flush keys and size of memdb.
* A new transaction commit path, which utilize flush and resolve lock requests.
  * The flush takes the same effect of prewrite and resolve lock is same to commit secondaries.
  * Add buffer batch get which can read flushed mutations.

Before implement this PR, main test results are from #1150's experiments.

See the integration test result in pingcap/tidb#50389

### Pipelined MemDB

After each row is processed, the `MayFlush` need to be invoked and the pipelined memdb will check if it needs to be flushed.
client-go-1114

### Flush

When flushing, the new `FlushRequest` protocol will be utilized. `FlushRequest` takes the same effect with `PrewriteRequest` in general, but we use a different protocol to devide them into different metrics and code paths.

When flush, a monotonic generation field is set, which is used to prevent retried request overwrite the later mutation.

Once flushing a memdb is failed after retry, the transaction aborts.

### Commit by Scan

Because we donot record the secondaries(the memory is not enough to store the keys of large DML), the common `CommonRequest` cannot be used. Thus we commit by `ResolveLockRequest`, which will actually scan & commit all locks in the target region.

The bounds of related keyspace is recorded during the execution, and only the `ResolveLockRequest` will be sent to related regions instead of scanning the global keyspace.


internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+		// Send the error to errCh after onFlushing status is set to false.
+		// this guarantees the onFlushing.Store(true) in another goroutine's MayFlush happens after onFlushing.Store(false) in this function.
+		p.errCh <- err
+	}()
+	return nil
+}
+
+// Flush will force flushingMemBuffer all the mutations, if there is an ongoing flush memdb, it will wait.
+// this should be called before running commit protocol.
+func (p *PipelinedMemDB) Flush() error {
```
ekexium, line 129: If we view the memdb in a layer under the transaction client, the function had better not assume anything on how it is used, like "this should be called before running commit protocol", at least the interface(name and comments) does not prevent upper users from using it elsewhere. The current implementation skips stuff like making a new memdb because of this assumption.
Maybe we could implement a function like doFlush(), and let MayFlush and ForceFlush both call it.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+//   - like MemDB, PipelinedMemDB also CANNOT be used concurrently
+type PipelinedMemDB struct {
+	// Like MemDB, this RWMutex only used to ensure memdbSnapGetter.Get will not race with
+	// concurrent memdb.Set, memdb.SetWithFlags, memdb.Delete and memdb.UpdateFlags.
+	sync.RWMutex
+	onFlushing        atomic.Bool
+	errCh             chan error
+	flushFunc         FlushFunc
+	memBuffer         *MemDB
+	flushingMemBuffer *MemDB
```
ekexium, line 38: It is a bit subtle to understand how concurrency affects the field, because of the existence of async flushes that run in other goroutines. 
Is it only mutated and accessed in the main thread of memdb since we only allow single-threaded usage, so there is no data race on this field?
Maybe we could make some comments explaning more about this.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+// Dirty returns whether the pipelined buffer is mutated.
+func (p *PipelinedMemDB) Dirty() bool {
+	return p.memBuffer.Dirty() || p.flushingMemBuffer != nil
+}
+
+// Get the value by given key, it returns tikverr.ErrNotExist if not exist.
+// The priority of the value is MemBuffer > flushingMemBuffer MemBuffer > Snapshot.
```
ekexium, line 64:
```suggestion
// The priority of the value is MemBuffer > flushingMemBuffer > Snapshot.
```

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+func (p *PipelinedMemDB) Get(k []byte) ([]byte, error) {
+	v, err := p.memBuffer.Get(k)
+	if p.flushingMemBuffer != nil && tikverr.IsErrNotFound(err) {
+		v, err = p.flushingMemBuffer.Get(k)
+	}
+	if err != nil {
```
ekexium, line 70: Do we need to check `tikverr.IsErrNotFound(err)` here? I suppose for each level of Get we should take both ErrNotFound and other errors into consideration

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+		return 0, err
+	}
+	return f, nil
+}
+
+// MayFlush is called during execution of a transaction, it does flush when there are enough keys and the ongoing flushingMemBuffer is done.
+func (p *PipelinedMemDB) MayFlush() error {
+	if p.flushFunc == nil {
+		return nil
```
ekexium, line 93: Shall we return an error here as no flushFunc is unexpected?

txnkv/transaction/pipelined_flush.go
```diff
@@ -0,0 +1,448 @@
...
+			RequestSource:          c.txn.GetRequestSource(),
+			ResourceControlContext: &kvrpcpb.ResourceControlContext{
+				ResourceGroupName: c.resourceGroupName,
+			},
+		},
+	)
+	return r
+}
+
+func (action actionPipelinedFlush) handleSingleBatch(
```
ekexium, line 101: It should not block this PR but there seems to be many unnecessary duplicated code for various actions...

txnkv/transaction/txn.go
```diff
@@ -162,6 +164,9 @@ type KVTxn struct {
 	aggressiveLockingDirty   atomic.Bool
 
 	forUpdateTSChecks map[string]uint64
+
+	isPipelined    bool
+	pipelinedMemDB *unionstore.PipelinedMemDB
```
ekexium, line 169: Is there any case that we may need both `us` and `pipelinedMemDB` in a same transaction? I thought we should make a new interface and make the original unionstore and PipelinedMemDB 2 implementations of the new interface. So only one field is needed here.

txnkv/transaction/txn.go
```diff
@@ -392,6 +402,138 @@ func (txn *KVTxn) IsPessimistic() bool {
...
+	txn.isPipelined = true
+	txn.snapshot.SetPipelined()
+	// TODO: set the correct sessionID
+	committer, err := newTwoPhaseCommitter(txn, 0)
+	if err != nil {
+		return err
+	}
+	txn.committer = committer
+	// disable 1pc and async commit for pipelined txn.
+	txn.committer.setOnePC(false)
```
ekexium, line 420: WARN logs are needed if any incompatible features are enabled by user as they are problematic usage

internal/unionstore/pipelined_memdb.go
you06: The comment "this should be called before running commit protocol" is for reminding client-go developer.

Upper users can use it when they want to flush all the mutations into stores and wait for it.

Maybe the interface can be:

```
MayFlush() error // may flush in another goroutine, returns error if there are any failed flushing.
ForceFlush() error // force flush when it's not empty in another goroutine, returns error if there are any failed flushing.
Wait() error // wait the flush result if there is onflushing memdb.
```

internal/unionstore/pipelined_memdb.go
you06: > Is it only mutated and accessed in the main thread of memdb since we only allow single-threaded usage, so there is no data race on this field?

Yes, the usage is same as the memdb, no concurrently read & write.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+	v, err := p.memBuffer.Get(k)
+	if p.flushingMemBuffer != nil && tikverr.IsErrNotFound(err) {
+		v, err = p.flushingMemBuffer.Get(k)
+	}
+	if err != nil {
+		return v, err
+	}
+	if len(v) == 0 {
+		return nil, tikverr.ErrNotExist
+	}
```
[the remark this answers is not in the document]
you06: Because `PipelinedMemDB` implement `Getter` interface, it doesn't accept a go context, it's more like a fast in-mem reader, I currently implement buffer_batch_get in upper layer.

Maybe we need to change the function to `func (p *PipelinedMemDB) Get(context.Context, []byte, tier int) ([]byte, error) {...}`?

metrics/metrics.go
```diff
@@ -726,6 +728,24 @@ func initMetrics(namespace, subsystem string, constLabels prometheus.Labels) {
 			Help:      "Counter of stale read requests bytes",
 		}, []string{LblResult, LblDirection})
 
+	TiKVPipelinedFlushLenHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "pipelined_flush_len",
+			Help:      "Bucketed histogram of length of pipelined flushed memdb",
+			Buckets:   prometheus.ExponentialBuckets(10000, 2, 13), // 10K ~ 40M
```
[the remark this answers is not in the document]
you06: Super long values may lead to flush batch with keys less than `MinFlushKeys`(10000), what about `prometheus.ExponentialBuckets(1000, 2, 16)`(1K to 32M)?

32M keys should be enough because the MaxFlushSize is 128MB, it's about 4 bytes per KV, even the size of key is longer than 4 bytes.

txnkv/transaction/txn.go
you06: > Is there any case that we may need both us and pipelinedMemDB in a same transaction?

No.

> make a new interface and make the original unionstore and PipelinedMemDB 2 implementations of the new interface.

There are 2 concerns:

* TiDB strongly depends on `GetUnionStore()` and `GetUnionStore().GetMemBuffer()`, with an interface, unionstore cannot be accessed directly.
* The `Get` function may introduce a tier parameter.

I'll try this abstraction.

txnkv/transaction/txn.go
you06: One pc and async commit fields are turned on in `committer.execute`.

Manually setting them to false is just a emphasis.

WARN logs only take effects when the `newTwoPhaseCommitter` returns committer with one-pc or async-commit on, which is impossible by now.

txnkv/transaction/txn.go
```diff
@@ -392,6 +402,138 @@ func (txn *KVTxn) IsPessimistic() bool {
...
+				return errors.New("invalid iterator")
+			}
+			endKey := it.Key()
+			if len(txn.committer.pipelinedEnd) == 0 || bytes.Compare(txn.committer.pipelinedEnd, endKey) < 0 {
+				txn.committer.pipelinedEnd = make([]byte, len(endKey))
+				copy(txn.committer.pipelinedEnd, endKey)
+			}
+			it.Close()
+		}
+		for it := memdb.IterWithFlags(nil, nil); it.Valid(); err = it.Next() {
```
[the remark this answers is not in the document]
you06: Yes, I wish it can reused, but tiny differences may be annoyed, let's leave a TODO.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+// PipelinedMemDB is a Store which contains
+//   - a mutable buffer for read and write
+//   - an immutable onflushing buffer for read
+//   - like MemDB, PipelinedMemDB also CANNOT be used concurrently
+type PipelinedMemDB struct {
+	// Like MemDB, this RWMutex only used to ensure memdbSnapGetter.Get will not race with
+	// concurrent memdb.Set, memdb.SetWithFlags, memdb.Delete and memdb.UpdateFlags.
+	sync.RWMutex
```
cfzjywxk, line 33: `MemDB` already has an internal `RWMutex`, what is this `RWMutex` used to protect? Is it better to group the to be protected fields inside the `RWMutex` like
```
sync.RWMutex  rwLock {
     ...
}
```

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+	onFlushing        atomic.Bool
+	errCh             chan error
+	flushFunc         FlushFunc
+	memBuffer         *MemDB
+	flushingMemBuffer *MemDB
+	len, size         int // len and size records the total flushed and onflushing memdb.
+}
+
+const (
+	MinFlushKeys = 10000
```
cfzjywxk, line 43: Need to add comments to explain about why choosing these values.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+	if p.flushingMemBuffer != nil && tikverr.IsErrNotFound(err) {
+		f, err = p.flushingMemBuffer.GetFlags(k)
+	}
+	if err != nil {
+		return 0, err
+	}
+	return f, nil
+}
+
+// MayFlush is called during execution of a transaction, it does flush when there are enough keys and the ongoing flushingMemBuffer is done.
```
cfzjywxk, line 90: It seems we could not distinguish if the flush is triggered or not if the returned error is `nil`? Because if there is ongoing flushing task, the returned value is also `nil`. I think It's better to return the exact results for a function call.

txnkv/transaction/txn.go
```diff
@@ -162,6 +164,9 @@ type KVTxn struct {
 	aggressiveLockingDirty   atomic.Bool
 
 	forUpdateTSChecks map[string]uint64
+
+	isPipelined    bool
```
cfzjywxk, line 168: Is it enough to just using one field `pipelinedMemDB *unionstore.PipelinedMemDB` and `pipelinedMemDB == nil` to indicate if it's pipelined?

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,159 @@
...
+	if p.memBuffer.Len() > 0 {
+		metrics.TiKVPipelinedFlushLenHistogram.Observe(float64(p.memBuffer.Len()))
+		metrics.TiKVPipelinedFlushSizeHistogram.Observe(float64(p.memBuffer.Size()))
+		return p.flushFunc(p.memBuffer)
+	}
+	return nil
+}
+
+// GetMemDB returns the mutable memdb for writing.
+func (p *PipelinedMemDB) GetMemDB() *MemDB {
```
cfzjywxk, line 145: Or is it needed to abstract the interfaces for `Memory DB` and just provide extrenal interface implementations, so the users may call `Set` or `Get` directly instead of exposing the internal `MemDB`. Maybe we would change the implementation doing further optimizations.

txnkv/transaction/pipelined_flush.go
```diff
@@ -0,0 +1,448 @@
...
+	"github.com/tikv/client-go/v2/internal/locate"
+	"github.com/tikv/client-go/v2/internal/logutil"
+	"github.com/tikv/client-go/v2/tikvrpc"
+	"github.com/tikv/client-go/v2/txnkv/txnlock"
+	"github.com/tikv/client-go/v2/util"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+type actionPipelinedFlush struct {
```
cfzjywxk, line 40: We should also add some integration test cases simualting how the `PipelinedMemDB` would used from the perspective like tidb execution, to cover the possible external interfaces usages and behaviours for it.

internal/unionstore/pipelined_memdb.go
you06: This mutex is only used to prevent concurrent read/write from upper layer, actually it protects the whole memdb, used as require. It's a patch to the origin memdb, without this embed mutex, this memdb does not satisfy the upper layer interface.

txnkv/transaction/txn.go
you06: I made an interface for `MemDB` and `PipelinedMemDB`, it's named `MemBuffer`. I think we can reuse the snapshot read code in `KVUnionStore`

See the changes in `union_store.go`.

The only change to user interface is `Get` now accept a context, this may break compile for projects which import client-go.

metrics/metrics.go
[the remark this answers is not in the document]
ekexium: Looks good

txnkv/transaction/txn.go
ekexium: Emm can we simply make assertions then?

internal/unionstore/pipelined_memdb.go
you06: I changed the interface to:

```go
// It always flush in another goroutine.
// Flush(true) should be invoked in the end of a transaction.
// Upper layer users can use Flush(false) or Flush(true) as they want.
Flush(force bool) (bool, error)
FlushWait() error
```

This leaves the optimize change for multi-statement usage:

* Call `Flush(true)` in the end of a DML
* Call `FlushWait()` before the next statement's execution
* Then we can ensure all the mutations are flushed into stores and skip union scan.

internal/unionstore/pipelined_memdb.go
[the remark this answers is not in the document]
you06: Add the buffer_batch_get, and we can skip it when lazy check existence when TiKV handling flush requests.

metrics/metrics.go
```diff
@@ -726,6 +728,24 @@ func initMetrics(namespace, subsystem string, constLabels prometheus.Labels) {
...
+		})
+
+	TiKVPipelinedFlushSizeHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "pipelined_flush_size",
+			Help:      "Bucketed histogram of size of pipelined flushed memdb",
+			Buckets:   prometheus.ExponentialBuckets(16777216, 1.2, 13), // 16M ~ 142M
```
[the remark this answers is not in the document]
you06: Current `MinFlushSize` and `MaxFlushSize` is 16M and 128M, so 142M is enough for batch size <= `MaxFlushSize`. Base by 1.2 is for more detailed buckets.

If we want larger value, `MaxFlushSize` should also be increased, unless the pipelined memdb blocks. I have no idea which value is the best, may make it configuration in the future.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+}
+
+// Dirty returns whether the pipelined buffer is mutated.
+func (p *PipelinedMemDB) Dirty() bool {
+	return p.memBuffer.Dirty() || p.len > 0
+}
+
+// GetMemDB returns the mutable memdb ONLY for writing.
+func (p *PipelinedMemDB) GetMemDB() *MemDB {
```
cfzjywxk, line 82: Is this still needed if the `PipelinedMemDB` has implemented the same interface with `MemDB`?

you06: It's a sideway to access the inner memdb, and should be unreachable for pipelined memdb.

Because the usual `memBufferMutations` need to be built from `MemDB` objects, the original memdb interface is provided.

https://github.com/tikv/client-go/blob/b8855b2e80fa0dc76d9b4f504f9e9b22d5590cd6/txnkv/transaction/2pc.go#L549-L551

cfzjywxk:
@you06 Please resolve the conflicts.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,286 @@
...
+	return p.memBuffer
+}
+
+// Get the value by given key, it returns tikverr.ErrNotExist if not exist.
+// The priority of the value is MemBuffer > flushingMemBuffer > Snapshot.
+func (p *PipelinedMemDB) Get(ctx context.Context, k []byte) ([]byte, error) {
+	v, err := p.memBuffer.Get(k)
+	if p.flushingMemBuffer != nil && tikverr.IsErrNotFound(err) {
```
cfzjywxk, line 90: If the returned error is not `NotFound`, the error should be returned direcrly, perhaps
```
if err != nil {
   if p.flushingMemBuffer != nil && tikverr.IsErrNotFound(err) {
      ...
   } else {
     return nil, err
   }
}
v, err = p.flushingMemBuffer.Get(k)
if err != nil {
   if tikverr.IsErrNotFound(err) && ctx.Value(pipelinedMemDBSkipRemoteBufferKey) == nil {
   }
}
```

is more clear and the error handing would not be forgotten.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,286 @@
...
+		// size < MinFlushSize, do not flush.
+		// MinFlushSize <= size < MaxFlushSize && keys < MinFlushKeys, do not flush.
+		// MinFlushSize <= size < MaxFlushSize && keys >= MinFlushKeys, flush.
+		// size >= MaxFlushSize, flush.
+		if size < MinFlushSize || (p.memBuffer.Len() < MinFlushKeys && size < MaxFlushSize) {
+			return false, nil
+		}
+		if !p.onFlushing.CompareAndSwap(false, true) && size < MaxFlushSize {
+			return false, nil
+		}
```
cfzjywxk, line 164: Better to abstract a `NeedFlush` function.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,286 @@
...
+func (p *PipelinedMemDB) Mem() uint64 {
+	return 0
+}
+
+// Reset the PipelinedMemDB to init state.
+func (p *PipelinedMemDB) Reset() {
+	p.memBuffer.Reset()
+	p.flushingMemBuffer = nil
+	p.len, p.size = 0, 0
+	p.onFlushing.Store(false)
```
cfzjywxk, line 262: Why the CAS operation is used in the `Flush` function but the `Store` is used directly here? I dont't quite get the strategy of the concurrency design and handling here.

If the `write` operations of `PipelinedMemDB` are designed to be handled under the protection of `mutex`, is the `CAS` operation unnecessary?

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,286 @@
...
+	default:
+	}
+}
+
+// Staging is not supported for PipelinedMemDB, it returns 0 handle.
+func (p *PipelinedMemDB) Staging() int {
+	return 0
+}
+
+// Cleanup implements MemBuffer interface.
```
cfzjywxk, line 274: Need to comment the belowing interfaces are not supported by the `PipelinedMemDB` as they do not return errors.

txnkv/transaction/2pc.go
```diff
@@ -1528,6 +1539,20 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
...
+		if _, err = c.txn.GetMemBuffer().Flush(true); err != nil {
+			return err
+		}
+		if err = c.txn.GetMemBuffer().FlushWait(); err != nil {
+			return err
+		}
+		if len(c.pipelinedCommitInfo.pipelinedStart) > 0 && len(c.pipelinedCommitInfo.pipelinedEnd) > 0 {
+			return c.commitFlushedMutations(bo)
+		} else {
+			panic("unreachable")
```
cfzjywxk, line 1552: We could return unexpected error later.

txnkv/transaction/txn.go
```diff
@@ -392,6 +404,148 @@ func (txn *KVTxn) IsPessimistic() bool {
 	return txn.isPessimistic
 }
 
+// IsPipelined returns true if it's a pipelined transaction.
+func (txn *KVTxn) IsPipelined() bool {
+	return txn.isPipelined
+}
+
+func (txn *KVTxn) SetPipelinedMemDB() (unionstore.MemBuffer, error) {
```
cfzjywxk, line 412: Perhaps `InitPipelinedMemDB` is better and the usage is 
```
if err := txn.InitPipelinedMemDB(); err != nil {
    return err.
}
```

txnkv/transaction/txn.go
```diff
@@ -392,6 +404,148 @@ func (txn *KVTxn) IsPessimistic() bool {
...
+	// TODO: set the correct sessionID
+	committer, err := newTwoPhaseCommitter(txn, 0)
+	if err != nil {
+		return nil, err
+	}
+	txn.committer = committer
+	// disable 1pc and async commit for pipelined txn.
+	if txn.committer.isOnePC() || txn.committer.isAsyncCommit() {
+		logutil.BgLogger().Warn("[pipelined dml] should not enable 1pc or async commit for pipelined txn",
+			zap.Bool("1pc", txn.committer.isOnePC()), zap.Bool("async commit", txn.committer.isAsyncCommit()))
```
cfzjywxk, line 427: Need to print the `start_ts` of the related transaction.

integration_tests/pipelined_memdb_test.go
```diff
@@ -0,0 +1,180 @@
...
+		s.True(flushed)
+		close(flushReturned)
+	}()
+
+	oneSec := time.After(time.Second)
+	select {
+	case <-flushReturned:
+		s.Fail("Flush should be blocked")
+	case <-oneSec:
+	}
```
cfzjywxk, line 155: When the flush is blocked, need to verify the latter write would be stalled when membuf  is full.

internal/unionstore/pipelined_memdb.go
you06: Like `MemDB`, `PipelinedMemDB` also does not allow parallel read/write operations, but the flush function runs in another goroutine, the atomic variable `onFlushing` avoids data races.

I remove the CAS operation by abstracting `needFlush` function and replace it with `atomic.Load`, the atomic variable is for quick checking if there is a ongoing flush task.

integration_tests/pipelined_memdb_test.go
you06: Latter write should not come during the flush blcok period.

The application may looks like this, no parallel `Flush` and `Set`:

```
for {
  txn.Set(...)
  txn.GetMemBuffer().Flush(false)
}

File: internal/unionstore/pipelined_memdb.go [hunk 28]
```diff
@@ -0,0 +1,313 @@
...
+//   - like MemDB, PipelinedMemDB also CANNOT be used concurrently
+type PipelinedMemDB struct {
+	// Like MemDB, this RWMutex only used to ensure memdbSnapGetter.Get will not race with
+	// concurrent memdb.Set, memdb.SetWithFlags, memdb.Delete and memdb.UpdateFlags.
+	sync.RWMutex
+	onFlushing        atomic.Bool
+	errCh             chan error
+	flushFunc         FlushFunc
+	bufferBatchGetter BufferBatchGetter
+	memBuffer         *MemDB
```
ekexium, line 41: Do we distinguish between "membuffer" and "memdb"? I thought memdb is the lower layer, and membuffer is a wrapper of it. If that's the convention, we should consider renaming the fields.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,313 @@
...
+		return 0, err
+	}
+	return f, nil
+}
+
+func (p *PipelinedMemDB) UpdateFlags(k []byte, ops ...kv.FlagsOp) {
+	p.memBuffer.UpdateFlags(k, ops...)
+}
+
+func (p *PipelinedMemDB) RemoveFromBuffer(key []byte) {}
```
ekexium, line 141: Let's panic here to avoid misuse

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// DeleteWithFlags deletes the key k in the MemBuffer with flags.
+func (p *PipelinedMemDB) DeleteWithFlags(key []byte, ops ...kv.FlagsOp) error {
+	return p.memBuffer.DeleteWithFlags(key, ops...)
+}
+
+// Flush is called during execution of a transaction, it does flush when there are enough keys and the ongoing flushingMemBuffer is done.
+// When the mutable memdb is too large, it blocks until the ongoing flush is done.
+func (p *PipelinedMemDB) Flush(force bool) (bool, error) {
```
ekexium, line 165: We'd better comment the meaning of the return values.

internal/unionstore/pipelined_memdb_test.go
```diff
@@ -0,0 +1,280 @@
...
+		memdb.GetMemDB().Set(key, value)
+	}
+	flushed, err := memdb.Flush(false)
+	require.Nil(t, err)
+	require.True(t, flushed)
+	require.True(t, memdb.OnFlushing())
+	require.Equal(t, memdb.memBuffer.Len(), 0)
+	require.Equal(t, memdb.memBuffer.Size(), 0)
+
+	// When size of memdb is greater than MaxFlushSize, MayFlush will be blocked.
```
ekexium, line 153: Are the comments and names including "MayFlush" stale?

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,313 @@
...
+	p.memBuffer = newMemDB()
+	p.generation++
+	go func(generation uint64) {
+		util.EvalFailpoint("beforePipelinedFlush")
+		metrics.TiKVPipelinedFlushLenHistogram.Observe(float64(p.flushingMemBuffer.Len()))
+		metrics.TiKVPipelinedFlushSizeHistogram.Observe(float64(p.flushingMemBuffer.Size()))
+		err := p.flushFunc(generation, p.flushingMemBuffer)
+		p.onFlushing.Store(false)
+		// Send the error to errCh after onFlushing status is set to false.
+		// this guarantees the onFlushing.Store(true) in another goroutine's MayFlush happens after onFlushing.Store(false) in this function.
```
ekexium, line 190: Comments need to be updated. There is no `MayFlush` now.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,313 @@
...
+	}(p.generation)
+	return true, nil
+}
+
+func (p *PipelinedMemDB) needFlush() bool {
+	size := p.memBuffer.Size()
+	// size < MinFlushSize, do not flush.
+	// MinFlushSize <= size < MaxFlushSize && keys < MinFlushKeys, do not flush.
+	// MinFlushSize <= size < MaxFlushSize && keys >= MinFlushKeys, flush.
+	// size >= MaxFlushSize, flush.
```
ekexium, line 201: `MaxFlushSize` implies a maximum size limit for a flush. However, its true purpose is to act as a threshold to trigger a flush. A more obvious name could be something like `ForceFlushSizeThreshold`.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,313 @@
...
+		err := <-p.errCh
+		// cleanup the flushingMemBuffer so the next call of FlushWait will not wait for the error channel.
+		p.flushingMemBuffer = nil
+		return err
+	}
+	return nil
+}
+
+// Iter implements the Retriever interface.
+func (p *PipelinedMemDB) Iter([]byte, []byte) (Iterator, error) {
```
ekexium, line 223: Do we need to return an emtpy iter here? What will happen during a UnionIter?

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,313 @@
...
+//   - a mutable buffer for read and write
+//   - an immutable onflushing buffer for read
+//   - like MemDB, PipelinedMemDB also CANNOT be used concurrently
+type PipelinedMemDB struct {
+	// Like MemDB, this RWMutex only used to ensure memdbSnapGetter.Get will not race with
+	// concurrent memdb.Set, memdb.SetWithFlags, memdb.Delete and memdb.UpdateFlags.
+	sync.RWMutex
+	onFlushing        atomic.Bool
+	errCh             chan error
+	flushFunc         FlushFunc
```
ekexium, line 39: Is there a specific reason to pass a flush func from outside rather than making it a method or function defined inside this module? From my perspective, it is more obscure to understand which variables are read by the flush func, and which variables it modifies. It makes understanding the concurrency correctness a bit harder (at least for me)?

internal/unionstore/pipelined_memdb.go
ekexium: Let's just panic.

examples/gcworker/go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	go.uber.org/multierr v1.11.0 // indirect
 	go.uber.org/zap v1.26.0 // indirect
 	golang.org/x/exp v0.0.0-20230711005742-c3f37128e5a4 // indirect
+<<<<<<< HEAD
```
ekexium, line 42: There are some merge conflicts.

internal/unionstore/pipelined_memdb.go
you06: Nice catch, IMO, membuffer is an interface and memdb is a implementation now, so better using `memDB` and `flushingMemDB` here, what do you think?

internal/unionstore/pipelined_memdb.go
you06: An empty iter can lead to wrong resultset. Image the following usage:

```
begin()
put(k, 1)
flush()
iter() // this equals to snapshot iter if an empty returned from PipelinedMemDB.Iter, put(k, 1) is missing
```

internal/unionstore/pipelined_memdb.go
you06: There are 2 reasons of passing a flush func:

1. I'm hesitate to introduce RPC calling code into the unionstore module, it sounds like an in-mem module.
2. If there are some other implementation of flush func(like S3 based flush), we can quickly replace it, so the flush func is a plugin module for pipelined memdb instead of core logic.

internal/unionstore/pipelined_memdb.go
ekexium: Looks good

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,313 @@
...
+	if p.flushingMemBuffer != nil {
+		v, err = p.flushingMemBuffer.Get(k)
+		if err == nil {
+			return v, nil
+		}
+		if !tikverr.IsErrNotFound(err) {
+			return nil, err
+		}
+	}
+	if ctx.Value(pipelinedMemDBSkipRemoteBufferKey) != nil {
```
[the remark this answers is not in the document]
you06: What about adding some methods for `MemBuffer` interface:

```go
type MemBuffer interface {
	...
	// Exist methods
	// Get gets the value for key k from the MemBuffer.
	Get(context.Context, []byte) ([]byte, error)
	// GetFlags gets the flags for key k from the MemBuffer.
	GetFlags([]byte) (kv.KeyFlags, error)

	// New methods
	GetFromTier(context.Context, []byte, tier int) ([]byte, error)
	GetFlagsFromTier([]byte) (kv.KeyFlags, error)
```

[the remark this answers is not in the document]
ekexium: Exposing the concept tier to users seems unnecessary. The usage of this method may be a cheap check of mutations in memory. So how about "GetLocal"? For vanilla memdb, it is aliased to `Get`.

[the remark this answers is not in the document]
ekexium: We can leave it as a TODO. It should be considered as an optimization.

internal/unionstore/pipelined_memdb.go
you06: `Iter` and `IterReverse` return error directly, and I just added an error iterator for `SnapshotIter` and `SnapshotIterReverse`, they should never be called for pipelined memdb.

In TiDB, those iterators are only used in unionscan, which never happens in autocommit statement.

https://github.com/tikv/client-go/blob/d480f284026dacd3d50e72e826fa2f1ace6c0fd3/internal/unionstore/pipelined_memdb.go#L261-L279

internal/unionstore/pipelined_memdb.go
ekexium: Seems to me the only meaning of the rwlock is to provide implement `RLock` and `RUnlock` of the interface. There is no actual usage of `Lock`, right?

internal/unionstore/pipelined_memdb.go
you06: Batch get returns empty map when the key is not found, so there will be no `tikverr.ErrNotExist` here.

And what ever the error is `tikverr.ErrNotExist` or other type error, we return it directly, seems meaningless check.

internal/unionstore/pipelined_memdb.go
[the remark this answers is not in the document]
you06: Added.

internal/unionstore/pipelined_memdb.go
you06: Done.

txnkv/transaction/txn.go
you06: Now we union the `MemDB` and `PipelinedMemDB` into `MemBuffer` interface, this field is required.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// Reset the PipelinedMemDB to init state.
+func (p *PipelinedMemDB) Reset() {
+	p.memDB.Reset()
+	p.flushingMemDB = nil
+	p.len, p.size = 0, 0
+	p.onFlushing.Store(false)
+	p.generation = 0
+	// wait for the error channel to be empty.
```
cfzjywxk, line 295: Should we wait for the flush to be finished first before the destruction? What are the use cases for `Reset`, when Reset is called and flush gets stuck, will it cause any issues?

metrics/metrics.go
```diff
@@ -105,6 +105,8 @@ var (
 	TiKVStaleReadCounter                     *prometheus.CounterVec
 	TiKVStaleReadReqCounter                  *prometheus.CounterVec
 	TiKVStaleReadBytes                       *prometheus.CounterVec
+	TiKVPipelinedFlushLenHistogram           prometheus.Histogram
+	TiKVPipelinedFlushSizeHistogram          prometheus.Histogram
```
cfzjywxk, line 109: A flush duration metric may also be needed.

txnkv/transaction/pipelined_flush.go
```diff
@@ -0,0 +1,448 @@
...
+			err = errors.Errorf("unexpected resolve err: %s", keyErr)
+			logutil.BgLogger().Error("resolveLock error", zap.Error(err))
+			return 0, err
+		}
+		return 1, nil
+	}
+}
+
+func (c *twoPhaseCommitter) resolveFlushedLocks(bo *retry.Backoffer, start, end []byte) {
+	const RESOLVE_CONCURRENCY = 8
```
cfzjywxk, line 380: The `BatchExecutor` already has a token limit to control concurrency, is it possible to re-use it?

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1025,6 +1078,17 @@ func (s *KVSnapshot) getResolveLockDetail() *util.ResolveLockDetail {
...
 
+// SetPipelined sets the snapshot to pipelined mode.
+func (s *KVSnapshot) SetPipelined(ts uint64) {
+	s.isPipelined = true
+	// In pipelined mode, some locks are flushed into stores during the execution.
+	// If a read request encounters these pipelined locks, it'll be a situation where lock.ts == start_ts.
+	// In order to allow the snapshot to proceed normally, we need to skip these locks.
+	// Otherwise, the transaction will attempt to resolve its own lock, leading to a mutual wait with the primary key TTL.
+	// Currently, we skip these locks by resolvedLocks mechanism.
+	s.resolvedLocks.Put(ts)
```
cfzjywxk, line 1089: Should this be handled at the TiKV side when, it is known if the lock is a pipelined lock?
/cc @ekexium

internal/unionstore/pipelined_memdb.go
you06: `Reset` should not be used IMO, because some flushed mutations can not be dropped by calling reset...

I don't see any usage of this function, what about just panic here?

txnkv/transaction/pipelined_flush.go
you06: Current `batchExecutor` is mutation based, since here we do not have any mutations, it's not applicable now.

I'll add a TODO which unify them by replace the mutation batch executor with task mutation batch executor.

internal/unionstore/pipelined_memdb.go
cfzjywxk: Or could we remove this function?

txnkv/transaction/pipelined_flush.go
cfzjywxk: Agree.

internal/unionstore/pipelined_memdb.go
you06: I'll try, I find a test in TiDB will be affected need to be fixed.

txnkv/transaction/pipelined_flush.go
you06: I think the range task runner is more suitable, it's used by GC resolve lock.

https://github.com/tikv/client-go/blob/0f8c594090485696741c84a1441cdacb2d14d597/txnkv/rangetask/range_task.go#L60-L73

txnkv/txnsnapshot/snapshot.go
you06: Here we just want to avoid txn resolve locks from itself, which becomes a deadlock.
The snapshot read in TiKV already handled the `resolved_locks` field, the behavior is same as ignoring non pipelined lock.

txnkv/transaction/txn.go
you06: Changed to fatal log.

txnkv/txnsnapshot/snapshot.go
ekexium: Utilizing resolved_locks is enough for stage 1. Otherwise we'll have to modify the protocol which introduces unnecessary complexity.

internal/unionstore/pipelined_memdb.go
you06: `Reset` is removed.

txnkv/transaction/pipelined_flush.go
you06: Now commit by range task runner.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,332 @@
...
+func (p *PipelinedMemDB) Flush(force bool) (bool, error) {
+	if p.flushFunc == nil {
+		return false, errors.New("flushFunc is not provided")
+	}
+	if !force && !p.needFlush() {
+		return false, nil
+	}
+	if p.flushingMemDB != nil {
+		if err := <-p.errCh; err != nil {
+			return false, err
```
zyguan, line 182: We may need to set p.flushingMemDB to nil before return, otherwise, `Flush` (and also `FlushWait`) must NOT be called once an error returned. Think about the following case:
- let p = { flushingMemDB: nil }
- p.Flush => p.flushingMemDB = x; go flush() { p.errCh <- e };
- p.Flush => return e (since p.flushingMemDB != nil and e in p.errCh)
- p.Flush => block forever (since p.flushingMemDB != nil but nothing in p.errCh)

txnkv/txnsnapshot/snapshot.go
cfzjywxk: In tikv is it possible to decide if the lock should be read through or not by
```
if lock.type == pipelined_lock && lock.start_ts == read_ts {
   // Read the lock content if the read is union get
   // Skip the lock if the read is snapshot get
}
```
Not a big problem, maybe we could discuss later.

cfzjywxk (approved):
It has been reviewed for some time, we could merge it first and finish todo tasks in later PRs.

We could try to split the tasks into smaller PRs In future development , and it would be easier to review and add unit test cases.

txnkv/transaction/pipelined_flush.go
```diff
@@ -0,0 +1,403 @@
...
+			}
+			err = c.doActionOnMutations(bo, actionPipelinedFlush{generation: action.generation}, batch.mutations)
+			return err
+		}
+		if resp.Resp == nil {
+			return errors.WithStack(tikverr.ErrBodyMissing)
+		}
+		flushResp := resp.Resp.(*kvrpcpb.FlushResponse)
+		keyErrs := flushResp.GetErrors()
+		var locks []*txnlock.Lock
```
ekexium, line 178: We can move the decalaration after the next if, so the else branch is not needed.

ekexium (commented):
`SetPessimistic` does not return an error. Can we put an assertion in it?
The rest LGTM.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,332 @@
...
+
+func NewPipelinedMemDB(bufferBatchGetter BufferBatchGetter, flushFunc FlushFunc) *PipelinedMemDB {
+	return &PipelinedMemDB{
+		memDB:             newMemDB(),
+		errCh:             make(chan error, 1),
+		flushFunc:         flushFunc,
+		bufferBatchGetter: bufferBatchGetter,
+		generation:        0,
+		entryLimit:        math.MaxUint64,
+		bufferLimit:       math.MaxUint64,
```
zyguan, line 82: How about to set it to `memDB.bufferSizeLimit` to make sure they are consistent, although both of them are inited with `math.MaxUint64` by now.

txnkv/transaction/pipelined_flush.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	minCommitTS := c.startTS + 1
+
+	req := &kvrpcpb.FlushRequest{
+		Mutations:   mutations,
+		PrimaryKey:  c.primary(),
+		StartTs:     c.startTS,
+		MinCommitTs: minCommitTS,
+		Generation:  generation,
```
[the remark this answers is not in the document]
you06: I use ` max(defaultLockTTL, ManagedLockTTL)` as the `lock_ttl`.

txnkv/transaction/pipelined_flush.go
you06: Done

you06:
There are some goroutine leaks in integration test, not severe I think.

internal/unionstore/pipelined_memdb.go
you06: In most cases there will be no concurrent read/write, may need to remove the internal mutex operations of MemDB in pipelined usages and lock in `PipelinedMemDB.Set`.

internal/unionstore/pipelined_memdb.go
you06: Since `p.Flush()` already return error, the transaction should be aborted and the final `p.Flush()` is unreachable.

May need to store the error status and return error directly for the `Flush` calls after first error is produced.

internal/unionstore/pipelined_memdb.go
```diff
@@ -0,0 +1,332 @@
...
+	return v, nil
+}
+
+func (p *PipelinedMemDB) GetFlags(k []byte) (kv.KeyFlags, error) {
+	f, err := p.memDB.GetFlags(k)
+	if p.flushingMemDB != nil && tikverr.IsErrNotFound(err) {
+		f, err = p.flushingMemDB.GetFlags(k)
+	}
+	if err != nil {
+		return 0, err
```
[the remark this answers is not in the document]
you06: This can be replaced by assertion checks in TiKV side when handling flush request.

`GetFlags` should not be called and return an error, let's leave it as future work.

internal/unionstore/pipelined_memdb.go
you06: Done.

PR #1115, opened 2024-01-15
Fix the issue that leader change with nea leader info cause invalidStore state and lead to unnecessary backoff

### LINKED ISSUE #1112: NotLeader error may cause unnecessary backoff by causing pseudo region error
When a `NotLeader` error carrying a new leader info is received from TiKV, we update the leader info of the region by updating the struct `Region` 's `store` field which is in type `RegionStore`). As the `Region` is a type of shared data managed by `RegionCache`, the `store` is updated by replacing atomic pointers, and thus `RegionStore` is used in copy-on-write pattern.

https://github.com/tikv/client-go/blob/abda04c22acf06d9922fbd20c5dc82b1764b62d0/internal/locate/region_cache.go#L2451-L2454

Then, #264 introduces this which is executed every time attempting to send a request:

https://github.com/tikv/client-go/blob/abda04c22acf06d9922fbd20c5dc82b1764b62d0/internal/locate/region_request.go#L1023-L1026

This forces the replicaSelector to be rebuilt when retrying on `NotLeader` error, by return nil rpcCtx:

https://github.com/tikv/client-go/blob/abda04c22acf06d9922fbd20c5dc82b1764b62d0/internal/locate/region_request.go#L894-L897

and then triggering pseudo region error:

https://github.com/tikv/client-go/blob/abda04c22acf06d9922fbd20c5dc82b1764b62d0/internal/locate/region_request.go#L1424-L1435

This will cause backoff to the caller of `SendReqCtx`, which is unnecessary in a `NotLeader` case where the new leader is immediately known.

MyonKeminta:
Fixes #1112

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -45,6 +45,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unsafe"
 
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
@@ -1008,6 +1009,10 @@ func (s *replicaSelector) proxyReplica() *replica {
 	return nil
 }
 
+func sliceIdentical[T any](a, b []T) bool {
+	return len(a) == len(b) && unsafe.SliceData(a) == unsafe.SliceData(b)
+}
+
 func (s *replicaSelector) refreshRegionStore() {
 	oldRegionStore := s.regionStore
 	newRegionStore := s.region.getStore()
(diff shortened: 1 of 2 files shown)


MyonKeminta:
@crazycs520 @zyguan @cfzjywxk PTAL
@crazycs520 I directly used your test code, if you don't mind 😥

internal/locate/region_request.go
```diff
@@ -1008,6 +1009,10 @@ func (s *replicaSelector) proxyReplica() *replica {
 	return nil
 }
 
+func sliceIdentical[T any](a, b []T) bool {
```
cfzjywxk, line 1012: Better to comment about this function.

cfzjywxk (approved):

zyguan (approved):

crazycs520 (approved):

crazycs520:
@MyonKeminta Could you cherry-pick this PR to [tidb-6.5-with-kv-timeout-feature](https://github.com/tikv/client-go/tree/tidb-6.5-with-kv-timeout-feature) branch?

PR #1117, opened 2024-01-16
Check time spent on attempting RPC to avoid spending too much time on retrying

MyonKeminta:
Ref https://github.com/pingcap/tidb/issues/50432


MyonKeminta:
I'm still considering whether the values passed to these `isExhausted` function calls are proper enough. Comments are welcome.

MyonKeminta:
PTAL while I'm testing how it works in a real cluster with io injection

zyguan (approved):

cfzjywxk (approved):

PR #1119, opened 2024-01-17
Support txn file transaction

coocood:
Txn File transaction is used for transactions of big size (>= 16MiB). By building chunks on external worker and then ingesting to TiKV, txn file transactions will reduce cost and load of TiKV comparing to normal transactions.


util/merge_ranges_test.go
```diff
@@ -0,0 +1,220 @@
...
+	// "container/list"
+	// "container/list"
+	"container/list"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMergeRangesSearchOverlapped(t *testing.T) {
```
ystaticy, line 27: How about add some comments to the test case？

kv/variables.go
```diff
@@ -57,6 +60,7 @@ func NewVariables(killed *uint32) *Variables {
 		BackoffLockFast: DefBackoffLockFast,
 		BackOffWeight:   DefBackOffWeight,
 		Killed:          killed,
+		EnableTxnFile:   true,
```
ystaticy, line 63: it’s better to use [GLOBAL variables](https://docs.pingcap.com/tidb/stable/system-variables) to control whether txn file transaction is enabled or not.
It allows user to try it out on some clusters first. We have `serverless version` in TiDB which is very convenient to turn on all switches.
You can see `serverless version` in 
https://github.com/tidbcloud/tidb-cse/blob/release-7.1-keyspace/session/bootstrap_serverless.go

pingyu: Yes it's designed to be used for variables. And here is the client-go part.

The tidb-cse part will be added after [here](https://github.com/tidbcloud/tidb-cse/blob/77d402036803ea1497a7950b3688e0256627a325/sessionctx/variable/sysvar.go#L1805):

```go
	{Scope: ScopeGlobal | ScopeSession, Name: TiDBBackoffLockFast, Value: strconv.Itoa(tikvstore.DefBackoffLockFast), Type: TypeUnsigned, MinValue: 1, MaxValue: math.MaxInt32, SetSession: func(s *SessionVars, val string) error {
		s.KVVars.BackoffLockFast = tidbOptPositiveInt32(val, tikvstore.DefBackoffLockFast)
		return nil
	}},
	{Scope: ScopeGlobal | ScopeSession, Name: TiDBBackOffWeight, Value: strconv.Itoa(tikvstore.DefBackOffWeight), Type: TypeUnsigned, MinValue: 0, MaxValue: math.MaxInt32, SetSession: func(s *SessionVars, val string) error {
		s.KVVars.BackOffWeight = tidbOptPositiveInt32(val, tikvstore.DefBackOffWeight)
		return nil
	}},
```

util/merge_ranges.go
```diff
@@ -0,0 +1,132 @@
...
+	if prev == nil {
+		return m.ranges.Front(), false
+	}
+	if bytes.Compare(key, prev.Value.(*keyRange).end) <= 0 {
+		return prev, true
+	}
+	return mark, false
+}
+
+func (m *MergeRanges) Covered(start []byte, end []byte) bool {
```
ystaticy, line 87:
```suggestion
// Covered return true if m.ranges completely covers the startkey to endkey range.
func (m *MergeRanges) Covered(start []byte, end []byte) bool {
```

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,936 @@
...
+	"github.com/tikv/client-go/v2/metrics"
+	"github.com/tikv/client-go/v2/tikvrpc"
+	"github.com/tikv/client-go/v2/txnkv/txnlock"
+	"github.com/tikv/client-go/v2/util"
+	atomicutil "go.uber.org/atomic"
+	"go.uber.org/zap"
+)
+
+var (
+	// BuildTxnFileMaxBackoff is max sleep time to build TxnFile.
```
ystaticy, line 49:
```suggestion
	// BuildTxnFileMaxBackoff is max sleep Millisecond to build TxnFile.
```

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,936 @@
...
+	region    *locate.KeyLocation
+	isPrimary bool
+}
+
+func (b chunkBatch) String() string {
+	return fmt.Sprintf("chunkBatch{region: %s, isPrimary: %t, txnChunkSlice: %s}", b.region.String(), b.isPrimary, b.txnChunkSlice.String())
+}
+
+// chunkBatch.txnChunkSlice should be sorted by smallest.
+func (b *chunkBatch) getSampleKeys() [][]byte {
```
ystaticy, line 72: How about rename to `getSortedChunkSliceSmallestKeys`, which better represents what the function returns.

pingyu: The method is named as its purpose, not implementation.

And the implementation would be changed in the future, e.g, we can store extra keys in the middle of chunk and use here, if necessary.

ystaticy (approved):

PR #1122, opened 2024-01-17
improve region reload strategy

zyguan:
This PR add a new reload strategy: if a region is marked as `needExpireAfterTTL`, we don't update `lastAccess` any more, thus the region will expire after RegionCacheTTL. Currently, we set this flag:
- if a region has down peers
- if a region has a store which is stale or not reachable

This ensure that those region will be reloaded every RegionCacheTTL. So the issues like https://github.com/tikv/client-go/issues/879 and https://github.com/pingcap/tidb/issues/35418 should be resolved.

This PR also optimize region fields and replace async-reload with delayed-reload.

Also ref: https://github.com/tikv/client-go/issues/1104

**Regression test for https://github.com/tikv/client-go/issues/879**

set region-cache-ttl and max-store-down-time to 5m, see the following timeline:
- 16:08 killed tikv-1@z3, the store turned to disconnect
- 16:13 store became down, pd started down-peer scheduling
- 16:13 ~ 16:33 regions were moved to tikv-0@z3, tidb reloaded regions periodically, cross-zone traffic decreased
- 16:33 ~ 16:45 all regions were health, no region reload
- 16:45 recovered tikv-1@z3, pd started to balance regions

**Regression test for https://github.com/tikv/client-go/pull/1029**

set region-cache-ttl and max-store-down-time to 5m, see the following timeline:
- 20:30 kill tiflash-0, the store turned to disconnect
- 20:35 store became down
- 20:35 ~ 20:42 no region reload since the cached regions had no down peer
- 20:42 exec `show table regions` to reload regions manually
- 20:42 ~ 21:02 reloaded regions every 5m since regions had a down peer after `show table regions`
- 21:00 recovered tiflash-0, load didn't blance due to down peer
- 21:02 load blanced after last region reload

**Regression test for https://github.com/tikv/client-go/pull/843**


cfzjywxk:
@crazycs520 PTAL

internal/locate/region_cache.go
```diff
@@ -1123,9 +1093,18 @@ func (c *RegionCache) findRegionByKey(bo *retry.Backoffer, key []byte, isEndKey
...
+		reloadOnAccess := flags&needReloadOnAccess > 0
+		var (
+			lr  *Region
+			err error
+		)
+		if reloadOnAccess {
+			lr, err = c.loadRegion(bo, key, isEndKey)
+		} else {
+			lr, err = c.loadRegionByID(bo, r.GetID())
+		}
```
cfzjywxk, line 1107: What's the difference here using by id or by key?

internal/locate/region_cache.go
```diff
@@ -120,8 +122,10 @@ func SetRegionCacheTTLSec(t int64) {
 }
 
 const (
-	updated  int32 = iota // region is updated and no need to reload.
-	needSync              // need sync new region info.
+	needReloadOnAccess     int32 = 1 << iota // indicates the region will be reloaded on next access
+	needExpireAfterTTL                       // indicates the region will expire after RegionCacheTTL (even when it's accessed continuously)
```
cfzjywxk, line 126: The strategy is to expire these regions with certain conditions like down peers, we may need to consider how to make the region reloads smoother as the current TTL is a constant value.

internal/locate/region_cache.go
zyguan: Here is the replacement of previous async reload introduced by https://github.com/tikv/client-go/pull/843, which uses `loadRegionByID`. I just try to keep the original implementation but do not very sure about why we use `loadRegionByID` actually. Maybe it's more efficient than `loadRegionByKey`.

internal/locate/region_cache.go
zyguan: I think it's relative fine for regions that have stale or unreachable peers, since the cache GC only scan 50 regions per seconds. For regions that have down peers, maybe we can add a random value to `lastAccess` when constructing them by `newRegion`?

cfzjywxk: Maybe we could do some improvements about smoothing region reloads in another PR, considering both expire reloading and region cache miss reloading. I rembember we've encountered lots of region reload caused by sudden TTL expire or something like it.

cfzjywxk (approved):
LGTM

internal/locate/region_cache.go
```diff
@@ -1123,9 +1093,18 @@ func (c *RegionCache) findRegionByKey(bo *retry.Backoffer, key []byte, isEndKey
 			c.insertRegionToCache(r, true, true)
 			c.mu.Unlock()
 		}
-	} else if r.checkNeedReloadAndMarkUpdated() {
+	} else if flags := r.resetSyncFlags(needReloadOnAccess | needAsyncReloadReady); flags > 0 {
```
you06, line 1096: Seems the async reload becomes sync here.

internal/locate/region_cache.go
```diff
@@ -2198,6 +2136,30 @@ func (c *RegionCache) cacheGC() {
...
+
+			// Check remaining regions and update sync flags
+			for _, region := range remaining {
+				syncFlags := region.getSyncFlags()
+				if syncFlags&needAsyncReloadReady > 0 {
+					// the region will be reload soon on access
+					continue
+				}
+				if syncFlags&needAsyncReloadPending > 0 {
+					region.setSyncFlags(needAsyncReloadReady)
```
you06, line 2148: May I understand the flag change here is to avoid too frequent meaningless reloading?

you06 (commented):
rest LGTM

internal/locate/region_cache.go
zyguan: Yes, it tries to limit the rate of reloading here.

internal/locate/region_cache.go
zyguan: Agreed. One idea I'm thinking about is: we may change `lastAccess` to `TTL` and push it as `now + RegionCacheTTL + RandDev` on access.

internal/locate/region_cache.go
you06: Rename or comment the flag, both OK to me.

The current comment indicates it will be reloaded in async mode, which is inaccurate.

```go
	needAsyncReloadPending                   // indicates the region will be reloaded in async mode
	needAsyncReloadReady                     // indicates the region is ready to be reloaded in async mode
```

zyguan: Renamed, PTAL.

you06 (approved):

cfzjywxk:
@zyguan 
An PR to update the client-go dependency of tidb is also needed.

zyguan:
> @zyguan An PR to update the client-go dependency of tidb is also needed.

I'll submit another PR soon, which tries to make reloading caused by TTL smooth (https://github.com/tikv/client-go/pull/1122#discussion_r1469009792) and adds some metrics to observe the reason of reloading.

PR #1124, opened 2024-01-17
ErrQueryInterrupted with parameters

ekexium:
Ref https://github.com/pingcap/tidb/issues/49643

We have to return the error with the signal to enable TiDB to convert it to the actual error type.


zyguan (approved):

error/error.go
```diff
@@ -96,11 +94,19 @@ var (
...
 	// ErrUnknown is the unknow error.
-	ErrUnknown = errors.New("unknow")
+	ErrUnknown = errors.New("unknown")
 	// ErrResultUndetermined is the error when execution result is unknown.
 	ErrResultUndetermined = errors.New("execution result undetermined")
 )
 
+type ErrQueryInterrupted struct {
+	Signal uint32
+}
```
MyonKeminta, line 104: Is the `Signal` one of some limited values? Can we consider adding some different error values for them? Or, can we add a new type `ErrQueryInterruptedWithSignal` for representing interrupted error with signal code, while keeping the previous `ErrQueryInterrupted` error value to avoid the compatibility problem?

ekexium: > Is the Signal one of some limited values? Can we consider adding some different error values for them?

It's not limited by client-go. TiDB can sends whatever it likes..

> Or, can we add a new type ErrQueryInterruptedWithSignal for representing interrupted error with signal code, while keeping the previous ErrQueryInterrupted error value to avoid the compatibility problem?

Yeah... Painful

cfzjywxk (approved):

PR #1127, opened 2024-01-22
fix the issue that health check may set liveness wrongly

### LINKED ISSUE #1111: An unready store might be marked as reachable
The health check loop will do reResolve periodically. If the store meta changed, a new store will be inserted and the old one will be marked as deleted. Both old store and new store are marked as reachable at the end of the health check loop. As a result, tidb may access this unready store and requests finally timeout.

zyguan:
close #1111


zyguan:
@cfzjywxk PTAL

internal/locate/region_cache.go
```diff
@@ -2618,6 +2618,8 @@ func (s *Store) reResolve(c *RegionCache) (bool, error) {
 	addr = store.GetAddress()
 	if s.addr != addr || !s.IsSameLabels(store.GetLabels()) {
 		newStore := &Store{storeID: s.storeID, addr: addr, peerAddr: store.GetPeerAddress(), saddr: store.GetStatusAddress(), storeType: storeType, labels: store.GetLabels(), state: uint64(resolved)}
+		newStore.livenessState = atomic.LoadUint32(&s.livenessState)
+		newStore.unreachableSince = s.unreachableSince
```
cfzjywxk, line 2622: How about abstracting a `newStore` function to force specify the `livenessState` and `unreachableSince` as the input parameters.

internal/locate/region_cache.go
```diff
@@ -2783,50 +2785,85 @@ func (s *Store) requestLivenessAndStartHealthCheckLoopIfNeeded(bo *retry.Backoff
 	// It may be already started by another thread.
 	if atomic.CompareAndSwapUint32(&s.livenessState, uint32(reachable), uint32(liveness)) {
 		s.unreachableSince = time.Now()
-		go s.checkUntilHealth(c)
+		reResolveInterval := 30 * time.Second
```
cfzjywxk, line 2788: Better to make it a constant value.

internal/locate/region_cache.go
```diff
@@ -2783,50 +2785,85 @@ func (s *Store) requestLivenessAndStartHealthCheckLoopIfNeeded(bo *retry.Backoff
...
 				lastCheckPDTime = time.Now()
 
 				valid, err := s.reResolve(c)
 				if err != nil {
 					logutil.BgLogger().Warn("[health check] failed to re-resolve unhealthy store", zap.Error(err))
 				} else if !valid {
-					logutil.BgLogger().Info("[health check] store meta deleted, stop checking", zap.Uint64("storeID", s.storeID), zap.String("addr", s.addr))
+					if s.getResolveState() == deleted {
+						// if the store is deleted, a new store with same id must be inserted (guaranteed by reResolve).
+						newStore, _ := c.getStore(s.storeID)
```
cfzjywxk, line 2826: Is it necessary to add an assertion `newStore != nil` here?

internal/locate/region_cache_test.go
```diff
@@ -1793,6 +1793,49 @@ func (s *testRegionCacheSuite) TestSlowScoreStat() {
...
+		}
+		return reachable
+	})
+
+	// start health check loop
+	atomic.StoreUint32(&store1.livenessState, store1Liveness)
+	go store1.checkUntilHealth(s.cache, livenessState(store1Liveness), 2*time.Second)
+
+	// update store meta after 1s
+	time.Sleep(time.Second)
```
cfzjywxk, line 1818: Could we use a shorter check interval to reduce sleep wait here?

internal/locate/region_cache.go
zyguan: The rawkv tests use failpoint to set it shorter.

internal/locate/region_cache.go
zyguan: It's guaranteed by `reResolve`, `changeToActiveStore` also rely on it.

internal/locate/region_cache.go
```diff
@@ -2783,50 +2785,85 @@ func (s *Store) requestLivenessAndStartHealthCheckLoopIfNeeded(bo *retry.Backoff
...
-			if l == reachable {
+			liveness = s.requestLiveness(c.ctx, c)
+			if liveness == reachable {
 				logutil.BgLogger().Info("[health check] store became reachable", zap.Uint64("storeID", s.storeID))
-
+				atomic.StoreUint32(&s.livenessState, uint32(reachable))
 				return
 			}
-			atomic.StoreUint32(&s.livenessState, uint32(l))
+			atomic.StoreUint32(&s.livenessState, uint32(liveness))
```
crazycs520, line 2845: how about move this after line#2839 and remove line 2842?

internal/locate/region_cache.go
crazycs520: how about 

```go
var DefReResolveInterval = 30 * time.Second // global scope.

func (s *Store) checkUntilHealth(c *RegionCache, liveness livenessState){
  reResolveInterval = DefReResolveInterval
  ...
}
```

and if test needs a shorter reResolveInterval, change `DefReResolveInterval` directly, no need to use the following failpoint.

zyguan: > how about
> 
> ```go
> var DefReResolveInterval = 30 * time.Second // global scope.
> 
> func (s *Store) checkUntilHealth(c *RegionCache, liveness livenessState){
>   reResolveInterval = DefReResolveInterval
>   ...
> }
> ```
> 
> and if test needs a shorter reResolveInterval, change `DefReResolveInterval` directly, no need to use the following failpoint.

Personally I do not prefer that. The global var is required to be handled carefully to avoid data race in ut. I've struggled with `SetRegionCacheTTLSec` in https://github.com/tikv/client-go/pull/1122 for about an hour.

internal/locate/region_cache.go
zyguan: How about to handle it later (as a part of https://github.com/tikv/client-go/issues/1104). There are 4 `&Store{...}` in region_cache.go now, I'd like to handle them togather according to the store lifecycle. Let this PR fix the corresponding issue by now.

internal/locate/region_cache_test.go
zyguan: Yes, it's actually 2s reResolveInterval makes the test slow.

cfzjywxk (approved):

crazycs520 (approved):

zyguan:
> @zyguan @cfzjywxk could we please cherry-pick it into release-7.5?

Sure, https://github.com/tikv/client-go/pull/1324 is merged.

PR #988, opened 2023-09-21
add replica selector init failed reason 

bufferflies:
before:
`
[2023/09/21 14:35:50.702 +08:00] [INFO] [region_request.go:1522] ["throwing pseudo region error due to no replica available"] [conn=1887493856] [session_alias=] [req-ts=444406212641423730] [req-type=Cop] [region="{ region id: 186911575, ver: 832, confVer: 407 }"] [replica-read-type=leader] [stale-read=false] [request-sender="{rpcError:<nil>,replicaSelector: replicaSelector{selectorStateStr: nil, cacheRegionIsValid: unknown, replicaStatus: []}}"] [retry-times=0] [total-backoff-ms=0] [total-backoff-times=0] [max-exec-timeout-ms=60000] [total-region-errors=]
`
after
`
[2023/09/21 19:07:34.787 +08:00] [INFO] [region_request.go:1533] ["throwing pseudo region error due to no replica available"] [conn=3749708330] [session_alias=] [req-ts=444411269215682567] [req-type=Cop] [region="{ region id: 8580178, ver: 192, confVer: 401 }"] [replica-read-type=leader] [stale-read=false] [request-sender="{rpcError:cache region is invalid,need synced:false,ttl expired:false,replicaSelector: replicaSelector{selectorStateStr: nil, cacheRegionIsValid: unknown, replicaStatus: []}}"] [retry-times=0] [total-backoff-ms=0] [total-backoff-times=0] [max-exec-timeout-ms=60000] [total-region-errors=]
`

PR #1142, opened 2024-01-26
replica selector refactor

### LINKED ISSUE #1144: replica selector refactor
# Background

Currently, the logic of `replicaSelector` is too complicated, and the code maintenance is a heavy burden. I guess no one can explain the common replica selection logic without looking at the code.

I think that the reason why replicaSelector is so complicated is because of the introduction of a state machine, which has many states and the transitions between states are also messy. The following is the state machine transition during the `tidb_replica_read = leader` strategy compiled by @zyguan, this is great, but still complicated, and this doesn't consider the situation when `enable-forwarding` is true, otherwise, it will be more complicated.

image

# Task

This task is trying to refactor `replicaSelector`, and has the following targets:

- Test replica selection strategies for various region error scenarios, and fix unreasonable behavior.
- Less code, clearer logic, easier to understand.
- Keep the `replicaSelector` logic/behavior the same as before.

Considering failure regression (since no one can guarantee that the new code is completely bug-free), the implementation is to introduce a new `replicaSelectorV2`, through the configuration file `enable-replica-selector-v2 = true` to use `replicaSelectorV2` by default, set `enable- replica-selector-v2 = false` to fall back to using the older version of `replicaSelector`.  After `replicaSelectorV2` has gone through 2 major versions stably, we can start to consider deleting the old version of `replicaSelector`

crazycs520:
close #1144

## What changed and how does it work?

Since the old version `replicaSelector` is hard to test and understand, This PR introduces a new replica selector implementation `replicaSelectorV2`, which is less code and easer to understand.

For rollback, I don't remove the old version `replicaSelector`, and add a new config `EnableReplicaSelectorV2` whose default value is true, which means use `replicaSelectorV2` by default, set it's value to false will use old version `replicaSelector`. After `replicaSelectorV2` is stable, maybe we can consider remove old version `replicaSelector` and config `EnableReplicaSelectorV2`.

# How to test?

- All the old tests are pass.
- I add many test case in `replica_selector_test.go` to make sure `EnableReplicaSelectorV2` has the same behavior with   old version `replicaSelector`.
  - TestReplicaReadAccessPathByCase
  - TestReplicaReadAccessPathByCase2
  - TestReplicaReadAccessPathByBasicCase
  - TestReplicaReadAccessPathByLeaderCase
  - TestReplicaReadAccessPathByFollowerCase
  - TestReplicaReadAccessPathByMixedAndPreferLeaderCase
  - TestReplicaReadAccessPathByStaleReadCase
  - TestReplicaReadAccessPathByTryIdleReplicaCase
  - TestReplicaReadAccessPathByFlashbackInProgressCase
  - TestReplicaReadAccessPathByProxyCase
  - TestReplicaReadAccessPathByLearnerCase
- I write a `TestReplicaReadAccessPathByGenError` test to generate various combinations of region errors to verify that `EnableReplicaSelectorV2` has the same behavior with   old version `replicaSelector`. From the following log, you can see this test generated a total 2200723 cases.

   ```log
   [2024/03/05 10:27:52.562 +08:00] [INFO] [replica_selector_test.go:2511] ["TestReplicaReadAccessPathByGenError Finished"] [total-case=2200723] [valid-case=857630] [invalid-case=1343093]
   ```
- The test coverage of `replicaSelectorV2` is more than 98%.
 
image

# Advice when reviewing PR

If you have any questions about the logic about replicaSelector, you can try to modify the logic code, and then run the test to see whether the test results are as expected.

```sh
cd internal/locate
go test -run=TestReplicaReadAccessPathBy.*Case   # run some test cases, this is fast and less than 1s on my m1 mbp.
go test > test.log   # run all pkg test, this may cost 230s on my m1 mbp.
```


zyguan (commented):
Overall LGTM. We may need to do some basic benchmarks and stablility tests to make sure there is no obvious regression.

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,486 @@
...
+		return false
+	}
+	return true
+}
+
+func (s *replicaSelectorV2) onNotLeader(
+	bo *retry.Backoffer, ctx *RPCContext, notLeader *errorpb.NotLeader,
+) (shouldRetry bool, err error) {
+	if s.target != nil {
+		s.target.notLeader = true
```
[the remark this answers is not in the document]
crazycs520: Yes. If not set `target.notLeader` to true, then will retry this replica, and this behavior is not compatible with before, since it should retry the remaining replica.

internal/locate/region_cache.go
```diff
@@ -2646,6 +2646,13 @@ func (s *StoreHealthStatus) markAlreadySlow() {
 	s.updateSlowFlag()
 }
 
+// resetSlowScore resets the slow score to 0. It's used for test.
+func (s *StoreHealthStatus) resetSlowScore() {
+	s.clientSideSlowScore.resetSlowScore()
+	s.updateTiKVServerSideSlowScore(0, time.Now())
```
MyonKeminta, line 2652: Can this be done in any other way? I think it might break the consistency of their internal states.
By the way, the minimal valid value (for the slow scores in both client side and tikv side) should be 1 instead of 0.

MyonKeminta: Maybe you can consider reset the whole `healthStatus` field in `Store` struct, if it won't cause races. Another choice is to fake `tick` calls repeatedly until the two scores decay to 1.

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,486 @@
...
+	proxy    *replica
+	attempts int
+}
+
+func newReplicaSelectorV2(
+	regionCache *RegionCache, regionID RegionVerID, req *tikvrpc.Request, opts ...StoreSelectorOption,
+) (*replicaSelectorV2, error) {
+	cachedRegion := regionCache.GetCachedRegionWithRLock(regionID)
+	if cachedRegion == nil || !cachedRegion.isValid() {
+		return nil, errors.New("cached region invalid")
```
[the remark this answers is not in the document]
crazycs520: This logic is introduce by https://github.com/tikv/client-go/pull/988/files#diff-0c36d07fbd1fc71c850ff57d323fe12bdf7f2c053ef71b7fec5bb44358f1070dR913 , I just try to compatible with this behavior.

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,554 @@
...
+
+func (s *ReplicaSelectMixedStrategy) calculateScore(r *replica, isLeader bool) int {
+	score := 0
+	if isLeader {
+		if s.preferLeader {
+			score += scoreOfPreferLeader
+		} else if s.tryLeader {
+			if len(s.labels) > 0 {
+				// when has match labels, prefer leader than not-matched peers.
+				score += scoreOfPreferLeader
```
MyonKeminta, line 342: * Should a not-slow normal follower (`NotSlow` + `NormalPeer` -> 2) have the same priority as a slow leader in prefer-leader mode (`PreferLeader` -> 2) ?

* Is there an assumption that in cases that leader must be expected (such as the first attempt in prefer-leader mode and the second attempt in stale read mode) must be done with the `ReplicaSelectLeaderStrategy` instead of using the `calculateScore`? If so, I think there should be some comments to explain that this function doesn't handle those cases.

* If `LabelMatch` always has higher priority than `PreferLeader`, and `PreferLeader` always has higher priority than `NormalPeer` and `NotSlow`, is it possible that we use different bits for these factors, so that the more significant bit a factor is at, the more dominant status the factor is? For factors that has the same priority, we reserve more bits for them. For example, we can have such definition:

    ```go
    const (
    	flagLabelMatches = 1 << 3
    	flagPreferLeader = 1 << 2
    	flagNormalPeer = 1
    	flagNotSlow = 1
    	
    	// So that we have definition of the final score:
    	// MSB                                                                               LSB
    	// [unused bits][1 bit: LabelMatches][1 bit: PreferLeader][2 bits: NormalPeer + NotSlow]
    )
    ```

    I'm not sure if this is possible to give the same result as your current code, but I think in this way, we can have less cases like:

   * Factor A < Factor C
   * Factor B < Factor C
   * Factor A + Factor B > Factor C

    which I'm afraid may make the priority calculation less comprehensible. Another benefit is that we can mostly know how a score comes from by reading its bits, which may help in diagnosing.

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,554 @@
...
+			return
+		}
+	}
+	preferLeader := req.ReplicaReadType == kv.ReplicaReadPreferLeader
+	if s.attempts > 1 {
+		if req.ReplicaReadType == kv.ReplicaReadMixed {
+			// For mixed read retry, prefer retry leader first.
+			preferLeader = true
+		}
+	}
```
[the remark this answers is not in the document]
crazycs520: No need to do this, since normally, in `ReplicaSelectMixedStrategy.isCandidate`, the replica `maxAttempt` is 1, so if the leader is attempted, then it won't be attempted again.

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,554 @@
...
+	if reloadRegion {
+		selector.region.setSyncFlags(needDelayedReloadPending)
+	}
+	if len(maxScoreIdxes) == 1 {
+		idx := maxScoreIdxes[0]
+		return replicas[idx]
+	} else if len(maxScoreIdxes) > 1 {
+		// if there are more than one replica with the same max score, we will randomly select one
+		// todo: consider use store slow score to select a faster one.
```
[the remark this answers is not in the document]
crazycs520: ok, I will change the comment.

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,554 @@
...
+	}
+	if s.busyThreshold > 0 {
+		// when can't find an idle replica, no need to invalidate region.
+		return nil
+	}
+
+	metrics.TiKVReplicaSelectorFailureCounter.WithLabelValues("exhausted").Inc()
+	for _, r := range replicas {
+		if r.deadlineErrUsingConfTimeout {
+			// when meet deadline exceeded error, do fast retry without invalidate region cache.
```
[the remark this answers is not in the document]
crazycs520: No. The logic is if have meet `deadlineErrUsingConfTimeout` error in this round, don't invalidate region to let up layer to remove kv_read_timeout then do fast retry.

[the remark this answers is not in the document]
crazycs520: This logic is same with the old version.

internal/locate/replica_selector.go
crazycs520: Great catch, done as your advice.

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,554 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NOTE: The code in this file is based on code from the
+// TiDB project, licensed under the Apache License v 2.0
+//
+// https://github.com/pingcap/tidb/tree/cc5e161ac06827589c4966674597c137cc9e809c/store/tikv/locate/region_cache.go
+//
+
+// Copyright 2016 PingCAP, Inc.
```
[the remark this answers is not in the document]
crazycs520: great, done.

internal/locate/region_request.go
```diff
@@ -633,7 +645,7 @@ func (state *accessByKnownProxy) onSendFailure(bo *retry.Backoffer, selector *re
 }
 
 func (state *accessByKnownProxy) onNoLeader(selector *replicaSelector) {
-	selector.state = &invalidLeader{}
+	selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx, fromAccessKnownLeader: true}
```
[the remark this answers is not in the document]
crazycs520: This is for https://github.com/tikv/client-go/issues/1207

Since I need to test `replicaSelectorV2 ` has some beha behavior with old `replicaSelector`, so I just change the old `replicaSelector` behavior here. If I don't change this, I'll have to modify the `replicaSelectorV2` code to accommodate this strange behavior.

internal/locate/region_request3_test.go
```diff
@@ -235,16 +235,17 @@ func (s *testRegionRequestToThreeStoresSuite) loadAndGetLeaderStore() (*Store, s
 }
 
 func (s *testRegionRequestToThreeStoresSuite) TestForwarding() {
-	s.regionRequestSender.regionCache.enableForwarding = true
+	sender := NewRegionRequestSender(s.cache, s.regionRequestSender.client)
+	sender.regionCache.enableForwarding = true
```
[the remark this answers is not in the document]
MyonKeminta: 🤔 I assume that the fields initialized in `SetupTest` should be done for each single test

zyguan (approved):

MyonKeminta (commented):
Rest LGTM

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,535 @@
...
+	flagNormalPeer   = 1
+	flagNotSlow      = 1
+)
+
+// calculateScore calculates the score of the replica.
+func (s *ReplicaSelectMixedStrategy) calculateScore(r *replica, isLeader bool) int {
+	score := 0
+	if isLeader {
+		if s.preferLeader {
+			score |= flagPreferLeader
```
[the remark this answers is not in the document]
MyonKeminta: Sorry I misunderstood it... nevermind

MyonKeminta (approved):
Rest LGTM

internal/locate/replica_selector.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+	if r.store.IsStoreMatch(s.stores) && r.store.IsLabelsMatch(s.labels) {
+		score |= flagLabelMatches
+	}
+	if !r.store.healthStatus.IsSlow() {
+		score += flagNotSlow
+	}
+	if score > 0 && r.attempts > 0 {
+		score = score - 1
+	}
```
MyonKeminta, line 345: I just noticed that how about making this rule a new bit (at the lowest position, and shift all other bits right)?
So that we can still take the benifit that a score value's bits can mostly show how it's calculated.

Btw consider adding some comment at line 341 to state that flagNotSlow and and flagNormalPeer is at the same position and 2 bits are reserved for them so that they can be added together.

crazycs520: Nice catch, I use following score flag now:

```go
	// The definition of the score is:
	// MSB                                                                               LSB
	// [unused bits][1 bit: LabelMatches][1 bit: PreferLeader][2 bits: NormalPeer + NotSlow]
	flagLabelMatches = 1 << 4
	flagPreferLeader = 1 << 3
	flagNormalPeer   = 1 << 2
	flagNotSlow      = 1 << 1
	flagNotAttempt   = 1
```

crazycs520: In fact, this is rare, usually each replica will only try 1 time, unless the replica has previously returned a dataNotReady error, and then there is a chance to try the replica again.

internal/locate/replica_selector_test.go
```diff
@@ -459,175 +674,1845 @@ func TestReplicaReadAccessPathByCase(t *testing.T) {
...
+	defer s.TearDownTest()
+
+	fakeEpochNotMatch := &errorpb.Error{EpochNotMatch: &errorpb.EpochNotMatch{}}
+	var ca replicaSelectorAccessPathCase
+	// Try idle replica strategy not support write request.
+	ca = replicaSelectorAccessPathCase{
+		reqType:         tikvrpc.CmdPrewrite,
+		readType:        kv.ReplicaReadLeader,
+		busyThresholdMs: 10,
+		accessErr:       []RegionErrorType{ServerIsBusyWithEstimatedWaitMsErr, ServerIsBusyWithEstimatedWaitMsErr},
```
you06, line 2023: Ditto, `CmdPrewrite` should not get `ServerIsBusyWithEstimatedWaitMsErr`, maybe change to `ServerIsBusyErr`?

internal/locate/replica_selector_test.go
```diff
@@ -459,175 +674,1845 @@ func TestReplicaReadAccessPathByCase(t *testing.T) {
...
+	}
+	s.True(s.runCaseAndCompare(ca))
+
+	ca = replicaSelectorAccessPathCase{
+		reqType:   tikvrpc.CmdPrewrite,
+		readType:  kv.ReplicaReadLeader,
+		staleRead: false,
+		timeout:   0,
+		label:     nil,
+		accessErr: []RegionErrorType{DeadLineExceededErr, NotLeaderErr, NotLeaderWithNewLeader2Err, ServerIsBusyWithEstimatedWaitMsErr, DeadLineExceededErr},
```
you06, line 818:
```suggestion
		accessErr: []RegionErrorType{DeadLineExceededErr, NotLeaderErr, NotLeaderWithNewLeader2Err, ServerIsBusyErr, DeadLineExceededErr},
```

I think `ServerIsBusyWithEstimatedWaitMsErr` nevers returns for write request.
But it's ok that always retry leader for `CmdPrewrite`.
Maybe change to `ServerIsBusyErr`?

you06 (commented):
rest LGTM

internal/locate/replica_selector_test.go
crazycs520: Great, done.

internal/locate/replica_selector_test.go
crazycs520: Great, done.

you06 (approved):

cfzjywxk (approved):

PR #1148, opened 2024-01-31
introduce a random jitter to region cache ttl

zyguan:
Ref https://github.com/tikv/client-go/pull/1122#discussion_r1469009792 and #1104 .

This PR tries to address https://github.com/tikv/client-go/pull/1122#discussion_r1469009792 by introducing a random jitter to region cache TTL.

To observe the reason of region reloading, the methods for searching cached regions are refactored.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1098,8 +1098,9 @@ func (c *RegionCache) LocateEndKey(bo *retry.Backoffer, key []byte) (*KeyLocatio
 }
 
 func (c *RegionCache) findRegionByKey(bo *retry.Backoffer, key []byte, isEndKey bool) (r *Region, err error) {
-	r = c.searchCachedRegion(key, isEndKey)
-	if r == nil {
+	var expired bool
+	r, expired = c.searchCachedRegionByKey(key, isEndKey)
+	if r == nil || expired {
 		// load region when it is not exists or expired.
 		lr, err := c.loadRegion(bo, key, isEndKey, pd.WithAllowFollowerHandle())
 		if err != nil {
@@ -1152,8 +1153,9 @@ func (c *RegionCache) findRegionByKey(bo *retry.Backoffer, key []byte, isEndKey 
 }
 
 func (c *RegionCache) tryFindRegionByKey(key []byte, isEndKey bool) (r *Region) {
-	r = c.searchCachedRegion(key, isEndKey)
-	if r == nil || r.checkSyncFlags(needReloadOnAccess) {
+	var expired bool
+	r, expired = c.searchCachedRegionByKey(key, isEndKey)
+	if r == nil || expired || r.checkSyncFlags(needReloadOnAccess) {
 		return nil
 	}
 	return r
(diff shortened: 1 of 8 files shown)


cfzjywxk:
@crazycs520 PTAL

internal/locate/region_cache.go
```diff
@@ -121,6 +122,29 @@ func SetRegionCacheTTLSec(t int64) {
...
+	regionCacheTTLJitterSec = jitter
+}
+
+// nextTTL returns a random TTL in range [ts+base, ts+base+jitter).
+func nextTTL(ts int64) int64 {
+	jitter := int64(0)
+	if regionCacheTTLJitterSec > 0 {
+		jitter = rand.Int63n(regionCacheTTLJitterSec)
+	}
+	return ts + regionCacheTTLSec + jitter
```
cfzjywxk, line 140: Is there a way to add some constraint checks to make sure the unit of the input `ts` is `second`, to avoid being used mistakenly?

zyguan: We pass `ts` here to avoid overhead of `time.Now()`. To validate it, maybe we can:
- record process start ts by a global var (eg. processStartTs)
- the ts might be invalid if ts - processStartTs > 100 years or ts < processStartTs
- if the ts is invalid, we can fallback to call `time.Now()`

But do we really need to do that? There are many mechanisms to invalidate a stale region, it might be not a big problem even when we call `nextTTL` with epoch millis, I guess.

Or, we can just change the function signature (also many releated functions like `checkRegionCacheTTL`), let it (them) accept an arg of type `time.Time`.

cfzjywxk: Maybe we could just add some comments to explain the constraints of the input `ts` in the comment if it's not a big problem when used mistakenly.

cfzjywxk (approved):
Rest LGTM

internal/locate/region_cache.go
```diff
@@ -366,28 +408,36 @@ func (r *Region) checkRegionCacheTTL(ts int64) bool {
...
-		if r.checkSyncFlags(needExpireAfterTTL) || atomic.CompareAndSwapInt64(&r.lastAccess, lastAccess, ts) {
+		// skip updating TTL when:
+		// 1. the region has been marked as `needExpireAfterTTL`
+		// 2. the TTL is far away from ts (still within jitter time)
+		if r.checkSyncFlags(needExpireAfterTTL) || ttl > ts+regionCacheTTLSec {
+			return true
+		}
+		if newTTL == 0 {
+			newTTL = nextTTL(ts)
+		}
```
you06, line 425:
```suggestion
		} else if newTTL <= ttl {
			return true
		}
```

Because `newTTL` is only calculated once
If the ttl is updated by another thread, it might be `newTLL < ttl`, we can skip updating ddl in this case.

```
// thread1
ttl1 := atomic.LoadInt64(&r.ttl)
newTTL1 = nextTTL(ts1)

// thread2
ttl2 := atomic.LoadInt64(&r.ttl)
newTTL2 = nextTTL(ts2) // newTTL2 > newTTL1
atomic.CompareAndSwapInt64(&r.ttl, ttl2, newTTL2) == true

// thread1
ttl1 := atomic.LoadInt64(&r.ttl) // second loop, ttl1 == newTTL2
atomic.CompareAndSwapInt64(&r.ttl, ttl1, newTTL1) // the ttl jumps back
```

you06 (commented):
rest LGTM

internal/locate/region_cache.go
zyguan: I think the constraint at L426 always holds. If ttl1(=newTTL2) > ts+regionCacheTTLSec, the function returns at L421, if not ttl1(=newTTL2) <= ts+regionCacheTTLSec <= ts+regionCacheTTLSec+randomJitter = newTTL1.

you06: Indeed.

you06 (approved):

PR #1149, opened 2024-01-31
txn: enable pessimistic region rollback

cfzjywxk:
Ref: https://github.com/tikv/tikv/issues/16158

Use region pessimistic rollback to speed up expired pessimistic lock cleanup.


zyguan (approved):

internal/mockstore/mocktikv/mvcc_leveldb.go
```diff
@@ -736,6 +736,32 @@ func (mvcc *MVCCLevelDB) PessimisticRollback(keys [][]byte, startTS, forUpdateTS
 	mvcc.mu.Lock()
 	defer mvcc.mu.Unlock()
 
+	// Scan the whole region for corresponding pessimistic locks.
+	if len(keys) == 0 {
```
MyonKeminta, line 740: Is there any related test using mocktikv?

MyonKeminta: To scan the current region correctly in mock tikv, it seems we need to get the region's `startKey` and `endKey` from the `kvHandler` (in method `handleKvPessimisticRollback`), and pass the `startKey` and `endKey` to `PessimisticRollback`.

MyonKeminta (commented):
Rest LGTM

internal/mockstore/mocktikv/mvcc_leveldb.go
cfzjywxk: Yes there are still cases using mock tikv, the key range is added, PTAL

MyonKeminta: It's ok though, I think passing `startKey` and `endKey` instead of the `session` is better, since it's consistent with other APIs such as `ResolveLock`.

MyonKeminta (approved):

internal/mockstore/mocktikv/mvcc_leveldb.go
cfzjywxk: Ok I've changed the input parameters to use start and end keys.

PR #1150, opened 2024-02-01
[DNM] Large dml tests

you06:
I did some research of pipelined DML(I call it large DML in the past) with this PR.

This is a tool that can invoke the TiDB executor to generate transaction mutations. I used it to estimate the size of transactions, the commit speed, and test the performance of the pipelined transaction (pipelined txn) strategy.

You can test other schemas with this tool.

```
cd examples/txnkv/large_commit && go build
```

Main conclusions from the experiments:

* With sysbench schema, inserting 10,000,000 rows, the size of memdb is about 2.34GB
* RES memory can be 2-3 times of the size of memdb, 6.64GB for sysbench 10M rows in this test
* In most cases, the wall time cost by prewrite phase is less than the execution phase, because execution can only utilize single core and easily be the bottleneck
  * The bottleneck becomes more severe in wide tables
  * Parallel execution is a optimization change in the future
* Even we only store keys in the memory, the memory can be unaffortable in schemas with many indexes
* With single core execution, the overall acceleration can reach **35%**


you06:
Close this PR because it's not intend to be merged, just for showcasing experimental results.

PR #1153, opened 2024-02-08
Support receving health feedback

MyonKeminta:
Ref: https://github.com/tikv/tikv/issues/16297


internal/locate/region_cache.go
```diff
@@ -3066,28 +3213,33 @@ func (c *RegionCache) asyncUpdateStoreSlowScore(interval time.Duration) {
...
+	healthDetails := make(map[uint64]HealthStatusDetail)
 	c.forEachStore(func(store *Store) {
-		store.updateSlowScoreStat()
-		slowScoreMetrics[store.storeID] = float64(store.getSlowScore())
+		store.healthStatus.update()
+		healthDetails[store.storeID] = store.healthStatus.GetHealthStatusDetail()
 	})
-	for store, score := range slowScoreMetrics {
-		metrics.TiKVStoreSlowScoreGauge.WithLabelValues(strconv.FormatUint(store, 10)).Set(score)
+	logutil.BgLogger().Info("checkAndUpdateStoreHealthStats: get health details", zap.Reflect("details", healthDetails))
```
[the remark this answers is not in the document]
MyonKeminta: Sorry, I added this for debugging.

internal/client/client.go
```diff
@@ -108,6 +108,13 @@ type Client interface {
 	CloseAddr(addr string) error
 	// SendRequest sends Request.
 	SendRequest(ctx context.Context, addr string, req *tikvrpc.Request, timeout time.Duration) (*tikvrpc.Response, error)
+	// SetEventListener registers an event listener for the Client instance. If called more than once, the previously
+	// set one will be replaced.
+	SetEventListener(listener ClientEventListener)
```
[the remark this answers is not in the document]
MyonKeminta: Yes, but I'm afraid that the assumption might be broken easily by mistake in the future...

cfzjywxk:
@crazycs520 PTAL

MyonKeminta:
As I noticed that the interface compatibility problem of unistore can be workarounded by `unistoreClientWrapper`, It's actually not necessary to split into multiple PRs. cc @zyguan

zyguan (approved):

internal/locate/region_cache.go
```diff
@@ -3164,6 +3339,10 @@ func (s *Store) recordReplicaFlowsStats(destType replicaFlowsType) {
 	atomic.AddUint64(&s.replicaFlowsStats[destType], 1)
 }
 
+func (s *Store) recordHealthFeedback(feedback *tikvpb.HealthFeedback) {
+	s.healthStatus.updateTiKVServerSideSlowScore(int64(feedback.GetSlowScore()), time.Now())
```
crazycs520, line 3343: feedback.FeedbackSeqNo is never used now?

crazycs520 (approved):
REST LGTM

internal/locate/region_cache.go
MyonKeminta: Yes, it's not checked for now. Maybe I'd better add a comment to note it.

internal/locate/region_cache.go
```diff
@@ -2584,6 +2578,169 @@ func (r *Region) ContainsByEnd(key []byte) bool {
...
+	defer s.tikvSideSlowScore.Unlock()
+
+	// Reload update time as it might be updated concurrently before acquiring mutex
+	lastUpdateTime = s.tikvSideSlowScore.lastUpdateTime.Load()
+	elapsed := now.Sub(*lastUpdateTime)
+	if elapsed < tikvSlowScoreUpdateFromPDInterval {
+		return
+	}
+
+	// TODO: Try to get store status from PD here. But it's not mandatory.
```
cfzjywxk, line 2675: The function would be executed inside the `batchRecvLoop`, would it affect the performance if we call `pdClient.func` here?

MyonKeminta: No. This function is not expected to be run in batchRecvLoop, but should be a periodic task executed in the background. 
 The function that might be called inside batchRecvLoop is `updateTiKVServerSideSlowScore`.

Here what we need to be careful about is the mutex.

cfzjywxk: Got it, it's better to add comments to functions that would be used in the critical performance path.

PR #1155, opened 2024-02-09
Configurable low res tso update interval (#1154)

### LINKED ISSUE #1154: Make low resolution timestamp update interval configurable
Currently it's a constant at 2s. This is a counterpart issue of https://github.com/pingcap/tidb/issues/51081.

ari-e:
Closes #1154 and is the first half of https://github.com/pingcap/tidb/issues/51081. 

Adds new function `SetLowResolutionTimestampUpdateInterval(time.Duration) error` to the Oracle interface and implements it in `pdOracle` to control how often timestamps are refreshed in the local cache. Default behavior remaining unchanged at 2s.

### DIFF
--- a/oracle/oracles/export_test.go
+++ b/oracle/oracles/export_test.go
@@ -36,12 +36,12 @@ package oracles
 
 import (
 	"context"
-	pd "github.com/tikv/pd/client"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/tikv/client-go/v2/oracle"
+	pd "github.com/tikv/pd/client"
 )
 
 // SetOracleHookCurrentTime exports localOracle's time hook to test.
@@ -74,8 +74,8 @@ func StartTsUpdateLoop(o oracle.Oracle, ctx context.Context, wg *sync.WaitGroup)
 		panic("expected pdOracle")
 	}
 	pd.quit = make(chan struct{})
+	wg.Add(1)
 	go func() {
-		wg.Add(1)
 		pd.updateTS(ctx)
 		wg.Done()
 	}()
--- a/oracle/oracles/export_test.go
+++ b/oracle/oracles/export_test.go
@@ -35,6 +35,9 @@
 package oracles
 
 import (
+	"context"
+	pd "github.com/tikv/pd/client"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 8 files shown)


oracle/oracles/export_test.go
```diff
@@ -35,6 +35,9 @@
 package oracles
 
 import (
+	"context"
+	pd "github.com/tikv/pd/client"
+	"sync"
 	"sync/atomic"
 	"time"
```
cfzjywxk, line 42:
```suggestion
	"context"
	"sync"
	"sync/atomic"
	"time"
	
	pd "github.com/tikv/pd/client"
```

oracle/oracles/pd.go
```diff
@@ -231,6 +237,10 @@ func (o *pdOracle) updateTS(ctx context.Context, interval time.Duration) {
 				o.setLastTS(ts, txnScope)
 				return true
 			})
+			if o.updateInterval != currentInterval {
+				currentInterval = o.updateInterval
```
cfzjywxk, line 241: There seems to be data race as the `updateInterval` is written in another thread. Perhaps atomic types need to be used for it.

oracle/oracles/pd.go
```diff
@@ -52,13 +53,16 @@ import (
 var _ oracle.Oracle = &pdOracle{}
 
 const slowDist = 30 * time.Millisecond
+const minTimestampUpdateInterval = 50 * time.Millisecond
+const maxTimestampUpdateInterval = 5 * time.Second
```
cfzjywxk, line 57: If the configuration is to be used as an system variable, the validation could be done in the tidb side.

oracle/oracles/pd.go
```diff
@@ -52,13 +53,16 @@ import (
...
 
 // pdOracle is an Oracle that uses a placement driver client as source.
 type pdOracle struct {
 	c pd.Client
 	// txn_scope (string) -> lastTSPointer (*atomic.Pointer[lastTSO])
-	lastTSMap sync.Map
-	quit      chan struct{}
+	lastTSMap      sync.Map
+	quit           chan struct{}
+	updateInterval time.Duration
```
you06, line 65: Not related to this PR, I think we can name `updateInterval` more clearly, like `lastTSUpdateInterval`, because `updateInterval` is a little vague about what to be updated.

ari-e: I can update for clearer naming

oracle/oracles/pd.go
ari-e: It will be a system variable and will have its own validation there, but the client should probably have its own validation to make sure it doesn't get into a bad state. For example passing in duration of 0 will cause it to panic without any checking of its own.

ari-e:
Thanks for the comments @cfzjywxk @you06 @zyguan, PR has been updated addressing them.

oracle/oracles/pd.go
ari-e: After thinking about this more I just made the validation check it won't panic and will let tidb do the additional validation as you suggested

tikv/kv.go
```diff
@@ -190,6 +191,15 @@ func WithPool(gp Pool) Option {
 	}
 }
 
+func WithUpdateInterval(updateInterval time.Duration) Option {
```
cfzjywxk, line 194: Need to add comments about exported functions.

cfzjywxk (approved):
Rest LGTM, thanks for the contribution!

cfzjywxk:
@zyguan @you06 
PTAL

you06 (approved):

ari-e:
Thanks for the reviews! @MyonKeminta can you PTAL? Also CI needs to be retriggered after my new commit that just added a comment :sigh:

PR #1171, opened 2024-02-22
enhance background job management of region cache

zyguan:
There are many background cron jobs in region cache. This PR tries to reduce boilerplate code (like the following) by `bgRunner`.
```go
ticker := time.NewTicker(interval)
r.wg.Add(1)
go func() {
	defer func() {
		r.wg.Done()
		ticker.Stop()
	}()
	for {
		select {
		case <-r.ctx.Done():
			return
		case <-ticker.C:
			// ...
		}
	}
}()
```

ref: #1104


cfzjywxk:
@crazycs520 PTAL

internal/locate/region_cache.go
```diff
@@ -508,6 +508,132 @@ func (mu *regionIndexMu) refresh(r []*Region) {
...
+	}
+	ticker := time.NewTicker(interval)
+	r.wg.Add(1)
+	go func() {
+		defer func() {
+			r.wg.Done()
+			ticker.Stop()
+		}()
+		triggerEnabled := trigger != nil
+		for triggerEnabled {
```
cfzjywxk, line 606: Why use a standalone for loop for the trigger events instead of putting it in the for loop below?

internal/locate/region_cache.go
```diff
@@ -556,35 +680,46 @@ func NewRegionCache(pdClient pd.Client) *RegionCache {
...
-		c.timelyRefreshCache(config.GetGlobalConfig().RegionsRefreshInterval)
+	var (
+		refreshStoreInterval = config.GetGlobalConfig().StoresRefreshInterval
+		needCheckStores      []*Store
+	)
+	c.bg.scheduleWithTrigger(func(ctx context.Context, t time.Time) bool {
+		filter := func(state resolveState) bool {
+			return state != unresolved && state != tombstone && state != deleted
+		}
+		if t.IsZero() {
```
cfzjywxk, line 704: We could comment this is triggered by store events.

internal/locate/region_cache.go
```diff
@@ -2169,88 +2251,81 @@ func (c *RegionCache) UpdateBucketsIfNeeded(regionID RegionVerID, latestBucketsV
...
-				c.mu.Lock()
-				for _, item := range expired {
-					c.mu.sorted.b.Delete(item)
-					c.mu.removeVersionFromCache(item.cachedRegion.VerID(), item.cachedRegion.GetID())
-				}
-				c.mu.Unlock()
+		// Reach the end of the region cache, start from the beginning
+		if count < limit {
+			cursor = beginning
+		}
```
cfzjywxk, line 2293: If the count equals the limit, the start cursor for the next round is still the `beginning := newBtreeSearchItem([]byte(""))`, should we record the last cursor?

internal/locate/region_cache.go
zyguan: The `trigger` is provided by the caller and we can not make sure it won't be closed by accident. If it closed, there will be a busy loop. Once `trigger` get closed, it should be considered asthere is  no more events.

internal/locate/region_cache.go
zyguan: I've put a comment at https://github.com/tikv/client-go/pull/1171/files#diff-07678f6dc4c3b8185168ba1e2c221da085c85e26b518cf52a288e5af726cdfdcR593 , do we need to add another comment here?

internal/locate/region_cache.go
cfzjywxk: So the purpose is to avoild `select` on an already closed channel right?

internal/locate/region_cache.go
cfzjywxk: I think it's better to make it more clear.

internal/locate/region_cache.go
zyguan: Yes. Actually `notifyCheckCh` never gets closed, just try to make the runner robust here.

internal/locate/region_cache.go
zyguan: Done

internal/locate/region_cache.go
zyguan: Fixed

cfzjywxk (approved):
LGTM

internal/locate/region_cache.go
```diff
@@ -508,6 +508,132 @@ func (mu *regionIndexMu) refresh(r []*Region) {
...
+		f(r.ctx)
+	}()
+}
+
+// schedule calls `f` every `interval`.
+func (r *bgRunner) schedule(f func(context.Context, time.Time) bool, interval time.Duration) {
+	if r.closed() {
+		return
+	}
+	ticker := time.NewTicker(interval)
```
crazycs520, line 573: How about moving this line after line 575? since the ticker is stop in another goroutine.

internal/locate/region_cache.go
```diff
@@ -508,6 +508,132 @@ func (mu *regionIndexMu) refresh(r []*Region) {
...
+		}
+	}()
+}
+
+// scheduleWithTrigger likes schedule, but also call `f` when `<-trigger`, in which case the time arg of `f` is zero.
+func (r *bgRunner) scheduleWithTrigger(f func(context.Context, time.Time) bool, interval time.Duration, trigger <-chan struct{}) {
+	if r.closed() {
+		return
+	}
+	ticker := time.NewTicker(interval)
```
crazycs520, line 598: ditto.

internal/locate/region_cache.go
zyguan: My point is: if the interval <= 0, it panic when calling schedule, which is able to be recovered by the caller; however if we move it into the goroutine, panic cannot be caught by users.

crazycs520: If so, we should check interval <= 0 here, instead of causing panic and recover to handle it?

zyguan: IMO, panic on interval <= 0 is reasonable, but the panic should be recoverable. I will update it as you said anyway.

crazycs520: I think we should try to check in advance to avoid panic, Instead of letting it panic, after all, this is a system software, users will panic when they see panic logs.

zyguan: I've changed it.
BTW, panic let you kown you've passed a wrong arg, however, keeping silent sometimes hide potential risks.

crazycs520 (approved):

PR #1178, opened 2024-02-23
refine onNotLeader logic when new leader is not available

### LINKED ISSUE #1169: refine onNotLeader logic when new leader is not available
# Issue

### 1. Minimal reproduce step

```go
		reqType:   tikvrpc.CmdGet,
		staleRead: true,
		timeout:   time.Second,
		label:     &metapb.StoreLabel{Key: "id", Value: "2"},
		accessErr: []RegionErrorType{DeadLineExceededErr, NotLeaderWithNewLeader2Err},
```

```go
			accessPath: []string{
				"{addr: store2, replica-read: false, stale-read: true}",
				"{addr: store1, replica-read: false, stale-read: false}",
```

```go
			accessPath: []string{
				"{addr: store2, replica-read: false, stale-read: true}",
				"{addr: store1, replica-read: false, stale-read: false}",
				"{addr: store2, replica-read: false, stale-read: false}"}, // first try store2 already got DeadLineExceededErr, don't retry it again!!
```

The request has user-defined timeout 1s, the replica selector logic is:

1. try store2 by stale-read, since store2 has match label, but got DeadLineExceededErr.
2. retry leader in store1, but got `NotLeaderWithNewLeader2Err`.
3. retry store2.

The problem is, store2 has been tried, and got DeadLineExceededErr, retry it may also got `DeadLineExceededErr`. we should try store3 by replica-read since store3 has never been tried.

crazycs520:
close #1169

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -413,7 +413,11 @@ func (state *accessKnownLeader) next(bo *retry.Backoffer, selector *replicaSelec
 		selector.state = &tryNewProxy{leaderIdx: state.leaderIdx}
 		return nil, stateChanged{}
 	}
-	if !state.isCandidate(leader) {
+	// If hibernate region is enabled and the leader is not reachable, the raft group
+	// will not be wakened up and re-elect the leader until the follower receives
+	// a request. So, before the new leader is elected, we should not send requests
+	// to the unreachable old leader to avoid unnecessary timeout.
+	if liveness != reachable || leader.isExhausted(maxReplicaAttempt, maxReplicaAttemptTime) {
 		selector.state = &tryFollower{leaderIdx: state.leaderIdx, lastIdx: state.leaderIdx, fromAccessKnownLeader: true}
 		return nil, stateChanged{}
 	}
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -434,6 +430,19 @@ func (state *accessKnownLeader) next(bo *retry.Backoffer, selector *replicaSelec
...
 }
 
+// check leader is candidate or not.
+func (state *accessKnownLeader) isCandidate(leader *replica) bool {
+	liveness := leader.store.getLivenessState()
+	// If hibernate region is enabled and the leader is not reachable, the raft group
+	// will not be wakened up and re-elect the leader until the follower receives
+	// a request. So, before the new leader is elected, we should not send requests
+	// to the unreachable old leader to avoid unnecessary timeout.
+	if liveness != reachable || leader.isExhausted(maxReplicaAttempt, maxReplicaAttemptTime) || leader.deadlineErrUsingConfTimeout {
```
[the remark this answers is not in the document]
crazycs520: Nice catch, I already remove this change here, since this need more discuss.

internal/locate/region_request.go
```diff
@@ -1200,7 +1200,11 @@ func (s *replicaSelector) updateLeader(leader *metapb.Peer) {
 				replica.attempts = maxReplicaAttempt - 1
 				replica.attemptedTime = 0
 			}
-			s.state = &accessKnownLeader{leaderIdx: AccessIndex(i)}
+			if !replica.deadlineErrUsingConfTimeout {
+				// if the new leader doesn't meet deadline exceeded error, switch to the new leader.
+				// Otherwise, keep the state unchanged and retry the request.
```
cfzjywxk, line 1205: This strategy may also need to be discussed with the `kv timeout feature` later. When timeout error is returned, there could be different situations, for example:
- The leader is busy or slow, it could not serve the request in time. In this case, trying follower using replica read is likely to fail too.
- There is some problem bettwen the kv-client and the leader peer. In this case, trying follower using replica read could be helpful.

To make better decisions maybe we could reference to the feedback slow information from the tikv nodes.

cfzjywxk (approved):
Maybe we could continue first

internal/locate/region_request.go
crazycs520: Now that the above logic has been met, you can take a look at the test case.

crazycs520:
> Maybe we could continue first

Done, and also add a test case for it.

zyguan (approved):

Issue #1175, opened 2024-02-23
For stale read request, we can retry DataIsNotReady replica by replica-read


crazycs520:
# Issue

### 1. Minimal reproduce step

In the following case, suppose the region leader is in `store1`.

```go
	ca = replicaSelectorAccessPathCase{
		reqType:   tikvrpc.CmdGet,
		readType:  kv.ReplicaReadMixed,
		staleRead: true,
		label:     &metapb.StoreLabel{Key: "id", Value: "2"},
		accessErr: []RegionErrorType{DataIsNotReadyErr, ServerIsBusyErr, ServerIsBusyErr, ServerIsBusyErr},
```

### 2. What did you expect to see?

```go
			accessPath: []string{
				"{addr: store2, replica-read: false, stale-read: true}",
				"{addr: store1, replica-read: false, stale-read: false}", // try leader with leader read.
				"{addr: store2, replica-read: true, stale-read: false}",
				"{addr: store3, replica-read: true, stale-read: false}",
			},
```

### 3. What did you see instead?

```go
			accessPath: []string{
				"{addr: store2, replica-read: false, stale-read: true}",
				"{addr: store1, replica-read: false, stale-read: false}", // try leader with leader read.
				"{addr: store3, replica-read: true, stale-read: false}",
```

Currently, the replica selector logic is:

1. try store2 by stale-read, since store2 has match label, but got `DataIsNotReadyErr`.
2. retry leader in store1, but got `ServerIsBusyErr`.
3. retry store3, but got `ServerIsBusyErr` too.
4. failed, since no replica available.

The problem is, store2 is available, it's meet `DataIsNotReadyErr`, and we can use replica-read to retry store2 again.

crazycs520:
should fix by https://github.com/crazycs520/client-go/commit/2d31d6f9b8f8f75ca4c41f9aab8c3f221d6c0cb9

crazycs520:
fix by #1181

PR #1181, opened 2024-02-26
don't retry same replica twice in a round, but for stale read, we can retry DataIsNotReady replica by replica-read

### LINKED ISSUE #1172: don't retry same replica twice in a round
# Issue

### 1. Minimal reproduce step

```go
		reqType:   tikvrpc.CmdGet,
		staleRead: true,
		timeout:   0,
		label:     &metapb.StoreLabel{Key: "id", Value: "3"},
		accessErr: []RegionErrorType{ServerIsBusyErr, ServerIsBusyErr},
```

```go
			accessPath: []string{
				"{addr: store3, replica-read: false, stale-read: true}",
				"{addr: store1, replica-read: false, stale-read: false}",
				"{addr: store2, replica-read: true, stale-read: false}"},
```

```go
			accessPath: []string{
				"{addr: store3, replica-read: false, stale-read: true}",
				"{addr: store1, replica-read: false, stale-read: false}",
```

Currently, the replica selector logic is:

1. try store3 by stale-read, since store3 has match label, but got ServerIsBusyErr.
2. retry leader in store1, but got `ServerIsBusyErr` too.
3. retry store3.

The problem is, store3 has been tried, and got ServerIsBusyErr, retry it may also got `ServerIsBusyErr`. we should try store2 by replica-read since store2 has never been tried.

crazycs520:
close #1172 ,  #1175


cfzjywxk (approved):
LGTM

zyguan (approved):

PR #1183, opened 2024-02-26
replica-read request with mixed strategy and with label, should be able to retry all remain replicas

### LINKED ISSUE #1176: replica-read request with mixed strategy and with label, should be able to retry all remain replicas
# Issue

### 1. Minimal reproduce step

```go
		reqType:   tikvrpc.CmdGet,
		staleRead: false,
		timeout:   time.Second,
		label:     &metapb.StoreLabel{Key: "id", Value: "2"},
		accessErr: []RegionErrorType{DeadLineExceededErr, ServerIsBusyErr},
```

```go
		expect: &accessPathResult{
			accessPath: []string{
			respErr:         "",
			respRegionError: nil,
```

```go
		expect: &accessPathResult{
			accessPath: []string{
				"{addr: store1, replica-read: true, stale-read: false}"}, // retry leader
			respErr:         "",
			respRegionError: fakeEpochNotMatch,  // return fake EpochNotMatchErr since no available replica.
```

Currently, the replica selector logic is:

1. try store2 by replica-read, since store2 has match label, but got `DeadLineExceededErr`.
2. retry leader in store1, but got `ServerIsBusyErr`.
3. return fake EpochNotMatchErr since no available replica.

The problem is, store3 is available, and we can use replica-read to retry store3, since store3 is never been tried.

crazycs520:
close #1176


zyguan (approved):

internal/locate/region_request.go
```diff
@@ -766,7 +766,8 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 			// If the leader is also unavailable, we can fallback to the follower and use replica-read flag again,
 			// The remote follower not tried yet, and the local follower can retry without stale-read flag.
 			// If leader tried and received deadline exceeded error, try follower.
-			if state.isStaleRead || leader.deadlineErrUsingConfTimeout {
+			// If has labels, some followers's labels may not match, then can't be candidate, need to try follower.
```
cfzjywxk, line 769:
```suggestion
			// If labels are used, some followers would be filtered by the labels and can't be candidates, they still need to be retried.
```

cfzjywxk (approved):

internal/locate/region_request.go
cfzjywxk: Chatgpt is useful to polish the sentenses.

PR #923, opened 2023-08-03
fallback to follower when leader is busy (#916)

you06:
Cherry pick #916 to master

PR #1184, opened 2024-02-27
reduce unnecessary tikvServerBusy backoff when able to try next replica

### LINKED ISSUE #1166: Reduce unnecessary tikvServerBusy backoff
# Issue

### 1. Minimal reproduce step

```go
		reqType:   tikvrpc.CmdGet,
		staleRead: true,
                 timeout:   time.Microsecond * 100,
		accessErr: []RegionErrorType{DataIsNotReadyErr, ServerIsBusyErr},
	}
```

No backoff.

```go
	backoffCnt:      0,
	backoffDetail:   []string{},
```

BTW, the expected access path is following, we should fast retry the next replica without backoff. Since a `tikvServerBusy` backoff may cost more than 1 second, and the request's timeout is `100ms`.

```go
	accessPath: []string{
		"{addr: store1, replica-read: false, stale-read: true}",
		"{addr: store3, replica-read: true, stale-read: false}", }
```

```go
	backoffCnt:      1,
	backoffDetail:   []string{"tikvServerBusy+1"},
```

crazycs520:
close #1166

PR [923](https://github.com/tikv/client-go/pull/923/files#diff-0c36d07fbd1fc71c850ff57d323fe12bdf7f2c053ef71b7fec5bb44358f1070dR1152 ) introduce the skip `tikvServerBusy` backoff logic, but it it only work for stale-read, and if `serverIsBusy.EstimatedWaitMs > 0` is true, it can't skip. 

This PR expanded the scope of skips `tikvServerBusy` backoff logic, make non-leader read requests can skip `tikvServerBusy` backoff, and if `serverIsBusy.EstimatedWaitMs > 0` is true, non-leader read requests can skip too.


internal/locate/region_request.go
```diff
@@ -1294,15 +1294,11 @@ func (s *replicaSelector) canFallback2Follower() bool {
...
-		return false
-	}
-	if !state.isStaleRead {
+	_, ok := s.state.(*accessKnownLeader)
+	if ok {
 		return false
 	}
-	// can fallback to follower only when the leader is exhausted.
-	return state.lastIdx == state.leaderIdx && state.IsLeaderExhausted(s.replicas[state.leaderIdx])
+	return true
```
crazycs520, line 1301: This change may reduce a lot of `tikvServerBusy` backoff, and may need to be discussed. @cfzjywxk @you06 @zyguan PTAL

internal/locate/region_request.go
```diff
@@ -1277,9 +1277,9 @@ func (s *replicaSelector) onServerIsBusy(
 		// Mark the server is busy (the next incoming READs could be redirect
 		// to expected followers. )
 		ctx.Store.markAlreadySlow()
-		if s.canFallback2Follower() {
-			return true, nil
-		}
+	}
+	if s.canFallback2Follower() {
```
cfzjywxk, line 1281: As @MyonKeminta has mentiond, there are situations the `backoff` is still needed, like:
- All the replicas in this region have reported busy or have other information indicating that the node is slow
- If the node that will be retried next is likely to be slow when retrying

It's better to clearly specify the conditions under which backoff can be skipped and then skip it, like
```
if conditions
   skip backoff
```
rather than deciding which situations require backoff, 
```
if conditions
  do backoff
```
to reduce the risk of changes.

crazycs520: Nice catch, done, PTAL

internal/locate/region_request.go
```diff
@@ -1288,14 +1288,24 @@ func (s *replicaSelector) onServerIsBusy(
 	return true, nil
 }
 
-// For some reasons, the leader is unreachable by now, try followers instead.
-// the state is changed in accessFollower.next when leader is unavailable.
-func (s *replicaSelector) canFallback2Follower() bool {
+// canSkipServerIsBusyBackoff returns true if the request can be sent to next replica and can skip ServerIsBusy backoff.
+func (s *replicaSelector) canSkipServerIsBusyBackoff() bool {
```
you06, line 1286: Maybe we can reuse this function handling region unavailable and changing it's name in the future. Not this PR's work.

you06 (approved):
LGTM

internal/locate/region_request.go
```diff
@@ -1288,21 +1282,24 @@ func (s *replicaSelector) onServerIsBusy(
...
+		// so don't skip since the leader is still busy.
 		return false
 	}
-	if !state.isStaleRead {
-		return false
+	for _, replica := range s.replicas {
+		if replica.attempts == 0 && replica.store.getLivenessState() == reachable &&
+			!replica.isEpochStale() && !replica.store.isSlow() {
+			return true
+		}
```
MyonKeminta, line 1300: Is it sure that the next attempt must be on such a node, no matter what selectorState it is currently and whether the state will change?

crazycs520: Strictly speaking, this is not guaranteed. I can't call `state.next()` to get next replica directly, because `state.next()` method is stateful, call it will affect internal `lastIdx`.

But fortunately, it misjudgment at most once in extreme cases.

Following is a misjudgment example with `ReplicaReadMixed`:

1. Try replica-1, got server is busy, skip backoff since replica-3 is good, but replica-2's store is slow.
2. Try replica-2, got server is busy, skip backoff since replica-3 is good.
3. Try replica-3, return success.

crazycs520:
> LGTM, but I'm thinking perhaps the better structure is to determine whether to backoff just before the next attempt, instead of just after encountering error, so that what @zyguan tried to do in #990 becomes a unified way. Btw I just noticed that it seems that this PR and #990 looks trying to solve the same problem 🤔 cc @zyguan

Nice catch. #990 introduces pending backoff for fast retry, which is great idea. After discussing with @zyguan, I introduce the  pending backoff idea  into this PR and unified the fast retry logic. And later, we can add more backoff to pending to do fast retry.

zyguan (approved):

internal/locate/region_request.go
```diff
@@ -2462,3 +2447,59 @@ func (s *replicaSelector) recordAttemptedTime(duration time.Duration) {
...
+	cfg *retry.Config
+	err error
+}
+
+// addPendingBackoff adds pending backoff for the store.
+func (s *replicaSelector) addPendingBackoff(store *Store, cfg *retry.Config, err error) {
+	storeId := uint64(0)
+	if store != nil {
+		storeId = store.storeID
+	}
```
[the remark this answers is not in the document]
crazycs520: I don't know, since there is a `ctx != nil && ctx.Store != nil` check in the up layer https://github.com/tikv/client-go/blob/ea5a5c4ba674f42a468510f4fa4168dad77196af/internal/locate/region_request.go#L1286. And with this check, look like `store` could be nil. But as I known, It should never be nil, I didn't delete this check because I was afraid that there would be a situation that I didn't know.

internal/locate/region_request.go
```diff
@@ -285,6 +285,8 @@ type replicaSelector struct {
 	// TiKV can reject the request when its estimated wait duration exceeds busyThreshold.
 	// Then, the client will receive a ServerIsBusy error and choose another replica to retry.
 	busyThreshold time.Duration
+	// pendingBackoffs records the pending backoff for fast retry.
```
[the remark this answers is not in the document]

internal/locate/region_request.go
```diff
@@ -285,7 +285,22 @@ type replicaSelector struct {
...
+	// Example-1, fast retry and success:
+	// 		1. send req to store 1, got ServerIsBusy region error, record `store1 -> BoTiKVServerBusy` backoff in pendingBackoffs and fast retry next replica.
+	// 		2. retry in store2, and success.
+	//      Since the request is success, we can skip the backoff and fast return result to user.
+	// Example-2: fast retry different replicas but all failed:
+	// 		1. send req to store 1, got ServerIsBusy region error, record `store1 -> BoTiKVServerBusy` backoff in pendingBackoffs and fast retry next replica.
+	// 		2. send req to store 2, got ServerIsBusy region error, record `store2 -> BoTiKVServerBusy` backoff in pendingBackoffs and fast retry next replica.
+	// 		3. send req to store 3, got ServerIsBusy region error, record `store3 -> BoTiKVServerBusy` backoff in pendingBackoffs and fast retry next replica.
+	//      4. no candidate since all stores are busy. But before return no candidate error to up layer, we need to call backoffOnNoCandidate function
+	//         to apply a max pending backoff, the backoff is to avoid frequent access and increase the pressure on the cluster.
```
MyonKeminta, line 298: The text looks not well aligned. How about considering using only spaces (no tabs) in comments?

MyonKeminta (commented):
Rest LGTM

internal/locate/region_cache.go
```diff
@@ -3252,6 +3252,21 @@ func (s *Store) GetPeerAddr() string {
...
+		estimatedWait := time.Duration(estimatedWaitMs) * time.Millisecond
+		// Update the estimated wait time of the store.
+		loadStats := &storeLoadStats{
+			estimatedWait:     estimatedWait,
+			waitTimeUpdatedAt: time.Now(),
+		}
+		s.loadStats.Store(loadStats)
+	} else {
+		// Mark the server is busy (the next incoming READs could be redirect to expected followers.)
+		s.healthStatus.markAlreadySlow()
```
cfzjywxk, line 3266: Why putting the `markAlreadySlow` in the `updateServerLoadStats`? I think it shoud remain in the `OnServerIsBusy` action function.

internal/locate/region_request.go
```diff
@@ -1487,6 +1477,11 @@ func (s *RegionRequestSender) SendReqCtx(
 			// TODO: Change the returned error to something like "region missing in cache",
 			// and handle this error like EpochNotMatch, which means to re-split the request and retry.
 			s.logSendReqError(bo, "throwing pseudo region error due to no replica available", regionID, retryTimes, req, totalErrors)
+			if s.replicaSelector != nil {
```
cfzjywxk, line 1480: Are the L1480 and L1514 strategy change that we should also backoff when there is no candidate or before retry, compared with the orginal logic?

crazycs520: The original logic is, whether there is a candidate or not, do backoff first when an region error is encountered.

compared with the original logic, you can see the `TestReplicaReadAccessPathByCase` change, some test-case's `backoffCnt` and `backoffDetail` are changed.

internal/locate/region_cache.go
crazycs520: done.

internal/locate/region_request.go
```diff
@@ -2462,3 +2465,59 @@ func (s *replicaSelector) recordAttemptedTime(duration time.Duration) {
 		proxyReplica.attemptedTime += duration
 	}
 }
+
+// canFastRetry returns true if the request can be sent to next replica.
+func (s *replicaSelector) canFastRetry() bool {
+	accessLeader, ok := s.state.(*accessKnownLeader)
```
cfzjywxk, line 2471: Should we just do fast retry when `tikv_client_read_timeout` is configured? It looks a little bit dangerous to use
```
if not fast retry
   return false
return true
```
as most of the request would be fast retried .

Besides, maybe we could consult slow information to decide whether fast retry should work. /cc @zyguan @MyonKeminta

crazycs520: I see the original logic of [canFallback2Follower](https://github.com/tikv/client-go/pull/923/files#diff-0c36d07fbd1fc71c850ff57d323fe12bdf7f2c053ef71b7fec5bb44358f1070dR1152-R1154) is trying to do fast backoff when retry next replica, so fast backoff is not only work for `tikv_client_read_timeout` is configured.
cc @you06

cfzjywxk: > I see the original logic of [canFallback2Follower](https://github.com/tikv/client-go/pull/923/files#diff-0c36d07fbd1fc71c850ff57d323fe12bdf7f2c053ef71b7fec5bb44358f1070dR1152-R1154) is trying to do fast backoff when retry next replica, so fast backoff is not only work for `tikv_client_read_timeout` is configured. cc @you06

@crazycs520 It the current logic the same? If so we could continue

crazycs520: Not same, this PR is for the fix issue #1166.

cfzjywxk (approved):

PR #1192, opened 2024-02-28
Don't invalid region in accessFollower when set user-defined req time out and meet DeadLineExceededErr

### LINKED ISSUE #1190: Don't invalid region in accessFollower when set user-defined req timeout and meet DeadLineExceededErr
# Issue

### 1. Minimal reproduce step

```go
		reqType:   tikvrpc.CmdGet,
		staleRead: false,
		timeout:   time.Second,
		label:     nil,
		accessErr: []RegionErrorType{ServerIsBusyErr, ServerIsBusyErr, DeadLineExceededErr},
```

```go
		expect: &accessPathResult{
			accessPath: []string{
			regionIsValid:   true,
```

```go
		expect: &accessPathResult{
			...
			regionIsValid:   false,
		},
```

The request has user-defined timeout 1s, the replica selector logic is:

1. try store1 by replica-read, but got ServerIsBusyErr.
2. retry store2, got `ServerIsBusyErr`.
3. retry store3, got `DeadLineExceededErr`, and then invalidate the region in cache.

The expected result is doesn't invalidate region in cache, and the do fast retry without user-defined timeout.

crazycs520:
close #1190

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -521,12 +521,11 @@ func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (
 
 	// If all followers are tried and fail, backoff and retry.
 	if selector.targetIdx < 0 {
-		if hasDeadlineExceededError(selector.replicas) {
-			// when meet deadline exceeded error, do fast retry without invalidate region cache.
-			return nil, nil
+		// when meet deadline exceeded error, do fast retry without invalidate region cache.
+		if !hasDeadlineExceededError(selector.replicas) {
+			metrics.TiKVReplicaSelectorFailureCounter.WithLabelValues("exhausted").Inc()
+			selector.invalidateRegion()
 		}
-		metrics.TiKVReplicaSelectorFailureCounter.WithLabelValues("exhausted").Inc()
-		selector.invalidateRegion()
 		return nil, nil
 	}
 	rpcCtx, err := selector.buildRPCContext(bo)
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -787,6 +785,10 @@ func (state *accessFollower) next(bo *retry.Backoffer, selector *replicaSelector
 				}
 				return nil, stateChanged{}
 			}
+			if hasDeadlineExceededError(selector.replicas) {
+				// when meet deadline exceeded error, do fast retry without invalidate region cache.
+				return nil, nil
+			}
 			metrics.TiKVReplicaSelectorFailureCounter.WithLabelValues("exhausted").Inc()
 			selector.invalidateRegion()
 			return nil, nil
```
cfzjywxk, line 794:
```suggestion
                        // when meet deadline exceeded error, do fast retry without invalidate region cache.
			if !hasDeadlineExceededError(selector.replicas) {
				metrics.TiKVReplicaSelectorFailureCounter.WithLabelValues("exhausted").Inc()
			        selector.invalidateRegion()
				
			}
			return nil, nil
```

internal/locate/region_request.go
```diff
@@ -521,11 +521,9 @@ func (state *tryFollower) next(bo *retry.Backoffer, selector *replicaSelector) (
 
 	// If all followers are tried and fail, backoff and retry.
 	if selector.targetIdx < 0 {
-		for _, replica := range selector.replicas {
-			if replica.deadlineErrUsingConfTimeout {
-				// when meet deadline exceeded error, do fast retry without invalidate region cache.
-				return nil, nil
-			}
+		if hasDeadlineExceededError(selector.replicas) {
```
cfzjywxk, line 524: Ditto below

internal/locate/region_request.go

cfzjywxk (approved):

PR #1196, opened 2024-02-29
region_cache: move store related code to a standalone file

zyguan:
The 2nd PR for extracting store management from region cache. (ref https://github.com/tikv/client-go/pull/1106 & https://github.com/tikv/client-go/issues/1104)


internal/locate/store_cache.go
MyonKeminta, line 1: Will there be a new type as the store cache in later work, or we are only going to split out a file?

MyonKeminta (approved):
LGTM

internal/locate/region_cache.go
```diff
@@ -69,16 +69,9 @@ import (
 	"github.com/tikv/client-go/v2/tikvrpc"
 	"github.com/tikv/client-go/v2/util"
 	pd "github.com/tikv/pd/client"
-	atomic2 "go.uber.org/atomic"
+	uatomic "go.uber.org/atomic"
```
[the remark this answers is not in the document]
zyguan: I think so. I will do it in this PR since it doesn't involve too many changes.

internal/locate/store_cache.go
zyguan: Yes, there will be a new type and store related fields will be move to it from `RegionCache`.

MyonKeminta (approved):

cfzjywxk (approved):

PR #1204, opened 2024-03-05
fix bug introduced by https://github.com/tikv/client-go/pull/1200

SeaRise:
Regions returned by `LocateKeyRange` must be ordered.

Has tested in https://github.com/pingcap/tidb/pull/51465


windtalker (approved):
LGTM

PR #1210, opened 2024-03-08
p-dml: fix unstopped resolve

you06:
When the resolve task crosses regions, it resolve the region repeatly without stop.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -1549,6 +1549,7 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		if err = c.txn.GetMemBuffer().FlushWait(); err != nil {
 			return err
 		}
+		c.txn.pipelinedCancel()
 		if len(c.pipelinedCommitInfo.pipelinedStart) == 0 || len(c.pipelinedCommitInfo.pipelinedEnd) == 0 {
 			return errors.Errorf("unexpected empty pipelinedStart(%s) or pipelinedEnd(%s)",
 				c.pipelinedCommitInfo.pipelinedStart, c.pipelinedCommitInfo.pipelinedEnd)
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -742,6 +742,8 @@ func (txn *KVTxn) Rollback() error {
 		}
 	}
 	if txn.IsPipelined() && txn.committer != nil {
+		// wait all flush to finish, this avoids data race.
+		txn.GetMemBuffer().FlushWait()
 		txn.committer.ttlManager.close()
 	}
 	txn.close()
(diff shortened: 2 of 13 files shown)


txnkv/transaction/txn.go
```diff
@@ -742,6 +742,8 @@ func (txn *KVTxn) Rollback() error {
 		}
 	}
 	if txn.IsPipelined() && txn.committer != nil {
+		// wait all flush to finish, this avoids data race.
+		txn.GetMemBuffer().FlushWait()
```
ekexium, line 746: Is there any convenient way to avoid this wait? I suppose Rollback operations had better be immediate. This is not a big problem any way. Don't be blocked by it.

ekexium (approved):

txnkv/transaction/txn.go
you06: It's not allowed that rollback and flush running concurrently.
Added a cancel function to avoid wait too long.

txnkv/transaction/2pc.go
```diff
@@ -1549,6 +1549,7 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		if err = c.txn.GetMemBuffer().FlushWait(); err != nil {
 			return err
 		}
+		c.txn.pipelinedCancel()
```
ekexium, line 1552: Is there a difference to cancel here?

you06: No difference, just avoid leaving uncanceled context.

txnkv/transaction/txn.go
cfzjywxk: > It's not allowed that rollback and flush running concurrently.

When the rollback has put the rollback record at the primary lock, it shoule be ok to be executed concurrently, the stale prewrite lock would be cleaned up by the rollback requests or they would encounter write conflict seeing the rollback record.

cfzjywxk (approved):

cfzjywxk:
@you06 The branch needs to be rebased.

txnkv/transaction/txn.go
you06: @cfzjywxk It's ok for the distribution-txn stores, but race may happens inside client-go.

`txn.diskFullOpt = kvrpcpb.DiskFullOpt_NotAllowedOnFull` is invoked inside `txn.close()`.

https://github.com/tikv/client-go/blob/b81a5ee7c43f5e34aced4a9aa72cb3701acd2217/txnkv/transaction/txn.go#L747

Meanwhile, `txn.diskFullOpt` may be read here.

https://github.com/tikv/client-go/blob/b81a5ee7c43f5e34aced4a9aa72cb3701acd2217/txnkv/transaction/pipelined_flush.go#L99

PR #1212, opened 2024-03-11
feat: introduce GetLocal() and BatchGet cache for PipelinedMemDB

ekexium:
Ref https://github.com/pingcap/tidb/issues/51625


you06:
I don't know what's wrong with golangci test. Fix it in this PR by adding `skip-pkg-cache` option, need to remove it before merge.

https://github.com/tikv/client-go/blob/d7f97008577a0352e788d4276103abe3a232b4e1/.github/workflows/test.yml#L51

Without this workaround, the test failed in prepare phase. e.g. https://github.com/tikv/client-go/actions/runs/8258621325/job/22591178277

internal/unionstore/pipelined_memdb.go
```diff
@@ -174,6 +176,62 @@ func (p *PipelinedMemDB) GetFlags(k []byte) (kv.KeyFlags, error) {
...
+			return nil, err
+		}
+		if len(v) > 0 {
+			m[string(k)] = v
+			p.prefetchCache[string(k)] = util.Some(v)
+		} else {
+			p.prefetchCache[string(k)] = util.None[[]byte]()
+		}
+	}
+	storageValues, err := p.bufferBatchGetter(ctx, shrinkKeys)
```
ekexium, line 201: BufferBatchGet doesn't distinguish NotExist and empty value?

internal/unionstore/pipelined_memdb.go
```diff
@@ -174,6 +176,62 @@ func (p *PipelinedMemDB) GetFlags(k []byte) (kv.KeyFlags, error) {
...
+				shrinkKeys = append(shrinkKeys, k)
+				continue
+			}
+			return nil, err
+		}
+		if len(v) > 0 {
+			m[string(k)] = v
+			p.prefetchCache[string(k)] = util.Some(v)
+		} else {
+			p.prefetchCache[string(k)] = util.None[[]byte]()
```
ekexium, line 198:
```suggestion
			p.prefetchCache[string(k)] = util.Some(v)
```
err == nil, so this is an empty value, right?
Should it be like this?

txnkv/transaction/batch_getter.go
```diff
@@ -66,31 +66,67 @@ func NewBufferBatchGetter(buffer BatchBufferGetter, snapshot BatchGetter) *Buffe
...
+	for _, key := range keys {
+		val, ok := bufferValues[string(key)]
+		if !ok {
+			shrinkKeys = append(shrinkKeys, key)
 		}
-		if !tikverr.IsErrNotFound(err) {
-			return nil, err
+		// the deleted key should be removed from the result, and also no need to snapshot read it again.
+		if len(val) == 0 {
+			delete(bufferValues, string(key))
```
ekexium, line 84: Should this `if` be put in an else branch of the above `if` statement?

We must explicitly state whether each BatchGet returns empty values.
For MemDBWithContext and PipelinedMemDB, they do.
For other `BatchGet`s, they probably don't?
Or shall we unify the behavior of the interface in a way that they all ignore empty values?

txnkv/transaction/batch_getter.go
```diff
@@ -66,31 +66,67 @@ func NewBufferBatchGetter(buffer BatchBufferGetter, snapshot BatchGetter) *Buffe
...
+	snapshot BatchGetter
+}
+
+// Prefetcher is the interface for Prefetch.
+type Prefetcher interface {
+	Prefetch(context.Context, [][]byte) (map[string][]byte, error)
+}
+
+// NewPrefetcher creates a new Prefetcher.
+func NewPrefetcher(buffer Prefetcher, snapshot BatchGetter) *BatchPrefetcher {
```
ekexium, line 108: I see no usage of this type. Can it be deleted?

internal/unionstore/pipelined_memdb.go
you06: See the collect function, NotExist won't be in the returned pairs, and empty values are. So when reading value from the map `storageValues`, we can distinguish this case.

https://github.com/ekexium/client-go/blob/925124f5b319d943813f29de30990bfe4572e8ef/txnkv/txnsnapshot/snapshot.go#L257-L266

txnkv/transaction/batch_getter.go
you06: Nice catch, this is a meaningless delete when `ok` is false.

`BatchGet` no more returns empty values in the current implementation, so we need to follow this rule.

https://github.com/tikv/client-go/blob/7bce6ea4dfa51c0ef98f16c9bbfc8c05144878b2/txnkv/transaction/batch_getter.go#L90-L92

txnkv/transaction/batch_getter.go
you06: Removed.

you06:
The test seems cannot exit. Not related to the main changes, will fix it before merge.

ekexium:
LGTM

you06:
The integration test with tikv is failed because the tested tikv binary is not updated to nightly yet(so it does not return empty values for buffer batch get).
image

txnkv/txnsnapshot/snapshot.go
```diff
@@ -255,7 +255,8 @@ func (s *KVSnapshot) BatchGetWithTier(ctx context.Context, keys [][]byte, readTi
 	// Create a map to collect key-values from region servers.
 	var mu sync.Mutex
 	err := s.batchGetKeysByRegions(bo, keys, readTier, func(k, v []byte) {
-		if len(v) == 0 {
+		// when read buffer tier, empty value means a delete record, should also collect it.
```
cfzjywxk, line 258: This is confusing, is it an extra fix?

cfzjywxk (commented):
Rest LGTM

txnkv/txnsnapshot/snapshot.go
you06: It's the different behaviors between snapshot read and buffer read. For a delete record, buffer read return empty value, e.g.:

```go
memdb := newMemDB()
memdb.Delete([]byte("k1"))
v, err := memdb.Get([]byte("k1")) // v == []byte{}, err == nil
``` 

The related [TiKV PR](https://github.com/tikv/tikv/pull/16646).

cfzjywxk (approved):

you06 (approved):

PR #1215, opened 2024-03-12
shouldn't try leader first when leader's store is slow in prefer-leader strategy in replicaSelectorV2

### LINKED ISSUE #1214: shouldn't try leader first when leader's store is slow in prefer-leader strategy in replicaSelectorV2
# Issue

### 1. Minimal reproduce step

Make config `EnableForwarding` set to true, which is enable forwarding.

In the following case, suppose the region leader is in `store1`, and `store1` is been marked slow since got `ServerIsBusyErr` before.

```go
			reqType:   tikvrpc.CmdGet,
			readType:  kv.ReplicaReadPreferLeader,
			staleRead: false,
			timeout:   0,
			accessErr: []RegionErrorType{ServerIsBusyErr, ServerIsBusyErr, ServerIsBusyErr},
```

```go
			expect: &accessPathResult{
				accessPath: []string{
					"{addr: store2, replica-read: true, stale-read: false}", // won't try leader in store1, since it is slow.
					"{addr: store3, replica-read: true, stale-read: false}",
				},
				respErr:         "",
				respRegionError: fakeEpochNotMatch,
				backoffCnt:      1,
				backoffDetail:   []string{"tikvServerBusy+1"},
				regionIsValid:   false,
			},
```

```go
			expect: &accessPathResult{
				accessPath: []string{
					"{addr: store3, replica-read: true, stale-read: false}",
				},
				respErr:         "",
				respRegionError: fakeEpochNotMatch,
				backoffCnt:      1,
				backoffDetail:   []string{"tikvServerBusy+1"},
				regionIsValid:   false,
			},
```

crazycs520:
close #1214

### DIFF
--- a/internal/locate/replica_selector_test.go
+++ b/internal/locate/replica_selector_test.go
@@ -1689,6 +1689,50 @@ func TestReplicaReadAccessPathByMixedAndPreferLeaderCase(t *testing.T) {
 		},
 	}
 	s.True(s.runMultiCaseAndCompare(cas))
+
+	cas = []replicaSelectorAccessPathCase{
+		{
+			reqType:   tikvrpc.CmdGet,
+			readType:  kv.ReplicaReadPreferLeader,
+			staleRead: false,
+			timeout:   0,
+			accessErr: []RegionErrorType{ServerIsBusyErr, ServerIsBusyErr},
+			expect: &accessPathResult{
+				accessPath: []string{
+					"{addr: store1, replica-read: true, stale-read: false}", // store1 will be marked already slow.
+					"{addr: store2, replica-read: true, stale-read: false}", // store2 will be marked already slow.
+					"{addr: store3, replica-read: true, stale-read: false}",
+				},
+				respErr:         "",
+				respRegionError: nil,
+				backoffCnt:      0,
+				backoffDetail:   []string{},
+				regionIsValid:   true,
+			},
+			afterRun: func() { /* don't invalid region */ },
+		},
+		{
+			reqType:   tikvrpc.CmdGet,
+			readType:  kv.ReplicaReadPreferLeader,
+			staleRead: false,
+			timeout:   0,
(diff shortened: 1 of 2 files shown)


MyonKeminta (approved):

internal/locate/replica_selector_test.go
```diff
@@ -1658,6 +1646,49 @@ func TestReplicaReadAccessPathByMixedAndPreferLeaderCase(t *testing.T) {
...
+			readType:  kv.ReplicaReadPreferLeader,
+			staleRead: false,
+			timeout:   0,
+			accessErr: []RegionErrorType{ServerIsBusyErr, ServerIsBusyErr, ServerIsBusyErr},
+			beforeRun: func() { /* don't resetStoreState */ },
+			expect: &accessPathResult{
+				accessPath: []string{
+					"{addr: store2, replica-read: true, stale-read: false}", // won't try leader in store1, since it is slow.
+					"{addr: store3, replica-read: true, stale-read: false}",
+					"{addr: store1, replica-read: true, stale-read: false}",
```
you06, line 1681: I'm wondering what will be if store2 is also slow. I think the access path shall be `store3 -> store1 -> store2`, but v1 seems don't try store2 after store1.
Is it an advantage of v2?

you06 (approved):
rest LGTM

internal/locate/replica_selector_test.go
cfzjywxk: @crazycs520 PTAL

crazycs520: Nice catch, It looks like v2 is not compatible with v1 in this case, I'm going to modify v2 to be compatible with v1 and add tests for that.

crazycs520: @you06 addressed, and add a test for it, PTAL.

you06 (approved):

cfzjywxk (approved):

PR #1224, opened 2024-03-13
feat: pipelined dml has its own max-txn-ttl of 24 hours

ekexium:
ref https://github.com/tikv/tikv/issues/16291
Pipelined txn tends to run long. The default max TTL of 1 hour might not be enough.


cfzjywxk (approved):
The max txn ttl hehaviour should be explained in the docment too.

you06 (approved):

Issue #1222, opened 2024-03-13
client_batch have race when enable-forwarding is true


crazycs520:
following is the data race. and the race will cause panic in https://github.com/pingcap/tidb/issues/39169

```go
==================
==================
WARNING: DATA RACE
Write at 0x00c0007eceb0 by goroutine 589:
  runtime.recvDirect()
      /usr/local/go/src/runtime/chan.go:348 +0x7c
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsEntry).error()
      /Users/cs/code/goread/src/github.com/pingcap/client-go/internal/client/client_batch.go:86 +0x130
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsClient).failPendingRequests.func1()
      /Users/cs/code/goread/src/github.com/pingcap/client-go/internal/client/client_batch.go:617 +0xe0
  sync.(*Map).Range()
      /usr/local/go/src/sync/map.go:476 +0x1a8
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsClient).failPendingRequests()
      /Users/cs/code/goread/src/github.com/pingcap/client-go/internal/client/client_batch.go:612 +0xd0
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsClient).recreateStreamingClient()
      /Users/cs/code/goread/src/github.com/pingcap/client-go/internal/client/client_batch.go:794 +0x10c
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsClient).batchRecvLoop()
      /Users/cs/code/goread/src/github.com/pingcap/client-go/internal/client/client_batch.go:702 +0x5c0
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsClient).initBatchClient.func1()
      /Users/cs/code/goread/src/github.com/pingcap/client-go/internal/client/client_batch.go:842 +0xb4

Previous read at 0x00c0007eceb0 by goroutine 601:
  runtime.chansend1()
      /usr/local/go/src/runtime/chan.go:146 +0x2c
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsClient).batchRecvLoop()
      /Users/cs/code/goread/src/github.com/pingcap/client-go/internal/client/client_batch.go:737 +0x8d4
  github.com/tikv/client-go/v2/internal/client.(*batchCommandsClient).initBatchClient.func1()
[...]

PR #1219, opened 2024-03-13
*: fix data race that may cause panic and concurrency request limit bug

crazycs520:
close #1222  , https://github.com/pingcap/tidb/issues/51728

PR #1226, opened 2024-03-14
fix issue that no available connections cause by concurrency request limit bug

### LINKED ISSUE #1225: no available connections cause by concurrency request limit bug
# Bug Report

#1098 introduced concurrency request limit.

https://github.com/tikv/client-go/blob/03bbadb2777ec8ab003123f24bc828dc48c03c31/internal/client/client_batch.go#L407

https://github.com/tikv/client-go/blob/03bbadb2777ec8ab003123f24bc828dc48c03c31/internal/client/client_batch.go#L569-L571

Following is the general situation about the upper code:

```go
hasHighPriorityTask = false  // most request don't have high priority.
c.maxConcurrencyRequestLimit = math.MaxInt64 // by default
c.sent >= 0
```

In general, the above code works fine. But it will have a bug with the data race in #1222 : 

- the data race will cause the `c.sent` to be less than 0
- then it will cause `c.available()` computation overflow
- then `c.available() > 0` will be false.
- then have `no available connections` issue, and the request could not be sent, resulting in a QPS drop to 0.  See issue in https://github.com/pingcap/tidb/issues/51728.

crazycs520:
Close #1225.

Data race issue #1222 should be fix in PR #1219.

### DIFF
--- a/internal/client/client_batch.go
+++ b/internal/client/client_batch.go
@@ -567,7 +567,14 @@ func (c *batchCommandsClient) isStopped() bool {
 }
 
 func (c *batchCommandsClient) available() int64 {
-	return c.maxConcurrencyRequestLimit.Load() - c.sent.Load()
+	limit := c.maxConcurrencyRequestLimit.Load()
+	sent := c.sent.Load()
+	// just in case, `sent` may less than 0, see https://github.com/tikv/client-go/issues/1225
+	// then maxInt64 - (-1) will overflow, which is unexpected.
+	if sent > 0 {
+		return limit - sent
+	}
+	return limit
 }
 
 func (c *batchCommandsClient) send(forwardedHost string, req *tikvpb.BatchCommandsRequest) {
--- a/internal/client/client_batch.go
+++ b/internal/client/client_batch.go
@@ -569,8 +569,7 @@ func (c *batchCommandsClient) isStopped() bool {
 func (c *batchCommandsClient) available() int64 {
 	limit := c.maxConcurrencyRequestLimit.Load()
 	sent := c.sent.Load()
-	// just in case, `sent` may less than 0, see https://github.com/tikv/client-go/issues/1225
-	// then maxInt64 - (-1) will overflow, which is unexpected.
+	//  The `sent` could be less than 0, see https://github.com/tikv/client-go/issues/1225 for details.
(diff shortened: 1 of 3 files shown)


internal/client/client_batch.go
```diff
@@ -567,7 +567,14 @@ func (c *batchCommandsClient) isStopped() bool {
 }
 
 func (c *batchCommandsClient) available() int64 {
-	return c.maxConcurrencyRequestLimit.Load() - c.sent.Load()
+	limit := c.maxConcurrencyRequestLimit.Load()
+	sent := c.sent.Load()
+	// just in case, `sent` may less than 0, see https://github.com/tikv/client-go/issues/1225
```
cfzjywxk, line 572:
```suggestion
	//  The `sent` could be less than 0, see https://github.com/tikv/client-go/issues/1225 for details.
```

internal/client/client_batch.go
```diff
@@ -567,7 +567,14 @@ func (c *batchCommandsClient) isStopped() bool {
...
+	// then maxInt64 - (-1) will overflow, which is unexpected.
+	if sent > 0 {
+		return limit - sent
```
cfzjywxk, line 575: Is it acceptable to return a negative result?

ekexium:
I didn't get it very well. Why are we fixing it by checking `sent > 0`? Seems to me it should be fixed by making sure `sent` is always non-negative?

crazycs520:
> I didn't get it very well. Why are we fixing it by checking `sent > 0`? Seems to me it should be fixed by making sure `sent` is always non-negative?

`sent` may be negative is cause by data race in issue #1222 , and I also have PR #1219 to do this fix, but I need to to more test to avoid regression as mentioned in https://github.com/tikv/client-go/pull/1219#discussion_r1523342977 , so I try to fix this bug without fix data race in this PR, since  code freezes.

internal/client/client_batch.go
crazycs520: It's ok here, negative result means not available.

crazycs520: Nice cache, it's better not return negative result here.

ekexium (approved):

MyonKeminta (approved):

PR #1229, opened 2024-03-14
enable replica-selector-v2 by default

crazycs520:
Add `TiKVClient .EnableReplicaSelectorV2` config and enable by default.

`TiKVClient.EnableReplicaSelectorV2` config should be removed after the new replica-selector-v2 is stable.

```toml
[tikv-client]
enable-replica-selector-v2 = true
```

### DIFF
--- a/config/client.go
+++ b/config/client.go
@@ -94,6 +94,7 @@ type TiKVClient struct {
 	// 0 means auto adjust by feedback.
 	MaxConcurrencyRequestLimit int64 `toml:"max-concurrency-request-limit" json:"max-concurrency-request-limit"`
 	// EnableReplicaSelectorV2 indicate whether to use the new replica-selector-v2.
+	// TODO(crazycs520): remove this config after the new replica-selector-v2 is stable.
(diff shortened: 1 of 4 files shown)


joey-yez (approved):

config/client.go
```diff
@@ -93,6 +93,9 @@ type TiKVClient struct {
 	// MaxConcurrencyRequestLimit is the max concurrency number of request to be sent the tikv
 	// 0 means auto adjust by feedback.
 	MaxConcurrencyRequestLimit int64 `toml:"max-concurrency-request-limit" json:"max-concurrency-request-limit"`
+	// EnableReplicaSelectorV2 indicate whether to use the new replica-selector-v2.
+	// TODO(crazycs520): remove this config after the new replica-selector-v2 is stable.
```
MyonKeminta, line 97: This TODO comment is removed. Do we still plan to remove this configuration when v2 is stable?

MyonKeminta (approved):
LGTM

PR #1230, opened 2024-03-15
txn: implement staging for PipelinedMemDB

ekexium:
Ref tikv/tikv#16291

Implement the staging interface for pipelined DML. Its correctness is guaranteed by that no stage can exist when a Flush() is called. Every stage lives inside a flush batch, which means the modifications all goes to the mutable memdb. Thus the staging of the whole PipeliendMemDB can be directly implemented by the memdb.


internal/unionstore/pipelined_memdb.go
```diff
@@ -375,17 +380,17 @@ func (p *PipelinedMemDB) SnapshotGetter() Getter {
 
 // Staging is not supported for PipelinedMemDB, it returns 0 handle.
```
cfzjywxk, line 381: The comments should be updated, better to include the correctness explanation as well.

cfzjywxk (approved):

you06 (approved):

PR #1235, opened 2024-03-18
implement rollback for pipelined dml

you06:
ref https://github.com/tikv/tikv/issues/16291

Now the locks of rollback pipelined DML need to be cleaned up by GC. With this PR, the locks can be cleanup faster.

### Flush phase

image

### Rollback phase

image

### SQL client

image


txnkv/transaction/pipelined_flush.go
```diff
@@ -392,22 +410,25 @@ func (c *twoPhaseCommitter) buildPipelinedResolveHandler(commit bool, resolved *
 	}, nil
 }
 
-func (c *twoPhaseCommitter) resolveFlushedLocks(bo *retry.Backoffer, start, end []byte) {
-	// TODO: implement cleanup.
+// resolveFlushedLocks resolves all locks in the given range [start, end) with the given status.
+// The resolve process is running in another goroutine so this function won't block.
+func (c *twoPhaseCommitter) resolveFlushedLocks(bo *retry.Backoffer, start, end []byte, commit bool) {
 	const RESOLVE_CONCURRENCY = 8
 	var resolved atomic.Uint64
```
you06, line 417: I'm not sure if we need to choose lower concurrency for rollback.

txnkv/transaction/pipelined_flush.go
```diff
@@ -334,16 +334,34 @@ func (c *twoPhaseCommitter) buildPipelinedResolveHandler(commit bool, resolved *
 		maxBackOff = CommitSecondaryMaxBackoff
 	}
 	regionCache := c.store.GetRegionCache()
+	// the handler function runs in a different goroutine, should copy the required values before it to avoid race.
+	var (
+		startTS           = c.startTS
```
ekexium, line 339: It seems they are mostly read-only after initialization. Can they cause data race

txnkv/transaction/pipelined_flush.go
```diff
@@ -392,22 +410,25 @@ func (c *twoPhaseCommitter) buildPipelinedResolveHandler(commit bool, resolved *
...
-		logutil.Logger(bo.GetCtx()).Error("[pipelined dml] commit transaction secondaries failed",
-			zap.Uint64("resolved regions", resolved.Load()),
-			zap.Error(err))
-	} else {
-		logutil.BgLogger().Info("[pipelined dml] commit transaction secondaries done",
-			zap.Uint64("resolved regions", resolved.Load()))
-	}
+	go func() {
+		if err = runner.RunOnRange(bo.GetCtx(), start, end); err != nil {
+			logutil.Logger(bo.GetCtx()).Error("[pipelined dml] commit transaction secondaries failed",
```
ekexium, line 426: Shall we change the log message as well

txnkv/transaction/pipelined_flush.go
you06: Actually only `c.txn.diskFullOpt` can cause data race now. But there are no guarantees for those read-only fields, it's a potential risk reading it in another goroutine.

What about init the context here and clone it by `proto.Clone(kvContext).(*kvrpcpb.Context)` in the resolve goroutine?

txnkv/transaction/pipelined_flush.go
you06: Nice catch.

txnkv/transaction/pipelined_flush.go
```diff
@@ -334,6 +335,18 @@ func (c *twoPhaseCommitter) buildPipelinedResolveHandler(commit bool, resolved *
 		maxBackOff = CommitSecondaryMaxBackoff
 	}
 	regionCache := c.store.GetRegionCache()
+	// the handler function runs in a different goroutine, should copy the required values before it to avoid race.
```
cfzjywxk, line 338: As dicussed in above comments, this may not be a perfect solution, we still need to find a way to avoid race completely.

cfzjywxk (approved):

txnkv/transaction/pipelined_flush.go
ekexium: Such guarantees can hardly be achieved in Golang ... I guess we usually ignore them until race is detected.
It's okay we can do this. It just seems a bit strange.

txnkv/transaction/pipelined_flush.go
you06: Maybe set `txn.committer` to `nil` which disable further accessing to rollbacked committer.

PR #1237, opened 2024-03-19
Upgrade PD client to fix the TSO hang bug

JmPotato:
Ref https://github.com/tikv/pd/issues/7849.


CabinfeverB:
Need to do some compatibility modification, like https://github.com/tikv/client-go/pull/1217

iosmanthus (approved):

PR #1240, opened 2024-03-19
txn: refactor set option to detect race risk

you06:
As [discussion](https://github.com/tikv/client-go/pull/1235#discussion_r1528127108) in #1235, we need some strategy to avoid data race in committer.

The fields currently used by the committer are lacking protection.


cfzjywxk:
@you06 Is this to be closed as the refactored would be done at tidb side?

PR #1248, opened 2024-03-26
cop: add kv read wall time to print result

cfzjywxk:
The kv read wall time which is the time consumed in tikv executors is not printted in the result. 
It is helpful to check time consumed in the tikv next operations, for example processing column with larger values.

### DIFF
--- a/util/execdetails.go
+++ b/util/execdetails.go
@@ -653,6 +653,13 @@ func (td *TimeDetail) String() string {
 		buf.WriteString("total_wait_time: ")
 		buf.WriteString(FormatDuration(td.WaitTime))
 	}
+	if td.KvReadWallTime > 0 {
+		if buf.Len() > 0 {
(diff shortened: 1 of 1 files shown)


util/execdetails.go
```diff
@@ -663,6 +670,17 @@ func (td *TimeDetail) String() string {
 	return buf.String()
 }
 
+// Merge merges the time detail into itself.
+func (td *TimeDetail) Merge(detail *TimeDetail) {
```
ekexium, line 674: Where is this used?

cfzjywxk: To be used here https://github.com/pingcap/tidb/blob/master/pkg/distsql/select_result.go#L531 to record the `TimeDetail` information like the `ScanDetail`.

zyguan (approved):

PR #1250, opened 2024-03-26
tikv: ensure safe-ts won't be max uint64

### LINKED ISSUE #1249: safe-ts is set to max uint64 after rolling restart tikv
After #793 , we try to get safe ts from pd http API. However, the semantics of the API returns `MaxUint64` when the safe ts is uninitalized, which is not same as previous `GetStoreSafeTS` RPC (of tikv). As a result, safe ts will be set to max uint64 after rolling restart tikv, which further leads to a large amount of stale read miss, because the calculated read ts will be now (the upper bound) in this case.

zyguan:
Try to fix #1249

with this PR, no stale read miss after rolling restart tikv:

without this fix:

### DIFF
--- a/tikv/kv_test.go
+++ b/tikv/kv_test.go
@@ -136,16 +136,6 @@ func (c *mockPDHTTPClient) GetMinResolvedTSByStoresIDs(ctx context.Context, stor
 	return c.Client.GetMinResolvedTSByStoresIDs(ctx, storeIDs)
 }
 
-func until(limit int, interval time.Duration, predicate func() bool) bool {
-	for i := 0; i < limit; i++ {
-		time.Sleep(interval)
-		if predicate() {
-			return true
-		}
-	}
-	return false
-}
-
 func (s *testKVSuite) TestMinSafeTsFromStores() {
 	mockClient := storeSafeTsMockClient{
 		Client:    s.store.GetTiKVClient(),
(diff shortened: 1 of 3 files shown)


cfzjywxk (approved):

tikv/kv_test.go
```diff
@@ -105,22 +124,127 @@ func (c *storeSafeTsMockClient) CloseAddr(addr string) error {
...
+}
+
+func (c *mockPDHTTPClient) GetMinResolvedTSByStoresIDs(ctx context.Context, storeIDs []uint64) (uint64, map[uint64]uint64, error) {
+	if f := c.mockGetMinResolvedTSByStoresIDs.Load(); f != nil {
+		return (*f)(ctx, storeIDs)
+	}
+	return c.Client.GetMinResolvedTSByStoresIDs(ctx, storeIDs)
+}
+
+func until(limit int, interval time.Duration, predicate func() bool) bool {
```
you06, line 139: Can we use `s.Eventually`?

tikv/kv_test.go
```diff
@@ -105,22 +124,127 @@ func (c *storeSafeTsMockClient) CloseAddr(addr string) error {
...
+			} else {
+				m[id] = math.MaxUint64
+			}
+		}
+		return math.MaxUint64, m, nil
+	})
+	until(10, time.Second, func() bool {
+		ts := s.store.GetMinSafeTS("z1")
+		s.Require().False(math.MaxUint64 == ts)
+		return ts == uint64(10)
```
you06, line 244: This `until` call will timeout and the predicate never returns true.

ekexium: Should we check the return value of `until`?

zyguan: Sorry, it should GetMinSafeTS from z2.

you06 (approved):

PR #1262, opened 2024-04-03
region_cache: filter out unresolved stores when GetTiFlashStores

zyguan:
Try to fix https://github.com/pingcap/tidb/issues/46602 .

`GetTiFlashStores` would return unresolved stores, which causes mpp probe keep checking tomebstone stores. `RegionCache` exports some methods for listing stores, like `GetAllStores`. There should/better be a contract that those methods do not return unresolved stores. Note that `GetTiFlashComputeStores` may still returns unresolved stores, however, the entire store list will be reloaded once invalidate method get called, that is, it's users' responsibility to reload tiflash compute stores.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -2191,7 +2191,8 @@ func (c *RegionCache) PDClient() pd.Client {
 	return c.pdClient
 }
 
-// GetTiFlashStores returns the information of all tiflash nodes.
+// GetTiFlashStores returns the information of all tiflash nodes. Like `GetAllStores`, the method only returns resolved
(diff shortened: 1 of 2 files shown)


internal/locate/region_cache.go
```diff
@@ -2194,7 +2194,7 @@ func (c *RegionCache) PDClient() pd.Client {
 // GetTiFlashStores returns the information of all tiflash nodes.
```
cfzjywxk, line 2194: The comments should be updated too.

cfzjywxk (approved):

you06 (approved):

PR #1267, opened 2024-04-07
Make slow store filtering the highest priority in replica selector v2

MyonKeminta:
Make slow store filtering the highest priority in replica selector v2.

With some minor changes:

* Add logs about health state change
* Disable health status tick of region cache in replica_selector_test
* Fix slow score resetting in test code
* Special type `storeSelectionScore` and `String` method which might help in logging
* A test about multiple replica in the same AZ, testing priority between label matching and retrying leader
* Minor renaming, typo fixing and comment fixing

There are some behavior inconsistencies found between v1 and v2 and maybe we need some discussion. I want to finish this PR within April 8 if possible so that this can be included in release-8.1 branch.

The inconsistencies are all about replica read mode without stale read and labels are specified, in which cases this branch in `calculateScore` is executed:

https://github.com/MyonKeminta/client-go/blob/05eec3f6f7a5e49b7aff92862716253afa882da5/internal/locate/replica_selector.go#L367-L370

And what it causes are:

1. In replica read mode without stale read, 1 store out of 3 has matching labels, and the first request to it failed. v2 retries on leader, but v1 may retry on any replica.
3. In replica read mode without stale read, 2 stores (1 of which is the leader) out of 6 have matching labels, then v2 will always send the first request to the leader, while v1 won't.
4. In replica read mode without stale read, 2 stores (both follower) out of 6 have matching labels, and requests to them both failed. Then the next attempt will always be sent to the leader in v2, while v1 won't.


internal/locate/replica_selector_test.go
```diff
@@ -2507,6 +2612,152 @@ func TestReplicaReadAccessPathByLearnerCase(t *testing.T) {
...
+					respRegionError: nil,
+					backoffCnt:      0,
+					backoffDetail:   []string{},
+					regionIsValid:   true,
+				},
+			}
+			s.True(s.runCaseAndCompare(ca))
+
+			s.T().Logf("test case: stale read: %v, with label: %v, slow: true", staleRead, withLabel)
+			expectedFirstStore := 2
```
cfzjywxk, line 2655: Could the `expectedFirstStore` also be `3` if labels are not set and store 1 is slow with `kv.ReplicaReadMixed`?

zyguan (approved):
The PR LGTM.

I'm wondering whether one single strategy can meet all user cases: what if some users are latency insensitive but very care about cross AZ traffic, or RTT is known to be unacceptable when cross AZ. Maybe we need to rethink the UI of replica selector (`tidb_replica_read`) later, how to make it reasonable and user-friendly.

MyonKeminta:
@crazycs520 @cfzjywxk @zyguan I'm considering removing `TestMultiReplicaInOneAZ` from this PR (which is irrelevant to the title of this PR) so that this PR can be merged within today hopefully. The problem revealed in that test case will be another topic.

cfzjywxk (approved):

PR #908, opened 2023-07-25
use context logger as much as possible

crazycs520:
Use `Logger(ctx)` instead of `BgLogger ` as much as possible.

Since `Logger(ctx)` maybe contain other information such as connection id, which is good for debug.

Here are some log before this PR:

```log
[2023/07/25 16:02:19.812 +08:00] [WARN] [region_request.go:612] ["unable to find stores with given labels"] [labels="[{\"key\":\"zone\",\"value\":\"z99\"}]"]
[2023/07/25 16:01:34.877 +08:00] [WARN] [region_request.go:1942] ["tikv reports `DataIsNotReady` retry later"] [store-id=4] [peer-id=8008] [region-id=1011] [safe-ts=443094687287869442] [ctx="region ID: 1011, meta: id:1011 start_key:\"t\\200\\000\\000\\000\\000\\000\\000`_r\\00100000000\\37700000000\\37700000000\\377051486\\000\\000\\375\\0010000-000\\3771\\000\\000\\000\\000\\000\\000\\000\\370\\003\\227s\\332eMU\\256\\350\" end_key:\"x\\000\\000\\000\" region_epoch:<conf_ver:77 version:63 > peers:<id:8006 store_id:1 > peers:<id:8007 store_id:14 > peers:<id:8008 store_id:4 > , peer: id:8008 store_id:4 , addr: 10.0.1.9:22160, idx: 0, reqStoreType: TiKvOnly, runStoreType: tikv"]
[...]

PR #1270, opened 2024-04-07
Allow setting session id of KVTxn

ekexium:
Self-explained by the comment.

### DIFF
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -397,8 +397,14 @@ func (txn *KVTxn) SetTxnSource(txnSource uint64) {
 
 // SetSessionID sets the session ID of the transaction.
 // If the committer is not initialized yet, the function won't take effect.
+// It is supposed to be set before performing any writes in the transaction to avoid data race.
+// It is designed to be called in ActivateTxn(), though subject to change.
 // It is especially useful for pipelined transactions, as its committer is initialized immediately
 // when the transaction is created.
+//
+// Note that commiter may also obtain a sessionID from context directly via sessionIDCtxKey.
+// TODO: consider unifying them.
+
 func (txn *KVTxn) SetSessionID(sessionID uint64) {
 	if txn.committer != nil {
 		txn.committer.sessionID = sessionID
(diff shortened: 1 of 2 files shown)


txnkv/transaction/txn.go
```diff
@@ -395,6 +395,16 @@ func (txn *KVTxn) SetTxnSource(txnSource uint64) {
 	txn.txnSource = txnSource
 }
 
+// SetSessionID sets the session ID of the transaction.
+// If the committer is not initialized yet, the function won't take effect.
+// It is especially useful for pipelined transactions, as its committer is initialized immediately
+// when the transaction is created.
```
cfzjywxk, line 401: Should also explain when it is supposed to be executed at tidb-side, is it `OnTxnActiviated`?

ekexium: > is it OnTxnActiviated?

Yes.
But as a lower level module I think it already provides a clear enough interface. It doesn't have to know (or restrict) how it will be used, which should be decided by the user.

cfzjywxk: But the `SessionID` is actually a tidb concept, and it's set in a tricky way using context from the upper layer.

ekexium: Yeah it's a pass-through variable, always annoying to deal with.

cfzjywxk (approved):

zyguan (approved):

txnkv/transaction/txn.go
```diff
@@ -395,6 +395,22 @@ func (txn *KVTxn) SetTxnSource(txnSource uint64) {
 	txn.txnSource = txnSource
 }
 
+// SetSessionID sets the session ID of the transaction.
+// If the committer is not initialized yet, the function won't take effect.
```
MyonKeminta, line 399: For the concern of completeness of the API, do we consider put it to another field temporarily if the committer is not initialized yet, and use it lazily when initializing the committer?

MyonKeminta (approved):

crazycs520:
https://github.com/tikv/client-go/pull/908  Can this PR help?

ekexium:
> #908 Can this PR help?

Yes. I tried to use ctx logger when possible. But it's not always feasible so SetSessionID is still needed for those cases.

crazycs520 (approved):

txnkv/transaction/txn.go
ekexium: It's worth consideration. But I don't want to do it now as it's not needed

PR #1271, opened 2024-04-08
Distinguish identifier and name in task runner

ekexium:
`name` is used in both metrics and logs. We separate name and identifier to avoid messy metrics and pertain useful logs.

### DIFF
--- a/txnkv/rangetask/range_task.go
+++ b/txnkv/rangetask/range_task.go
@@ -93,6 +93,16 @@ type TaskHandler = func(ctx context.Context, r kv.KeyRange) (TaskStat, error)
 // will be canceled.
 func NewRangeTaskRunner(
 	name string,
+	store storage,
+	concurrency int,
+	handler TaskHandler,
+) *Runner {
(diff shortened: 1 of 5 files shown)


txnkv/rangetask/range_task.go
```diff
@@ -90,12 +93,14 @@ type TaskHandler = func(ctx context.Context, r kv.KeyRange) (TaskStat, error)
...
 func NewRangeTaskRunner(
 	name string,
+	identifier string,
 	store storage,
 	concurrency int,
 	handler TaskHandler,
 ) *Runner {
 	return &Runner{
 		name:            name,
+		identifier:      identifier,
```
you06, line 103: What about `if identifier == "" { identifier = name }`?

you06 (commented):
Need to fix other `NewRangeTaskRunner` usages, rest LGTM.

you06 (approved):

cfzjywxk (approved):

PR #1278, opened 2024-04-09
txn: cache result of snapshot.Get

you06:
Now snapshot cache can only be generated by `snapshot.BatchGet` which doesn't optimize the case get the same key multi times in a txn.

This PR cache the result of `snapshot.Get` to optimize this case.

### Before this PR

image

### With this PR

image

## Benchmark

Test with sysbench point select, but wrap it in txn:

```
begin;
point select with same id n times;
commit;
```

|n|Before thisPR(txn/s)|With this PR(txn/s)|
|-|-|-|
|2|59029|68541|
|3|43032|57330|
|4|34583|49275|

image

### DIFF
--- a/txnkv/txnsnapshot/snapshot.go
+++ b/txnkv/txnsnapshot/snapshot.go
@@ -192,35 +192,6 @@ func (s *KVSnapshot) SetSnapshotTS(ts uint64) {
 	s.resolvedLocks = util.TSSet{}
 }
 
-// UpdateSnapshotCache sets the values of cache, for further fast read with same keys.
-func (s *KVSnapshot) UpdateSnapshotCache(keys [][]byte, m map[string][]byte) {
-	// Update the cache.
-	s.mu.Lock()
-	if s.mu.cached == nil {
-		s.mu.cached = make(map[string][]byte, min(len(keys), 8))
-	}
-	for _, key := range keys {
-		val := m[string(key)]
-		s.mu.cachedSize += len(key) + len(val)
-		s.mu.cached[string(key)] = val
-	}
-
-	const cachedSizeLimit = 10 << 30
-	if s.mu.cachedSize >= cachedSizeLimit {
-		for k, v := range s.mu.cached {
-			if _, needed := m[k]; needed {
-				continue
-			}
-			delete(s.mu.cached, k)
-			s.mu.cachedSize -= len(k) + len(v)
-			if s.mu.cachedSize < cachedSizeLimit {
-				break
-			}
-		}
-	}
-	s.mu.Unlock()
-}
-
 // IsInternal returns if the KvSnapshot is used by internal executions.
 func (s *KVSnapshot) IsInternal() bool {
 	return util.IsRequestSourceInternal(s.RequestSource)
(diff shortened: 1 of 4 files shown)


cfzjywxk:
@you06 
We could do some simple benchmark tests for this optimization, like
begin;
point get
point get
commit

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1025,6 +1005,43 @@ func (s *KVSnapshot) SnapCache() map[string][]byte {
...
+// UpdateSnapshotCache sets the values of cache, for further fast read with same keys.
+func (s *KVSnapshot) UpdateSnapshotCache(keys [][]byte, m map[string][]byte) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.mu.cached == nil {
+		s.mu.cached = make(map[string][]byte, min(len(keys), 8))
+	}
+	for _, key := range keys {
+		val := m[string(key)]
+		s.mu.cachedSize += len(key) + len(val)
```
ekexium, line 1017: If the cache already contains the key, should it minus the original value length here?

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1025,6 +1005,43 @@ func (s *KVSnapshot) SnapCache() map[string][]byte {
...
+			}
+		}
+	}
+}
+
+// CleanCache cleans the cache for given keys. Only for test.
+func (s *KVSnapshot) CleanCache(keys [][]byte) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for _, key := range keys {
```
ekexium, line 1040: Let's also reset `cachedSize` here.

you06:
> do some simple benchmark tests

Benchmark result added.

txnkv/txnsnapshot/snapshot.go
```diff
@@ -623,12 +600,29 @@ func (s *KVSnapshot) Get(ctx context.Context, k []byte) ([]byte, error) {
...
+	// Check the cached values first.
+	if s.mu.cached != nil {
+		if value, ok := s.mu.cached[string(k)]; ok {
+			atomic.AddInt64(&s.mu.hitCnt, 1)
+			s.mu.RUnlock()
+			if len(value) == 0 {
+				return nil, tikverr.ErrNotExist
+			}
+			return value, nil
+		}
```
cfzjywxk, line 613: The `s.mu` is not released in the else branch.

txnkv/txnsnapshot/snapshot.go
```diff
@@ -645,37 +639,23 @@ func (s *KVSnapshot) Get(ctx context.Context, k []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	// Update the cache.
+	s.UpdateSnapshotCache([][]byte{k}, map[string][]byte{string(k): val})
```
cfzjywxk, line 643: Would the `string(k)` invovle an extra allocation in current golang?

txnkv/txnsnapshot/snapshot.go
```diff
@@ -645,37 +639,23 @@ func (s *KVSnapshot) Get(ctx context.Context, k []byte) ([]byte, error) {
...
-	if _, err := util.EvalFailpoint("snapshot-get-cache-fail"); err == nil {
-		if bo.GetCtx().Value("TestSnapshotCache") != nil {
-			s.mu.RUnlock()
-			panic("cache miss")
-		}
-	}
 
 	cli := NewClientHelper(s.store, &s.resolvedLocks, &s.committedLocks, true)
-
+	s.mu.RLock()
```
cfzjywxk, line 658: The `Unlock` is not called, is it a mistake editing?

txnkv/txnsnapshot/snapshot.go
you06: It's released in [L632](https://github.com/tikv/client-go/blob/90f9b060de894153ace9e6c16be1d79efe6d7c57/txnkv/txnsnapshot/snapshot.go#L632)

txnkv/txnsnapshot/snapshot.go
you06: Maybe we need to use [unsafe cast](https://github.com/pingcap/tidb/blob/69d7770335a7611d05bb2f3a0888c79afb03e92a/pkg/util/hack/hack.go#L29-L34) like tidb.

cfzjywxk (approved):

PR #1279, opened 2024-04-09
region_cache: extract store related fields to store cache

zyguan:
ref https://github.com/tikv/client-go/issues/1104

The rest part for extracting store management from region cache


zyguan:
@cfzjywxk @MyonKeminta PTAL

crazycs520 (approved):

PR #1284, opened 2024-04-10
only set MinSafeTS to 0 when all stores' are 0

zyguan:
https://github.com/tikv/client-go/pull/1277 introduce a regression: there might be a store whose safe ts is 0 for a long time, in this case, flashback will be blocked (ref: https://github.com/tikv/tikv/issues/13675). We should `setMinSafeTS(0)` only when all (NOT any) of `store.safe_ts = 0`.

### DIFF
--- a/internal/logutil/log.go
+++ b/internal/logutil/log.go
@@ -36,6 +36,7 @@ package logutil
 
 import (
 	"context"
+	"testing"
 
 	"github.com/pingcap/log"
 	"go.uber.org/zap"
@@ -60,3 +61,11 @@ type ctxLogKeyType struct{}
 // CtxLogKey is the key to retrieve logger from context.
 // It can be assigned to another value.
 var CtxLogKey interface{} = ctxLogKeyType{}
+
+// AssertWarn panics when in testing mode, and logs a warning msg otherwise.
+func AssertWarn(logger *zap.Logger, msg string, fields ...zap.Field) {
+	if testing.Testing() {
+		logger.Panic(msg, fields...)
+	}
+	logger.Warn(msg, fields...)
+}
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -44,7 +44,6 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
-	"testing"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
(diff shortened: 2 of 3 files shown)


tikv/kv.go
```diff
@@ -646,13 +637,17 @@ func (s *KVStore) updateMinSafeTS(txnScope string, storeIDs []uint64) {
...
+			if safeTS != 0 && safeTS < minSafeTS {
 				minSafeTS = safeTS
 			}
 		} else {
 			minSafeTS = 0
 		}
 	}
+	// if minSafeTS is still math.MaxUint64, that means all store safe ts are 0, then we set minSafeTS to 0.
+	if minSafeTS == math.MaxUint64 {
+		minSafeTS = 0
```
cfzjywxk, line 649: Or could we set it to something like `safepoint` ts ?

zyguan: Sound reasonable, let me think about if it would break some constraints.

Defined2014 (approved):

internal/logutil/log.go
```diff
@@ -60,3 +61,11 @@ type ctxLogKeyType struct{}
 // CtxLogKey is the key to retrieve logger from context.
 // It can be assigned to another value.
 var CtxLogKey interface{} = ctxLogKeyType{}
+
+// AssertWarn panics when in testing mode, and logs a warning msg otherwise.
+func AssertWarn(logger *zap.Logger, msg string, fields ...zap.Field) {
+	if testing.Testing() {
```
you06, line 67: This condition is never "true" when running with bazel.
It's ok if it's used in client-go's test only.

zyguan: Interesting, I tired to compile the following code by `go test -c ...` and run the binary, it also showed true.
```go
func TestDemo(t *testing.T) {
	fmt.Println(">>> in test?", testing.Testing())
}
```

Although the function is in internal package, I want it works in tidb ut. I'll figure out how bazel build/run test binaries.

you06: FYI the only work way I find is TiDB pass the `intest.InTest` to client-go via an interface, not zero cost, and greatly impact the the client's interface.
If you use intest flag condition build in client-go, bazel compile will fail...

zyguan: https://github.com/bazelbuild/rules_go/issues/3686 I realize the issue has be reported by @tiancaiamao . Maybe just let it only works in client-go ut by now.

you06: LGTM

cfzjywxk (approved):

PR #1286, opened 2024-04-11
Use Mem() instead of Size() to estimate pipelined-memdb size

ekexium:
ref tikv/tikv#16291

By default, a pipelined-txn can use more than 1GB memory, most of which are consumed by memdb. 
Use `Mem()` as an estimate to more accurately control the total memory usage.

### DIFF
--- a/internal/unionstore/pipelined_memdb.go
+++ b/internal/unionstore/pipelined_memdb.go
@@ -69,9 +69,9 @@ const (
 )
 
 type flushOption struct {
-	MinFlushKeys            int
-	MinFlushSize            int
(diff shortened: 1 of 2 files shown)


ekexium:
In my local test, I've seen max usage of 768MB with this PR. Given the default session quota is 1GB, shall we consider reducing the default (and fixed) threshold from 128MB to 64MB?

cfzjywxk (approved):
LGTM.

cfzjywxk:
I think it's find to increase the `session quota` if it's not enough running large transactions.

you06 (approved):

PR #1287, opened 2024-04-11
Support actively requesting update health feedback information by calling RPC to TiKV

MyonKeminta:
Tests will be added later. Let's see if this approach is acceptable now...

It seems possible to use the [`BatchCommandsEmptyRequest`](https://github.com/pingcap/kvproto/blob/c699538f7aa18394ae5e09cd1291209079b391d6/proto/tikvpb.proto#L187) to do this, but as it's designed for test purpose, and there are some special handlings that I don't understand in client-go (such as `CallRPC` function in client-go directly returns an empty response for it without accessing TiKV), I think a better idea would be adding a new API, so I added a ~~`RequestHealthFeedback`~~ RPC call to TiKV. A drawback of this solution is that this might make it harder to include this to branch release-8.1...

**Update**: The new RPC request is finally named `GetHealthFeedback` and is now a complete API that supports to be used in both batched & non-batched mode. And it has a special behavior, which is how it's being made use of in current implementation: If it's called in batched RPC mode, the response batch (`BatchCommandsResponse`) will also carry the feedback info.

Ref: https://github.com/tikv/tikv/issues/16297

Requires:
* https://github.com/pingcap/kvproto/pull/1239
* https://github.com/tikv/tikv/pull/17008


MyonKeminta:
cc @crazycs520 PTAL
The integration test may fail due to TiKV image not updated to latest version. It should work tomorrow without building manually.

tikv/kv.go
```diff
@@ -235,14 +235,38 @@ func loadOption(store *KVStore, opt ...Option) {
...
+	// carried in the `resp` here.
+	resp, err := tikvClient.SendRequest(ctx, addr, tikvrpc.NewRequest(tikvrpc.CmdGetHealthFeedback, &kvrpcpb.GetHealthFeedbackRequest{}), getHealthFeedbackTimeout)
+	if err != nil {
+		return err
+	}
+	regionErr, err := resp.GetRegionError()
+	if err != nil {
+		return err
+	}
+	if regionErr != nil {
```
cfzjywxk, line 254: Would `region error` be returned in the current feedback implementation, and should the request be retried?

MyonKeminta: Currently TiKV won't return region error here. Whether it need to be retried can be left to be considered in the future if someday we make TiKV possible to return the region error.

cfzjywxk (approved):
LGTM

MyonKeminta:
@crazycs520 @zyguan PTAL

zyguan (approved):

PR #1291, opened 2024-04-12
log: log the resolve progress of pipelined-dml every 30s

ekexium:


txnkv/transaction/pipelined_flush.go
```diff
@@ -453,6 +453,7 @@ func (c *twoPhaseCommitter) resolveFlushedLocks(bo *retry.Backoffer, start, end
 		RESOLVE_CONCURRENCY,
 		handler,
 	)
+	runner.SetStatLogInterval(30 * time.Second)
```
cfzjywxk, line 456: It would also be helpful to log this information in `show processlist` like statements.

cfzjywxk (approved):
Please also add ref issue or descriptions

txnkv/transaction/pipelined_flush.go
ekexium: This is usually async so it is not supposed to be found in `show processlist`?

cfzjywxk: If the result is already returned it would not be displayed in `show processlist`, we need a place to view the progress of these async tasks, maybe some internal system tables like `show xxx`?

ekexium: Yeah that's a good idea. We may have a new syntax `show commit jobs` for such purposes.

PR #1295, opened 2024-04-15
log: only log the first lock in every flush or prewrite response

### LINKED ISSUE #595: `prewrite encounters lock` log is noisy
```
root@cloud-ecosystem-00:~/cluster/cloud/sys-test-16/tidb-0# wc -l tidb-2022-10-10T23-40-54.065.log
640243 tidb-2022-10-10T23-40-54.065.log
root@cloud-ecosystem-00:~/cluster/cloud/sys-test-16/tidb-0# grep 'prewrite encounters lock' tidb-2022-10-10T23-40-54.065.log | wc -l
548084
```

ekexium:
close #595 
1 lock is enough for most of the cases.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -1219,7 +1219,9 @@ func keepAlive(
 				logutil.Logger(bo.GetCtx()).Info("ttlManager live up to its lifetime",
 					zap.Uint64("txnStartTS", c.startTS),
 					zap.Uint64("uptime", uptime),
-					zap.Uint64("maxTxnTTL", config.GetGlobalConfig().MaxTxnTTL))
+					zap.Uint64("maxTxnTTL", config.GetGlobalConfig().MaxTxnTTL),
+					zap.Bool("isPipelinedTxn", isPipelinedTxn),
+				)
 				metrics.TiKVTTLLifeTimeReachCounter.Inc()
 				// the pessimistic locks may expire if the ttl manager has timed out, set `LockExpired` flag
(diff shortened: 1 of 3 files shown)


cfzjywxk (approved):
Actually, this does help us in some corner case investigations before.
Currently, without persisted lock view it seems to be difficult to analyze conflict problems.
/cc @MyonKeminta

ekexium:
> Actually, this does help us in some corner case investigations before. Currently, without persisted lock view it seems to be difficult to analyze conflict problems. /cc @MyonKeminta

There may be other solutions to that balance log verbosity and information-richness. One worth consideration might be "for each conflicting ts, only log 1 key". How about that

you06 (approved):

cfzjywxk:
> > Actually, this does help us in some corner case investigations before. Currently, without persisted lock view it seems to be difficult to analyze conflict problems. /cc @MyonKeminta
> 
> There may be other solutions that balance log verbosity and information-richness. One worth consideration might be "for each conflicting ts, only log 1 key". How about that?

It looks fine.

cfzjywxk:
> > > Actually, this does help us in some corner case investigations before. Currently, without persisted lock view it seems to be difficult to analyze conflict problems. /cc @MyonKeminta
> > 
> >

> > There may be other solutions that balance log verbosity and information-richness. One worth consideration might be "for each conflicting ts, only log 1 key". How about that?
> 
> 


It looks fine.

cfzjywxk (approved):

PR #1296, opened 2024-04-16
feat: stop flush when first error is met, like prewrite

ekexium:
ref tikv/tikv#16291
If a flush request is known to fail, don't continue sending other parts.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -2049,16 +2049,18 @@ func (batchExe *batchExecutor) process(batches []batchMutations) error {
 		return err
 	}
 
-	// For prewrite, stop sending other requests after receiving first error.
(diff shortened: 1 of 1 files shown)


txnkv/transaction/2pc.go
```diff
@@ -2049,16 +2049,18 @@ func (batchExe *batchExecutor) process(batches []batchMutations) error {
 		return err
 	}
 
-	// For prewrite, stop sending other requests after receiving first error.
+	// For prewrite and flush, stop sending other requests after receiving first error.
```
cfzjywxk, line 2052: If there is such case for `prewrite`, we could add one for `flush` too.

ekexium: Yes. I've checked all type assertions. Seems the only thing left that make them "unequal" is stats.

cfzjywxk (approved):

PR #1298, opened 2024-04-16
reverts PR  #988 to fix transaction is undetermined cause by region invalid error

crazycs520:
This reverts PR  #988

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1015,14 +1015,9 @@ func newReplicaSelector(
 	regionCache *RegionCache, regionID RegionVerID, req *tikvrpc.Request, opts ...StoreSelectorOption,
 ) (*replicaSelector, error) {
 	cachedRegion := regionCache.GetCachedRegionWithRLock(regionID)
-	if cachedRegion == nil {
-		return nil, errors.New("cached region not found")
-	} else if cachedRegion.checkSyncFlags(needReloadOnAccess) {
-		return nil, errors.New("cached region need reload")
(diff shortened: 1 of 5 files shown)


internal/locate/region_request.go
```diff
@@ -1423,10 +1422,9 @@ func (s *RegionRequestSender) getRPCContext(
 	switch et {
 	case tikvrpc.TiKV:
 		if s.replicaSelector == nil {
-			selector, err := NewReplicaSelector(s.regionCache, regionID, req, opts...)
-			if err != nil {
-				s.rpcError = err
-				return nil, nil
+			selector, err := NewReplicaSelector(s.regionCache, regionID, req, opts...) //nolint:staticcheck // ignore SA4023, never returns a nil interface value
+			if selector == nil || !selector.isValid() || err != nil {                  //nolint:staticcheck // ignore SA4023, never returns a nil interface value
```
MyonKeminta, line 1426: I didn't understand, why SA4023 is triggered here?

crazycs520: If I remove `selector == nil` check here, the `lint staticcheck ` will pass, otherwise will have following lint error:

image

MyonKeminta: 😥

MyonKeminta (approved):

internal/locate/region_request.go
crazycs520: But I don't want to remove `selector == nil` check here, Since I always afraid to call `selector.isValid()` maybe panic without check`selector == nil`.

zyguan (approved):

PR #1299, opened 2024-04-18
region_cache: fix issue that LocateKey may returns a wrong region

zyguan:
https://github.com/tikv/client-go/pull/1122#discussion_r1469008167 here we introduce an issue: if a region is marked as NeedDelayedReload, then we try to reload it by id and return the new region directly without checking the range, which may cause `LocateKey` returns a wrong region when the region is just splitted. To fix this issue, we can:
1. always use `loadRegion` (by key)
2. check the range of the new region (returned by `loadRegionByID`), call `loadRegion` if it doesn't contain the key

Here we prefer the first solution for simplicity.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1318,18 +1318,12 @@ func (c *RegionCache) findRegionByKey(bo *retry.Backoffer, key []byte, isEndKey 
 		}
 	} else if flags := r.resetSyncFlags(needReloadOnAccess | needDelayedReloadReady); flags > 0 {
 		// load region when it be marked as need reload.
-		reloadOnAccess := flags&needReloadOnAccess > 0
 		var (
 			lr  *Region
 			err error
 		)
-		if reloadOnAccess {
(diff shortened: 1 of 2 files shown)


crazycs520 (approved):

internal/locate/region_cache.go
```diff
@@ -1318,25 +1318,20 @@ func (c *RegionCache) findRegionByKey(bo *retry.Backoffer, key []byte, isEndKey
 		}
 	} else if flags := r.resetSyncFlags(needReloadOnAccess | needDelayedReloadReady); flags > 0 {
 		// load region when it be marked as need reload.
```
cfzjywxk, line 1320: The comment should be updated too.

cfzjywxk (approved):

internal/locate/region_cache.go
zyguan: Added

MyonKeminta (approved):

PR #1316, opened 2024-04-22
Check primary key when handling CommitTsExpired error

ekexium:
This is simply a defensive check. We don't expect it to happen.

### DIFF
--- a/txnkv/transaction/commit.go
+++ b/txnkv/transaction/commit.go
@@ -35,6 +35,7 @@
 package transaction
 
 import (
+	"bytes"
 	"encoding/hex"
 	"time"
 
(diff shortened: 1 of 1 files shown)


txnkv/transaction/commit.go
```diff
@@ -158,7 +159,17 @@ func (action actionCommit) handleSingleBatch(c *twoPhaseCommitter, bo *retry.Bac
...
 				logutil.Logger(bo.GetCtx()).Info("2PC commitTS rejected by TiKV, retry with a newer commitTS",
 					zap.Uint64("txnStartTS", c.startTS),
-					zap.Stringer("info", logutil.Hex(rejected)))
+					zap.Stringer("info", logutil.Hex(rejected)),
+				)
+
+				if !batch.isPrimary || !bytes.Equal(rejected.Key, c.primary()) {
+					logutil.Logger(bo.GetCtx()).Error("2PC commitTS rejected by TiKV, but the key is not the primary key",
+						zap.Uint64("txnStartTS", c.startTS),
+						zap.String("key", hex.EncodeToString(rejected.Key)),
```
cfzjywxk, line 168: Do we have redact encode interfaces in client-go?

ekexium: No

ekexium: We may consider not logging the keys, so that we're not making the situation worse

txnkv/transaction/commit.go
```diff
@@ -158,7 +159,17 @@ func (action actionCommit) handleSingleBatch(c *twoPhaseCommitter, bo *retry.Bac
 			if rejected := keyErr.GetCommitTsExpired(); rejected != nil {
 				logutil.Logger(bo.GetCtx()).Info("2PC commitTS rejected by TiKV, retry with a newer commitTS",
 					zap.Uint64("txnStartTS", c.startTS),
-					zap.Stringer("info", logutil.Hex(rejected)))
+					zap.Stringer("info", logutil.Hex(rejected)),
+				)
+
+				if !batch.isPrimary || !bytes.Equal(rejected.Key, c.primary()) {
```
MyonKeminta, line 165: Is the left part of the condition `!batch.isPrimary` necessary?

MyonKeminta (approved):

txnkv/transaction/commit.go
ekexium: We definitely expect this is the primary batch, so just in case.

cfzjywxk (approved):

PR #1317, opened 2024-04-22
txn: abort pipelined dml when pk rollback

you06:
ref pingcap/tidb#52760, tikv/tikv#16291

Once the pipelined dml's ttl manager is stopped, fail the pipelined dml to avoid GC blocking and writing meaningless locks.

```sql
MySQL [test1]> insert into sbtest4_gc_max select * from sbtest4;
ERROR 1105 (HY000): ttl manager is closed
```

### DIFF
--- a/integration_tests/pipelined_memdb_test.go
+++ b/integration_tests/pipelined_memdb_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"strconv"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


integration_tests/pipelined_memdb_test.go
```diff
@@ -401,3 +426,41 @@ func (s *testPipelinedMemDBSuite) TestPipelinedPrefetch() {
...
+
+	s.Eventuallyf(func() bool {
+		return !txnProbe.GetCommitter().IsTTLRunning()
+	}, 5*time.Second, 100*time.Millisecond, "ttl manager should stop after primary lock is resolved")
+
+	txn.Set([]byte("key2"), []byte("value2"))
+	flushed, err = txn.GetMemBuffer().Flush(true)
+	s.Nil(err)
+	s.True(flushed)
+	s.NotNil(txn.GetMemBuffer().FlushWait())
```
ekexium, line 465: Can replace it with "ErrorContains" to describe our expectation.

ekexium (approved):

txnkv/transaction/txn.go
```diff
@@ -467,6 +467,9 @@ func (txn *KVTxn) InitPipelinedMemDB() error {
 	pipelinedMemDB := unionstore.NewPipelinedMemDB(func(ctx context.Context, keys [][]byte) (map[string][]byte, error) {
 		return txn.snapshot.BatchGetWithTier(ctx, keys, txnsnapshot.BatchGetBufferTier)
 	}, func(generation uint64, memdb *unionstore.MemDB) (err error) {
+		if atomic.LoadUint32((*uint32)(&txn.committer.ttlManager.state)) == uint32(stateClosed) {
+			return errors.New("ttl manager is closed")
```
cfzjywxk, line 471: Should we also print a `WARN` log here?

you06: This will lead to query failure, and the log in TiDB be like.

image

cfzjywxk (approved):

PR #1320, opened 2024-04-24
txn: do not cache value with snapshot version maxuint64

you06:
The snapshot version maxuint64 doesn't satisfy the total order of TSO, do not cache it to avoid anomaly of missing latest data.


cfzjywxk:
@crazycs520 PTAL

zyguan (approved):

crazycs520 (approved):

PR #1325, opened 2024-04-28
pick #1250 to tidb-7.5

zyguan:
pick #1250 to tidb-7.5, also include #1277 & #1284.


zyguan:
@cfzjywxk @MyonKeminta @crazycs520 PTAL

cfzjywxk (approved):

PR #1331, opened 2024-05-09
[WIP] Benchmark tool for TiKV

LykxSassinator:
## Description
This pull request introduces a benchmark tool designed to evaluate the performance of a specific TiKV cluster without the necessity of deploying TiDB.
The motivation of this pr derives from the challenges associated with performance testing TiKV clusters, which typically require deploying one or more TiDB nodes. This process can be time-consuming, and issues related to performance regressions may not be promptly communicated to developers.

Therefore, this pull request offers a simplified implementation of the benchmark tool to address these concerns.


ekexium:
Is a go.sum missing? Do we need tests for this tool? We may also need a manual or readme for it.
I think it is not the best option to put the tool in examples, maybe we can have a bin or tools directory for its, or even a standalone repo.

examples/benchtool/main.go
```diff
@@ -0,0 +1,74 @@
...
+	workloads.GlobalContext, cancel = context.WithCancel(context.Background())
+
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc,
+		syscall.SIGHUP,
+		syscall.SIGINT,
+		syscall.SIGTERM,
+		syscall.SIGQUIT)
+
+	closeDone := make(chan struct{}, 1)
```
ekexium, line 54: What is it for?

LykxSassinator:
> Is a go.sum missing? Do we need tests for this tool? We may also need a manual or readme for it. I think it is not the best option to put the tool in examples, maybe we can have a bin or tools directory for its, or even a standalone repo.

FYI, the reason why this tool is set under `examples` is used to make this tool accessible and easy to use, where the `README.md` is already specified.

And as for tests, the basic and offline tests have been done for it, which can be reviewed in `https://pingcap-cn.feishu.cn/wiki/QtoIwKWqNiiYp7kTujIcjw4RnEh`.

Meanwhile, it's acceptable to move this tool into an individual repo if we can make an agreement about this advice.

examples/benchtool/main.go
LykxSassinator: Redundant. Already cleaned.

PR #1098, opened 2023-12-28
client: Cache tikv request in tidb client side

bufferflies:
close #1099
```
1. all the requests will put into the priority queue.
2. requests can't be canceled if the client is not available tempory, it remove from the queue if the request has been canceled by the caller such as timeout or server is stop.
3. add new config to limit the concurrency of requests that are sent one tikv.
```

## Manual Test
rg1 is high priority, rg2 is medium priorit.
the connection of high priority always is 5,
the conncetion of medium priority is [5,20,50,100,200]
default

image

max-concurrency-request-limit =30
image

performance between master and this pr
image

The high priority sql run faster than the medium after set max-concurrency-request-limit =30

performance comapre 
img_v3_027k_2c98e0cb-2d07-419a-9183-7c270b8f521g

PR #1339, opened 2024-05-11
fix issue of doesn't fast fail request when no available connections

crazycs520:
Ref https://github.com/pingcap/tidb/issues/53222

https://github.com/tikv/client-go/pull/1098/files#diff-289d0ab3e9da9a88db8dcf8ed7095af6756da3ea8da38212b352742a83d93ba2L383

If do not cancel the request in time, the upper layer will wait until it times out, and we should cancel requests when no available connections and let the upper layer try another replica again.

Before This PR:

Benchmark test with stop 1 tikv store. Many slow query execution time exceed 1 min.

image

This PR:

Many slow query execution time is less than 25s.

image

### DIFF
--- a/internal/client/client_batch.go
+++ b/internal/client/client_batch.go
@@ -418,6 +418,12 @@ func (a *batchConn) getClientAndSend() {
 	if cli == nil {
 		logutil.BgLogger().Info("no available connections", zap.String("target", target), zap.Any("reasons", reasons))
 		metrics.TiKVNoAvailableConnectionCounter.Inc()
+		if config.GetGlobalConfig().TiKVClient.MaxConcurrencyRequestLimit == config.DefMaxConcurrencyRequestLimit {
+			// Only cancel requests when MaxConcurrencyRequestLimit feature is not enabled, to be compatible with the behavior of older versions.
+			// TODO: But when MaxConcurrencyRequestLimit feature is enabled, the requests won't be canceled and will wait until timeout.
+			// This behavior may not be reasonable, as the timeout is usually 40s or 60s, which is too long to retry in time.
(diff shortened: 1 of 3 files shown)


cfzjywxk (approved):
LGTM

internal/client/client_batch.go
```diff
@@ -418,6 +418,12 @@ func (a *batchConn) getClientAndSend() {
 	if cli == nil {
 		logutil.BgLogger().Info("no available connections", zap.String("target", target), zap.Any("reasons", reasons))
 		metrics.TiKVNoAvailableConnectionCounter.Inc()
+		if config.GetGlobalConfig().TiKVClient.MaxConcurrencyRequestLimit == config.DefMaxConcurrencyRequestLimit {
+			// Only cancel request when MaxConcurrencyRequestLimit feature is not enabled, for compatible with the behavior of older versions.
+			// TODO: But when MaxConcurrencyRequestLimit feature is enable, the request won't be canceled and will wait timeout,
+			// this behavior may not be reasonable, the timeout is usually 40s or 60s, which is too long to retry in time.
```
MyonKeminta, line 424:
```suggestion
			// Only cancel requests when MaxConcurrencyRequestLimit feature is not enabled, to be compatible with the behavior of older versions.
			// TODO: But when MaxConcurrencyRequestLimit feature is enabled, the requests won't be canceled and will wait until timeout.
			// This behavior may not be reasonable, as the timeout is usually 40s or 60s, which is too long to retry in time.
```

MyonKeminta (commented):
Rest LGTM

MyonKeminta (approved):

PR #1300, opened 2024-04-18
improve region request log for diagnose

crazycs520:
Close #1315.

### What changed and how does it work?

#### 1. Reduce `region_request` log

- add a slow log threshold 100ms, avoid print too much log.

**Before This PR**

After 1 tikv server down when benchmark test, tidb will output tens of thousands log about `region_request` in a matter of minutes

```sql
> rg 'region_request.go' tidb.log | wc
  26917 1839370 26737091
```

**This PR**

```sql
> rg 'region_request.go' tidb.log | wc
   2675  285852 3440758
```

#### 2. make log more **friendly to diagnostic slow queries.**

Such as use  `Connection ID` and `txn timestamp` to grep tidb log, then got following log:

```sql
> rg 449392618766336000 tidb.log  | rg 1023410562
[...]

PR #1340, opened 2024-05-13
improve region request log for diagnose (1300)

crazycs520:
cherry-pick 
- https://github.com/tikv/client-go/pull/1300
- https://github.com/tikv/client-go/pull/908
- https://github.com/tikv/client-go/pull/1344

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1516,61 +1516,6 @@ func (s *RegionRequestSender) SendReqCtx(
 	}
 }
 
-//<<<<<<< HEAD
-//func (s *RegionRequestSender) logSendReqError(bo *retry.Backoffer, msg string, regionID RegionVerID, retryTimes int, req *tikvrpc.Request, totalErrors map[string]int) {
-//	var replicaStatus []string
-//	replicaSelectorState := "nil"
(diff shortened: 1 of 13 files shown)


internal/locate/region_request.go
```diff
@@ -35,9 +35,9 @@
 package locate
 
 import (
-	"bytes"
 	"context"
 	"fmt"
+	"maps"
```
MyonKeminta, line 40: > package maps is not in GOROOT (/opt/hostedtoolcache/go/1.18.10/x64/src/maps)

Looks like this package doesn't exist in go 1.18

internal/locate/region_request_test.go
MyonKeminta, line 1: Is the test `TestRegionRequestStats` added in #1300 not included here?

MyonKeminta (commented):
Rest LGTM

internal/locate/region_request_test.go
crazycs520: Nice catch, I missed it before, and now I've added it.

internal/locate/region_request.go
crazycs520: Nice catch, done.

cfzjywxk (approved):

PR #1342, opened 2024-05-14
feat: record flush_wait_ms in TxnInfo

ekexium:
ref https://github.com/pingcap/tidb/issues/50215
This is to show performance characteristics of pipelined dml.

### DIFF
--- a/internal/unionstore/union_store.go
+++ b/internal/unionstore/union_store.go
@@ -237,8 +237,12 @@ type MemBuffer interface {
 	Flush(force bool) (bool, error)
 	// FlushWait waits for the flushing task done and return error.
 	FlushWait() error
-	// FlushWaitDuration returns the total duration spent on FlushWait()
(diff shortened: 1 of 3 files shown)


internal/unionstore/union_store.go
```diff
@@ -236,6 +237,8 @@ type MemBuffer interface {
 	Flush(force bool) (bool, error)
 	// FlushWait waits for the flushing task done and return error.
 	FlushWait() error
+	// FlushWaitDuration returns the total duration spent on FlushWait()
+	FlushWaitDuration() time.Duration
```
cfzjywxk, line 241: In case we may need more execution details like `flushDuration`, do we need to make the interface something like `FlushDetails`?

ekexium: That looks reasonable

cfzjywxk (approved):

zyguan (approved):

PR #1344, opened 2024-05-16
*: fix panic in get cause error

### LINKED ISSUE #1343: nil panic once meet quota limit error
image

image

```
[2024/05/15 17:57:51.732 +08:00] [ERROR] [coprocessor.go:1152] ["copIteratorWork meet panic"] [r="runtime error: invalid memory address or nil pointer dereference"] ["stack trace"="gi
thub.com/pingcap/tidb/pkg/store/copr.(*copIteratorWorker).handleTask.func1\n\t/data/nvme0n1/csn/program/go/tidb/pkg/store/copr/coprocessor.go:1154\nruntime.gopanic\n\t/snap/go/10585/s
rc/runtime/panic.go:770\nruntime.panicmem\n\t/snap/go/10585/src/runtime/panic.go:261\nruntime.sigpanic\n\t/snap/go/10585/src/runtime/signal_unix.go:881\ngithub.com/tikv/client-go/v2/i
nternal/locate.(*RegionRequestSender).sendReqToRegion\n\t/data/nvme0n1/csn/program/go/client-go/internal/locate/region_request.go:2118\ngithub.com/tikv/client-go/v2/internal/locate.(*
RegionRequestSender).SendReqCtx\n\t/data/nvme0n1/csn/program/go/client-go/internal/locate/region_request.go:1796\ngithub.com/tikv/client-go/v2/txnkv/txnsnapshot.(*ClientHelper).SendRe
qCtx\n\t/data/nvme0n1/csn/program/go/client-go/txnkv/txnsnapshot/client_helper.go:146\ngithub.com/pingcap/tidb/pkg/store/copr.(*copIteratorWorker).handleTaskOnce\n\t/data/nvme0n1/csn/
program/go/tidb/pkg/store/copr/coprocessor.go:1275\ngithub.com/pingcap/tidb/pkg/store/copr.(*copIteratorWorker).handleTask\n\t/data/nvme0n1/csn/program/go/tidb/pkg/store/copr/coproces
sor.go:1165\ngithub.com/pingcap/tidb/pkg/store/copr.(*copIteratorWorker).run\n\t/data/nvme0n1/csn/program/go/tidb/pkg/store/copr/coprocessor.go:828"]
```

the error.Cause may retrun nil:

https://github.com/tikv/client-go/blob/c40432e3abccd374ea0f6b945f95b925004b4e49/internal/locate/region_request.go#L2118

and 

https://github.com/tikv/client-go/blob/c40432e3abccd374ea0f6b945f95b925004b4e49/internal/locate/region_request.go#L2214

crazycs520:
close #1343

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -2222,7 +2222,9 @@ func (s *RegionRequestSender) onSendFail(bo *retry.Backoffer, ctx *RPCContext, r
 			}
 		}
 	}
-	metrics.TiKVRPCErrorCounter.WithLabelValues(getErrMsg(err), storeLabel).Inc()
+	if errStr := getErrMsg(err); len(errStr) > 0 {
+		metrics.TiKVRPCErrorCounter.WithLabelValues(errStr, storeLabel).Inc()
+	}
 
 	if ctx.Store != nil && ctx.Store.storeType == tikvrpc.TiFlashCompute {
 		s.regionCache.InvalidateTiFlashComputeStoresIfGRPCError(err)
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -180,6 +180,17 @@ func (r *RequestErrorStats) RecordRPCErrorStats(errLabel string) {
 	}
 }
 
+// getErrMsg returns error message. if the error has cause error, then return cause error message.
+func getErrMsg(err error) string {
+	if err == nil {
+		return ""
+	}
+	if causeErr := errors.Cause(err); causeErr != nil {
+		return causeErr.Error()
+	}
+	return err.Error()
+}
+
 // String implements fmt.Stringer interface.
 func (r *RegionRequestRuntimeStats) String() string {
 	if r == nil {
@@ -2115,7 +2126,7 @@ func (s *RegionRequestSender) sendReqToRegion(
 	if err != nil {
 		s.rpcError = err
 		if s.Stats != nil {
-			errStr := errors.Cause(err).Error()
+			errStr := getErrMsg(err)
 			s.Stats.RecordRPCErrorStats(errStr)
 			s.recordRPCAccessInfo(req, rpcCtx, errStr)
 		}
@@ -2211,7 +2222,7 @@ func (s *RegionRequestSender) onSendFail(bo *retry.Backoffer, ctx *RPCContext, r
 			}
 		}
 	}
-	metrics.TiKVRPCErrorCounter.WithLabelValues(errors.Cause(err).Error(), storeLabel).Inc()
+	metrics.TiKVRPCErrorCounter.WithLabelValues(getErrMsg(err), storeLabel).Inc()
 
 	if ctx.Store != nil && ctx.Store.storeType == tikvrpc.TiFlashCompute {
 		s.regionCache.InvalidateTiFlashComputeStoresIfGRPCError(err)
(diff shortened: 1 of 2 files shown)


internal/locate/region_request.go
```diff
@@ -2211,7 +2222,7 @@ func (s *RegionRequestSender) onSendFail(bo *retry.Backoffer, ctx *RPCContext, r
 			}
 		}
 	}
-	metrics.TiKVRPCErrorCounter.WithLabelValues(errors.Cause(err).Error(), storeLabel).Inc()
+	metrics.TiKVRPCErrorCounter.WithLabelValues(getErrMsg(err), storeLabel).Inc()
```
cfzjywxk, line 2225: Should it be like this?
```
errStr := getErrMsg(err)
if len(errStr) > 0 {
    metrics.TiKVRPCErrorCounter.WithLabelValues(getErrMsg(err), storeLabel).Inc()
}
```
Or to check if the input `err` is nil?
It's strange to use an empty string as a metric label.

crazycs520: But normally, the `errStr` shouldn't be an empty string, since the `err` can't be nil here.

cfzjywxk: If the input `err` is required not to be nil, then should we return directly at L2173, or add some assertions?

But how the panic in https://github.com/tikv/client-go/issues/1343 happen? Is the input `err` nil in that case?

crazycs520: #1343 is caused by some error implement `Cause` interface, but `Cause` method return `nil`. See the test in this PR.

cfzjywxk: So the 
```
if len(errStr) > 0 { metrics.TiKVRPCErrorCounter.WithLabelValues(getErrMsg(err), storeLabel).Inc() }
```
seems to be necessary, as the cause string could be empty and it should not be used as label.

Or we use `unknown` for this situation.

cfzjywxk: Skipping recording error counting if the `cause` is empty, may lead to incorrect total error number counting.
Perhaps `Unknown` is better?

crazycs520: In `getErrMsg`, if the error's cause is nil, will return the `error.Error()`, unless the error message itself is empty string, otherwise it wouldn't be empty string here.

cfzjywxk: It's better to use some defensive programming style but not to assume if it could be nil or not.
Actually if it's c/c++ code, each pointer should be checked first before using.

cfzjywxk (approved):

MyonKeminta (approved):

internal/locate/region_request.go
```diff
@@ -2211,7 +2222,11 @@ func (s *RegionRequestSender) onSendFail(bo *retry.Backoffer, ctx *RPCContext, r
 			}
 		}
 	}
-	metrics.TiKVRPCErrorCounter.WithLabelValues(errors.Cause(err).Error(), storeLabel).Inc()
+	if errStr := getErrMsg(err); len(errStr) > 0 {
+		metrics.TiKVRPCErrorCounter.WithLabelValues(getErrMsg(err), storeLabel).Inc()
```
nolouch, line 2226: will the error message be too long?

crazycs520: It's possible, but I haven't seen any error message very long yet.

nolouch: my issue error message is:


image

crazycs520: Great, I think this error length is acceptable.

PR #1095, opened 2023-12-28
client: replace array by priority queue to support cache queue

bufferflies:
close #1094

PR #1353, opened 2024-05-22
revert #1095 #1098 #1226 to fix performance regression

crazycs520:
revert #1095 #1098 #1226


crazycs520:
After test, those PR don't have too much performance regression.

PR #1355, opened 2024-05-29
Fix the problem that statement being oom-killed within DoneAggressiveLocking causing the transaction still in aggressive locking state

MyonKeminta:
This is a quick fix to avoid the transaction being left in aggressive locking state after panicking in `DoneAggressiveLocking`. While I haven't decide how to fix it in the most elegant approach, there should be a quick fix to avoid the problem, so that we have time to find the final solution more carefully.

Ref: https://github.com/pingcap/tidb/issues/53540#issuecomment-2132568080 , https://github.com/pingcap/tidb/issues/53540#issuecomment-2138089140

### DIFF
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -957,6 +957,11 @@ func (txn *KVTxn) CancelAggressiveLocking(ctx context.Context) {
 	if txn.aggressiveLockingContext == nil {
 		panic("Trying to cancel aggressive locking while it's not started")
 	}
+
+	defer func() {
+		txn.aggressiveLockingContext = nil
+	}()
+
 	txn.cleanupAggressiveLockingRedundantLocks(context.Background())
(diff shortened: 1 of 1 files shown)


ekexium (approved):

cfzjywxk (commented):
We could add some comments about it, maybe we could restore it after the more elegant solution is merged.

txnkv/transaction/txn.go
```diff
@@ -957,6 +957,11 @@ func (txn *KVTxn) CancelAggressiveLocking(ctx context.Context) {
 	if txn.aggressiveLockingContext == nil {
 		panic("Trying to cancel aggressive locking while it's not started")
 	}
+
+	defer func() {
```
cfzjywxk, line 961: We could add some comments about it, maybe we could restore it after the more elegant solution is merged.

cfzjywxk (approved):

PR #1356, opened 2024-05-30
add func for new test keyspace tikv store

ystaticy:


tikv/test_util.go
```diff
@@ -93,3 +94,41 @@ func NewTestTiKVStore(client Client, pdClient pd.Client, clientHijack func(Clien
...
+		client = clientHijack(client)
+	}
+
+	if pdClientHijack != nil {
+		pdCli = pdClientHijack(pdCli)
+	}
+
+	// Make sure the uuid is unique.
+	uid := uuid.New().String()
+	spkv := NewMockSafePointKV()
```
iosmanthus, line 125: Should this spkv add a prefix too?

ystaticy: It is a `map[string]string` ，not a etcd storage at here.
So we don't need the prefix, right?

tikv/test_util.go
```diff
@@ -93,3 +94,41 @@ func NewTestTiKVStore(client Client, pdClient pd.Client, clientHijack func(Clien
 	tikvStore.mock = true
 	return tikvStore, err
 }
+
+// NewTestTiKVStore creates a test store with Option
+func NewTestKeyspaceTiKVStore(client Client, pdClient pd.Client, clientHijack func(Client) Client, pdClientHijack func(pd.Client) pd.Client, txnLocalLatches uint, keyspaceMeta keyspacepb.KeyspaceMeta, opt ...Option) (*KVStore, error) {
```
iosmanthus, line 99: Are these test utils provided for TiDB?

ystaticy: Yes, it's for TiDB.

iosmanthus:
Rest LGTM

iosmanthus (approved):

PR #1362, opened 2024-06-06
Introduce a new method MemHookSet

ekexium:
part of [pingcap/tidb#53832](https://github.com/pingcap/tidb/pull/53764)
Add a method to show whether the hook is set, to avoid redundantly setting it.


cfzjywxk (approved):
LGTM

crazycs520 (approved):

PR #1366, opened 2024-06-19
add support for aggressive batching

zyguan:
Currently, we can only batch more requests when the tikv is detected as overload. This PR allow such an aggressive batching in normal cases by setting `overload-threshold` to 0. Workloads with high throughput (like sysbench oltp point select) can benfit a lot from it because batching more requests reduces the grpc overhead and the number of syscalls. Refer to the following benchmark result.

https://clinic.pingcap.com.cn/portal/#/orgs/541/clusters/7381654358775158872?from=1718762265&to=1718762865

  | avg lat | p95 lat | throughput | througput diff
-- | -- | -- | -- | --
baseline | 1.52 | 3.02 | 65654.51 | -
batch-wait-time=100us | 1.36 | 2.43 | 73432.90 | 11.85%

image


zyguan:
resolved by #1411

PR #1367, opened 2024-06-19
region cache: add `BatchLocateKeyRanges` interface

you06:
ref pingcap/tidb#53850

The current client-go doesn't support batch load non-continuous key ranges, this PR implement this methods which outperforms the former [`LocateKey`](https://github.com/tikv/client-go/blob/ac8fa1d73a0c55e8526a1ae9b88670b1b64970e6/internal/locate/region_cache.go#L1207) a lot.

The test result:

1. split table into 1000 regions by `SPLIT TABLE sbtest1 BETWEEN (0) AND (1000000) REGIONS 1000;`
2. wait until region cache expired.
3. run a simple query `explain analyze select * from sbtest1 limit 1;`

- Without this interface: `...build_task_duration: 81.5ms...`, statmenet execution takes 0.11 sec
- With this interface: `...build_task_duration: 2.7ms...`, statement execution takes 0.02 sec


internal/locate/pd_codec.go
```diff
@@ -141,6 +142,28 @@ func (c *CodecPDClient) ScanRegions(ctx context.Context, startKey []byte, endKey
 	return regions, nil
 }
 
+// BatchScanRegions encodes the key before send requests to pd-server and decodes the
+// returned StartKey && EndKey from pd-server.
```
cfzjywxk, line 146: Better to also explain the usage of `limit` parameters.

internal/locate/region_cache.go
```diff
@@ -1203,6 +1204,149 @@ func (c *RegionCache) LocateKeyRange(bo *retry.Backoffer, startKey, endKey []byt
 	return res, nil
 }
 
+type BatchLocateKeyRangesOption struct {
+	needLeader  bool // whether load leader only, if it's set to true, regions without leader will be skipped.
```
cfzjywxk, line 1208: `needRegionWithValidLeader` is more understandable.

internal/locate/region_cache.go
```diff
@@ -1203,6 +1204,149 @@ func (c *RegionCache) LocateKeyRange(bo *retry.Backoffer, startKey, endKey []byt
...
+			uncachedRanges = uncachedRanges[mid:]
+			if bytes.Compare(endKey, uncachedRanges[0].StartKey) > 0 {
+				uncachedRanges[0].StartKey = endKey
+			}
+		}
+	}
+	// append the rest cache hit regions
+	for i := cachedIdx; i < len(cachedRegions); i++ {
+		appendRegion(cachedRegions[i])
+	}
```
cfzjywxk, line 1346: Better to wrap this merge operation into a standalone function and write unit-tests to cover all the possible combinations.

internal/locate/region_cache.go
```diff
@@ -1559,6 +1703,33 @@ func (c *RegionCache) BatchLoadRegionsWithKeyRange(bo *retry.Backoffer, startKey
...
+		err = errors.Errorf("PD returned no region, first range start_key: %q, last range end_key: %q, encode_start_key: %q, encode_end_key: %q",
+			util.HexRegionKeyStr(startKey), util.HexRegionKeyStr(endKey),
+			util.HexRegionKeyStr(c.codec.EncodeRegionKey(startKey)), util.HexRegionKeyStr(c.codec.EncodeRegionKey(endKey)))
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for _, region := range regions {
```
cfzjywxk, line 1727: Do we have a specific `update` interface of the `RegionCache`?

internal/locate/region_cache.go
```diff
@@ -1962,29 +2134,93 @@ func (c *RegionCache) scanRegions(bo *retry.Backoffer, startKey, endKey []byte,
...
+		pdOpts := []pd.GetRegionOption{pd.WithAllowFollowerHandle()}
+		if opt.needBuckets {
+			pdOpts = append(pdOpts, pd.WithBuckets())
+		}
+		regionsInfo, err := c.pdClient.BatchScanRegions(ctx, keyRanges, limit, pdOpts...)
+		metrics.LoadRegionCacheHistogramWithRegions.Observe(time.Since(start).Seconds())
+		if err != nil {
+			if apicodec.IsDecodeError(err) {
+				return nil, errors.Errorf("failed to decode region range key, startKey: %q, limit: %d, err: %v, encode_start_key: %q",
+					util.HexRegionKeyStr(startKey), limit, err, util.HexRegionKeyStr(c.codec.EncodeRegionKey(startKey)))
```
cfzjywxk, line 2176: Should the `Key` be red-act in error messages?

internal/locate/region_cache.go
```diff
@@ -1962,29 +2134,93 @@ func (c *RegionCache) scanRegions(bo *retry.Backoffer, startKey, endKey []byte,
...
+			}
+			metrics.RegionCacheCounterWithBatchScanRegionsError.Inc()
+			backoffErr = errors.Errorf(
```
cfzjywxk, line 2179: Ditto.

internal/locate/region_cache_test.go
```diff
@@ -2519,3 +2520,88 @@ func BenchmarkInsertRegionToCache2(b *testing.B) {
...
+
+	toRanges := func(keys ...string) []kv.KeyRange {
+		ranges := make([]kv.KeyRange, 0, len(keys)/2)
+		for i := 0; i < len(keys); i += 2 {
+			ranges = append(ranges, kv.KeyRange{StartKey: []byte(keys[i]), EndKey: []byte(keys[i+1])})
+		}
+		return ranges
+	}
+
+	// nil --- 'a' --- 'b' --- 'c' --- 'd' --- 'e' --- 'f' --- 'g' --- nil
```
cfzjywxk, line 2574: We could add test to assume the region information merge happens at
The beginning
2. The middle
3. The end

Or there are stale region information in the region cache from these positions.

internal/locate/region_cache.go
you06: The leader is not empty even it's an invalid or outdated leader. For a new splitted regions, they are reported to PD after leaders are elected, so the leaders are also not nil.

So this parameter won't affect anything from what I know, just keep it to keep the compatibility.

I think `needRegionWithLeaderPeer` a proper name.

internal/locate/region_cache_test.go
you06: The cases are difficult to test in the integration `BatchLocateKeyRanges`, I add the coverage in `TestBatchScanRegionsMerger` and `TestSplitKeyRanges`.

internal/locate/region_cache.go
you06: No, there are only `UpdateLeader` and `UpdateBucketsIfNeeded`

BTW, I'm not sure if `UpdateRegion` is resonable, because there can be multi intersected regions when you want to update one region, the name `update` looks like only one old region info will be replaced by the new one.

internal/locate/region_cache.go
you06: Seems red-act is not supported in current client-go, maybe remove this log and add a TODO by now.

internal/locate/region_cache.go
```diff
@@ -1962,27 +2145,91 @@ func (c *RegionCache) scanRegions(bo *retry.Backoffer, startKey, endKey []byte,
...
 	}
+	if len(regions) == 0 {
+		return nil, errors.New("receive Regions with no peer")
+	}
+	if len(regions) < len(regionsInfo) {
+		logutil.Logger(context.Background()).Debug(
+			"regionCache: scanRegion finished but some regions has no leader.")
+	}
+	return regions, nil
+
```
zyguan, line 2221: Here is a redundant empty line.

internal/locate/region_cache.go
```diff
@@ -1956,35 +2138,92 @@ func (c *RegionCache) scanRegions(bo *retry.Backoffer, startKey, endKey []byte,
...
+		start := time.Now()
+		pdOpts := []pd.GetRegionOption{pd.WithAllowFollowerHandle()}
+		if opt.needBuckets {
+			pdOpts = append(pdOpts, pd.WithBuckets())
 		}
-		if len(regions) < len(regionsInfo) {
-			logutil.Logger(context.Background()).Debug(
-				"regionCache: scanRegion finished but some regions has no leader.")
+		regionsInfo, err := c.pdClient.BatchScanRegions(ctx, keyRanges, limit, pdOpts...)
+		metrics.LoadRegionCacheHistogramWithRegions.Observe(time.Since(start).Seconds())
```
zyguan, line 2174: Should we use a new series with a tag like `batch_scan_regions`?

integration_tests/split_test.go
```diff
@@ -235,6 +235,17 @@ func (c *mockPDClient) ScanRegions(ctx context.Context, startKey []byte, endKey
...
 
+func (c *mockPDClient) BatchScanRegions(ctx context.Context, keyRanges []pd.KeyRange, limit int, opts ...pd.GetRegionOption) ([]*pd.Region, error) {
+	c.RLock()
+	defer c.RUnlock()
+
+	if c.stop {
+		return nil, errors.WithStack(errStopped)
+	}
+	return c.client.BatchScanRegions(ctx, keyRanges, limit, opts...)
+
```
zyguan, line 246:
```suggestion

```

internal/locate/region_cache.go
```diff
@@ -1203,6 +1204,160 @@ func (c *RegionCache) LocateKeyRange(bo *retry.Backoffer, startKey, endKey []byt
...
+		if lastEndKey := m.lastEndKey.Inner(); lastEndKey != nil && bytes.Compare(*lastEndKey, m.cachedRegions[m.cachedIdx].EndKey()) >= 0 {
+			// skip the cached region that is covered by the uncached region.
+			continue
+		}
+		m.appendKeyLocation(m.cachedRegions[m.cachedIdx])
+	}
+	return m.mergedLocations
+}
+
+func nextRanges(keyRanges []pd.KeyRange, locatedKey []byte) []pd.KeyRange {
```
cfzjywxk, line 1341: Better to explain the input parameters for this function, also the returned slice is referenced but not cloned. The `nextRanges` is more understandable as a local function or closure, as a standalone function how about `RangesAfterKey`.

internal/locate/region_cache.go
```diff
@@ -1203,6 +1204,160 @@ func (c *RegionCache) LocateKeyRange(bo *retry.Backoffer, startKey, endKey []byt
...
+func (m *batchLocateRangesMerger) appendRegion(uncachedRegion *Region) {
+	defer func() {
+		endKey := uncachedRegion.EndKey()
+		if len(endKey) == 0 {
+			m.cachedIdx = len(m.cachedRegions)
+		} else {
+			m.lastEndKey = util.Some(uncachedRegion.EndKey())
+		}
+	}()
+	if len(uncachedRegion.StartKey()) == 0 {
```
cfzjywxk, line 1306: Does this mean a single range [-inf, inf)? Better to add comments for it. 
If so we should check the result does excatly has one leff range as the result, here or when the `build` is called.

internal/locate/region_cache.go
```diff
@@ -1203,6 +1204,160 @@ func (c *RegionCache) LocateKeyRange(bo *retry.Backoffer, startKey, endKey []byt
...
+
+// BatchLocateKeyRanges lists regions in the given ranges.
+func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.KeyRange, opts ...BatchLocateKeyRangesOpt) ([]*KeyLocation, error) {
+	uncachedRanges := make([]pd.KeyRange, 0, len(keyRanges))
+	cachedRegions := make([]*Region, 0, len(keyRanges))
+	// 1. find regions from cache
+	for _, keyRange := range keyRanges {
+		for {
+			r := c.tryFindRegionByKey(keyRange.StartKey, false)
+			if r == nil {
```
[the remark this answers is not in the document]
you06: That's a further optimization, let me leave a TODO here.

internal/locate/region_cache.go
```diff
@@ -1203,6 +1204,160 @@ func (c *RegionCache) LocateKeyRange(bo *retry.Backoffer, startKey, endKey []byt
...
+			m.cachedIdx = len(m.cachedRegions)
+		} else {
+			m.lastEndKey = util.Some(uncachedRegion.EndKey())
+		}
+	}()
+	if len(uncachedRegion.StartKey()) == 0 {
+		m.appendKeyLocation(uncachedRegion)
+		return
+	}
+	lastEndKey := m.lastEndKey.Inner()
```
cfzjywxk, line 1310: Could the `util.Option` be avoided? The check is still whether the pointer is not nil.

internal/locate/region_cache.go
you06: A closure can not be tested independently, let me rename and comment it.

internal/locate/region_cache.go
you06: Ok, `len(StartKey) == 0` means the range is `(-inf, ?)`.

The region (whether it's cached or not) with lower left bound always add to result earlier, I think we don't need to check left bound in `build` again.

zyguan (approved):

cfzjywxk (approved):
LGTM

cfzjywxk:
@ekexium Sorry I did not notice the updated comments..

ekexium:
> @ekexium Sorry I did not notice the updated comments..

It's OK. It's not a blocker.

internal/locate/region_cache.go
```diff
@@ -1203,6 +1204,160 @@ func (c *RegionCache) LocateKeyRange(bo *retry.Backoffer, startKey, endKey []byt
...
 
+type batchLocateKeyRangesOption struct {
+	// whether load leader only, if it's set to true, regions without leader will be skipped.
+	// Note there is leader even the leader is invalid or outdated.
+	needRegionHasLeaderPeer bool
+	// whether load buckets, if it's set to true, more traffic will be consumed.
+	needBuckets bool
+}
+
+type BatchLocateKeyRangesOpt func(*batchLocateKeyRangesOption)
```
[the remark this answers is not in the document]
you06: IMO, for a long time, we follow this pattern a good practice, and such design is everywhere over the tidb codebase.

```golang
opt := NewBatchLodateKeyRangeOption().WithNeedBuckets().WithNeedRegionHasLeaderPeer()
```

I think the chain style config is a good design with a quick glance.

PR #1377, opened 2024-07-02
region cache: check if the pd returned regions covers the ranges

### LINKED ISSUE #1376: Need to check result of `ScanRegions` and `BatchScanRegions`
ref pingcap/tidb#54375

- Reproduce test: https://github.com/you06/client-go/commit/d7d994c6956fb5960219c13edbc31d5e80430390

The failpoint mocks the split region which is not reported to PD yet, thus the returned result of scan regions can have some gaps inside it. The PD's API caller must check if the returned regions can fully cover the given range or ranges.

The batch load functions are used by tikv coprocessor since 8.2 and tiflash coprocessor in earlier versions, which affects the correctness.

you06:
close #1376

Either `len(regionsInfo) == 0` or gaps inside `regionsInfo` are **valid** cases, we should handle it. Unless we may suffer correctness issue if we read from regions with gap.

This PR adds the checks if regions can fully cover the given ranges in `scanRegions` and `batchScanRegions`. Since this is a rare case, I think we can just simply retry all the ranges if there are gaps.


internal/locate/region_cache.go
```diff
@@ -2207,15 +2212,77 @@ func (c *RegionCache) batchScanRegions(bo *retry.Backoffer, keyRanges []pd.KeyRa
...
+				"PD returned regions have gaps, range num: %d, limit: %d",
+				len(keyRanges), limit,
+			)
+			continue
 		}
 		return c.handleRegionInfos(bo, regionsInfo, opt.needRegionHasLeaderPeer)
 	}
 }
 
+func regionsHaveGap(ranges []pd.KeyRange, regionsInfo []*pd.Region, limit int) bool {
```
lance6716, line 2232: we have a similar function in lightning/BR use case

https://github.com/pingcap/tidb/blob/1480bb127a9363e31e8b2e9f3808289ea54d9d28/br/pkg/streamhelper/regioniter.go#L103

Do you think it's better to let PD client do the retry rather than different callers? cc @okjiang

okJiang: We are kind to accept your feature request. Please feel free to file it to PD leader.

you06: Ok, I'll fix it in PD client.

you06: After discussion, we decides leave it as a TODO in the future.

internal/locate/region_cache.go
```diff
@@ -2207,15 +2212,78 @@ func (c *RegionCache) batchScanRegions(bo *retry.Backoffer, keyRanges []pd.KeyRa
...
+			continue
 		}
 		return c.handleRegionInfos(bo, regionsInfo, opt.needRegionHasLeaderPeer)
 	}
 }
 
+// regionsHaveGap checks if the loaded regions can fully cover the key ranges.
+// If there are any gaps between the regions, it returns true, then the requests might be retried.
+// TODO: remove this function after PD client supports gap detection and handling it.
+func regionsHaveGap(ranges []pd.KeyRange, regionsInfo []*pd.Region, limit int) bool {
```
cfzjywxk, line 2235: The name seems not able to cover the actual usages which is more complex

internal/locate/region_cache.go
```diff
@@ -2207,15 +2212,78 @@ func (c *RegionCache) batchScanRegions(bo *retry.Backoffer, keyRanges []pd.KeyRa
...
+	if len(regionsInfo) == 0 {
+		return true
+	}
+	checkIdx := 0                  // checked index of ranges
+	checkKey := ranges[0].StartKey // checked key of ranges
+	for _, r := range regionsInfo {
+		if r.Meta == nil {
+			return true
+		}
+		if len(r.Meta.StartKey) != 0 && bytes.Compare(r.Meta.StartKey, checkKey) > 0 {
```
MyonKeminta, line 2251: The left half of the condition seems unnecessary?

internal/locate/region_cache.go
```diff
@@ -2154,7 +2154,12 @@ func (c *RegionCache) scanRegions(bo *retry.Backoffer, startKey, endKey []byte,
 		metrics.RegionCacheCounterWithScanRegionsOK.Inc()
 
 		if len(regionsInfo) == 0 {
-			return nil, errors.Errorf("PD returned no region, limit: %d", limit)
+			backoffErr = errors.Errorf("PD returned no region, limit: %d", limit)
+			continue
+		}
+		if regionsHaveGap([]pd.KeyRange{{StartKey: startKey, EndKey: endKey}}, regionsInfo, limit) {
+			backoffErr = errors.Errorf("PD returned regions have gaps, limit: %d", limit)
+			continue
```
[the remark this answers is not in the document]
you06: Yes, this should be picked back to earlier versions, which fixes the gap correctness issue for tiflash coprocessor.

internal/locate/region_cache.go
```diff
@@ -2207,15 +2212,78 @@ func (c *RegionCache) batchScanRegions(bo *retry.Backoffer, keyRanges []pd.KeyRa
...
+		if len(r.Meta.StartKey) != 0 && bytes.Compare(r.Meta.StartKey, checkKey) > 0 {
+			// there is a gap between returned region's start_key and current check key
+			return true
+		}
+		if len(r.Meta.EndKey) == 0 {
+			// the current region contains all the rest ranges.
+			return false
+		}
+		checkKey = r.Meta.EndKey
+		for len(ranges[checkIdx].EndKey) > 0 && bytes.Compare(checkKey, ranges[checkIdx].EndKey) >= 0 {
```
MyonKeminta, line 2260: Not very familiar with the usage of it. Is the ranges here guaranteed to be already sorted by their start key?

internal/locate/region_cache.go
```diff
@@ -2207,15 +2212,78 @@ func (c *RegionCache) batchScanRegions(bo *retry.Backoffer, keyRanges []pd.KeyRa
...
+			// the end_key of returned region can cover multi ranges.
+			checkIdx++
+			if checkIdx == len(ranges) {
+				// all ranges are covered.
+				return false
+			}
+		}
+		if bytes.Compare(checkKey, ranges[checkIdx].StartKey) < 0 {
```
cfzjywxk, line 2268: What if `len(ranges[checkIdx].StartKey) == 0`, or the key range is ["", "")?

internal/locate/region_cache.go
```diff
@@ -2207,15 +2212,78 @@ func (c *RegionCache) batchScanRegions(bo *retry.Backoffer, keyRanges []pd.KeyRa
...
+	checkKey := ranges[0].StartKey // checked key of ranges
+	for _, r := range regionsInfo {
+		if r.Meta == nil {
+			return true
+		}
+		if len(r.Meta.StartKey) != 0 && bytes.Compare(r.Meta.StartKey, checkKey) > 0 {
+			// there is a gap between returned region's start_key and current check key
+			return true
+		}
+		if len(r.Meta.EndKey) == 0 {
```
cfzjywxk, line 2255: We could move the corner case check above.

internal/locate/region_cache.go
```diff
@@ -2207,15 +2212,78 @@ func (c *RegionCache) batchScanRegions(bo *retry.Backoffer, keyRanges []pd.KeyRa
...
+		return false
+	}
+	if checkIdx < len(ranges)-1 {
+		// there are still some ranges not covered.
+		return true
+	}
+	if len(checkKey) == 0 {
+		return false
+	} else if len(ranges[checkIdx].EndKey) == 0 {
+		return true
```
cfzjywxk, line 2285: Is it possible the end key of the last region is also empty, should the `true` be returned in this case?

MyonKeminta (approved):

internal/locate/region_cache.go
you06: Yes, the input ranges are sorted and cannot be overlapped.

internal/locate/region_cache.go
MyonKeminta: The start key doesn't have special meaning when it's empty and it still fits in lexicographic order. Then the condition will be false and `checkKey` won't be forwarded.
If the `ranges` can be guaranteed to be ordered by their start keys, this looks correct to me

you06: If `len(start_key) == 0`, the condition just not stands, checkKey will be  the current region's endKey.

internal/locate/region_cache.go
MyonKeminta: This is the `else` branch of `if len(checkKey) == 0`, so here the `checkKey` is not empty, meaning that the returned regions didn't cover up to the very end.

internal/locate/region_cache.go
you06: Both `len(start_key) == 0` and `["", "")` key ranges are tested.

internal/locate/region_cache.go
you06: Nice catch.

internal/locate/region_cache.go
you06: We can't move it before start key check.

e.g. the first range is `[a, c)` but the first region is `[b, "")`, if we return false when `len(region.end) == 0`, it get wrong.

internal/locate/region_cache.go
MyonKeminta: Ah, sorry, it still looks fine.
But I suggest adding some test cases where there are only one requested range and only one region in the result, and  the range is half unbounded (empty startKey + non-empty endKey and non-empty startKey + empty endKey).

crazycs520 (approved):

internal/locate/region_cache.go
you06: Half unbounded cases added.

MyonKeminta (approved):

cfzjywxk (approved):

PR #1380, opened 2024-07-08
region cache: check if the pd returned regions covers the ranges (#1377)

you06:
cherry pick #1377 to tidb-7.1


windtalker (approved):
LGTM

PR #1392, opened 2024-07-16
Let pipelined memdb ignore bufferLimit

ekexium:
For `MemDB`, bufferLimit is passed from TiDB's config `txn-total-size-limit`. It's to avoid OOM.
For a pipelined memdb, its "size" should not be limited. Besides, it's inner `MemDB`'s size is already limited by `ForceFlushMemSizeThreshold`. 
To reduce the complexity, I suggest that pipelined memdb should ignore the config `txn-total-size-limit`


internal/unionstore/pipelined_memdb.go
```diff
@@ -306,7 +306,8 @@ func (p *PipelinedMemDB) Flush(force bool) (bool, error) {
 	p.len += p.flushingMemDB.Len()
 	p.size += p.flushingMemDB.Size()
 	p.memDB = newMemDB()
-	p.memDB.SetEntrySizeLimit(p.entryLimit, p.bufferLimit)
+	// buffer size is limited by ForceFlushMemSizeThreshold. Do not set bufferLimit
+	p.memDB.SetEntrySizeLimit(p.entryLimit, math.MaxUint64)
```
cfzjywxk, line 310: Better to wrap a constant variable `const MemUnlimited = math.MaxUint64`.

zyguan (approved):

cfzjywxk (approved):

PR #1393, opened 2024-07-22
memdb: fix memdb snapshot get/iter is not actually snapshot

### LINKED ISSUE #1394: memdb snapshot getter and iter returns latest data instead
The snapshot getter and iter of memdb should read bypass the staging data. But the current snapshot read the latest data instead, see the unit test in PR.

TiDB unionscan executors read data from snapshot:
- https://github.com/pingcap/tidb/blob/cdd7c9e3044e2f4cc1dce611d0786a4f467bbeb4/pkg/executor/union_scan.go#L113
- https://github.com/pingcap/tidb/blob/7e73ddc91b5f9f089e84f0dc645cf71a27b2ad50/pkg/executor/mem_reader.go#L578-L582

you06:
close #1394
ref pingcap/tidb#54929

If I understand correctly, the snapshot get/iter always read data from non-staging snapshot, thus the added test should passed, but the currently memdb mistakenly read the latest data...


internal/unionstore/memdb_arena.go
```diff
@@ -353,7 +353,7 @@ func (l *memdbVlog) getSnapshotValue(addr memdbArenaAddr, snap *MemDBCheckpoint)
 	if result.isNull() {
 		return nil, false
 	}
-	return l.getValue(addr), true
```
zyguan, line 356: Interesting, we may need to check if there is any caller relying on this mistake behaviour.

zyguan (commented):
Interesting,

internal/unionstore/memdb_arena.go
you06: Yeah, I don't know if the tidb users are affected by this mistake, it should only happens for the queries which write data and do unionscan in the same time.

ekexium (approved):

internal/unionstore/memdb_arena.go
you06: There is, I have a regression test in pingcap/tidb#54929.

zyguan (approved):

you06:
The TiDB test passed with this patch.

I check the usages of snapshot getter/iter, there is no code relies on this mistake behavior from what I see, it's just the exist usage of TiDB cannot specify the difference between snapshot getter/iter and latest getter/iter.

@cfzjywxk I think we can merge this PR and patch it to the LTS versions.

PR #1400, opened 2024-07-29
[DNM] memdb: replace the current implementation with ART(adaptive radix tree)

you06:
ref pingcap/tidb#55287

This PR introduce the ART(adaptive radix tree) as a faster replacement of the current memdb. In the micro bench, this implementation outperforms the current memdb in every case, faster in single thread as well as lower total CPU utilization.

The implementation is inspired by [plar/go-adaptive-radix-tree](https://github.com/plar/go-adaptive-radix-tree), with some additional work:

- Support memory arena, which reduces the allocation cost in GC language.
- Support vlog, and cascade transaction(aka. staging/release/cleanup)
- Support iterator.
- Support tracking memory usage.

In a word, the ART has the same interface as current memdb.


internal/unionstore/art/node.go
```diff
@@ -0,0 +1,741 @@
...
+		n48 := n.node48(a)
+		return n48.children[n48.keys[idx]]
+	case typeNode256:
+		return n.node256(a).children[idx]
+	default:
+		panic("invalid node kind")
+	}
+}
+
+func (n4 *node4) init() {
```
ekexium, line 150: Can we somehow reduce duplicated code in all `init`s?
Or can we consider using generics to reduce more duplicated code? Maybe a benchmark is needed.

internal/unionstore/art/node.go
```diff
@@ -0,0 +1,741 @@
...
+
+func (n *artNode) node48(a *artAllocator) *node48 {
+	return a.getNode48(n.addr)
+}
+
+func (n *artNode) node256(a *artAllocator) *node256 {
+	return a.getNode256(n.addr)
+}
+
+func (an artNode) matchDeep(a *artAllocator, key Key, depth uint32) uint32 /* mismatch index*/ {
```
ekexium, line 362: Should we unify all receiver types to use pointers?

internal/unionstore/art/node.go
```diff
@@ -0,0 +1,741 @@
...
+	return -2, nullArtNode
+}
+
+func (an artNode) _findChild16(a *artAllocator, c byte) (int, artNode) {
+	n16 := an.node16(a)
+	for idx := 0; idx < int(n16.nodeNum); idx++ {
+		if n16.keys[idx] == c {
+			return idx, n16.children[idx]
+		}
+	}
```
ekexium, line 508: IIRC the paper uses SIMD or binary search here. We may benchmark the difference between the linear search and a binary search

go.mod
```diff
@@ -59,3 +60,5 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/plar/go-adaptive-radix-tree => github.com/you06/go-adaptive-radix-tree v0.0.0-20240523051018-0278e8bfcd2b
```
ekexium, line 64: Do we also need to review this? Shall we consider merge the changes to upstream?

internal/unionstore/art_memdb.go
```diff
@@ -0,0 +1,265 @@
+package unionstore
```
ekexium, line 1: Is this file still needed?

internal/unionstore/art/arena.go
```diff
@@ -0,0 +1,370 @@
...
+func (f *artAllocator) getLeaf(addr nodeAddr) *leaf {
+	if addr.isNull() {
+		return nil
+	}
+	data := f.nodeAllocator.getData(addr)
+	return (*leaf)(unsafe.Pointer(&data[0]))
+}
+
+// memArena get all the data, DO NOT access others data.
+func (a *memArena) getData(addr nodeAddr) []byte {
```
ekexium, line 237: The usage of this function is always converted to a pointer. So I suppose the function does not have to exist? We may have a direct function converting address to pointers.

internal/unionstore/art/art.go
```diff
@@ -0,0 +1,644 @@
...
+	var t Art
+	t.root = nullArtNode
+	t.stages = make([]ARTCheckpoint, 0, 2)
+	t.entrySizeLimit = math.MaxUint64
+	t.bufferSizeLimit = math.MaxUint64
+	//t.allocator.init()
+	t.allocator.nodeAllocator.freeNode4 = make([]nodeAddr, 0, 1<<4)
+	t.allocator.nodeAllocator.freeNode16 = make([]nodeAddr, 0, 1<<3)
+	t.allocator.nodeAllocator.freeNode48 = make([]nodeAddr, 0, 1<<2)
+	t.allocator.nodeAllocator.freeNode256 = make([]nodeAddr, 0, 1)
```
ekexium, line 38: Why do they have different capacity?
Can we assure the length of these slices are always limited, i.e. there won't be an unlimited-long slice of free nodes?

internal/unionstore/art/art.go
```diff
@@ -0,0 +1,644 @@
...
+	return t.len
+}
+
+// Size returns the size of the MemBuffer.
+func (t *Art) Size() int {
+	return t.size
+}
+
+// ARTCheckpoint is the checkpoint of memory DB.
+type ARTCheckpoint struct {
```
ekexium, line 372: Can merge it with MemDBCheckPoint, maybe

internal/unionstore/art/art.go
```diff
@@ -0,0 +1,644 @@
...
+		var hdr vlogHdr
+		hdr.load(block[hdrOff:])
+		lf := t.allocator.getLeaf(hdr.nodeAddr)
+		lf.vAddr = hdr.oldValue
+		t.size -= int(hdr.valueLen)
+		if hdr.oldValue.isNull() {
+			keptFlags := lf.getKeyFlags()
+			keptFlags = keptFlags.AndPersistent()
+			if keptFlags == 0 {
+				lf.markDelete()
```
ekexium, line 491: Why do we only mark it deleted, but not freeing the node as well?

internal/unionstore/art/node.go
```diff
@@ -0,0 +1,741 @@
...
+)
+
+const (
+	// if we use 16-bit idx for nodeAddr, and break nodeAddr into embed fields, the artNodeSize can be 8 bits.
+	// By reducing 4 bits for nodeAddr, we can save about 1/4-1/3 memories for node.
+	artNodeSize = 12
+	node4size   = 80
+	node16size  = 236
+	node48size  = 888
+	node256size = 3096
```
ekexium, line 38: Is it better to replace it with `node256Size=Unsafe.SizeOf(node256{})`?

internal/unionstore/art/node.go
```diff
@@ -0,0 +1,741 @@
...
+
+func (n4 *node4) nextPresentIdx(start int) int {
+	mask := math.MaxUint8 - (uint8(1) << start) + 1 // e.g. offset=3 => 0b11111000
+	present := n4.present & mask
+	present |= 0xf0
+	return bits.TrailingZeros8(present)
+}
+
+func (n4 *node4) prevPresentIdx(start int) int {
+	mask := uint8(1<<(start+1) - 1) // e.g. start=3 => 0b000...0001111
```
ekexium, line 167: Nit: It might be clearer to add another pair of parentheses

internal/unionstore/art/node.go
```diff
@@ -0,0 +1,741 @@
...
+	present := n4.present & mask
+	zeros := bits.LeadingZeros8(present)
+	return 8 - zeros - 1
+}
+
+func (n16 *node16) init() {
```
ekexium, line 173: Can we somehow reduce duplicated code in all `init`s?
Or can we consider using generics to reduce more duplicated code? Maybe a benchmark is needed.

internal/unionstore/art_memdb.go
you06: There are some benchmark comparations between `plar/go-adaptive-radix-tree` and ours in `memdb_bench_test.go`, it should be removed later.

go.mod
you06: Maybe the memory arena without vlog is useful for them, but I don't have any performance data for it, what about discussing with them to see if they would like to adopt our improvement, which is not related to our implementation in client-go(I think client-go's requirements are quite unique).

internal/unionstore/art/arena.go
```diff
@@ -0,0 +1,358 @@
...
+
+func (f *artAllocator) getValue(valAddr nodeAddr) []byte {
+	hdrOff := valAddr.off - memdbVlogHdrSize
+	block := f.vlogAllocator.blocks[valAddr.idx].buf
+	valLen := endian.Uint32(block[hdrOff:])
+	if valLen == 0 {
+		return tombstone
+	}
+	valOff := hdrOff - valLen
+	return block[valOff:hdrOff:hdrOff]
```
ekexium, line 357: I know this line is copied from memdbArena. I'm just wondering why its capacity is specified as `hdrOff`

internal/unionstore/art/node.go
```diff
@@ -0,0 +1,719 @@
...
+const (
+	node4size   = int(unsafe.Sizeof(node4{}))
+	node16size  = int(unsafe.Sizeof(node16{}))
+	node48size  = int(unsafe.Sizeof(node48{}))
+	node256size = int(unsafe.Sizeof(node256{}))
+	leafSize    = int(unsafe.Sizeof(leaf{}))
+)
+
+var (
+	// nullNode48 is used to initialize the node 256
```
ekexium, line 40:
```suggestion
	// nullNode48 is used to initialize node48
```

internal/unionstore/art/arena.go
```diff
@@ -0,0 +1,358 @@
...
+	"math"
+	"sync/atomic"
+	"unsafe"
+
+	"github.com/tikv/client-go/v2/internal/logutil"
+	"go.uber.org/zap"
+)
+
+const (
+	alignMask = 1<<32 - 8 // 29 bit 1 and 3 bit 0.
```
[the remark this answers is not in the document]
you06: I copied it from [memdb's arena](https://github.com/tikv/client-go/blob/cd64e24de8d1a4edb9852930fdeddc4abbe07692/internal/unionstore/memdb_arena.go#L49).
My understanding is because [the address](https://github.com/tikv/client-go/blob/cd64e24de8d1a4edb9852930fdeddc4abbe07692/internal/unionstore/memdb_arena.go#L63) is 32bit integer, 32 bit mask is enough.

internal/unionstore/art/arena.go
```diff
@@ -0,0 +1,358 @@
...
+	nullAddr = nodeAddr{math.MaxUint32, math.MaxUint32}
+	endian   = binary.LittleEndian
+)
+
+type nodeAddr struct {
+	idx uint32
+	off uint32
+}
+
+func (addr nodeAddr) isNull() bool {
```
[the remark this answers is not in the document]
you06: What an inline trick.

internal/unionstore/art/art_iterator.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+	return it.prev()
+}
+
+func (it *baseIter) close() {
+	it.nodes = it.nodes[:0]
+	it.idxes = it.idxes[:0]
+}
+
+type ArtMemKeyHandle struct {
```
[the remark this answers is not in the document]
you06: Use `MemKeyHandle` will lead to cyclic imports now actually.

`ArtMemKeyHandle` is same to `MemKeyHandle` because the mem arena is same, but I think they can be different for different MemBuffer implementation.

The best choice is associated type. But associated type is not supported, maybe try generic type in `memBufferMutations` if we want it to support all the MemBuffer implementation.

internal/unionstore/art/arena.go
you06: See [Full slice expressions](https://go.dev/ref/spec#Full_slice_expressions), the capacity is set to `max - low`(`hdrOff - valOff` in our code).

internal/unionstore/art/art.go
you06: `freeNode256` is not required actually, I'll remove it.

Don't worry about the unlimited-long slice of free nodes, it's reused [here](https://github.com/you06/client-go/blob/90e7e39c64bf24309d6a8aeb401e5519ffa1dbe2/internal/unionstore/art/arena.go#L107-L109).

When a node4 is grown to node16, the old one will be put into the `freeNode4` and the next `allocNode4` can reuse it, I support the node with less capacity is allocated with higher frequency and have more chance to be recycled, so the free space of node4 is the longest by default.

internal/unionstore/art/art.go
you06: Free the node needs to change the parent of it, and I'm not inclined to store the parent address in leaf (which consumes more memory).

Note delete node in the original memdb actually do not release any space, see [comment](https://github.com/tikv/client-go/blob/cd64e24de8d1a4edb9852930fdeddc4abbe07692/internal/unionstore/memdb_arena.go#L282), the only benefit of it is the suppress the height of tree, which is not a issue for ART.

So I think mark it as deleted won't make it worse at least.

Like the [memdb's comment](https://github.com/tikv/client-go/blob/cd64e24de8d1a4edb9852930fdeddc4abbe07692/internal/unionstore/memdb_arena.go#L282), I do not implement reusing leaf not memdb by now, because of the difficulty of length-variety.

internal/unionstore/art/arena.go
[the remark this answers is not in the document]
ekexium: My point is will it overflow in a 32-bit machine?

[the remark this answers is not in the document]
ekexium: However, modifying the type would necessitate changes in numerous locations where int is currently implemented. In my opinion, we should avoid using `int` in these places... I'm comfortable with simply declaring that we don't support 32-bit systems.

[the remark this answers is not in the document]
you06: > My point is will it overflow in a 32-bit machine?

The `maxBlockSize` is set to 128MB(128 << 20), which is far below maxuint32. The offset of address is less than `maxBlockSize` so it won't overflow.

----

We can declare it don't support 32-bit systems, but I still want to use uint32 address 1 it reduces the memory usage for address. Is `const alignMask uint32 = 0xFFFFFFF8` what you mean by avoiding using `int`.

cfzjywxk:
@you06 
Please change it to draft PR or adding [DNM] flag on the title.

you06:
All the ART works are finished by other PRs.

PR #1402, opened 2024-07-29
fix issue that store's liveness may incorrectly marked as unreachable when the store restarts with label changed

### LINKED ISSUE #1401: store's liveness may incorrectly marked as unreachable when the store restarts with label changed
### Introduce

store's liveness may incorrectly marked as unreachable when the store restarts with label changed, then RPC with leader strategy on that store will always got `Region is unavailable` error.

### How it happened?

When store stopped, the request on this sore will failed and then start health check by [checkUntilHealth](https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L2622) in a separate goroutine.

https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L2665-L2669 

And there is another background worker [asyncCheckAndResolveLoop](https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L458) to do [store.reResolve](https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L2428) timely in a separate goroutine with following logic:

https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L2456-L2463

When the store **restarts with label changes**, and if the execution order is the following:

1. A new store with **same id but different label** is added, and the new store's liveness is derived from  the old one which is still `unreachable`

https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L2456-L2462

2. [checkUntilHealth](https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L2622) check the old store become reachable and return:

https://github.com/tikv/client-go/blob/f9fbc4c8f5785442a686230bf453f9292e1acd09/internal/locate/region_cache.go#L2665-L2669

3. The old store's liveness became reachable, but the new store's liveness is still unreachable, and it will always be unreachable. Actually,  the new store is just restarted and available, while the client-go will not send request to this store since it's liveness is unreachable, then all region RPC request which leader on this store will failed and return `Region is unavailable` to user.

### Affetc Branch

master: fce0abfd82b6dbc28f8950e62bec74d971839877
tidb-8.1

tidb-6.5-with-kv-timeout-feature

crazycs520:
close #1401

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -2455,12 +2455,20 @@ func (s *Store) reResolve(c *RegionCache) (bool, error) {
 	addr = store.GetAddress()
 	if s.addr != addr || !s.IsSameLabels(store.GetLabels()) {
 		newStore := &Store{storeID: s.storeID, addr: addr, saddr: store.GetStatusAddress(), storeType: storeType, labels: store.GetLabels(), state: uint64(resolved)}
-		newStore.livenessState = atomic.LoadUint32(&s.livenessState)
 		newStore.unreachableSince = s.unreachableSince
 		c.storeMu.Lock()
+		// set new store liveness state in mutex to avoid problem in https://github.com/tikv/client-go/issues/1401
+		newStore.livenessState = atomic.LoadUint32(&s.livenessState)
 		c.storeMu.stores[newStore.storeID] = newStore
 		c.storeMu.Unlock()
 		s.setResolveState(deleted)
+		logutil.BgLogger().Info("store address or labels changed, add new store and mark old store deleted",
+			zap.Uint64("store", s.storeID),
+			zap.String("old-addr", s.addr),
+			zap.Any("old-labels", s.labels),
+			zap.String("new-addr", newStore.addr),
+			zap.Any("new-labels", newStore.labels),
+			zap.String("liveness", newStore.getLivenessState().String()))
 		return false, nil
 	}
 	s.changeResolveStateTo(needCheck, resolved)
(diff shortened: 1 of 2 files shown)


internal/locate/region_cache.go
```diff
@@ -2455,12 +2455,20 @@ func (s *Store) reResolve(c *RegionCache) (bool, error) {
 	addr = store.GetAddress()
 	if s.addr != addr || !s.IsSameLabels(store.GetLabels()) {
 		newStore := &Store{storeID: s.storeID, addr: addr, saddr: store.GetStatusAddress(), storeType: storeType, labels: store.GetLabels(), state: uint64(resolved)}
-		newStore.livenessState = atomic.LoadUint32(&s.livenessState)
 		newStore.unreachableSince = s.unreachableSince
 		c.storeMu.Lock()
+		// set new store liveness state in mutex to avoid problem in https://github.com/tikv/client-go/issues/1401
+		newStore.livenessState = atomic.LoadUint32(&s.livenessState)
```
crazycs520, line 2461: Another way to fix it is always set newStore's livenessState to reachable.

zyguan: Aways setting the newStore to reachable may lead to issues like https://github.com/tikv/client-go/issues/1111 .

It seems that this PR cannot handle the case of changing of the store addr.

The root cause of the bug is that an unreachable new store can be inserted without starting health check, because the reResovle can be called by health-check-loop and check-and-resolve-loop and only health-check-loop will start health check for new store when it's unreachable. So, how about to start health check in reResolve when a unreachable new store inserted?

zyguan: LGTM, except for that the code of triggering health check in health-check-loop need to be removed.

zyguan: Besides, we may also check the resolveState in health check loop, return if the current store is deleted.

crazycs520: done, PTAL

zyguan (approved):

mittalrishabh:
can you also change the message from region unavailable to store unavailable if it can not able to connect to store. It will help to root cause these type of issues.

crazycs520:
> can you also change the message from region unavailable to store unavailable if it can not able to connect to store. It will help to root cause these type of issues.

Thanks for the suggestion, we may change this error messge later, but not in this PR, since It's not easy to change.

cfzjywxk (approved):

PR #1408, opened 2024-07-31
refine region request debug log

crazycs520:
Log region request round failed log when total backoff duration is more than 1s.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -829,7 +829,7 @@ func (s *RegionRequestSender) SendReqCtx(
 				if err := s.replicaSelector.backoffOnNoCandidate(bo); err != nil {
 					return nil, nil, retryTimes, err
 				}
-				if cost := time.Since(startTime); cost > slowLogSendReqTime || cost > timeout {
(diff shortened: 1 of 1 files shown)


internal/locate/region_request.go
```diff
@@ -829,7 +829,7 @@ func (s *RegionRequestSender) SendReqCtx(
 				if err := s.replicaSelector.backoffOnNoCandidate(bo); err != nil {
 					return nil, nil, retryTimes, err
 				}
-				if cost := time.Since(startTime); cost > slowLogSendReqTime || cost > timeout {
+				if cost := time.Since(startTime); cost > slowLogSendReqTime || cost > timeout || bo.GetTotalSleep() > 1000 {
```
cfzjywxk, line 832: Is it necessary to make the backoff time `1000` configurable?

zyguan (approved):

internal/locate/region_request.go
crazycs520: Add a config or session variable need PM agree, I don't think this is a strong demand currently.

cfzjywxk: I thinks it's ok to have a configuration for the log-threshold, but it introduces a lot of extra work like discussing with PM..

cfzjywxk (approved):

PR #1410, opened 2024-07-31
region cache: batch find regions by key ranges from cache

wshwsh12:
Currently, client-go only supports batch reading of region information from pd, but not from region cache. 
If there is a large table with 200k regions, even if all regions are in memory, it still takes 0.8-0.9s to read all regions. 
This PR optimizes the querying times to the B-tree, the time consumption reduces by about 40%.

```
This PR
Running tool: /gosdk/go1.22.1/bin/go test -benchmem -run=^$ -bench ^BenchmarkBatchLocateKeyRanges$ github.com/tikv/client-go/v2/internal/locate --tags=intest --count=1 -v

goos: linux
goarch: amd64
pkg: github.com/tikv/client-go/v2/internal/locate
cpu: 13th Gen Intel(R) Core(TM) i9-13900KF
BenchmarkBatchLocateKeyRanges
BenchmarkBatchLocateKeyRanges-32    	    1116	   1305628 ns/op	 1412405 B/op	   10921 allocs/op
PASS
ok  	github.com/tikv/client-go/v2/internal/locate	1.987s

Running tool: /gosdk/go1.22.1/bin/go test -benchmem -run=^$ -bench ^BenchmarkBatchLocateKeyRanges$ github.com/tikv/client-go/v2/internal/locate --tags=intest --count=1 -v

Master Branch
goos: linux
goarch: amd64
pkg: github.com/tikv/client-go/v2/internal/locate
cpu: 13th Gen Intel(R) Core(TM) i9-13900KF
BenchmarkBatchLocateKeyRanges
BenchmarkBatchLocateKeyRanges-32    	     502	   2367271 ns/op	 1512372 B/op	   20020 allocs/op
PASS
ok  	github.com/tikv/client-go/v2/internal/locate	1.831s
```

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1259,6 +1259,7 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
 		keyRange.StartKey = r.EndKey()
 		// Batch load rest regions from Cache.
 		batchSize := 100
+		containsAll := false
 	outer:
 		for {
 			batchRegionInCache, err := c.scanRegionsFromCache(bo, keyRange.StartKey, keyRange.EndKey, batchSize)
@@ -1267,21 +1268,23 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
 			}
 			for _, r = range batchRegionInCache {
 				if !r.Contains(keyRange.StartKey) { // uncached hole, load the rest regions
-					uncachedRanges = append(uncachedRanges, pd.KeyRange{StartKey: keyRange.StartKey, EndKey: keyRange.EndKey})
 					break outer
 				}
 				cachedRegions = append(cachedRegions, r)
 				lastRegion = r
 				if r.ContainsByEnd(keyRange.EndKey) {
 					// the range is fully hit in the region cache.
+					containsAll = true
 					break outer
 				}
 				keyRange.StartKey = r.EndKey()
 			}
 			if len(batchRegionInCache) < batchSize { // region cache miss, load the rest regions
-				uncachedRanges = append(uncachedRanges, pd.KeyRange{StartKey: keyRange.StartKey, EndKey: keyRange.EndKey})
 				break outer
 			}
+		}
+		if !containsAll {
+			uncachedRanges = append(uncachedRanges, pd.KeyRange{StartKey: keyRange.StartKey, EndKey: keyRange.EndKey})
 		}
 	}
 
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1240,24 +1240,48 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
 				keyRange.StartKey = lastRegion.EndKey()
 			}
 		}
+		// TODO: find all the cached regions in the range.
+		// now we only check if the region is cached from the lower bound, if there is a uncached hole in the middle,
+		// we will load the rest regions even they are cached.
+		r := c.tryFindRegionByKey(keyRange.StartKey, false)
+		lastRegion = r
+		if r == nil {
+			// region cache miss, add the cut range to uncachedRanges, load from PD later.
+			uncachedRanges = append(uncachedRanges, pd.KeyRange{StartKey: keyRange.StartKey, EndKey: keyRange.EndKey})
+			continue
+		}
+		// region cache hit, add the region to cachedRegions.
+		cachedRegions = append(cachedRegions, r)
+		if r.ContainsByEnd(keyRange.EndKey) {
+			// the range is fully hit in the region cache.
+			continue
+		}
+		keyRange.StartKey = r.EndKey()
+		// Batch load rest regions from Cache.
+		batchSize := 100
+	outer:
 		for {
-			// TODO: find all the cached regions in the range.
-			// now we only check if the region is cached from the lower bound, if there is a uncached hole in the middle,
-			// we will load the rest regions even they are cached.
-			r := c.tryFindRegionByKey(keyRange.StartKey, false)
-			lastRegion = r
-			if r == nil {
-				// region cache miss, add the cut range to uncachedRanges, load from PD later.
+			batchRegionInCache, err := c.scanRegionsFromCache(bo, keyRange.StartKey, keyRange.EndKey, batchSize)
+			if err != nil {
+				return nil, err
+			}
+			for _, r = range batchRegionInCache {
+				if !r.Contains(keyRange.StartKey) { // uncached hole, load the rest regions
+					uncachedRanges = append(uncachedRanges, pd.KeyRange{StartKey: keyRange.StartKey, EndKey: keyRange.EndKey})
+					break outer
+				}
+				cachedRegions = append(cachedRegions, r)
+				lastRegion = r
+				if r.ContainsByEnd(keyRange.EndKey) {
+					// the range is fully hit in the region cache.
+					break outer
+				}
+				keyRange.StartKey = r.EndKey()
+			}
+			if len(batchRegionInCache) < batchSize { // region cache miss, load the rest regions
 				uncachedRanges = append(uncachedRanges, pd.KeyRange{StartKey: keyRange.StartKey, EndKey: keyRange.EndKey})
-				break
+				break outer
 			}
-			// region cache hit, add the region to cachedRegions.
-			cachedRegions = append(cachedRegions, r)
-			if r.ContainsByEnd(keyRange.EndKey) {
-				// the range is fully hit in the region cache.
-				break
-			}
-			keyRange.StartKey = r.EndKey()
 		}
 	}
 
@@ -2078,7 +2102,8 @@ func (c *RegionCache) loadRegionByID(bo *retry.Backoffer, regionID uint64) (*Reg
 	}
 }
 
-// TODO(youjiali1995): for optimizing BatchLoadRegionsWithKeyRange, not used now.
+// For optimizing BatchLocateKeyRanges, scanRegionsFromCache scans at most `limit` regions from cache.
+// The first region's startKey is `startKey`, all regions returned are continuous and have no hole.
 func (c *RegionCache) scanRegionsFromCache(bo *retry.Backoffer, startKey, endKey []byte, limit int) ([]*Region, error) {
 	if limit == 0 {
 		return nil, nil
@@ -2089,9 +2114,6 @@ func (c *RegionCache) scanRegionsFromCache(bo *retry.Backoffer, startKey, endKey
 	defer c.mu.RUnlock()
 	regions = c.mu.sorted.AscendGreaterOrEqual(startKey, endKey, limit)
 
-	if len(regions) == 0 {
-		return nil, errors.New("no regions in the cache")
-	}
 	return regions, nil
 }
 
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1258,11 +1258,10 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
 		}
 		keyRange.StartKey = r.EndKey()
 		// Batch load rest regions from Cache.
-		batchSize := 100
 		containsAll := false
 	outer:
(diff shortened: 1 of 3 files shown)


internal/locate/region_cache.go
```diff
@@ -1240,24 +1240,48 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
 				keyRange.StartKey = lastRegion.EndKey()
 			}
 		}
+		// TODO: find all the cached regions in the range.
+		// now we only check if the region is cached from the lower bound, if there is a uncached hole in the middle,
+		// we will load the rest regions even they are cached.
+		r := c.tryFindRegionByKey(keyRange.StartKey, false)
```
you06, line 1246: Can we remove this get single region?

internal/locate/region_cache.go
```diff
@@ -2078,20 +2083,14 @@ func (c *RegionCache) loadRegionByID(bo *retry.Backoffer, regionID uint64) (*Reg
...
-	}
-
+// For optimizing BatchLocateKeyRanges, scanRegionsFromCache scans all regions from cache.
+func (c *RegionCache) scanRegionsFromCache(bo *retry.Backoffer, startKey, endKey []byte) ([]*Region, error) {
 	var regions []*Region
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	regions = c.mu.sorted.AscendGreaterOrEqual(startKey, endKey, limit)
+	regions = c.mu.sorted.DescendLessOrEqual(startKey, endKey)
+	slices.Reverse(regions)
```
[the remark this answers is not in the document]
wshwsh12: just use ascend scan maybe lost the first region.    For example, search range [b,d)  from region [a, c) [c, e),  use ascend scan can only get region [c, e)

internal/locate/sorted_btree.go
```diff
@@ -91,6 +92,32 @@ func (s *SortedRegions) AscendGreaterOrEqual(startKey, endKey []byte, limit int)
...
+	f := func(item *btreeItem) bool {
+		region := item.cachedRegion
+		if len(startKey) > 0 && bytes.Compare(region.StartKey(), startKey) <= 0 {
+			if !region.checkRegionCacheTTL(now) {
+				return false
+			}
+			regions = append(regions, region)
+			return false
+		}
+		if !region.checkRegionCacheTTL(time.Now().Unix()) {
```
you06, line 107: There is also a now~

internal/locate/sorted_btree.go
```diff
@@ -80,24 +80,40 @@ func (s *SortedRegions) SearchByKey(key []byte, isEndKey bool) (r *Region) {
...
 		}
-		if !region.checkRegionCacheTTL(time.Now().Unix()) {
+		regions = append(regions, region)
+		return len(regions) < limit
+	})
+	return regions
+}
+
+// DescendLess returns all items that are less than or equal to the key.
+func (s *SortedRegions) DescendLessOrEqual(startKey, endKey []byte) (regions []*Region) {
```
you06, line 96: Another question, why or-equal for this location func? IMO, `endKey` should be excluded bound.

wshwsh12: BTree doesn't provide the function DescendLess...

internal/locate/region_cache.go
```diff
@@ -1240,25 +1240,30 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
...
+		// TODO: find all the cached regions in the range.
+		// now we only check if the region is cached from the lower bound, if there is a uncached hole in the middle,
+		// we will load the rest regions even they are cached.
+		batchRegionInCache, err := c.scanRegionsFromCache(bo, keyRange.StartKey, keyRange.EndKey)
+		if err != nil {
+			return nil, err
+		}
+		hit := false
+		for _, r := range batchRegionInCache {
+			if !r.Contains(keyRange.StartKey) { // uncached hole, load the rest regions
```
windtalker, line 1252: I don't understand this code. If `[keyRange.StartKey, keyRange.Endkey)` contains multiple regions, in my understanding only the first region can contain `keyRange.StartKey`?

internal/locate/region_cache.go
```diff
@@ -1240,25 +1240,30 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
...
 				break
 			}
-			// region cache hit, add the region to cachedRegions.
 			cachedRegions = append(cachedRegions, r)
+			lastRegion = r
 			if r.ContainsByEnd(keyRange.EndKey) {
 				// the range is fully hit in the region cache.
+				hit = true
```
windtalker, line 1259: maybe rename `hit` to `containsAll` can be more understandable?

internal/locate/sorted_btree.go
windtalker: I'm a little concened here that this function holds a global lock, if it takes too long time, maybe this sql will affect the other sqls, especially the oltp queries.

internal/locate/region_cache.go
wshwsh12: After processing a region, I set keyRange.StartKey = r.endkey(). So the next region also can contain the  keyRange.StartKey.

internal/locate/region_cache.go
wshwsh12: Addressed.

internal/locate/region_cache.go
wshwsh12: No. If we need to retrieve all regions by AscendGreaterOrEqual, since the Key of B-Tree is StartKey, we must retrieve the first region by SearchByKey.

internal/locate/region_cache.go
```diff
@@ -1240,24 +1240,51 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
...
+		}
+		// region cache hit, add the region to cachedRegions.
+		cachedRegions = append(cachedRegions, r)
+		if r.ContainsByEnd(keyRange.EndKey) {
+			// the range is fully hit in the region cache.
+			continue
+		}
+		keyRange.StartKey = r.EndKey()
+		// Batch load rest regions from Cache.
+		batchSize := 100
```
windtalker, line 1261: I think you can use `defaultRegionsPerBatch` as the batchSize?

internal/locate/region_cache.go
```diff
@@ -1240,24 +1240,51 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
...
+				lastRegion = r
+				if r.ContainsByEnd(keyRange.EndKey) {
+					// the range is fully hit in the region cache.
+					containsAll = true
+					break outer
+				}
+				keyRange.StartKey = r.EndKey()
+			}
+			if len(batchRegionInCache) < batchSize { // region cache miss, load the rest regions
+				break outer
```
windtalker, line 1283: looks like just use `break` here is ok

internal/locate/region_cache.go
```diff
@@ -2078,7 +2105,8 @@ func (c *RegionCache) loadRegionByID(bo *retry.Backoffer, regionID uint64) (*Reg
 	}
 }
 
-// TODO(youjiali1995): for optimizing BatchLoadRegionsWithKeyRange, not used now.
+// For optimizing BatchLocateKeyRanges, scanRegionsFromCache scans at most `limit` regions from cache.
+// The first region's startKey is `startKey`, all regions returned are continuous and have no hole.
```
windtalker, line 2109: maybe you need to emphasize that it is the caller's responsibility to make sure that `startKey` is a node in the B-tree,  otherwise, the `startKey` will not be included in the return regions?

internal/locate/sorted_btree.go
```diff
@@ -79,12 +80,22 @@ func (s *SortedRegions) SearchByKey(key []byte, isEndKey bool) (r *Region) {
...
+	lastStartKey := startKey
 	s.b.AscendGreaterOrEqual(newBtreeSearchItem(startKey), func(item *btreeItem) bool {
 		region := item.cachedRegion
 		if len(endKey) > 0 && bytes.Compare(region.StartKey(), endKey) >= 0 {
 			return false
 		}
+		if !region.checkRegionCacheTTL(now) {
+			return false
+		}
+		if len(lastStartKey) > 0 && !region.Contains(lastStartKey) { // uncached hole
```
windtalker, line 95: when will `len(lastStartKey) == 0`?

wshwsh12: Some test cases have the key, for -inf to +inf.

you06: The case `len(lastStartKey) == 0` should never happen in current usage(start key is assigned [here](https://github.com/tikv/client-go/blob/9e5fff01aef323cbb5b38699e65d64278a6ebb7c/internal/locate/region_cache.go#L1259), and it's never empty if the code runs here).

When `len(lastStartKey) == 0` and the first region is not start from -inf, we should stop scan regions, see the following test.

```go
func (s *testRegionRequestToSingleStoreSuite) TestRegionCacheStartNonEmpty() {
	_ = s.cache.refreshRegionIndex(s.bo)
	r, _ := s.cache.scanRegionsFromCache(s.bo, []byte{}, nil, 10)
	s.Equal(len(r), 1)

	region, _ := s.cache.LocateRegionByID(s.bo, s.region)
	v2 := region.Region.confVer + 1
	r2 := metapb.Region{Id: region.Region.id, RegionEpoch: &metapb.RegionEpoch{Version: region.Region.ver, ConfVer: v2}, StartKey: []byte{1}}
	st := newUninitializedStore(s.store)

	s.cache.mu.Lock()
	s.cache.mu.sorted.Clear()
	s.cache.mu.Unlock()
	// region cache after clear: []

	s.cache.insertRegionToCache(&Region{meta: &r2, store: unsafe.Pointer(st), ttl: nextTTLWithoutJitter(time.Now().Unix())}, true, true)
	// region cache after insert: [[1, +inf)]

	r, _ = s.cache.scanRegionsFromCache(s.bo, []byte{}, nil, 10)
	s.Equal(len(r), 0) // fail in current implementation
}
```

wshwsh12: I add the test `TestRegionCacheStartNonEmpty` and remove the `len(lastStartKey) == 0` check.  Now the code stop scan regions when `len(lastStartKey) == 0` and the first region is not start from -inf.   PTAL again.

windtalker:
lgtm

internal/locate/region_cache_test.go
```diff
@@ -2259,13 +2260,35 @@ func (s *testRegionRequestToSingleStoreSuite) TestRefreshCache() {
...
+	s.cache.mu.Lock()
+	s.cache.mu.sorted.Clear()
+	s.cache.mu.Unlock()
+	// region cache after clear: []
+
+	s.cache.insertRegionToCache(&Region{meta: &r2, store: unsafe.Pointer(st), ttl: nextTTLWithoutJitter(time.Now().Unix())}, true, true)
+	// region cache after insert: [[1, +inf)]
+
+	r, _ = s.cache.scanRegionsFromCache(s.bo, []byte{}, nil, 10)
+	s.Equal(len(r), 0) // fail in current implementation
```
you06, line 2289: Remove the comment `// fail ...`

internal/locate/region_cache_test.go
```diff
@@ -2259,13 +2260,35 @@ func (s *testRegionRequestToSingleStoreSuite) TestRefreshCache() {
 	s.cache.insertRegionToCache(&Region{meta: &r2, store: unsafe.Pointer(st), ttl: nextTTLWithoutJitter(time.Now().Unix())}, true, true)
 
 	r, _ = s.cache.scanRegionsFromCache(s.bo, []byte{}, nil, 10)
-	s.Equal(len(r), 2)
+	s.Equal(len(r), 1)
```
you06, line 2263: This is not related to this PR I think.

Since region cache doesn't remove the first intersected region(it scan intersected region by `AscendGreaterOrEqual`), the outdated region (-inf, inf) is still alive and the new inserted valid region [{1}, inf) is ignored because the first seen region (-inf, inf) contains all the required ranges... We finally got outdated region though we know some valid regions.

Can you add a TODO in this test?

you06 (commented):
rest LGTM

you06 (approved):

PR #1411, opened 2024-08-01
improve batch efficiency for high throughput workloads

zyguan:
Currently, we can only batch more requests when the tikv is detected as overload. This PR allow adaptively batching more requests according to request arrival intervals. Workloads with high throughput (like sysbench oltp point select) can benfit a lot from it because batching more requests reduces the grpc overhead and the number of syscalls.

ref: https://github.com/tikv/client-go/pull/1366 & https://github.com/tikv/client-go/pull/1373


internal/client/client_batch.go
```diff
@@ -321,16 +376,101 @@ func (a *batchConn) fetchMorePendingRequests(
...
+type turboBatchOptions struct {
+	V int     `json:"v"`
+	N int     `json:"n,omitempty"`
+	T float64 `json:"t,omitempty"`
+	W float64 `json:"w,omitempty"`
+	P float64 `json:"p,omitempty"`
+	Q float64 `json:"q,omitempty"`
+}
+
+type turboBatchTrigger struct {
```
cfzjywxk, line 414: Need to add comments from the design doc here to introduce the `turboBatchTrigger` and `turboBatchOptions`  concepts briefly.

The detailed design document is also neede to be referenced in the github issue.

internal/client/client_batch.go
```diff
@@ -72,6 +75,10 @@ type batchCommandsEntry struct {
 	canceled int32
 	err      error
 	pri      uint64
+
+	start   time.Time
```
cfzjywxk, line 79:
```suggestion
        // start indicates when the batch commands entry is generated and sent to the batch conn channel.
	start   time.Time
```

cfzjywxk (approved):
LGTM

internal/client/client_batch.go
```diff
@@ -296,23 +342,33 @@ func (a *batchConn) fetchMorePendingRequests(
...
+	} else {
+		a.fetchMoreTimer.Reset(maxWaitTime)
+	}
 	for a.reqBuilder.len() < batchWaitSize {
 		select {
 		case entry := <-a.batchCommandsCh:
 			if entry == nil {
+				if !a.fetchMoreTimer.Stop() {
+					<-a.fetchMoreTimer.C
+				}
```
crazycs520, line 356: Why wait timeout here, because the following line will be returned anyway, and no more entries will be fetched after waiting

internal/client/client_batch.go
```diff
@@ -344,11 +502,20 @@ func (a *batchConn) batchSendLoop(cfg config.TiKVClient) {
 		}
 	}()
 
-	bestBatchWaitSize := cfg.BatchWaitSize
+	trigger, ok := newTurboBatchTriggerFromPolicy(cfg.BatchPolicy)
+	if !ok {
+		initBatchPolicyWarn.Do(func() {
+			logutil.BgLogger().Warn("fallback to default batch policy due to invalid value", zap.String("value", cfg.BatchPolicy))
```
crazycs520, line 508: When the configuration is incorrect, this log may be printed too frequently.

ekexium (approved):

internal/client/client_batch.go
ekexium: It's in a `sync.Once`

ekexium:
Some thoughts: Have we considered node load metrics like CPU utilization (for TiDB, and possibly TiKV) as factors in the algorithm? It might be worth exploring an adaptive approach that accounts for these variables.

internal/client/client_batch.go
```diff
@@ -357,27 +524,43 @@ func (a *batchConn) batchSendLoop(cfg config.TiKVClient) {
...
-				a.fetchMorePendingRequests(int(cfg.MaxBatchSize), int(bestBatchWaitSize), cfg.MaxBatchWaitTime)
+				a.fetchMorePendingRequests(int(cfg.MaxBatchSize), int(cfg.BatchWaitSize), cfg.MaxBatchWaitTime)
+			} else if turboBatchWaitTime > 0 && headArrivalInterval > 0 && trigger.needFetchMore(headArrivalInterval) {
+				n, m := math.Modf(avgBatchWaitSize)
+				batchWaitSize := int(n)
+				if trigger.opts.V == 0 {
+					batchWaitSize = int(cfg.BatchWaitSize)
+				} else if m >= trigger.opts.Q {
+					batchWaitSize++
+				}
```
crazycs520, line 539: Could you extract this into a function and add tests to it?

internal/client/client_batch.go
```diff
@@ -321,16 +377,118 @@ func (a *batchConn) fetchMorePendingRequests(
...
+	ok = true
+	return
+}
+
+func (t *turboBatchTrigger) turboWaitSeconds() float64 {
+	return t.opts.T
+}
+
+func (t *turboBatchTrigger) turboWaitTime() time.Duration {
+	return time.Duration(t.opts.T * float64(time.Second))
```
MyonKeminta, line 457: Thanks golang...

zyguan:
> Some thoughts: Have we considered node load metrics like CPU utilization (for TiDB, and possibly TiKV) as factors in the algorithm? We have a special branch that handle the case where TiKV gRPC pool utilization exceeds 200%, which seems a bit sub-optimal to me.
> 
> It might be worth exploring an adaptive approach that accounts for these variables.

Yes, the load and cpu util have been considered. Since the feature is preferred to be turned on by default, the most important task is to eliminate the perf regression under light workloads. The main reason for the regression is the extra wait time introduced by `fetchMoreRequests`, and it's directly related to the request arrival interval. If the interval is much longer than the max wait time, then it barely waits for more requests. However, it's hard to modeling the relationship between the regression and the load / cpu util.

MyonKeminta (approved):

PR #1414, opened 2024-08-02
txn-file: Fix miss slice after region error is met or commit failed

pingyu:
### Change

- Sort and deduplicate txn chunks slice after region error is met.
- Fix the ignorance of commit error.


pingyu:
The lint error will be fixed by #1415.

coocood (approved):

PR #1423, opened 2024-08-14
region_request: remove backoff for stale read

you06:
ref pingcap/tidb#55388

We always retry stale read with snapshot read, so the backoff for it can be removed, because wait for safe ts catching up is meaningless for snapshot read.


crazycs520 (approved):

cfzjywxk (approved):
LGTM

cfzjywxk:
@Tema 
Yes it would be picked to all the LTS version.

PR #1424, opened 2024-08-14
transaction: allow cleanup after the connection is killed (WIP, DNM)

YangKeao:
This PR fixed two issues:

1. The `cleanup` logic doesn't take effect for the commit in async commit stage. (Though, I'm not sure whether it's a mistake or by design. It seems good now.)
2. Allow `cleanup` to run after the connection is killed. This makes it possible to clean up the lock before TiDB shutdown.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -1075,9 +1075,17 @@ func (c *twoPhaseCommitter) doActionOnBatches(
 					status,
 				),
 			)
-			// TODO: There might be various signals besides a query interruption,
-			// but we are unable to differentiate them, because the definition is in TiDB.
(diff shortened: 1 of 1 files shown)


txnkv/transaction/2pc.go
```diff
@@ -1638,6 +1644,10 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		return err
 	}
 
+	if c.txn.resourceGroupName == "testslowcommit" {
```
YangKeao, line 1647: Some tests dirty codes. @YangKeao please remember to remove it later.

YangKeao:
It's better to add hooks to wait for background goroutines. https://github.com/tikv/client-go/pull/1432.

This PR doesn't work, and I'm not going to investigate it. Close now.

PR #1426, opened 2024-08-20
membuffer: refactor the memdb to support multi implementations

you06:
ref pingcap/tidb#55287

This PR mainly contains 3 changes:
- Split the memory arena and vlog from the red-black tree memory arena implementation.
- Add the empty framework of ART.
- Make ART, RBT and memory arena different packages.


internal/unionstore/art.go
```diff
@@ -0,0 +1,210 @@
...
+import (
+	"math"
+	"sync"
+
+	tikverr "github.com/tikv/client-go/v2/error"
+	"github.com/tikv/client-go/v2/kv"
+)
+
+type ART struct {
+	sync.RWMutex
```
cfzjywxk, line 27: What are the fields that should be protected by the mutex?

internal/unionstore/art.go
```diff
@@ -0,0 +1,210 @@
...
+	skipMutex       bool
+	allocator       artAllocator
+	root            artNode
```
cfzjywxk, line 30: Perhaps wrapping the art related fields together, and it would be conveneient if we want to abstract it into a standalone library in the future.

internal/unionstore/art_node.go
```diff
@@ -0,0 +1,54 @@
...
+	typeARTNode48  artNodeKind = 3
+	typeARTNode256 artNodeKind = 4
+	typeARTLeaf    artNodeKind = 5
+)
+
+var nullArtNode = artNode{kind: typeARTInvalid, addr: nullAddr}
+
+type artKey []byte
+
+type artNode struct {
```
cfzjywxk, line 36: Ditto, the direcotry structure could be:
```
art_based_memdb.go  // MemDB implementation based on the art, including memdb related functionalities.
art_dir
   art.go
   art_node.go
   art_iterator.go
   art_test.go
rbt_based_memdb.go
rbt_dir
   rbt.go
   rbt_iterator.go
   rbt_test.go
memdb_test.go
memdb_bench.go
```

zyguan (approved):

internal/unionstore/art.go
you06: It's required by the membuffer interface.

https://github.com/tikv/client-go/blob/75e3705e58f1c9f68cce3467de4429ea959b59b7/internal/unionstore/union_store.go#L171-L174

In TiDB, some SQL executors will read and write membuffer concurrently, the membuffer needs to be guarded by mutex in this case.

cfzjywxk: As it's a memdb requirement/constraint instead of an `ART` one, how about
```
type ARTMemDB struct {
     sync.RWMutex
     artImpl  ART // ART is a single threaded data strcuture implementation.
}
```

cfzjywxk (approved):

PR #1389, opened 2024-07-13
Let memdb cache last traversed node

ekexium:
A common usage in TiDB is to access a same key multiple times consecutively. We can save subsequent traversals when this happen.

### Sysbench (standard DML)

I tested oltp_write_only, no diff found compared with master.

### YCSB benchmark (pipelined DML)

In a YCSB 3M rows benchmark, the computation duration (total duration - flush wait) decreases, proving the optimization works as expected. We also observe a significant increase in flush_wait when applying this optimization. That should be a standalone problem to track and solve, IMO.

Master
insert 47.73-9.74=37.99s
update 30.74-2.78=27.96s
delete 23.81-5.98=17.83s
```
[pipelined_flush.go:304] ["[pipelined dml] start to commit transaction"] [conn=2669674502] [session_alias=] [keys=5999998] [flush_wait_duration=9.742798374s] [total_duration=47.735029816s] [size=3.506GB] [startTS=451525108111769604]

[pipelined_flush.go:304] ["[pipelined dml] start to commit transaction"] [conn=2669674506] [session_alias=] [keys=2999999] [flush_wait_duration=2.787723191s] [total_duration=30.74270092s] [size=3.243GB] [startTS=451525199285190657]

[pipelined_flush.go:304] ["[pipelined dml] start to commit transaction"] [conn=2669674510] [session_alias=] [keys=5999998] [flush_wait_duration=5.983734665s] [total_duration=23.811122881s] [size=198MB] [startTS=451525285989580803]
```

This PR
insert 43.12-10.50=32.62s
update 31.76-4.96=26.8s
delete 28.82-15.06=13.76s
```
[pipelined_flush.go:304] ["[pipelined dml] start to commit transaction"] [conn=3839885318] [session_alias=] [keys=5999998] [flush_wait_duration=10.507358104s] [total_duration=43.121563977s] ["memdb cache hit count"=11999996] ["memdb cache miss count"=11999996] [size=3.506GB] [startTS=451516433638621187]

[pipelined_flush.go:304] ["[pipelined dml] start to commit transaction"] [conn=3839885320] [session_alias=] [keys=2999999] [flush_wait_duration=4.96714099s] [total_duration=31.767803066s] ["memdb cache hit count"=5999998] ["memdb cache miss count"=2999999] [size=3.243GB] [startTS=451516523593334791]

[...]

PR #1453, opened 2024-08-27
memdb: use atomic u64 addr to reduce allocation

you06:
After #1389, we cache the last traversed node in memdb, but it's an atomic pointer, which need a heap allocation every time we update it, so the benchmark has little regression.

```
go test ./internal/unionstore -v -benchmem -bench=BenchmarkMemDbBufferSequential -run=^a -benchtime=1x

before  #1389
BenchmarkMemDbBufferSequential-8               1          42312917 ns/op        12576704 B/op         31 allocs/op

after #1389
BenchmarkMemDbBufferSequential-8               1          46407916 ns/op        15776704 B/op     200031 allocs/op

this pr
BenchmarkMemDbBufferSequential-8               1          44018167 ns/op        12576704 B/op         31 allocs/op
```

This PR also fixes some bugs of memdb:
- The refactor leave a mistake of the snapshot reverse iterator.
- The cached addr should be reset when memdb is reset.

### DIFF
--- a/internal/unionstore/rbt/rbt.go
+++ b/internal/unionstore/rbt/rbt.go
@@ -72,8 +72,9 @@ type RBT struct {
 	dirty       bool
 	stages      []arena.MemDBCheckpoint
 
-	// The lastTraversedNode must exist
-	lastTraversedNode atomic.Pointer[MemdbNodeAddr]
+	// The lastTraversedNode stores addr in uint64 of the last traversed node.
+	// Compare to atomic.Pointer, atomic.Uint64 can avoid allocation so it's more efficient.
+	lastTraversedNode atomic.Uint64
 	hitCount          atomic.Uint64
 	missCount         atomic.Uint64
 }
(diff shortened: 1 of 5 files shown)


internal/unionstore/rbt/rbt.go
```diff
@@ -85,24 +86,26 @@ func New() *RBT {
 	db.stages = make([]arena.MemDBCheckpoint, 0, 2)
 	db.entrySizeLimit = unlimitedSize
 	db.bufferSizeLimit = unlimitedSize
-	db.lastTraversedNode.Store(&nullNodeAddr)
+	db.lastTraversedNode.Store(math.MaxUint64)
```
ekexium, line 89: We can define a const var like `NullU64Addr` next to `NullAddr`

ekexium (approved):

you06:
@cfzjywxk PTAL

cfzjywxk (approved):

PR #1458, opened 2024-09-05
feat: Broadcast min_commit_ts for pipelined transactions

ekexium:
Ref: https://github.com/tikv/tikv/issues/17459
Prerequires https://github.com/pingcap/kvproto/pull/1265

Changes
1. Periodic updates of min_commit_ts
2. Broadcast min_commit_ts
3. Periodic updates of full store list

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -728,7 +728,7 @@ func NewRegionCache(pdClient pd.Client, opt ...RegionCacheOpt) *RegionCache {
 	}
 	c.bg.schedule(
 		func(ctx context.Context, _ time.Time) bool {
-			refreshFullStoreList(ctx, c.PDClient(), c.stores)
+			refreshFullStoreList(ctx, c.stores)
 			return false
 		}, refreshStoreListInterval,
 	)
@@ -736,8 +736,8 @@ func NewRegionCache(pdClient pd.Client, opt ...RegionCacheOpt) *RegionCache {
 }
 
 // Try to refresh full store list. Errors are ignored.
-func refreshFullStoreList(ctx context.Context, pdClient pd.Client, stores storeCache) {
-	storeList, err := pdClient.GetAllStores(ctx)
+func refreshFullStoreList(ctx context.Context, stores storeCache) {
+	storeList, err := stores.fetchAllStores(ctx)
 	if err != nil {
 		logutil.Logger(ctx).Info("refresh full store list failed", zap.Error(err))
 		return
@@ -747,17 +747,16 @@ func refreshFullStoreList(ctx context.Context, pdClient pd.Client, stores storeC
 		if exist {
 			continue
 		}
-		s := stores.getOrInsertDefault(store.GetId())
 		// GetAllStores is supposed to return only Up and Offline stores.
 		// This check is being defensive and to make it consistent with store resolve code.
 		if store == nil || store.GetState() == metapb.StoreState_Tombstone {
-			s.setResolveState(tombstone)
 			continue
 		}
 		addr := store.GetAddress()
 		if addr == "" {
 			continue
 		}
+		s := stores.getOrInsertDefault(store.GetId())
 		// TODO: maybe refactor this, together with other places initializing Store
 		s.addr = addr
 		s.peerAddr = store.GetPeerAddress()
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -2686,7 +2686,7 @@ func (c *RegionCache) UpdateBucketsIfNeeded(regionID RegionVerID, latestBucketsV
 
 const cleanCacheInterval = time.Second
 const cleanRegionNumPerRound = 50
-const refreshStoreListInterval = time.Second
+const refreshStoreListInterval = 10 * time.Second
 
 // gcScanItemHook is only used for testing
 var gcScanItemHook = new(atomic.Pointer[func(*btreeItem)])
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -726,7 +726,44 @@ func NewRegionCache(pdClient pd.Client, opt ...RegionCacheOpt) *RegionCache {
 		// cache GC is incompatible with cache refresh
 		c.bg.schedule(c.gcRoundFunc(cleanRegionNumPerRound), cleanCacheInterval)
 	}
+	c.bg.schedule(
+		func(ctx context.Context, _ time.Time) bool {
+			refreshFullStoreList(ctx, c.PDClient(), c.stores)
+			return false
+		}, refreshStoreListInterval,
+	)
 	return c
+}
+
+// Try to refresh full store list. Errors are ignored.
+func refreshFullStoreList(ctx context.Context, pdClient pd.Client, stores storeCache) {
+	storeList, err := pdClient.GetAllStores(ctx)
+	if err != nil {
+		logutil.Logger(ctx).Info("refresh full store list failed", zap.Error(err))
+		return
+	}
+	for _, store := range storeList {
+		s, exist := stores.get(store.GetId())
+		if exist {
+			continue
+		}
+		s = stores.getOrInsertDefault(store.GetId())
+		if store == nil || store.GetState() == metapb.StoreState_Tombstone {
+			s.setResolveState(tombstone)
+			return
+		}
+		addr := store.GetAddress()
+		if addr == "" {
+			return
+		}
+		// TODO: maybe refactor this, together with other places initializing Store
+		s.addr = addr
+		s.peerAddr = store.GetPeerAddress()
+		s.saddr = store.GetStatusAddress()
+		s.storeType = tikvrpc.GetStoreTypeByMeta(store)
+		s.labels = store.GetLabels()
+		s.changeResolveStateTo(unresolved, resolved)
+	}
 }
 
 // only used fot test.
@@ -2649,6 +2686,7 @@ func (c *RegionCache) UpdateBucketsIfNeeded(regionID RegionVerID, latestBucketsV
 
 const cleanCacheInterval = time.Second
 const cleanRegionNumPerRound = 50
+const refreshStoreListInterval = time.Second
 
 // gcScanItemHook is only used for testing
 var gcScanItemHook = new(atomic.Pointer[func(*btreeItem)])
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -750,11 +750,11 @@ func refreshFullStoreList(ctx context.Context, pdClient pd.Client, stores storeC
 		s := stores.getOrInsertDefault(store.GetId())
 		if store == nil || store.GetState() == metapb.StoreState_Tombstone {
 			s.setResolveState(tombstone)
-			return
+			continue
 		}
 		addr := store.GetAddress()
 		if addr == "" {
-			return
+			continue
 		}
 		// TODO: maybe refactor this, together with other places initializing Store
 		s.addr = addr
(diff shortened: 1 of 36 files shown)


txnkv/transaction/2pc.go
```diff
@@ -164,7 +164,7 @@ type twoPhaseCommitter struct {
 	}
 
 	useAsyncCommit    uint32
-	minCommitTS       uint64
+	minCommitTS       *minCommitTsManager
```
cfzjywxk, line 167:
```suggestion
	minCommitTSMgr       *minCommitTsManager
```

txnkv/transaction/2pc.go
```diff
@@ -1140,6 +1141,69 @@ const (
 	stateClosed
 )
 
+// WriteAccessLevel represents the level of write access required to modify the value
```
cfzjywxk, line 1144: Is `phase` better than `level` in the definition?

txnkv/transaction/2pc.go
```diff
@@ -1268,19 +1346,105 @@ func keepAlive(
...
+	}
+	req.Context.ResourceGroupTag = resourceGroupTag
+
+	var wg sync.WaitGroup
+	errChan := make(chan error, len(stores))
+	semaphore := make(chan struct{}, broadcastMaxConcurrency)
+
+	for _, s := range stores {
+		wg.Add(1)
+		go func(addr string) {
```
cfzjywxk, line 1407: Is it possible to unify the goroutine usage from the `KVTxn.spawnWithStorePool` to make use of the pool?

txnkv/transaction/2pc.go
```diff
@@ -1268,19 +1346,105 @@ func keepAlive(
...
+	req.Context.ResourceControlContext = &kvrpcpb.ResourceControlContext{
+		ResourceGroupName: resourceGroupName,
+	}
+	req.Context.ResourceGroupTag = resourceGroupTag
+
+	var wg sync.WaitGroup
+	errChan := make(chan error, len(stores))
+	semaphore := make(chan struct{}, broadcastMaxConcurrency)
+
+	for _, s := range stores {
```
cfzjywxk, line 1405: Perhaps a more common pattern is 
```
// Create the store task channel from the slice.
taskChan := make(chan *store, len(stores))  

// Launch a limited number of workers, or use the gorouinte pool.
for i := 0; i < broadcastMaxConcurrency; i++ {
	wg.Add(1)
	go func() {
		defer wg.Done()
		for s := range taskChan {  // Receive *store from taskChan
			_, err := store.GetTiKVClient().SendRequest(
				bo.GetCtx(),
				s.GetAddr(),
				req,
				broadcastRpcTimeout,
			)
			if err != nil {
				errChan <- err
			}
		}
	}()
}
```

txnkv/transaction/2pc.go
ekexium: I don't get it. Did you mean using `c.store.Go()`?

cfzjywxk: Yes, there are several wrappers based on the `kvstore.Go` interface, like the `KVTxn.spawnWithStorePool`. As a result there are different ways of spwaning goroutines in client-go by now, it's better to unify the usages if possible

txnkv/transaction/2pc.go
```diff
@@ -1268,19 +1346,109 @@ func keepAlive(
...
+	req.Context.ResourceControlContext = &kvrpcpb.ResourceControlContext{
+		ResourceGroupName: resourceGroupName,
+	}
+	req.Context.ResourceGroupTag = resourceGroupTag
+
+	var wg sync.WaitGroup
+	errChan := make(chan error, len(stores))
+	taskChan := make(chan *locate.Store, len(stores))
+
+	for i := 0; i < broadcastMaxConcurrency; i++ {
```
you06, line 1405:
```suggestion
	concurrency := min(broadcastMaxConcurrency, len(stores))
	taskChan := make(chan *locate.Store, concurrency)

	for i := 0; i < concurrency; i++ {
```

txnkv/transaction/pipelined_flush.go
```diff
@@ -338,6 +338,23 @@ func (c *twoPhaseCommitter) commitFlushedMutations(bo *retry.Backoffer) error {
 		zap.Uint64("startTS", c.startTS),
 		zap.Uint64("commitTS", commitTS),
 	)
+	broadcastToAllStores(
```
you06, line 341: The function call will wait for the broadcast done in current goroutine, though this won't affect too much for p-dml, shall we put it in an individual goroutine instead of blocking the statement from returning success to client?

you06 (commented):
BTW, this is not related to this PR, can you change the `RLock/RUnlock` to `Lock/Unlock` [here](https://github.com/tikv/client-go/blob/ad0302eef1bf688a90020b52884ed64437844b95/txnkv/transaction/pipelined_flush.go#L333-L335)?

I have a question about the future plan, have you considered apply this broadcast also for slow or large 2pc transactions?

internal/locate/region_cache.go
```diff
@@ -2649,6 +2686,7 @@ func (c *RegionCache) UpdateBucketsIfNeeded(regionID RegionVerID, latestBucketsV
 
 const cleanCacheInterval = time.Second
 const cleanRegionNumPerRound = 50
+const refreshStoreListInterval = time.Second
```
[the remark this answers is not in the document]
cfzjywxk: Make sense, it could be lowered.

txnkv/transaction/2pc.go
cfzjywxk: +1

internal/locate/region_cache.go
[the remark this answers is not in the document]
ekexium: It affects the max resolved-ts lag when new store comes online. I've changed it to 10s.

txnkv/transaction/pipelined_flush.go
ekexium: Good idea

txnkv/transaction/2pc.go
ekexium: Gotcha. I didn't find `spawnWithStorePool` as it is recently merged

ekexium:
> BTW, this is not related to this PR, can you change the `RLock/RUnlock` to `Lock/Unlock` [here](https://github.com/tikv/client-go/blob/ad0302eef1bf688a90020b52884ed64437844b95/txnkv/transaction/pipelined_flush.go#L333-L335)?

Done.

> I have a question about the future plan, have you considered apply this broadcast also for slow or large 2pc transactions?

It's possible. But I think we need more empirical data before implementing that. It's unclear now how this optimization would benefit normal large transactions.

txnkv/transaction/2pc.go
ekexium: In my initial implementation I used `phase` too. However, I switched to `level` as it's more intuitive. `level` is self-explanatory, whereas `phase` might require additional context to understand.

internal/locate/region_cache.go
```diff
@@ -726,9 +726,46 @@ func NewRegionCache(pdClient pd.Client, opt ...RegionCacheOpt) *RegionCache {
...
+	}
+	for _, store := range storeList {
+		_, exist := stores.get(store.GetId())
+		if exist {
+			continue
+		}
+		s := stores.getOrInsertDefault(store.GetId())
+		if store == nil || store.GetState() == metapb.StoreState_Tombstone {
+			s.setResolveState(tombstone)
+			return
```
crazycs520, line 753:
```suggestion
			continue
```

txnkv/transaction/pipelined_flush.go
```diff
@@ -439,13 +457,19 @@ func (c *twoPhaseCommitter) resolveFlushedLocks(bo *retry.Backoffer, start, end
 	const RESOLVE_CONCURRENCY = 8
 	var resolved atomic.Uint64
 	handler, err := c.buildPipelinedResolveHandler(commit, &resolved)
+	var commitTs uint64
+	if commit {
+		commitTs = atomic.LoadUint64(&c.commitTS)
+	} else {
+		commitTs = 0
+	}
```
crazycs520, line 465:
```suggestion
	commitTs := uint64(0)
	if commit {
		commitTs = atomic.LoadUint64(&c.commitTS)
	}
```

txnkv/transaction/prewrite.go
```diff
@@ -413,14 +413,14 @@ func (action actionPrewrite) handleSingleBatch(
 						return nil
 					}
 					logutil.Logger(bo.GetCtx()).Warn(
-						"async commit cannot proceed since the returned minCommitTS is zero, "+
+						"async commit cannot proceed since the returned minCommitTSMgr is zero, "+
```
crazycs520, line 416: No need to change this?

internal/locate/region_cache.go
```diff
@@ -726,9 +726,46 @@ func NewRegionCache(pdClient pd.Client, opt ...RegionCacheOpt) *RegionCache {
...
+			continue
+		}
+		s := stores.getOrInsertDefault(store.GetId())
+		if store == nil || store.GetState() == metapb.StoreState_Tombstone {
+			s.setResolveState(tombstone)
+			continue
+		}
+		addr := store.GetAddress()
+		if addr == "" {
+			continue
```
cfzjywxk, line 757: Better to explain in the comments when the `pdClient.GetAllStores(ctx)` may return a store list containg `nil` store and `emptry address` and how should they be treated. For example, dose it mean the returned  `store list` is  invalid and `pdClient.GetAllStores(ctx)` should be retried when there invalid element in the list?

ekexium: I believe that `GetAllStores` is supposed to return only stores in Up and Offline states. This additional check serves as a defensive measure and ensures consistency with the store resolve procedure

internal/locate/region_cache.go
```diff
@@ -726,9 +726,48 @@ func NewRegionCache(pdClient pd.Client, opt ...RegionCacheOpt) *RegionCache {
...
+		// GetAllStores is supposed to return only Up and Offline stores.
+		// This check is being defensive and to make it consistent with store resolve code.
+		if store == nil || store.GetState() == metapb.StoreState_Tombstone {
+			s.setResolveState(tombstone)
+			continue
+		}
+		addr := store.GetAddress()
+		if addr == "" {
+			continue
+		}
```
zyguan, line 760: Can we move this before `stores.getOrInsertDefault`?  `stores.getOrInsertDefault(store.GetId())` will insert a store{id:0} if `store == nil`, and inserting a tombstone or a store whose addr is unknown seems pointless.

txnkv/transaction/2pc.go
```diff
@@ -1265,19 +1343,119 @@ func keepAlive(
...
+
+		for _, s := range stores {
+			taskChan <- s
+		}
+
+		close(taskChan)
+		wg.Wait()
+	}
+
+	if err := txn.spawnWithStorePool(broadcastFunc); err != nil {
```
zyguan, line 1441: Since we've already used a goroutine pool, how about spawning broadcast tasks like the following:
```go
	stores := store.GetRegionCache().GetStoresByType(tikvrpc.TiKV)
	rateLimit := make(chan struct{}, min(broadcastMaxConcurrency, len(stores)))
	for _, s := range stores {
		rateLimit <- struct{}{}
		target := s
		err := txn.spawnWithStorePool(func() {
			defer func() {
				<-rateLimit
			}()
			// send request to target
		})
		if err != nil {
			<-rateLimit
		}
	}
```

zyguan (approved):

cfzjywxk (approved):

PR #1465, opened 2024-09-19
 membuffer: support staging & checkpoint for ART

you06:
ref https://github.com/pingcap/tidb/issues/55287

This PR implements `Staging/Release/Cleanup` and `Checkpoint/RevertToCheckpoint` for ART, the behavior is exactly same with RBT.

### DIFF
--- a/internal/unionstore/art/art.go
+++ b/internal/unionstore/art/art.go
@@ -398,6 +398,9 @@ func (t *ART) checkpoint() arena.MemDBCheckpoint {
 
 func (t *ART) RevertNode(hdr *arena.MemdbVlogHdr) {
 	lf := t.allocator.getLeaf(hdr.NodeAddr)
+	if lf == nil {
+		panic("revert an invalid node")
+	}
 	lf.vAddr = hdr.OldValue
 	t.size -= int(hdr.ValueLen)
 	if hdr.OldValue.IsNull() {
--- a/internal/unionstore/art/art.go
+++ b/internal/unionstore/art/art.go
@@ -396,7 +396,7 @@ func (t *ART) checkpoint() arena.MemDBCheckpoint {
 	return t.allocator.vlogAllocator.Checkpoint()
 }
 
-func (t *ART) RevertNode(hdr *arena.MemdbVlogHdr) {
+func (t *ART) RevertVAddr(hdr *arena.MemdbVlogHdr) {
 	lf := t.allocator.getLeaf(hdr.NodeAddr)
 	if lf == nil {
 		panic("revert an invalid node")
@@ -446,6 +446,10 @@ func (t *ART) Staging() int {
 }
 
 func (t *ART) Release(h int) {
+	if h == 0 {
+		// 0 is the invalid and no-effect handle.
+		return
+	}
 	if h != len(t.stages) {
 		panic("cannot release staging buffer")
 	}
@@ -459,6 +463,10 @@ func (t *ART) Release(h int) {
 }
 
 func (t *ART) Cleanup(h int) {
+	if h == 0 {
+		// 0 is the invalid and no-effect handle.
+		return
+	}
 	if h > len(t.stages) {
 		return
 	}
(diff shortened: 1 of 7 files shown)


internal/unionstore/art/art.go
```diff
@@ -392,51 +393,86 @@ func (t *ART) Size() int {
 }
 
 func (t *ART) checkpoint() arena.MemDBCheckpoint {
-	panic("unimplemented")
+	return t.allocator.vlogAllocator.Checkpoint()
 }
 
 func (t *ART) RevertNode(hdr *arena.MemdbVlogHdr) {
-	panic("unimplemented")
+	lf := t.allocator.getLeaf(hdr.NodeAddr)
```
cfzjywxk, line 400: Better to check the validness before using it, so the possbile wrong state would not pass around.
```
if lf == nil
   panic("To revert invalid node")
```

cfzjywxk (approved):

internal/unionstore/art/art.go
```diff
@@ -392,51 +393,89 @@ func (t *ART) Size() int {
...
+	if h > len(t.stages) {
+		return
+	}
+	if h < len(t.stages) {
+		panic(fmt.Sprintf("cannot cleanup staging buffer, h=%v, len(db.stages)=%v", h, len(t.stages)))
+	}
 
-func (t *ART) truncate(snap *arena.MemDBCheckpoint) {
-	panic("unimplemented")
+	cp := &t.stages[h-1]
```
ekexium, line 469: Is it possible that `len(stages)=0` and `h=0`? Shall we test that?

internal/unionstore/art/art.go
```diff
@@ -392,51 +393,89 @@ func (t *ART) Size() int {
...
-
-func (t *ART) revertToCheckpoint(cp *arena.MemDBCheckpoint) {
-	panic("unimplemented")
-}
-
-func (t *ART) moveBackCursor(cursor *arena.MemDBCheckpoint, hdr *arena.MemdbVlogHdr) {
-	panic("unimplemented")
-}
+	if h > len(t.stages) {
+		return
```
ekexium, line 463: I wonder whether there are such usages...

internal/unionstore/art/art.go
```diff
@@ -392,51 +393,89 @@ func (t *ART) Size() int {
 }
 
 func (t *ART) checkpoint() arena.MemDBCheckpoint {
-	panic("unimplemented")
+	return t.allocator.vlogAllocator.Checkpoint()
 }
 
 func (t *ART) RevertNode(hdr *arena.MemdbVlogHdr) {
```
ekexium, line 399: Is "RevertVLogEntry" or something similar a more appropriate name? `RevertNode` looks like we are deleting the node.

ekexium (commented):
Can we have unit tests that test multi-level stages? Current tests only includes 1 levels and simple staging or cleanup.
The rest LGTM.

internal/unionstore/art/art.go
you06: `h == 0` means staging is not support in [TiDB](https://github.com/pingcap/tidb/blob/adbc0034ffb4470b42dc9f75ea595fae7499008b/pkg/table/tables/mutation_checker.go#L88-L90), I think we can just return for this case.

I'll add test for it.

internal/unionstore/art/art.go
you06: TiDB always call [`defer memBuffer.Clean`](https://github.com/pingcap/tidb/blob/fb9b8a194f8716beaee035cd0c285c30e29a9d64/pkg/table/tables/tables.go#L798) after a group of mutations (even `memBuffer.Release` is already invoked), so this case is expected and valid.

internal/unionstore/art/art.go
you06: Rename it to `RevertVAddr`.

ekexium (approved):

PR #1466, opened 2024-09-23
membuffer: support iterator for ART

you06:
ref pingcap/tidb#55287

This PR implements [`Iterator`](https://github.com/tikv/client-go/blob/271945f3453afde2a990eec4f86b52c6f041ea7b/internal/unionstore/union_store.go#L45-L52) for ART.


internal/unionstore/memdb_test.go
```diff
@@ -928,6 +955,7 @@ func testMemDBCheckpoint(t *testing.T, buffer MemBuffer) {
 
 func TestBufferLimit(t *testing.T) {
 	testBufferLimit(t, newRbtDBWithContext())
+	testBufferLimit(t, newArtDBWithContext())
 }
 
 func testBufferLimit(t *testing.T, buffer MemBuffer) {
```
cfzjywxk, line 961: Is all the previous memdb test executed for both `art` and `rbt` implementations by now?

cfzjywxk (approved):

internal/unionstore/memdb_test.go
you06: Snapshot read is not supported yet, all the other tests are running for both `art` and `rbt`.

internal/unionstore/art/art_iterator.go
```diff
@@ -14,18 +14,489 @@
...
+func (t *ART) iter(lowerBound, upperBound []byte, reverse, includeFlags bool) (*Iterator, error) {
+	it := &Iterator{
+		tree:         t,
+		reverse:      reverse,
+		valid:        true,
+		includeFlags: includeFlags,
+		inner: &baseIter{
+			allocator: &t.allocator,
+		},
+		currAddr: arena.BadAddr, // the default value of currAddr is not equal to any valid address
```
ekexium, line 53: Can we document how `BadAddr` is different from `NullAddr`?

PR #1467, opened 2024-09-24
membuffer: implement snapshot get and iterator for ART

you06:
ref pingcap/tidb#55287

This PR implements [Snapshot](https://github.com/tikv/client-go/blob/7d0f0bc93b5bae761fa1e7c9c4994686e6107cc3/internal/unionstore/union_store.go#L200-L205) and other methods of `MemBuffer` interface for ART.

### DIFF
--- a/internal/unionstore/art/art_iterator.go
+++ b/internal/unionstore/art/art_iterator.go
@@ -84,6 +84,15 @@ func (it *Iterator) Value() []byte {
 	return it.tree.allocator.vlogAllocator.GetValue(it.currLeaf.vAddr)
 }
 
+// HasValue returns false if it is flags only.
+func (it *Iterator) HasValue() bool {
+	return !it.isFlagsOnly()
+}
+
+func (it *Iterator) isFlagsOnly() bool {
+	return it.currLeaf != nil && it.currLeaf.vAddr.IsNull()
+}
+
 func (it *Iterator) Next() error {
 	if !it.valid {
 		// iterate is finished
--- a/internal/unionstore/art/art_iterator.go
+++ b/internal/unionstore/art/art_iterator.go
@@ -50,7 +50,10 @@ func (t *ART) iter(lowerBound, upperBound []byte, reverse, includeFlags bool) (*
 		inner: &baseIter{
 			allocator: &t.allocator,
 		},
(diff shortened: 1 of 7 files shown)


internal/unionstore/art/art_snapshot.go
```diff
@@ -14,30 +14,111 @@
...
+			return err
+		}
+		if i.setValue() {
+			return nil
+		}
+	}
+	return nil
+}
+
+func (i *SnapIter) setValue() bool {
```
cfzjywxk, line 115: For boolean returning value function, better to name it like `IsValueSet`.

ekexium (approved):

internal/unionstore/art/art_iterator.go
```diff
@@ -50,7 +50,10 @@ func (t *ART) iter(lowerBound, upperBound []byte, reverse, includeFlags bool) (*
 		inner: &baseIter{
 			allocator: &t.allocator,
 		},
-		currAddr: arena.BadAddr, // the default value of currAddr is not equal to any valid address
+		// the default value of currAddr is not equal to any valid address
+		// arena.BadAddr's idx is maxuint32 - 1, which is impossible in common cases,
```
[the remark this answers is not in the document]
you06: The `currAddr` will be set to valid or null address when first time call `Next`, so the only purpose of `BadAddr` is to avoid `currAddr == endAddr` at the beginning.

`BadAddr` can never appear in valid usages, because the [max size](https://github.com/tikv/client-go/blob/527f80a18663674d8d86bfcd7bf7d0d249853ac8/internal/unionstore/arena/arena.go#L49) of block is set, so the offset `maxuint32` is impossible.

internal/unionstore/art/art_snapshot.go
you06: This function actually set the value, and the return boolean of this function is whether set value successful.

What about rename it to `trySetValue`?

cfzjywxk: Perhaps `trySetValueRef` is better, also renaming the `value` field of `SnapIter` to `valueRef`, because it dose not own the value content.

cfzjywxk: Not a big problem, `trySetValue` is ok too.

cfzjywxk (approved):

PR #1470, opened 2024-10-03
membuffer: implement cache for ART

you06:
ref pingcap/tidb#55287

ART can also benefit from the cache of last visited leaf, this PR implement the cache, it's almost same as RBT's implementation.

The test result of new added `BenchmarkMemBufferCache`:

```text
// RBT, with cache, 1 cache-miss and 10 cache-hits for each key
BenchmarkMemBufferCache/RBT
BenchmarkMemBufferCache/RBT-32          10000000               185.9 ns/op             0 B/op          0 allocs/op

// master, no cache, 11 cache-miss for each key
BenchmarkMemBufferCache/ART
BenchmarkMemBufferCache/ART-32          10000000               223.3 ns/op             0 B/op          0 allocs/op

// this PR, with cache, 1 cache-miss and 10 cache-hits for each key
BenchmarkMemBufferCache/ART
BenchmarkMemBufferCache/ART-32          10000000               126.6 ns/op             0 B/op          0 allocs/op
```

### DIFF
--- a/examples/gcworker/go.mod
+++ b/examples/gcworker/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/opentracing/opentracing-go v1.2.0 // indirect
 	github.com/pingcap/errors v0.11.5-0.20211224045212-9687c2b0f87c // indirect
 	github.com/pingcap/failpoint v0.0.0-20220801062533-2eaa32854a6c // indirect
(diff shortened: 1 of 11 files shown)


cfzjywxk (approved):

you06:
> A benchmark result of its effect in TiDB's use case would be more convincing.

Little improvement, see [TiDB's PR](https://github.com/pingcap/tidb/pull/56473).

PR #1471, opened 2024-10-09
Do not broadcast txn status when heartbeat fail

ekexium:
This ensures that PK is the single source of truth for transaction status on the TiKV side. It prevents potential inconsistencies where the cached information might be more up-to-date than the info stored in the PK.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -1343,13 +1343,16 @@ func keepAlive(
 					}
 					return
 				}
-			} else {
-				keepFail = 0
-				metrics.TxnHeartBeatHistogramOK.Observe(time.Since(startTime).Seconds())
+				continue
 			}
 
-			// broadcast to all stores
-			if err != nil && isPipelinedTxn {
+			keepFail = 0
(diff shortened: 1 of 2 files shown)


cfzjywxk (approved):

txnkv/transaction/2pc.go
```diff
@@ -1349,7 +1349,7 @@ func keepAlive(
 			}
 
 			// broadcast to all stores
-			if isPipelinedTxn {
+			if err != nil && isPipelinedTxn {
```
you06, line 1352: Isn't it `if err == nil ...`?

Maybe add a comment to explain why skip txn status broadcast when heartbeat failed.

Maybe this way is better:

```go
if err != nil {
	... // logs
	if stopHeartBeat || keepFail > maxConsecutiveFailure {
		return
	}
	continue
}

// heartbeat success
keepFail = 0
metrics.TxnHeartBeatHistogramOK.Observe(time.Since(startTime).Seconds())
if isPipelinedTxn {
	... // broadcast
}

you06 (approved):

PR #1477, opened 2024-10-16
membuffer: improve ART's code and test coverage

you06:
ref pingcap/tidb#55287
This PR tries to make the code simplify and increase the test coverage (only some unreachable paths are not covered).

### DIFF
--- a/error/error.go
+++ b/error/error.go
@@ -212,6 +212,14 @@ func (e *ErrTxnTooLarge) Error() string {
 	return fmt.Sprintf("txn too large, size: %v.", e.Size)
 }
 
+type ErrKeyTooLarge struct {
+	KeySize int
+}
+
+func (e *ErrKeyTooLarge) Error() string {
+	return fmt.Sprintf("key size too large, size: %v.", e.KeySize)
+}
+
 // ErrEntryTooLarge is the error when a key value entry is too large.
 type ErrEntryTooLarge struct {
 	Limit uint64
--- a/internal/unionstore/art/art.go
+++ b/internal/unionstore/art/art.go
@@ -557,25 +557,6 @@ func (t *ART) InspectStage(handle int, f func([]byte, kv.KeyFlags, []byte)) {
 	t.allocator.vlogAllocator.InspectKVInLog(t, &head, &tail, f)
 }
 
-// SelectValueHistory select the latest value which makes `predicate` returns true from the modification history.
-func (t *ART) SelectValueHistory(key []byte, predicate func(value []byte) bool) ([]byte, error) {
-	_, x := t.search(key)
-	if x == nil {
-		return nil, tikverr.ErrNotExist
-	}
-	if x.vAddr.IsNull() {
-		// A flags only key, act as value not exists
-		return nil, tikverr.ErrNotExist
-	}
-	result := t.allocator.vlogAllocator.SelectValueHistory(x.vAddr, func(addr arena.MemdbArenaAddr) bool {
-		return predicate(t.allocator.vlogAllocator.GetValue(addr))
-	})
-	if result.IsNull() {
-		return nil, nil
-	}
-	return t.allocator.vlogAllocator.GetValue(result), nil
-}
-
 func (t *ART) SetMemoryFootprintChangeHook(hook func(uint64)) {
 	innerHook := func() {
 		hook(t.allocator.nodeAllocator.Capacity() + t.allocator.vlogAllocator.Capacity())
--- a/internal/unionstore/art/art.go
+++ b/internal/unionstore/art/art.go
@@ -68,6 +68,11 @@ func New() *ART {
 }
 
 func (t *ART) Get(key []byte) ([]byte, error) {
+	if t.vlogInvalid {
+		// panic for easier debugging.
+		panic("vlog is resetted")
+	}
+
 	// 1. search the leaf node.
 	_, leaf := t.traverse(key, false)
 	if leaf == nil || leaf.vAddr.IsNull() {
@@ -90,12 +95,18 @@ func (t *ART) GetFlags(key []byte) (kv.KeyFlags, error) {
 }
 
 func (t *ART) Set(key artKey, value []byte, ops ...kv.FlagsOp) error {
+	if t.vlogInvalid {
+		// panic for easier debugging.
+		panic("vlog is reset")
+	}
+
 	if len(key) > MaxKeyLen {
 		return &tikverr.ErrEntryTooLarge{
 			Limit: MaxKeyLen,
 			Size:  uint64(len(value)),
 		}
 	}
+
 	if value != nil {
 		if size := uint64(len(key) + len(value)); size > t.entrySizeLimit {
 			return &tikverr.ErrEntryTooLarge{
@@ -104,6 +115,7 @@ func (t *ART) Set(key artKey, value []byte, ops ...kv.FlagsOp) error {
 			}
 		}
 	}
+
 	if len(t.stages) == 0 {
 		t.dirty = true
 	}
@@ -294,11 +306,7 @@ func (t *ART) expandLeaf(key artKey, depth uint32, prev, current artNode) (arena
 	newAn.addChild(&t.allocator, l2Key.charAt(int(depth)), !l2Key.valid(int(depth)), leaf2Addr)
 
 	// swap the old leaf with the new node4.
-	if prev == nullArtNode {
-		t.root = newAn
-	} else {
-		prev.replaceChild(&t.allocator, key.charAt(prevDepth), newAn)
-	}
+	prev.replaceChild(&t.allocator, key.charAt(prevDepth), newAn)
 	return leaf2Addr.addr, leaf2
 }
 
@@ -340,11 +348,7 @@ func (t *ART) expandNode(key artKey, depth, mismatchIdx uint32, prev, current ar
 	// insert the artLeaf into new node
 	newLeafAddr, newLeaf := t.newLeaf(key)
 	newAn.addChild(&t.allocator, key.charAt(int(depth+mismatchIdx)), !key.valid(int(depth+mismatchIdx)), newLeafAddr)
-	if prev == nullArtNode {
-		t.root = newAn
-	} else {
-		prev.replaceChild(&t.allocator, key.charAt(prevDepth), newAn)
-	}
+	prev.replaceChild(&t.allocator, key.charAt(prevDepth), newAn)
 	return newLeafAddr.addr, newLeaf
 }
 
--- a/internal/unionstore/art/art.go
+++ b/internal/unionstore/art/art.go
@@ -47,7 +47,7 @@ type ART struct {
 	len             int
 	size            int
 
-	// The lastTraversedNode stores addr in uint64 of the last traversed node, include search and recursiveInsert.
+	// The lastTraversedNode stores addr in uint64 of the last traversed node, includes search and recursiveInsert.
 	// Compare to atomic.Pointer, atomic.Uint64 can avoid heap allocation, so it's more efficient.
 	lastTraversedNode atomic.Uint64
 	hitCount          atomic.Uint64
@@ -69,7 +69,7 @@ func New() *ART {
 
 func (t *ART) Get(key []byte) ([]byte, error) {
 	// 1. search the leaf node.
-	_, leaf := t.search(key)
+	_, leaf := t.traverse(key, false)
 	if leaf == nil || leaf.vAddr.IsNull() {
 		return nil, tikverr.ErrNotExist
 	}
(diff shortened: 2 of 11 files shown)


internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,150 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
+	}
+}
+
+type baseIter struct {
+	allocator *artAllocator
+	idxes     []int
+	nodes     []artNode
+}
+
+// seekInf seeks the boundary of the tree, reverse
```
cfzjywxk, line 207: The naming `seekInf` is confusing, is it better to use something like `seekToFirst`.

cfzjywxk, line 207:
```suggestion
// seekInf seeks the boundary of the tree, sequential or reverse.
```

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,150 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
 	if len(key) == 0 {
-		// if the seek key is empty, it means -inf or +inf, return root node directly.
-		nodes = append(nodes, curr)
-		if it.reverse {
-			idxes = append(idxes, node256cap)
-		} else {
-			idxes = append(idxes, inplaceIndex)
-		}
-		return idxes, nodes
+		panic("empty key is not allowed")
```
cfzjywxk, line 226:
```suggestion
		panic("seek with empty key is not allowed")
```

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,150 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
 		if node.prefixLen > 0 {
-			mismatchIdx := node.matchDeep(&it.tree.allocator, &curr, key, depth)
+			mismatchIdx := node.matchDeep(it.allocator, &curr, key, depth)
 			if mismatchIdx < node.prefixLen {
 				// no leaf node is match with the seek key
-				leafNode := minimum(&it.tree.allocator, curr)
-				leafKey := leafNode.asLeaf(&it.tree.allocator).GetKey()
+				leafNode := minimum(it.allocator, curr)
+				leafKey := leafNode.asLeaf(it.allocator).GetKey()
 				if mismatchIdx+depth == uint32(len(key)) || key[depth+mismatchIdx] < leafKey[depth+mismatchIdx] {
```
cfzjywxk, line 249: What doest it mean for this branch like `mismatchIdx+depth == uint32(len(key))`, could you add somments?

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,150 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
+}
+
+// seekInf seeks the boundary of the tree, reverse
+func (it *baseIter) seekInf(root artNode, reverse bool) {
+	// if the seek key is empty, it means -inf or +inf, return root node directly.
+	it.nodes = []artNode{root}
+	if reverse {
+		it.idxes = []int{node256cap}
+	} else {
+		it.idxes = []int{inplaceIndex}
```
cfzjywxk, line 214: Should the names of the constant variables start with captital letter like `Node256Cap` and `InplaceIndex`. Besides, the `InplaceIndex` is a little bit confusing, better to add comments about it.

internal/unionstore/art/art.go
```diff
@@ -90,6 +95,18 @@ func (t *ART) GetFlags(key []byte) (kv.KeyFlags, error) {
 }
 
 func (t *ART) Set(key artKey, value []byte, ops ...kv.FlagsOp) error {
+	if t.vlogInvalid {
+		// panic for easier debugging.
+		panic("vlog is reset")
+	}
+
+	if len(key) > MaxKeyLen {
+		return &tikverr.ErrEntryTooLarge{
```
ekexium, line 104: `ErrKeyTooLarge`?

internal/unionstore/art/art.go
```diff
@@ -300,11 +306,7 @@ func (t *ART) expandLeaf(key artKey, depth uint32, prev, current artNode) (arena
 	newAn.addChild(&t.allocator, l2Key.charAt(int(depth)), !l2Key.valid(int(depth)), leaf2Addr)
 
 	// swap the old leaf with the new node4.
-	if prev == nullArtNode {
```
ekexium, line 303: Why can we remove the check?

internal/unionstore/art/art_iterator.go
you06: I think these constants are private, just avoid exporting them.

internal/unionstore/art/art.go
you06: Because the root node can't be a leaf, so when we expanding a leaf, the prev node is definitely not null.

Consider this case:
```
1. t.root == nullArtNode
2. insert a key [1, 2], t.root becomes a node4
3. insert a key [1, 3], t.root.findChild(1) is a leaf, so expandLeaf will be invoked, prev is t.root now
```

internal/unionstore/memdb_test.go
```diff
@@ -41,6 +41,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	tikverr "github.com/tikv/client-go/v2/error"
```
cfzjywxk, line 44: The imports are not sorted.

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,161 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
-func (it *Iterator) seek(key artKey) ([]int, []artNode) {
-	curr := it.tree.root
+// seek the first node and index that >= key, nodes[0] is the root node
+func (it *baseIter) seek(root artNode, key artKey) {
+	curr := root
 	depth := uint32(0)
-	idxes := make([]int, 0, 8)
-	nodes := make([]artNode, 0, 8)
+	it.idxes = make([]int, 0, 8)
+	it.nodes = make([]artNode, 0, 8)
```
cfzjywxk, line 228: Better to do the allocation lazily before its usages.

And move the check
```
if len(key) == 0 {
    panic(...)
}
```
above

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,161 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
+			mismatchIdx := node.matchDeep(it.allocator, &curr, key, depth)
 			if mismatchIdx < node.prefixLen {
-				// no leaf node is match with the seek key
-				leafNode := minimum(&it.tree.allocator, curr)
-				leafKey := leafNode.asLeaf(&it.tree.allocator).GetKey()
-				if mismatchIdx+depth == uint32(len(key)) || key[depth+mismatchIdx] < leafKey[depth+mismatchIdx] {
+				// no leaf node is match with the seek key, we need to check whether the seek key is less than the prefix.
+				// If the seek key is less than the prefix, all the children are located on the right side of the seek key,
+				// unless the children are located on the left side of the seek key.
+				var prefix []byte
```
cfzjywxk, line 253:
```suggestion
				// Check whether the seek key is smaller than the prefix, as no leaf node matches the seek key.
				// If the seek key is smaller than the prefix, all the children are located on the right side of the seek key.
				// Othewise,  the children are located on the left side of the seek key.
				var prefix []byte
```

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,161 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
+					prefix = leafNode.asLeaf(it.allocator).getKeyDepth(depth)
+				}
+				if mismatchIdx+depth == uint32(len(key)) || key[depth+mismatchIdx] < prefix[mismatchIdx] {
 					// key < leafKey, set index to -1 means all the children are larger than the seek key
-					idxes = append(idxes, -1)
+					it.idxes = append(it.idxes, -1)
 				} else {
-					// key > leafKey, set index to 256 means all the children are less than the seek key
-					idxes = append(idxes, node256cap)
+					// key > prefix, set index to 256 means all the children are less than the seek key
```
cfzjywxk, line 264:
```suggestion
					// key > prefix, set index to 256 means all the children are smaller than the seek key.
```

error/error.go
```diff
@@ -212,6 +212,14 @@ func (e *ErrTxnTooLarge) Error() string {
 	return fmt.Sprintf("txn too large, size: %v.", e.Size)
 }
 
+type ErrKeyTooLarge struct {
```
ekexium, line 215: Kind reminder: TiDB should have corresponding conversions.

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,161 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
+				// If the seek key is less than the prefix, all the children are located on the right side of the seek key,
+				// unless the children are located on the left side of the seek key.
+				var prefix []byte
+				if mismatchIdx < maxPrefixLen {
+					prefix = node.prefix[:]
+				} else {
+					leafNode := minimum(it.allocator, curr)
+					prefix = leafNode.asLeaf(it.allocator).getKeyDepth(depth)
+				}
+				if mismatchIdx+depth == uint32(len(key)) || key[depth+mismatchIdx] < prefix[mismatchIdx] {
```
cfzjywxk, line 260: What does it mean if the `mismatchIdx+depth == uint32(len(key))` hold? 
As well as  `mismatchIdx+depth > uint32(len(key))`  or  `mismatchIdx+depth < uint32(len(key))`.

ekexium (approved):
Some minor grammar tweaks would make these comments better. May let LLM do this.

internal/unionstore/art/art_iterator.go
you06: If the `mismatchIdx+depth == uint32(len(key))` holds, it means the seek key is the prefix of the any leaf key in this node, and the leaf key is longer and larger than the seek key.

internal/unionstore/art/art_iterator.go
```diff
@@ -318,197 +338,166 @@ func (it *baseIter) compare(other *baseIter) int {
...
-		n16 := it.allocator.getNode16(curr.addr)
-		idx = min(idx, int(n16.nodeNum)-1)
-		if idx >= 0 {
-			it.idxes[depth] = idx
-			child := n16.children[idx]
-			if child.kind == typeLeaf {
-				return child
+		case typeNode48:
+			n48 := it.allocator.getNode48(curr.addr)
+			if idx >= 0 {
```
cfzjywxk, line 462: Anything needs for the `else` branch of this condition, like `panic`?

internal/unionstore/art/art_iterator.go
```diff
@@ -318,197 +338,166 @@ func (it *baseIter) compare(other *baseIter) int {
...
-			idx = n48.prevPresentIdx(idx)
-		}
-		if idx >= 0 {
-			it.idxes[depth] = idx
-			child := n48.children[n48.keys[idx]]
-			if child.kind == typeLeaf {
-				return child
+		case typeNode256:
+			n256 := it.allocator.getNode256(curr.addr)
+			if idx >= 0 {
```
cfzjywxk, line 476: Ditto, or add some comments to explain why the else branch does not need to be checked.

internal/unionstore/art/art_iterator.go
```diff
@@ -156,140 +156,160 @@ func (it *Iterator) init(lowerBound, upperBound []byte) {
...
+		it.valid = false
+	}
 }
 
-// seek the first node and index that >= key, return the indexes and nodes of the lookup path
-// nodes[0] is the root node
-func (it *Iterator) seek(key artKey) ([]int, []artNode) {
-	curr := it.tree.root
-	depth := uint32(0)
-	idxes := make([]int, 0, 8)
```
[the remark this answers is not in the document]
you06: `seek` is only called once, so the `capacity` is always 0 here. I think it's ok to leave the allocation later.

If we want to choose a capacity here, the best value is the height of the tree, which is not tracked. Maybe we can leave a rather small capacity initialization(like `it.idxes, it.nodes = make([]int, 0, 2), make([]artNode, 0, 2)`) here.

The `capacity == 1` is enough for tiny transactions, which only have several leaves in the root node.
Meanwhile, for large transactions, the more calling of `Next` already takes a long time, the time cost of capacity growth can be ignored.

internal/unionstore/art/art_iterator.go
you06: The else branch means(`idx == inplaceIndex`), it will be handled later. Added some comment.

cfzjywxk: Could the `idx` be other negative values, like in situations the memory is corruted?
```
If 
else if
else
   panic("xx")
```
How about also adding panic for these unrechable branches to make it more clear.

cfzjywxk (approved):

cfzjywxk (approved):

cfzjywxk:
@ekexium PTAL again

ekexium (approved):

PR #1478, opened 2024-10-16
add retry limiter to backoff function

Tema:
Current backoff policy does not help much to limit TiDB reties to retrieve Region from PD when there are issues with Region Metadata in PD:

image

This PR adds an ability to configure global retry limiter to Backoff function per each config. It also creates a new Backoff config dedicated to PD Region Metadata calls which will be used in TiDB in separate PR:
```
BoPDRegionMetadata = NewConfigWithRetryLimit("pdRegionMetadata", &metrics.BackoffHistogramPD, NewBackoffFnCfg(500, 3000, EqualJitter), NewRetryRateLimiter(10, 0.1), tikverr.NewErrPDServerTimeout(""))
```

The config above allows a single retry per each 10 previous successful call (0.1), but limit overall retry budget to 10. It always start with full budget of retries.

### DIFF
--- a/config/retry/config.go
+++ b/config/retry/config.go
@@ -177,7 +177,7 @@ var (
 	BoTiFlashRPC       = NewConfig("tiflashRPC", &metrics.BackoffHistogramRPC, NewBackoffFnCfg(100, 2000, EqualJitter), tikverr.ErrTiFlashServerTimeout)
 	BoTxnLock          = NewConfig("txnLock", &metrics.BackoffHistogramLock, NewBackoffFnCfg(100, 3000, EqualJitter), tikverr.ErrResolveLockTimeout)
 	BoPDRPC            = NewConfig("pdRPC", &metrics.BackoffHistogramPD, NewBackoffFnCfg(500, 3000, EqualJitter), tikverr.NewErrPDServerTimeout(""))
-	BoPDRegionMetadata = NewConfigWithRetryLimit("pdRegionMetadata", &metrics.BackoffHistogramPD, NewBackoffFnCfg(500, 3000, EqualJitter), NewRetryRateLimiter(10, 10), tikverr.NewErrPDServerTimeout(""))
+	BoPDRegionMetadata = NewConfigWithRetryLimit("pdRegionMetadata", &metrics.BackoffHistogramPD, NewBackoffFnCfg(500, 3000, EqualJitter), NewRetryRateLimiter(100, 1), tikverr.NewErrPDServerTimeout(""))
 	// change base time to 2ms, because it may recover soon.
 	BoRegionMiss               = NewConfig("regionMiss", &metrics.BackoffHistogramRegionMiss, NewBackoffFnCfg(2, 500, NoJitter), tikverr.ErrRegionUnavailable)
 	BoRegionScheduling         = NewConfig("regionScheduling", &metrics.BackoffHistogramRegionScheduling, NewBackoffFnCfg(2, 500, NoJitter), tikverr.ErrRegionUnavailable)
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -2059,7 +2059,7 @@ func (c *RegionCache) loadRegion(bo *retry.Backoffer, key []byte, isEndKey bool,
 	opts = append(opts, pd.WithBuckets())
 	for {
 		if backoffErr != nil {
-			err := bo.Backoff(retry.BoPDRPC, backoffErr)
+			err := bo.Backoff(retry.BoPDRegionMetadata, backoffErr)
 			if err != nil {
 				return nil, errors.WithStack(err)
 			}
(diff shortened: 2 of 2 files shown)


cfzjywxk:
It's ok to introduce a rate-limiting mechanism for the kv-client. Finding the optimal balance between error handling, retry success, and avoiding overloading PD could be challenging, or selecting a suitable default value that works for most scenarios.

Another approach is similar to TiKV health control feedback (as discussed in [tikv/tikv#16297](https://github.com/tikv/tikv/issues/16297)), where some processing capacity information is carried in PD responses and fed back to the KV client. Based on this feedback, the kv-client can then decide its concurrency control and rate-limiting strategy accordingly.

config/retry/config.go
```diff
@@ -96,6 +98,50 @@ func NewConfig(name string, metric *prometheus.Observer, backoffFnCfg *BackoffFn
 	}
 }
 
+type RetryRateLimiter struct {
```
cfzjywxk, line 101: Need comments like for exported functions and type definitions, 
```
// RetryRateLimiter is used to limit retry times for PD requests.
```
or the lint check would fail.

config/retry/config.go
```diff
@@ -96,6 +98,50 @@ func NewConfig(name string, metric *prometheus.Observer, backoffFnCfg *BackoffFn
...
+	tokenCount                 int32
+	allowedRetryToSuccessRatio float32
+	cap                        int32
+}
+
+func NewRetryRateLimiter(cap int32, ratio float32) *RetryRateLimiter {
```
cfzjywxk, line 107: Ditto for the comments, and better to explain the meaning of the input parameters.

Besides, would it be less expensive using `int` or `uint` type for `ratio` instead of float values?

config/retry/config.go
```diff
@@ -96,6 +98,50 @@ func NewConfig(name string, metric *prometheus.Observer, backoffFnCfg *BackoffFn
...
+	return &RetryRateLimiter{
+		cap, // always init with full bucket to allow retries on startup
+		ratio,
+		cap,
+	}
+}
+
+// add a token to the rate limiter bucket according to configured retry to success ratio and cap
```
cfzjywxk, line 115: The comment format is like
```
// addRetryToken is a ...
```
, needs to start with the function name

config/retry/config.go
```diff
@@ -96,6 +98,50 @@ func NewConfig(name string, metric *prometheus.Observer, backoffFnCfg *BackoffFn
...
+func (r *RetryRateLimiter) addRetryToken() {
+	if rand.Float32() < r.allowedRetryToSuccessRatio {
```
cfzjywxk, line 117: As metioned above, would it be less expensive to use integer random values?

config/retry/config.go
```diff
@@ -96,6 +98,50 @@ func NewConfig(name string, metric *prometheus.Observer, backoffFnCfg *BackoffFn
...
+func (r *RetryRateLimiter) takeRetryToken() bool {
+	if atomic.LoadInt32(&r.tokenCount) > 0 {
+		// it is ok to go below 0, because consumed token will still match added one at the end
+		atomic.AddInt32(&r.tokenCount, -1)
+		return true
+	}
+	return false
+}
+
+func NewConfigWithRetryLimit(name string, metric *prometheus.Observer, backoffFnCfg *BackoffFnCfg, retryRateLimiter *RetryRateLimiter, err error) *Config {
```
cfzjywxk, line 135: Ditto for the comments of exported function.

config/retry/config.go
```diff
@@ -119,10 +165,11 @@ const txnLockFastName = "txnLockFast"
...
-	BoTxnLock    = NewConfig("txnLock", &metrics.BackoffHistogramLock, NewBackoffFnCfg(100, 3000, EqualJitter), tikverr.ErrResolveLockTimeout)
-	BoPDRPC      = NewConfig("pdRPC", &metrics.BackoffHistogramPD, NewBackoffFnCfg(500, 3000, EqualJitter), tikverr.NewErrPDServerTimeout(""))
+	BoTiKVRPC          = NewConfig("tikvRPC", &metrics.BackoffHistogramRPC, NewBackoffFnCfg(100, 2000, EqualJitter), tikverr.ErrTiKVServerTimeout)
+	BoTiFlashRPC       = NewConfig("tiflashRPC", &metrics.BackoffHistogramRPC, NewBackoffFnCfg(100, 2000, EqualJitter), tikverr.ErrTiFlashServerTimeout)
+	BoTxnLock          = NewConfig("txnLock", &metrics.BackoffHistogramLock, NewBackoffFnCfg(100, 3000, EqualJitter), tikverr.ErrResolveLockTimeout)
+	BoPDRPC            = NewConfig("pdRPC", &metrics.BackoffHistogramPD, NewBackoffFnCfg(500, 3000, EqualJitter), tikverr.NewErrPDServerTimeout(""))
+	BoPDRegionMetadata = NewConfigWithRetryLimit("pdRegionMetadata", &metrics.BackoffHistogramPD, NewBackoffFnCfg(500, 3000, EqualJitter), NewRetryRateLimiter(10, 0.1), tikverr.NewErrPDServerTimeout(""))
```
cfzjywxk, line 172: As discussed in the previous comments, it would be challendge to choose a default value for all kinds of scenarios? Do we have some tests for the choice of `10, 0.1`?

PR #1479, opened 2024-10-17
membuffer: fix data race when `IsReadOnly` and Set may be concurrently invoked

you06:
ref pingcap/tidb#56178

TiDB concurrently call `IsReadOnly` and some write operations in union statement, so we need to avoid data race by mutex.

### DIFF
--- a/internal/unionstore/memdb_art.go
+++ b/internal/unionstore/memdb_art.go
@@ -138,6 +138,15 @@ func (db *artDBWithContext) BatchGet(ctx context.Context, keys [][]byte) (map[st
 	return m, nil
 }
 
+// Dirty implements the MemBuffer interface.
+func (db *artDBWithContext) Dirty() bool {
+	if !db.skipMutex {
+		db.RLock()
+		defer db.RUnlock()
+	}
+	return db.ART.Dirty()
+}
+
 // GetMetrics implements the MemBuffer interface.
 func (db *artDBWithContext) GetMetrics() Metrics { return Metrics{} }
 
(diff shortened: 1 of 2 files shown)


ekexium:
I understand it fixes the data race. I'm just wondering whether it means we are using it correctly. What is the expected behavior when `dirty` is accessed concurrently?

I haven't checked the case yet but a first rough thought might be like:
```
Thread-1 gets dirty = false
Thread-2 writes dirty = true
Thread-1 does something that depends on dirty = false? // this can be problematic?
```

you06:
> Is there a test case in tidb to verify the fix?

There is a test case in the issue.

you06:
> I understand it fixes the data race. I'm just wondering whether it means we are using it correctly. What is the expected behavior when `dirty` is accessed concurrently?
> 
> I haven't checked the case yet but a first rough thought might be like:
> 
> ```
> Thread-1 gets dirty = false
> Thread-2 writes dirty = true
> Thread-1 does something that depends on dirty = false? // this can be problematic?
> ```

For this question, we're just avoiding the race between [`UpdateFlags`](https://github.com/tikv/client-go/blob/691e80ae0ea98eb1f1f6734a2ecd41c139f246bc/internal/unionstore/union_store.go#L188-L189) and `IsDirty`, if `IsDirty() == true`, the result from membuffer will be not-exist error and it finally read from the store.

This is a large topic because TiDB used to pretend it's executor is running in serial mode. So there may be some hidden issues with high possibility.

PR #1482, opened 2024-10-28
membuffer: compare the keys of ART by chunk

you06:
ref pingcap/tidb#55287

In order to perform path compression in ART, we need to find the longest common prefix the keys in many places, when the common prefix is long, this can be slow.

This PR compares the keys by chunk. Because a chunk is a uint64 which can contains 8 bytes, this can improve the performance a lot when the common prefix is long.

- Long Common Prefix
```
# before
BenchmarkMemBufferSetGetLongKey
BenchmarkMemBufferSetGetLongKey/ART
BenchmarkMemBufferSetGetLongKey/ART-32           1000000               874.3 ns/op          2149 B/op          0 allocs/op

# this PR
BenchmarkMemBufferSetGetLongKey
BenchmarkMemBufferSetGetLongKey/ART
BenchmarkMemBufferSetGetLongKey/ART-32           1000000               524.5 ns/op          2149 B/op          0 allocs/op
```

- Common Cases

```
# before
BenchmarkPut
BenchmarkPut/ART
BenchmarkPut/ART-32              1000000                91.70 ns/op          335 B/op          0 allocs/op
BenchmarkPutRandom
BenchmarkPutRandom/ART
BenchmarkPutRandom/ART-32                1000000               182.1 ns/op           404 B/op          0 allocs/op
BenchmarkGet
BenchmarkGet/ART
BenchmarkGet/ART-32                      1000000                31.31 ns/op            0 B/op          0 allocs/op
BenchmarkGetRandom
BenchmarkGetRandom/ART
BenchmarkGetRandom/ART-32                1000000               142.7 ns/op    

# this PR
BenchmarkPut
BenchmarkPut/ART
BenchmarkPut/ART-32              1000000                89.18 ns/op          335 B/op          0 allocs/op
BenchmarkPutRandom
BenchmarkPutRandom/ART
BenchmarkPutRandom/ART-32                1000000               175.8 ns/op           404 B/op          0 allocs/op
BenchmarkGet
BenchmarkGet/ART
BenchmarkGet/ART-32                      1000000                31.74 ns/op            0 B/op          0 allocs/op
BenchmarkGetRandom
BenchmarkGetRandom/ART
BenchmarkGetRandom/ART-32                1000000               132.5 ns/op   
```

### DIFF
--- a/internal/unionstore/art/art_node.go
+++ b/internal/unionstore/art/art_node.go
@@ -16,7 +16,6 @@ package art
 
 import (
 	"bytes"
-	"encoding/binary"
 	"math"
 	"math/bits"
 	"runtime"
@@ -59,22 +58,6 @@ const (
 	node256size = int(unsafe.Sizeof(node256{}))
 	leafSize    = int(unsafe.Sizeof(artLeaf{}))
 )
-
-var nativeEndian binary.ByteOrder
-
-func init() {
-	buf := [2]byte{}
-	*(*uint16)(unsafe.Pointer(&buf[0])) = uint16(0xABCD)
-
-	switch buf {
-	case [2]byte{0xCD, 0xAB}:
-		nativeEndian = binary.LittleEndian
-	case [2]byte{0xAB, 0xCD}:
-		nativeEndian = binary.BigEndian
-	default:
-		panic("Could not determine native endianness.")
-	}
-}
 
 var nullArtNode = artNode{kind: typeInvalid, addr: arena.NullAddr}
 
--- a/internal/unionstore/art/art_node.go
+++ b/internal/unionstore/art/art_node.go
@@ -385,8 +385,9 @@ func longestCommonPrefixByChunk(l1Key, l2Key artKey, depth uint32) uint32 {
 			xor := *(*uint64)(p1) ^ *(*uint64)(p2)
 			return limit - remaining + uint32(bits.TrailingZeros64(xor)>>3) - depth
 		}
-		p1 = unsafe.Pointer(uintptr(p1) + 8)
-		p2 = unsafe.Pointer(uintptr(p2) + 8)
+
+		p1 = unsafe.Add(p1, 8)
+		p2 = unsafe.Add(p2, 8)
 		remaining -= 8
 	}
 
--- a/internal/unionstore/art/art_node.go
+++ b/internal/unionstore/art/art_node.go
@@ -16,8 +16,10 @@ package art
 
 import (
 	"bytes"
+	"encoding/binary"
 	"math"
 	"math/bits"
+	"runtime"
 	"sort"
 	"testing"
 	"unsafe"
@@ -57,6 +59,22 @@ const (
 	node256size = int(unsafe.Sizeof(node256{}))
 	leafSize    = int(unsafe.Sizeof(artLeaf{}))
 )
+
+var nativeEndian binary.ByteOrder
+
+func init() {
+	buf := [2]byte{}
+	*(*uint16)(unsafe.Pointer(&buf[0])) = uint16(0xABCD)
+
+	switch buf {
+	case [2]byte{0xCD, 0xAB}:
+		nativeEndian = binary.LittleEndian
+	case [2]byte{0xAB, 0xCD}:
+		nativeEndian = binary.BigEndian
+	default:
+		panic("Could not determine native endianness.")
+	}
+}
 
 var nullArtNode = artNode{kind: typeInvalid, addr: arena.NullAddr}
 
@@ -310,14 +328,7 @@ func (n *nodeBase) setPrefix(key artKey, prefixLen uint32) {
 // Node if the nodeBase.prefixLen > maxPrefixLen and the returned mismatch index equals to maxPrefixLen,
 // key[maxPrefixLen:] will not be checked by this function.
 func (n *nodeBase) match(key artKey, depth uint32) uint32 /* mismatch index */ {
-	idx := uint32(0)
-	limit := min(min(n.prefixLen, maxPrefixLen), uint32(len(key))-depth)
-	for ; idx < limit; idx++ {
-		if n.prefix[idx] != key[idx+depth] {
-			return idx
-		}
-	}
-	return idx
+	return longestCommonPrefix(key[depth:], n.prefix[:min(n.prefixLen, maxPrefixLen)], 0)
 }
 
 // matchDeep returns the mismatch index of the key and the node's prefix.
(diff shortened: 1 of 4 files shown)


ekexium (dismissed):
~Some minor grammar tweaks would make these comments better. May let LLM do this.~ I commented on the wrong PR 😆

internal/unionstore/art/art_node.go
```diff
@@ -251,6 +252,14 @@ func (k artKey) valid(pos int) bool {
 	return pos < len(k)
 }
 
+const CHUNK_SIZE = 8
```
ekexium, line 255: `ChunkSize` as per Go convention？

internal/unionstore/art/art_node.go
```diff
@@ -251,6 +252,14 @@ func (k artKey) valid(pos int) bool {
...
+
+// u64Chunk returns the uint64 chunk of 8 bytes starting from the given index.
+// Compare the key by chunks(merge 8 bytes into a uint64 chunk) is faster than by byte.
+func (k artKey) u64Chunk(idx int) uint64 {
```
ekexium, line 259: It can accept uint32 to save unnecessary type conversions

ekexium (commented):
Would you like to compare it with something like this?
```golang
    // Use unsafe for faster byte comparison
    p1 := unsafe.Pointer(&l1Key[depth])
    p2 := unsafe.Pointer(&l2Key[depth])
    
    // Compare 8 bytes at a time
    remaining := minLen - depth
    for remaining >= 8 {
        if *(*uint64)(p1) != *(*uint64)(p2) {
            // Find first different byte using trailing zeros
            xor := *(*uint64)(p1) ^ *(*uint64)(p2)
            return depth + uint32(bits.TrailingZeros64(xor) >> 3)
        }
        p1 = unsafe.Pointer(uintptr(p1) + 8)
        p2 = unsafe.Pointer(uintptr(p2) + 8)
        depth += 8
        remaining -= 8
}
```

you06:
> Would you like to compare it with something like this?

The unsafe implementation.

- Long Common Prefix

That's about +15% speed.

```
BenchmarkMemBufferSetGetLongKey
BenchmarkMemBufferSetGetLongKey/ART
BenchmarkMemBufferSetGetLongKey/ART-32           1000000               458.1 ns/op          2149 B/op          0 allocs/op
```

The common cases aren't that stable on my PC, and even the master code runs much slower than when the result in the description now (there's not much difference anyway).

The unsafe implementation runs faster, primarily due to better CPU pipeline utilization and more accurate prediction through chunk comparison. However, it seems to only work on little-endian machines, and I'm not sure if we need to worry about big-endian architectures...

ekexium:
> The unsafe implementation runs faster, the only concern is it seems to only work on little-endian machines, and I'm not sure if we need to worry about big-endian architectures...

We can check internal/goarch, and fallback to other methods for big-endian architectures.

internal/unionstore/art/art_node.go
```diff
@@ -350,10 +360,34 @@ func (an *artNode) asNode256(a *artAllocator) *node256 {
...
-	// TODO: possible optimization
-	// Compare the key by loop can be very slow if the final LCP is large.
-	// Maybe optimize it by comparing the key in chunks if the limit exceeds certain threshold.
+	idx, limit := depth, uint32(min(len(l1Key), len(l2Key)))
+
+	if idx == limit {
+		return 0
+	}
+
+	p1 := unsafe.Pointer(&l1Key[depth])
```
cfzjywxk, line 369: On some architectures (e.g., older ARM architectures), unaligned memory access is significantly slower than aligned access. 

Skip considering it  is generally safe and keeps code simpler.

cfzjywxk (approved):

internal/unionstore/art/art_node.go
ekexium: It is said that unaligned access may even crash. Shall we consider an alignment check like
`isAligned := (uintptr(p1) & 7) == 0 && (uintptr(p2) & 7) == 0`?

internal/unionstore/art/art_node.go
```diff
@@ -355,10 +349,49 @@ func (an *artNode) asNode256(a *artAllocator) *node256 {
 // longestCommonPrefix returns the length of the longest common prefix of two keys.
 // the LCP is calculated from the given depth, you need to guarantee l1Key[:depth] equals to l2Key[:depth] before calling this function.
 func longestCommonPrefix(l1Key, l2Key artKey, depth uint32) uint32 {
+	switch runtime.GOARCH {
```
ekexium, line 352: We can introduce a const boolean and do the comparison only once

you06:
/hold

The client-go's test seems ok, but the memory is violated when running with tidb.

```
fatal error: checkptr: pointer arithmetic result points to invalid allocation

goroutine 21334 gp=0xc002051a40 m=12 mp=0xc00088aa08 [running]:
runtime.throw({0x97a1a64?, 0x216ae1d?})
	GOROOT/src/runtime/panic.go:1067 +0x48 fp=0xc004e40608 sp=0xc004e405d8 pc=0x21dc848
runtime.checkptrArithmetic(0xc005f7bc0f?, {0xc004e406a8, 0x1, 0x219e3e9?})
	GOROOT/src/runtime/checkptr.go:69 +0x9c fp=0xc004e40638 sp=0xc004e40608 pc=0x216afbc
github.com/tikv/client-go/v2/internal/unionstore/art.longestCommonPrefixByChunk({0xc00603a9d8, 0x8, 0x18?}, {0xc004f97138, 0x9, 0x21e7bd9?}, 0x0)
	external/com_github_tikv_client_go_v2/internal/unionstore/art/art_node.go:388 +0xfc fp=0xc004e406c0 sp=0xc004e40638 pc=0x3db30dc
github.com/tikv/client-go/v2/internal/unionstore/art.longestCommonPrefix({0xc00603a9d8, 0x8, 0x8}, {0xc004f97138, 0x9, 0x14}, 0x0)
	external/com_github_tikv_client_go_v2/internal/unionstore/art/art_node.go:355 +0x69 fp=0xc004e40710 sp=0xc004e406c0 pc=0x3db2f69
github.com/tikv/client-go/v2/internal/unionstore/art.(*nodeBase).match(...)
	external/com_github_tikv_client_go_v2/internal/unionstore/art/art_node.go:314
```

cfzjywxk:
> /hold
> 
> The client-go's test seems ok, but the memory is violated when running with tidb.
> 
> ```
> fatal error: checkptr: pointer arithmetic result points to invalid allocation
> 
> goroutine 21334 gp=0xc002051a40 m=12 mp=0xc00088aa08 [running]:
> runtime.throw({0x97a1a64?, 0x216ae1d?})
> 	GOROOT/src/runtime/panic.go:1067 +0x48 fp=0xc004e40608 sp=0xc004e405d8 pc=0x21dc848
> runtime.checkptrArithmetic(0xc005f7bc0f?, {0xc004e406a8, 0x1, 0x219e3e9?})
> 	GOROOT/src/runtime/checkptr.go:69 +0x9c fp=0xc004e40638 sp=0xc004e40608 pc=0x216afbc
> github.com/tikv/client-go/v2/internal/unionstore/art.longestCommonPrefixByChunk({0xc00603a9d8, 0x8, 0x18?}, {0xc004f97138, 0x9, 0x21e7bd9?}, 0x0)
> 	external/com_github_tikv_client_go_v2/internal/unionstore/art/art_node.go:388 +0xfc fp=0xc004e406c0 sp=0xc004e40638 pc=0x3db30dc
> github.com/tikv/client-go/v2/internal/unionstore/art.longestCommonPrefix({0xc00603a9d8, 0x8, 0x8}, {0xc004f97138, 0x9, 0x14}, 0x0)
> 	external/com_github_tikv_client_go_v2/internal/unionstore/art/art_node.go:355 +0x69 fp=0xc004e40710 sp=0xc004e406c0 pc=0x3db2f69
> github.com/tikv/client-go/v2/internal/unionstore/art.(*nodeBase).match(...)
> 	external/com_github_tikv_client_go_v2/internal/unionstore/art/art_node.go:314
> ```

@you06 Is this a new bug introduced in https://github.com/tikv/client-go/pull/1477?

ekexium (approved):

you06:
> @you06 Is this a new bug introduced in #1477?

#1477 introduced a bug in the `baseIter.next`.

The error `arithmetic checkptr: pointer arithmetic result points to invalid allocation` is caused by `unsafe.Pointer(uintptr(p1) + 8)`, we should use `unsafe.Add` instead.

The CI in TiDB can pass after fixing both bugs.

cfzjywxk (approved):

PR #1483, opened 2024-10-31
membuffer: fix memory leak in red-black tree

### LINKED ISSUE #1375: MemDB does not reuse free nodes
MemDB cannot [reuse free nodes](https://github.com/tikv/client-go/blob/4a72526f6c309afb72c5a5910002a3c6d30730e5/internal/unionstore/memdb_arena.go#L282), which may lead to mem leak in some cases. If a DML writes many keys and it has been retried multiple times due to pessimistic lock error, MemDB would re-alloc nodes for each try. Refer to the following demo code, the mem usage of allocator keeps growing.

you06:
close #1375, ref pingcap/tidb#56837

The issue #1375 figures out that the RBT membuffer cannot reuse free nodes which causes a memory leak. This pr uses marking the node as deleted to replace the actual deletion.

Pros:

- Because the memory arena cannot handle the memory fragmentation well, we avoid the memory leak by this mark-delete strategy.

Cons:

- In certain cases, the mark-delete tree may be taller than the actual deletion, which can slightly impact the performance.

### DIFF
--- a/internal/unionstore/rbt/rbt.go
+++ b/internal/unionstore/rbt/rbt.go
@@ -349,8 +349,7 @@ func (db *RBT) Set(key []byte, value []byte, ops ...kv.FlagsOp) error {
 	// the NeedConstraintCheckInPrewrite flag is temporary,
 	// every write to the node removes the flag unless it's explicitly set.
 	// This set must be in the latest stage so no special processing is needed.
-	var flags kv.KeyFlags
-	flags = x.GetKeyFlags()
+	flags := x.GetKeyFlags()
 	if flags == 0 && x.vptr.IsNull() && x.isDeleted() {
 		db.count++
 		db.size += int(x.klen)
@@ -365,9 +364,6 @@ func (db *RBT) Set(key []byte, value []byte, ops ...kv.FlagsOp) error {
 		db.dirty = true
 	}
 	x.setKeyFlags(flags)
-	if x.isDeleted() {
-		panic("")
-	}
 
 	if value == nil {
 		return nil
--- a/internal/unionstore/rbt/rbt.go
+++ b/internal/unionstore/rbt/rbt.go
@@ -911,7 +911,7 @@ func (n *memdbNode) markDelete() {
 	n.flags = (nodeColorBit & n.flags) | deleteFlag
 }
 
-func (n *memdbNode) markUndelete() {
+func (n *memdbNode) unmarkDelete() {
 	n.flags &= ^deleteFlag
 }
 
--- a/internal/unionstore/rbt/rbt.go
+++ b/internal/unionstore/rbt/rbt.go
@@ -124,8 +124,8 @@ func (db *RBT) RevertVAddr(hdr *arena.MemdbVlogHdr) {
 			db.count--
 			db.size -= int(node.klen)
 		} else {
-			node.setKeyFlags(keptFlags)
 			db.dirty = true
+			node.setKeyFlags(keptFlags)
 		}
 	} else {
 		db.size += len(db.vlog.GetValue(hdr.OldValue))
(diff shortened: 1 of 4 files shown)


you06:
@cfzjywxk I test this PR (with RBT enabled) is ok to pass TiDB's CI, PTAL.

zyguan (approved):

internal/unionstore/rbt/rbt.go
```diff
@@ -347,12 +349,16 @@ func (db *RBT) Set(key []byte, value []byte, ops ...kv.FlagsOp) error {
 	// the NeedConstraintCheckInPrewrite flag is temporary,
 	// every write to the node removes the flag unless it's explicitly set.
 	// This set must be in the latest stage so no special processing is needed.
-	var flags kv.KeyFlags
+	flags := x.GetKeyFlags()
+	if flags == 0 && x.vptr.IsNull() && x.isDeleted() {
+		db.count++
+		db.size += int(x.klen)
```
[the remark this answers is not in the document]
you06: The delete flag is dropped in `setKeyFlags`, called in L366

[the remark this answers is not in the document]
cfzjywxk: From a maintainability perspective, calling `unmarkDelete` here is better, as it keeps symmetry with `rbt.go:L123-L125` and avoids implicit dependencies on subsequent cleanup operations.  

Besides, `setKeyFlags` could be renamed to `ResetKeyFlags`, as `setKeyFlags` might be misleading and make people think it's an operation for adding flag bits.

crazycs520, line 355: When reusing the deleted node, doesn't need to clean the key/value inside the node?

you06: The value is already cleaned by `RevertToCheckpoint`, so `x.vptr.IsNull() == true` here.

https://github.com/tikv/client-go/blob/70049ae310bfcc02bbd8f3929ec4fd779d697e7f/internal/unionstore/arena/arena.go#L347-L357

And the key is unchanged, we just reuse it.

internal/unionstore/rbt/rbt.go
```diff
@@ -893,10 +903,22 @@ func (n *memdbNode) getKeyFlags() kv.KeyFlags {
...
-func (n *memdbNode) setKeyFlags(f kv.KeyFlags) {
+func (n *memdbNode) resetKeyFlags(f kv.KeyFlags) {
 	n.flags = (^nodeFlagsMask & n.flags) | uint16(f)
 }
 
+func (n *memdbNode) markDelete() {
+	n.flags = (nodeColorBit & n.flags) | deleteFlag
+}
+
+func (n *memdbNode) markUndelete() {
```
cfzjywxk, line 914:
```suggestion
func (n *memdbNode) unmarkDelete() {
```

cfzjywxk (commented):
@you06 
Do we have a benchmark test with tidb to verify the memory leak issue is resolved?

crazycs520 (commented):
REST LGTM

you06:
> Do we have a benchmark test with tidb to verify the memory leak issue is resolved?

This test can reproduce the memory leak issue in TiDB side, will add it to TiDB.

```go
func TestInsertRollbackMemoryLeak(t *testing.T) {
	store := testkit.CreateMockStore(t)
	tk := testkit.NewTestKit(t, store)
	tk.MustExec("use test")
	tk.MustExec("create table t(a varchar(255) primary key)")
	key1 := strings.Repeat("a", 255)
	key2 := strings.Repeat("b", 255)
	tk.MustExec(`set global tidb_mem_oom_action='cancel'`)
	tk.MustExec("set session tidb_mem_quota_query=10240")
	tk.MustExec("begin")
	tk.MustExec("insert into t values(?)", key2)
	for i := 0; i < 100; i++ {
		err := tk.ExecToErr("insert into t values(?), (?)", key1, key2)
		require.Error(t, err)
		if strings.Contains(err.Error(), "Duplicate") {
			continue
		}
		require.NoError(t, err)
	}
	tk.MustExec("commit")
}
```

PR #1484, opened 2024-11-01
Support adaptive update interval for low resolution ts

MyonKeminta:
Ref: https://github.com/pingcap/tidb/issues/56809

### DIFF
--- a/oracle/oracle.go
+++ b/oracle/oracle.go
@@ -51,6 +51,12 @@ type Oracle interface {
 	GetLowResolutionTimestamp(ctx context.Context, opt *Option) (uint64, error)
 	GetLowResolutionTimestampAsync(ctx context.Context, opt *Option) Future
 	SetLowResolutionTimestampUpdateInterval(time.Duration) error
+	// GetStaleTimestamp generates a timestamp based on the recently fetched timestamp and the elapsed time since
+	// when that timestamp was fetched. The result is expected to be about `prevSecond` seconds before the current
+	// time.
+	// WARNING: This method does not guarantee whether the generated timestamp is legal for accessing the data.
+	// Neither is it safe to use it for verifying the legality of another calculated timestamp.
+	// Be sure to validate the timestamp before using it to access the data.
 	GetStaleTimestamp(ctx context.Context, txnScope string, prevSecond uint64) (uint64, error)
 	IsExpired(lockTimestamp, TTL uint64, opt *Option) bool
 	UntilExpired(lockTimeStamp, TTL uint64, opt *Option) int64
@@ -61,6 +67,13 @@ type Oracle interface {
 
 	// GetAllTSOKeyspaceGroupMinTS gets a minimum timestamp from all TSO keyspace groups.
 	GetAllTSOKeyspaceGroupMinTS(ctx context.Context) (uint64, error)
+
+	// ValidateSnapshotReadTS verifies whether it can be guaranteed that the given readTS doesn't exceed the maximum ts
+	// that has been allocated by the oracle, so that it's safe to use this ts to perform snapshot read, stale read,
+	// etc.
+	// Note that this method only checks the ts from the oracle's perspective. It doesn't check whether the snapshot
+	// has been GCed.
+	ValidateSnapshotReadTS(ctx context.Context, readTS uint64, opt *Option) error
 }
 
 // Future is a future which promises to return a timestamp.
--- a/oracle/oracles/pd.go
+++ b/oracle/oracles/pd.go
@@ -462,6 +462,9 @@ func (o *pdOracle) nextUpdateInterval(now time.Time, requiredStaleness time.Dura
 	} else {
 		newInterval = nextState(checkUnadjustable, checkAdapting, checkNormal, checkRecovering)
 	}
+
+	metrics.TiKVLowResolutionTSOUpdateIntervalSecondsGauge.Set(newInterval.Seconds())
+
 	return newInterval
 }
 
@@ -638,7 +641,7 @@ func (o *pdOracle) GetExternalTimestamp(ctx context.Context) (uint64, error) {
 
 func (o *pdOracle) getCurrentTSForValidation(ctx context.Context, opt *oracle.Option) (uint64, error) {
 	ch := o.tsForValidation.DoChan(opt.TxnScope, func() (interface{}, error) {
-		//metrics.ValidateReadTSFromPDCount.Inc()
+		metrics.TiKVValidateReadTSFromPDCount.Inc()
 
 		// If the call that triggers the execution of this function is canceled by the context, other calls that are
 		// waiting for reusing the same result should not be canceled. So pass context.Background() instead of the
--- a/oracle/oracles/pd.go
+++ b/oracle/oracles/pd.go
@@ -612,7 +612,7 @@ func (o *pdOracle) getCurrentTSForValidation(ctx context.Context, opt *oracle.Op
 
 func (o *pdOracle) ValidateSnapshotReadTS(ctx context.Context, readTS uint64, opt *oracle.Option) error {
 	latestTS, err := o.GetLowResolutionTimestamp(ctx, opt)
-	// If we fail to get latestTS or the readTS exceeds it, get a timestamp from PD to double check.
+	// If we fail to get latestTS or the readTS exceeds it, get a timestamp from PD to double-check.
 	// But we don't need to strictly fetch the latest TS. So if there are already concurrent calls to this function
 	// loading the latest TS, we can just reuse the same result to avoid too many concurrent GetTS calls.
 	if err != nil || readTS > latestTS {
--- a/oracle/oracles/pd.go
+++ b/oracle/oracles/pd.go
@@ -37,7 +37,6 @@ package oracles
 import (
 	"context"
 	"fmt"
-	"golang.org/x/sync/singleflight"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -49,6 +48,7 @@ import (
 	"github.com/tikv/client-go/v2/oracle"
 	pd "github.com/tikv/pd/client"
 	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
 )
 
 var _ oracle.Oracle = &pdOracle{}
--- a/oracle/oracles/pd.go
+++ b/oracle/oracles/pd.go
@@ -55,11 +55,37 @@ var _ oracle.Oracle = &pdOracle{}
 
 const slowDist = 30 * time.Millisecond
 
-const minAllowedAdaptiveUpdateTSInterval = 500 * time.Millisecond
-const adaptiveUpdateTSIntervalShrinkingPreserve = 100 * time.Millisecond
-const adaptiveUpdateTSIntervalResetThreshold = 200 * time.Millisecond
-const adaptiveUpdateTSIntervalResetPerSecond = 20 * time.Millisecond
-const adaptiveUpdateTSIntervalResetTimeout = 5 * time.Minute
+type adaptiveUpdateTSIntervalState int
+
+const (
+	adaptiveUpdateTSIntervalStateNormal adaptiveUpdateTSIntervalState = iota
+	adaptiveUpdateTSIntervalStateAdapting
+	adaptiveUpdateTSIntervalStateRecovering
+	adaptiveUpdateTSIntervalStateUnadjustable
+)
+
+func (s adaptiveUpdateTSIntervalState) String() string {
+	switch s {
+	case adaptiveUpdateTSIntervalStateNormal:
+		return "normal"
+	case adaptiveUpdateTSIntervalStateAdapting:
+		return "adapting"
+	case adaptiveUpdateTSIntervalStateRecovering:
+		return "recovering"
+	case adaptiveUpdateTSIntervalStateUnadjustable:
+		return "unadjustable"
+	default:
+		return "unknown"
+	}
+}
+
+const (
+	minAllowedAdaptiveUpdateTSInterval        = 500 * time.Millisecond
+	adaptiveUpdateTSIntervalShrinkingPreserve = 100 * time.Millisecond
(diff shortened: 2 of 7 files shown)


oracle/oracles/pd.go
```diff
@@ -241,28 +320,145 @@ func (o *pdOracle) getLastTSWithArrivalTS(txnScope string) (*lastTSO, bool) {
...
+				zap.Duration("newAdaptiveUpdateInterval", currentAdaptiveInterval),
+				zap.Stringer("prevState", o.adaptiveUpdateIntervalState.state),
+				zap.Stringer("newState", adaptiveUpdateTSIntervalStateAdapting))
+			o.adaptiveUpdateIntervalState.state = adaptiveUpdateTSIntervalStateAdapting
+		}
+		return currentAdaptiveInterval
+	}
+
+	lastReachDropThresholdTime := time.UnixMilli(o.adaptiveUpdateIntervalState.lastReachDropThresholdTime.Load())
+	if now.Sub(lastReachDropThresholdTime) < adaptiveUpdateTSIntervalDelayBeforeRecovering && currentAdaptiveInterval != configuredInterval {
```
cfzjywxk, line 371: Should we check the `currentAdaptiveInterval != configuredInterval` first as `currentAdaptiveInterval` equals `configuredInterval` most of the time?

oracle/oracles/pd.go
```diff
@@ -241,28 +320,145 @@ func (o *pdOracle) getLastTSWithArrivalTS(txnScope string) (*lastTSO, bool) {
...
+		// There is a recent request that requires a short staleness. Keep the current adaptive interval.
+
+		if o.adaptiveUpdateIntervalState.state != adaptiveUpdateTSIntervalStateAdapting {
```
cfzjywxk, line 374: How could the `currentAdaptiveInterval` is changed but it's not adapting state? A little bit confused, could you add more comments about the state management?

oracle/oracle.go
```diff
@@ -51,6 +51,12 @@ type Oracle interface {
 	GetLowResolutionTimestamp(ctx context.Context, opt *Option) (uint64, error)
 	GetLowResolutionTimestampAsync(ctx context.Context, opt *Option) Future
 	SetLowResolutionTimestampUpdateInterval(time.Duration) error
+	// GetStaleTimestamp generates a timestamp based on the recently fetched timestamp and the elapsed time since
+	// when that timestamp was fetched. The result is expected to be about `prevSecond` seconds before the current
+	// time.
+	// WARNING: This method does not guarantee whether the generated timestamp is legal for accessing the data.
+	// Neither is it safe to use it for verifying the legality of another calculated timestamp.
+	// Be sure to validate the timestamp before using it to access the data.
```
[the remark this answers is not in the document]
MyonKeminta: I've considered this. It makes sence, but it can't be easily done due to the current usage in TiDB repo.

oracle/oracles/pd.go
```diff
@@ -48,19 +48,84 @@ import (
...
+	// The actual interval to update the low resolution ts. If the configured one is too large to satisfy the
+	// requirement of the stale read or snapshot read, the actual interval can be automatically set to a shorter
+	// value than lastTSUpdateInterval.
+	// This value is also possible to be updated by SetLowResolutionTimestampUpdateInterval, which may happen when
+	// user adjusting the update interval manually.
+	adaptiveLastTSUpdateInterval atomic.Int64
+
+	adaptiveUpdateIntervalState struct {
+		// The mutex to avoid racing between updateTS goroutine and SetLowResolutionTimestampUpdateInterval.
+		mu sync.Mutex
```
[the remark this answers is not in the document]
MyonKeminta: But it's actually not protected. There are also some accesses without locking.🤔

oracle/oracles/pd.go
MyonKeminta: It can also be the recovering state, and a new read operation requesting short staleness stops the recovering.

oracle/oracles/pd.go
```diff
@@ -366,3 +610,81 @@ func (o *pdOracle) SetExternalTimestamp(ctx context.Context, ts uint64) error {
...
+			return errors.Errorf("cannot set read timestamp to a future time")
+		}
+	} else {
+		estimatedCurrentTS, err := o.getStaleTimestamp(opt.TxnScope, 0)
+		if err != nil {
+			logutil.Logger(ctx).Warn("failed to estimate current ts by getSlateTimestamp for auto-adjusting update low resolution ts interval",
+				zap.Error(err), zap.Uint64("readTS", readTS), zap.String("txnScope", opt.TxnScope))
+		} else {
+			o.adjustUpdateLowResolutionTSIntervalWithRequestedStaleness(readTS, estimatedCurrentTS, time.Now())
+		}
```
crazycs520, line 658: I'm sorry, I don't understand why we need those logic?

oracle/oracle.go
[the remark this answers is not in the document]
crazycs520: Can we modify `GetStaleTimestamp` to return the latest fetched timestamp, without adding the arrival duration?

oracle/oracles/pd.go
```diff
@@ -241,28 +344,145 @@ func (o *pdOracle) getLastTSWithArrivalTS(txnScope string) (*lastTSO, bool) {
...
+// requesting a short staleness (i.e., the read operation specifies the version it wants to read instead of allocating
+// from PD, and the required version is very close to the current time). In this case, the requested staleness should be
+// passed to this function as the requiredStaleness parameter.
+func (o *pdOracle) nextUpdateInterval(now time.Time, requiredStaleness time.Duration) time.Duration {
+	o.adaptiveUpdateIntervalState.mu.Lock()
+	defer o.adaptiveUpdateIntervalState.mu.Unlock()
+
+	configuredInterval := time.Duration(o.lastTSUpdateInterval.Load())
+	currentAdaptiveInterval := time.Duration(o.adaptiveLastTSUpdateInterval.Load())
+	if configuredInterval <= minAllowedAdaptiveUpdateTSInterval {
```
crazycs520, line 358: If `configuredInterval <= minAllowedAdaptiveUpdateTSInterval` is true, then `adaptiveLastTSUpdateInterval` can be set to `configuredInterval`. The variable `minAllowedAdaptiveUpdateTSInterval` doesn't behave like its name.

How about defining a `minUpdateTSInterval` and both `lastTSUpdateInterval` and `adaptiveLastTSUpdateInterval` should less than `minUpdateTSInterval`.

oracle/oracles/pd.go
```diff
@@ -48,19 +48,108 @@ import (
...
+		return "unadjustable"
+	default:
+		return "unknown"
+	}
+}
+
+const (
+	// minAllowedAdaptiveUpdateTSInterval is the lower bound of the adaptive update ts interval for avoiding an abnormal
+	// read operation causing the update interval to be too short.
+	minAllowedAdaptiveUpdateTSInterval = 500 * time.Millisecond
```
crazycs520, line 97: The minimal value of tidb variable `tidb_low_resolution_tso_update_interval` is 10ms, why `minAllowedAdaptiveUpdateTSInterval` is 500ms?

oracle/oracle.go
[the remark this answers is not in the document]
cfzjywxk: After discussion with @MyonKeminta , the refactor done could be done in another PR as it requires a lot of work.

oracle/oracles/pd.go
```diff
@@ -366,3 +610,81 @@ func (o *pdOracle) SetExternalTimestamp(ctx context.Context, ts uint64) error {
 func (o *pdOracle) GetExternalTimestamp(ctx context.Context) (uint64, error) {
 	return o.c.GetExternalTimestamp(ctx)
 }
+
+func (o *pdOracle) getCurrentTSForValidation(ctx context.Context, opt *oracle.Option) (uint64, error) {
+	ch := o.tsForValidation.DoChan(opt.TxnScope, func() (interface{}, error) {
+		//metrics.ValidateReadTSFromPDCount.Inc()
```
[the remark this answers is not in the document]
MyonKeminta: Ah, there should be a metrics here.

oracle/oracles/pd.go
```diff
@@ -366,3 +610,81 @@ func (o *pdOracle) SetExternalTimestamp(ctx context.Context, ts uint64) error {
...
+
+		// If the call that triggers the execution of this function is canceled by the context, other calls that are
+		// waiting for reusing the same result should not be canceled. So pass context.Background() instead of the
+		// current ctx.
+		res, err := o.GetTimestamp(context.Background(), opt)
+		// After finishing the current call, allow the next call to trigger fetching a new TS.
+		o.tsForValidation.Forget(opt.TxnScope)
```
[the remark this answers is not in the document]
MyonKeminta: No it's not needed. I just realized that I misunderstood the usage of singleflight.

oracle/oracles/pd.go
```diff
@@ -366,3 +610,81 @@ func (o *pdOracle) SetExternalTimestamp(ctx context.Context, ts uint64) error {
...
 }
+
+func (o *pdOracle) getCurrentTSForValidation(ctx context.Context, opt *oracle.Option) (uint64, error) {
+	ch := o.tsForValidation.DoChan(opt.TxnScope, func() (interface{}, error) {
+		//metrics.ValidateReadTSFromPDCount.Inc()
+
+		// If the call that triggers the execution of this function is canceled by the context, other calls that are
+		// waiting for reusing the same result should not be canceled. So pass context.Background() instead of the
+		// current ctx.
+		res, err := o.GetTimestamp(context.Background(), opt)
```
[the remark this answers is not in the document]
MyonKeminta: Using context.Background() here is expected, as said in the above comments.

oracle/oracles/pd.go
MyonKeminta: We estimate the gap between the current time and the readTS, to see if the gap is shorter than the update interval, so that we need to shrink the update interval.

oracle/oracle.go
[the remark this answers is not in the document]
MyonKeminta: > Can we modify `GetStaleTimestamp` to return the latest fetched timestamp, without adding the arrival duration?

What you said is exactly what `GetLowResolutionTSO` does. If we deprecate the `GetStaleTimestamp` method, the change can be complictated and difficult considering the current usage in TiDB repo. Simply changing all usages of `GetStaleTimestamp` to `GetLowResolutionTSO` may significantly reduce the precision of user specified staleness.

oracle/oracles/pd.go
MyonKeminta: The name `minAllowedAdaptiveUpdateTSInterval` expected to limit the *adaptive* update ts interval, to avoid it automatically performs the update too frequently. If the user configures a short interval intentionally, we adopt the user's choice.

oracle/oracles/pd.go
MyonKeminta: To avoid the adaptive update interval mechanism automatically and silently chooses a too low interval can causing too high frequency of getting ts from PD. But we do not reject it if the user configures it explicitly and the low update interval is just what the user expects.

oracle/oracles/pd.go
MyonKeminta: By the way I tried to refine the code of the nextUpdateInterval function, which extracted each branches into small functions and unified the log printing. Please see if it's better than the previous one.

oracle/oracles/pd.go
```diff
@@ -241,28 +345,172 @@ func (o *pdOracle) getLastTSWithArrivalTS(txnScope string) (*lastTSO, bool) {
...
+
+		return adaptiveUpdateTSIntervalStateRecovering, newInterval
+	}
+
+	// Check the specified states in order, until the state becomes determined.
+	// If it's still undetermined after all checks, keep the previous state.
+	nextState := func(checkFuncs ...func() (adaptiveUpdateTSIntervalState, time.Duration)) time.Duration {
+		for _, f := range checkFuncs {
+			state, newInterval := f()
+			if state == none {
```
cfzjywxk, line 425: Should it be returned here if the `state` is `adaptiveUpdateTSIntervalStateUnadjustable` instead of continue processing?

cfzjywxk (commented):
Rest LGTM

oracle/oracles/pd.go
MyonKeminta: If state is not none, it should return at line 454

cfzjywxk (approved):

oracle/oracles/pd.go
```diff
@@ -366,3 +638,84 @@ func (o *pdOracle) SetExternalTimestamp(ctx context.Context, ts uint64) error {
...
+		}
+		return res.Val.(uint64), nil
+	}
+}
+
+func (o *pdOracle) ValidateSnapshotReadTS(ctx context.Context, readTS uint64, opt *oracle.Option) error {
+	latestTS, err := o.GetLowResolutionTimestamp(ctx, opt)
+	// If we fail to get latestTS or the readTS exceeds it, get a timestamp from PD to double-check.
+	// But we don't need to strictly fetch the latest TS. So if there are already concurrent calls to this function
+	// loading the latest TS, we can just reuse the same result to avoid too many concurrent GetTS calls.
```
you06, line 667: The current implementation is enough I think. This is just a minor suggestion. 

Use singleflight to reduce the GetTS calls is reasonable, but it's possible that a valid ts is invalid in this case.

1. thread 1 validate current_ts and send a get ts request, PD allocates its latest TSO and the response is slow.
2. thread 2 validate current_ts and the singleflight will reuse the thread 1's TSO.
3. The GetTS request of thread 1's is responded.

Then thread 2 can get a stale TSO from PD which increases the failure possibility.

Maybe an enhancement is to resend a singleflight GetTS request if `readTS > latestTS` to recheck the timestamp in this case.

crazycs520 (approved):

MyonKeminta:
(Note: further cherry picks needs to carry #1502 at the same time)

PR #1486, opened 2024-11-08
clean up locks when p-dml is failed

you06:
Ref: tikv/tikv#17459

When the pipelined transaction is failed in commit phase, we should also broadcast the txn status and cleanup failed locks.

## Test

Manually inject error when committing primary key of pipelined transaction. 

```sql
MySQL [test]> set session tidb_dml_type=bulk;
Query OK, 0 rows affected (0.000 sec)

MySQL [test]> insert into _sbtest1 select * from sbtest1;
ERROR 1105 (HY000): [pipelined dml] mock error
```

- Before, the locks can block resolved  ts.

image

- This PR, the resolved ts is not blocked.

image

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -1608,8 +1608,32 @@ func (c *twoPhaseCommitter) cleanup(ctx context.Context) {
 		cleanupKeysCtx := context.WithValue(c.store.Ctx(), retry.TxnStartKey, ctx.Value(retry.TxnStartKey))
 		var err error
 		if c.txn.IsPipelined() {
-			// TODO: cleanup pipelined txn
-			// TODO: broadcast txn status
+			if len(c.pipelinedCommitInfo.pipelinedStart) != 0 && len(c.pipelinedCommitInfo.pipelinedEnd) != 0 {
+				broadcastToAllStores(
+					c.txn,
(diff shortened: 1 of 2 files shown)


txnkv/transaction/2pc.go
```diff
@@ -1608,8 +1608,32 @@ func (c *twoPhaseCommitter) cleanup(ctx context.Context) {
 		cleanupKeysCtx := context.WithValue(c.store.Ctx(), retry.TxnStartKey, ctx.Value(retry.TxnStartKey))
 		var err error
 		if c.txn.IsPipelined() {
-			// TODO: cleanup pipelined txn
-			// TODO: broadcast txn status
+			if len(c.pipelinedCommitInfo.pipelinedStart) != 0 && len(c.pipelinedCommitInfo.pipelinedEnd) != 0 {
```
ekexium, line 1611: Should we close the TTL manager here?

you06: It's closed here.
https://github.com/tikv/client-go/blob/70049ae310bfcc02bbd8f3929ec4fd779d697e7f/txnkv/transaction/txn.go#L720

ekexium (approved):

cfzjywxk (approved):

Issue #1488, opened 2024-11-11
Export CommitTS of KVTxn


b6g:
The `CommitTS` of an AsyncCommit transaction is determined from TiKV's `max_ts`. We need it to check the order of the timestamps in TiDB, as described in https://github.com/pingcap/tidb/issues/57165.

cfzjywxk:
The already exists [`tidb_last_txn_info`](https://docs.pingcap.com/tidb/stable/system-variables#tidb_last_txn_info-new-in-v409) could be used to track the last transaction information in current session.

PR #1489, opened 2024-11-11
Export commitTS of KVTxn

b6g:
Ref: https://github.com/pingcap/tidb/issues/57165
Close: https://github.com/tikv/client-go/issues/1488

Export KVTxn.CommitTS to make it available in TiDB.

### DIFF
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -1704,6 +1704,10 @@ func (txn *KVTxn) StartTS() uint64 {
 	return txn.startTS
 }
 
+func (txn *KVTxn) CommitTS() uint64 {
+	return txn.commitTS
+}
+
 // Valid returns if the transaction is valid.
 // A transaction become invalid after commit or rollback.
 func (txn *KVTxn) Valid() bool {
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -1704,6 +1704,7 @@ func (txn *KVTxn) StartTS() uint64 {
 	return txn.startTS
 }
 
+// CommitTS returns the commit timestamp of the already committed transaction, or zero if it's not committed yet.
 func (txn *KVTxn) CommitTS() uint64 {
 	return txn.commitTS
 }
--- a/integration_tests/1pc_test.go
+++ b/integration_tests/1pc_test.go
@@ -112,12 +112,12 @@ func (s *testOnePCSuite) Test1PC() {
 	s.Equal(txn.GetCommitter().GetOnePCCommitTS(), txn.GetCommitter().GetCommitTS())
 	s.Greater(txn.GetCommitter().GetOnePCCommitTS(), txn.StartTS())
 	// Check keys are committed with the same version
-	s.mustGetFromSnapshot(txn.GetCommitTS(), k3, v3)
-	s.mustGetFromSnapshot(txn.GetCommitTS(), k4, v4)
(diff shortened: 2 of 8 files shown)


b6g:
ping

txnkv/transaction/txn.go
```diff
@@ -1700,6 +1700,10 @@ func (txn *KVTxn) StartTS() uint64 {
 	return txn.startTS
 }
 
+func (txn *KVTxn) CommitTS() uint64 {
```
you06, line 1703: There are still some tests using `txn.GetCommitTS` in `1pc_test.go`.

```bash
cd integration_tests && go test ./...
```

you06 (commented):
Export `CommitTS` LGTM.

I also take a look at pingcap/tidb#57305 which guarantee `txn.commit_ts > txn.start_ts` and `txn.commit_ts > prev_txn.commit_ts` in session.

To ensure the `start_ts` is greater than any exist `commit_ts` cross sessions, we may have a field `last_commit_ts` in [store level](https://github.com/tikv/client-go/blob/05d115b3e88bef1af68b57addd5d1b1799e947da/tikv/kv.go#L108) and set it to the latest commit_ts. Each time, we fetch start_ts like this:

```
last_commit_ts = store.last_commit_ts.Load()
start_ts = oracle.GetTSO()
if start_ts < last_commit_ts {
  panic(...)
}
``` 

This may be another task, @cfzjywxk what's your idea.

txnkv/transaction/txn.go
cfzjywxk: @you06 Seems to be an enhancement to add a check before and after the  `ts allocation`, how about filing an issue for it?

b6g: The integration_tests are fixed.

Sounds a good idea to check store's last_commit_ts is smaller than start_ts, because store's last_commit_ts is not directly allocated from pd.

txnkv/transaction/txn.go
```diff
@@ -1704,6 +1704,10 @@ func (txn *KVTxn) StartTS() uint64 {
...
```
cfzjywxk, line 1707:
```suggestion
// CommitTS returns the commit timestamp of the already committed transaction, zero if it's not committed yet.
func (txn *KVTxn) CommitTS() uint64 {
```

Exported functions need explicity comments from the ci & lint.

txnkv/transaction/txn.go
you06: > how about filing an issue for it?

@cfzjywxk  It's mentioned in pingcap/tidb#57165 IMO.

> In theory, all read requests occurring after the write transaction should use a `read_ts` greater than the previous known transaction's `commit_ts`.

cfzjywxk (approved):
LGTM

b6g:
The Integration Test failed because of timeout. Is it expected? How to trigger it to re-run?

b6g:
@you06 Can I get another approval to merge the PR? Thanks!

you06 (approved):

cfzjywxk:
@b6g 
Thanks for the contribution!
For the tidb pr, besides the checking within the same session perhaps we could consider @you06 has proposed 
```
last_commit_ts = store.last_commit_ts.Load()
start_ts = oracle.GetTSO()
if start_ts < last_commit_ts {
  panic(...)
}
```
checking the last commit timestamp after fetching a new ts.

PR #1494, opened 2024-11-13
Allow setting concurrency for pipelined flush and resolveLocks

ekexium:
For tidb variable `tidb_pipelined_dml_resource_policy`


ekexium:
@you06 @cfzjywxk PTAL when you have time, thanks

cfzjywxk (approved):
LGTM

cfzjywxk:
@you06 PTAL

txnkv/transaction/txn.go
```diff
@@ -638,6 +665,34 @@ func (txn *KVTxn) InitPipelinedMemDB() error {
 	return nil
 }
 
+func (txn *KVTxn) throttle() {
```
you06, line 668: What about naming the function with "p-txn"?

```suggestion
func (txn *KVTxn) pipelinedTxnThrottle() {
```

txnkv/transaction/pipelined_flush.go
```diff
@@ -481,7 +480,7 @@ func (c *twoPhaseCommitter) resolveFlushedLocks(bo *retry.Backoffer, start, end
 		fmt.Sprintf("pipelined-dml-%s", status),
 		fmt.Sprintf("pipelined-dml-%s-%d", status, c.startTS),
 		c.store,
-		RESOLVE_CONCURRENCY,
+		c.txn.pipelinedResolveLockConcurrency,
```
you06, line 483: The throttle flow control works when flushing.

And the resolve-lock phase can also consume many resources, and it may also needs to be controlled by throttle, what about adding a TODO here.

txnkv/transaction/txn.go
```diff
@@ -638,6 +665,34 @@ func (txn *KVTxn) InitPipelinedMemDB() error {
...
+	if sleepMs == 0 {
+		return
+	}
+	logutil.BgLogger().Info(
+		"[pipelined dml] throttle",
+		zap.Int("sleepMs", sleepMs),
+		zap.Float64("writeThrottleRatio", txn.writeThrottleRatio),
+		zap.Uint64("session", txn.committer.sessionID),
+		zap.Uint64("startTS", txn.startTS),
+	)
```
you06, line 692: Logger util in client-go will not print arguments in reversed order.

```suggestion
	logutil.BgLogger().Info(
		"[pipelined dml] throttle",
		zap.Uint64("session", txn.committer.sessionID),
		zap.Uint64("startTS", txn.startTS),
		zap.Int("sleepMs", sleepMs),
		zap.Float64("writeThrottleRatio", txn.writeThrottleRatio),
	)
```

examples/gcworker/go.mod
```diff
@@ -1,6 +1,6 @@
 module gcworker
 
-go 1.23
+go 1.23.2
```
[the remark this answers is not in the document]
ekexium: I forget the reason to do so, probably because of compiling or CI stuff. I will revert it

txnkv/transaction/pipelined_flush.go
ekexium: Throttling the resolve lock goroutine is in our current plan I think. For now we rely on the concurrency to constrain it. If we want a better control of its resource usage we might need to consider a more comprehensive mechanism, e.g. one that helps reduce the overhead of resolving locks of multiple p-txns running in parallel.

txnkv/transaction/txn.go
```diff
@@ -109,12 +111,20 @@ func (e *tempLockBufferEntry) trySkipLockingOnRetry(returnValue bool, checkExist
...
+
 // TxnOptions indicates the option when beginning a transaction.
 // TxnOptions are set by the TxnOption values passed to Begin
 type TxnOptions struct {
-	TxnScope       string
-	StartTS        *uint64
-	PipelinedMemDB bool
+	TxnScope     string
+	StartTS      *uint64
+	PipelinedTxn PipelinedTxnOptions
```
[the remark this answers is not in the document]
ekexium: Manually sepecifying default values seems a bit strange to me. These fields must be correctly specified by upper layers if they are used. And they have a valid range check when used.
I add validity checks for them before usage. PTAL

you06 (approved):

PR #1509, opened 2024-11-22
retry as stale read on replica

mittalrishabh:
Issue #17422
We have seen that retrying stale read as follower read caused raft channel clogging which increases the blast radius and impact other regions. 
This would be the flow with this change
1. Try on replica in same AZ
2. Retry in leader if original request is not on leader and leader is marked serverIsBusy and DeadlineExceeded
3. Retry on replica as stale read

I expect requests to be succeeded in leader read if the leader is not slow. We should retry it as follower read if leader already slow. 

Unit Test
1. I injected the error between TiDB-->TikV and confirmed that reads are retried as stale reads. 

Test results

Screenshot 2024-11-22 at 10 36 54 AM
Screenshot 2024-11-22 at 10 34 02 AM


internal/locate/replica_selector.go
```diff
@@ -185,14 +185,28 @@ func (s *replicaSelector) nextForReplicaReadMixed(req *tikvrpc.Request) {
 		if s.isStaleRead && s.attempts == 1 {
 			// stale-read request first access.
 			if !s.target.store.IsLabelsMatch(s.option.labels) && s.target.peer.Id != s.region.GetLeaderPeerID() {
-				// If the target replica's labels is not match and not leader, use replica read.
```
cfzjywxk, line 188: The comments could be kept.

internal/locate/replica_selector.go
```diff
@@ -300,6 +315,15 @@ func (s *ReplicaSelectMixedStrategy) next(selector *replicaSelector) *replica {
 	return nil
 }
 
+func (s *ReplicaSelectMixedStrategy) canSendReplicaRead(selector *replicaSelector) bool {
```
cfzjywxk, line 318: Another approach is to integrate the following logic with 
```
if !s.target.store.IsLabelsMatch(s.option.labels) && s.target.peer.Id != s.region.GetLeaderPeerID()
```
,  and abstract a method like
```
func (s replicaSelector*) DecideReplicaReadType() ReadType {
    ...
}

if readType == staleRead {
    req.StaleRead = true
    req.ReplicaRead = false
} else if read Type == followerRead {
    ....
} else {
   ...
}
```

cfzjywxk (commented):
The strategy it worth considering, similar to https://github.com/tikv/client-go/pull/1109 before.

Please add a unit test case to verify its functionality and fix the failed unit test cases.

crazycs520 (approved):

mittalrishabh:
[zyguan](https://github.com/zyguan), [you06](https://github.com/you06) please review these changes

zyguan (approved):

PR #1512, opened 2024-12-03
Handle error invalid_max_ts_update

ekexium:
Ref https://github.com/tikv/tikv/issues/17916

The error will directly propagate to user and abort current stmt.

The string match is more suitable for cherry-picking to older versions, we may consider introducing a new error type in master branch.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1432,12 +1432,18 @@ func regionErrorToLabel(e *errorpb.Error) string {
 		return "mismatch_peer_id"
 	} else if e.GetBucketVersionNotMatch() != nil {
 		return "bucket_version_not_match"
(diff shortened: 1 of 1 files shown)


internal/locate/region_request.go
```diff
@@ -1735,6 +1741,10 @@ func (s *RegionRequestSender) onRegionError(
 		return false, nil
 	}
 
+	if isInvalidMaxTsUpdate(regionErr) {
```
cfzjywxk, line 1744: How about logging the request information here too?

ekexium: Makes sense

cfzjywxk (approved):

you06 (approved):

PR #1516, opened 2024-12-11
pdclient: Add caller info to pd client

okJiang:
ref https://github.com/tikv/pd/issues/8593


internal/locate/pd_codec.go
```diff
@@ -71,7 +72,7 @@ func NewCodecPDClientWithKeyspace(mode apicodec.Mode, client pd.Client, keyspace
 		return nil, err
 	}
 
-	return &CodecPDClient{client, codec}, nil
+	return &CodecPDClient{client.WithCallerComponent("codec-pd-client"), codec}, nil
```
JmPotato, line 75: Should we also extract this as a constant to reuse it in line 61?

okJiang: LGTM

tikv/kv.go
```diff
@@ -285,7 +285,7 @@ func NewKVStore(uuid string, pdClient pd.Client, spkv SafePointKV, tikvclient Cl
 		clusterID:       pdClient.GetClusterID(context.TODO()),
 		uuid:            uuid,
 		oracle:          o,
-		pdClient:        pdClient,
+		pdClient:        pdClient.WithCallerComponent("kv-store"),
```
bufferflies, line 288: How about defining and describing one enum not using string?

util/pd_interceptor.go
```diff
@@ -64,6 +65,10 @@ type InterceptedPDClient struct {
 	pd.Client
 }
 
+func NewInterceptedPDClient(client pd.Client) *InterceptedPDClient {
```
bufferflies, line 68: needs comment for public.

bufferflies (commented):
rest lgtm

tikv/kv.go
okJiang: How about defining them in pd repo? If you agree, I can open a pr to do it.

bufferflies (approved):

JmPotato (approved):

PR #1523, opened 2024-12-16
region_request: tiny refactor - init function failpointSendReqResult

AndreMouche:
Wrap the logic of handling failpoint `tikvStoreSendReqResult` into a function to make the main logic of `SendReqCtx` function more concise and friendly to maintain.

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -730,43 +730,8 @@ func (s *RegionRequestSender) SendReqCtx(
 		bo.SetCtx(opentracing.ContextWithSpan(bo.GetCtx(), span1))
 	}
 
-	if val, err := util.EvalFailpoint("tikvStoreSendReqResult"); err == nil {
-		if s, ok := val.(string); ok {
-			switch s {
-			case "timeout":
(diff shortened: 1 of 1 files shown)


internal/locate/region_request.go
```diff
@@ -1929,3 +1894,62 @@ func (s *baseReplicaSelector) backoffOnNoCandidate(bo *retry.Backoffer) error {
...
+	err error,
+) {
+	val, e := util.EvalFailpoint("tikvStoreSendReqResult")
+	if e != nil {
+		return
+	}
+	errMsg, ok := val.(string)
+	if !ok {
+		return
+	}
```
MyonKeminta, line 1910: It's ok to just follow the original pattern  `if val, err := util.EvalFailpoint("tikvStoreSendReqResult"); err == nil { ...`, instead of reformatting it. It's actually the expanded version of the `failpoint.Inject` pattern.

MyonKeminta (approved):

MyonKeminta (approved):

ekexium (approved):

PR #1525, opened 2024-12-17
Upgrade the PD client to adopt the latest refactor

JmPotato:
Integration testing can only resume normal operation after TiDB has completed its update.


integration_tests/pd_api_test.go
```diff
@@ -49,7 +50,7 @@ type apiTestSuite struct {
 func (s *apiTestSuite) SetupTest() {
 	require := s.Require()
 	addrs := strings.Split(*pdAddrs, ",")
-	pdClient, err := pd.NewClient(addrs, pd.SecurityOption{})
+	pdClient, err := pd.NewClient(caller.Component("client-go-integration-test"), addrs, pd.SecurityOption{})
```
rleungx, line 53: @okJiang, do we have a standard for the definition of the component?

okJiang: If it is in the test, we can use caller.TestComponent. Otherwise, we can put it empty now, I will update them in https://github.com/tikv/client-go/pull/1516

okJiang: BTW, how about adding a `DefaultComponent = ""` in pd repo?

you06 (approved):

rleungx (approved):

PR #1526, opened 2024-12-17
grpc: 1.63.2 -> 1.69.0

iosmanthus:
`grpc.NewSharedBufferPool` is enabled by default in grpc `1.69.0`, we could remove the `GrpcSharedBufferPool` option of client-go.

Fore more information, see: https://pkg.go.dev/google.golang.org/grpc/experimental#WithBufferPool


hawkingrei:
ref https://github.com/pingcap/tidb/issues/50627

disksing (approved):

heavycrystal:
does this PR need work or can it be merged? 
`google.golang.org/grpc` is now at `v1.72.0`

benmeadowcroft (approved):

tscolari:
Any change we could get these in?
As it stands it's a bit limiting the combination of external dependencies versions necessary to build this package.

Just a note that it's not that it's enabled by default, it was completely removed on later versions, which makes it impossible to build as part of any system that is up to date with the grpc libraries.

PR #1528, opened 2024-12-18
tikvrpc: avoid attaching ctx twice

zyguan:
Currently `AttachContext` will be called at least twice for each request, which is unnecessary and leads to more mem allocs.

image


zyguan:
included in #1555

PR #1529, opened 2024-12-18
txnkv/transaction/prewrite: tiny refactor HandleSingleBatch

AndreMouche:
Do some tiny refactor for `handleSingleBatch` 
Extracted some functions to make more friendly to reviewers and maintainers, meanwhile make the unit testing easier to write too


txnkv/transaction/prewrite.go
```diff
@@ -226,286 +227,368 @@ func (action actionPrewrite) handleSingleBatch(
...
+	if handler.resolvingRecordToken != nil {
+		handler.committer.store.GetLockResolver().ResolveLocksDone(handler.committer.startTS, *handler.resolvingRecordToken)
+	}
+}
+
+func (handler *prewrite1BatchReqHandler) beforeSend() {
+	handler.attempts++
+	if handler.action.hasRpcRetries {
+		handler.req.IsRetryRequest = true
+	}
```
you06, line 349: What about
```suggestion
	}
	if handler.attempts == 1 {
		return
	}
```

txnkv/transaction/prewrite.go
```diff
@@ -226,286 +227,368 @@ func (action actionPrewrite) handleSingleBatch(
...
+	reqBegin := time.Now()
+	if reqBegin.Sub(handler.begin) > slowRequestThreshold {
+		logutil.BgLogger().Warn(
+			"slow prewrite request",
+			zap.Uint64("startTS", handler.committer.startTS),
+			zap.Stringer("region", &handler.batch.region),
+			zap.Int("attempts", handler.attempts),
+		)
+		handler.begin = time.Now()
```
you06, line 358:
```suggestion
		handler.begin = reqBegin
```

txnkv/transaction/prewrite.go
```diff
@@ -226,286 +227,368 @@ func (action actionPrewrite) handleSingleBatch(
...
+	return nil
+}
+
+// handleSingleBatchSucceed handles the response when the prewrite request is successful.
+func (handler *prewrite1BatchReqHandler) handleSingleBatchSucceed(prewriteResp *kvrpcpb.PrewriteResponse) error {
+	// Clear the RPC Error since the request is evaluated successfully.
+	handler.sender.SetRPCError(nil)
+
+	// Update CommitDetails
+	reqDuration := time.Since(handler.begin)
```
you06, line 525: `handler.begin` is not [`reqBegin`](https://github.com/tikv/client-go/blob/6e324f2e07c93680aa08a7ca0a8dca89a3f70e30/txnkv/transaction/prewrite.go#L350).

txnkv/transaction/prewrite.go
```diff
@@ -226,286 +227,368 @@ func (action actionPrewrite) handleSingleBatch(
...
+		prewriteResp.ExecDetailsV2,
+	)
+
+	if handler.batch.isPrimary {
+		// After writing the primary key, if the size of the transaction is larger than 32M,
+		// start the ttlManager. The ttlManager will be closed in tikvTxn.Commit().
+		// In this case 1PC is not expected to be used, but still check it for safety.
+		if int64(handler.committer.txnSize) > config.GetGlobalConfig().TiKVClient.TTLRefreshedTxnSize &&
+			prewriteResp.OnePcCommitTs == 0 {
+			handler.committer.run(handler.committer, nil, false)
```
you06, line 539: Though `ttlManager` is embed in `twoPhaseCommitter`, call `committer.run` is little weird. Better to change it to:

```suggestion
			handler.ttlManager.committer.run(handler.committer, nil, false)
```

txnkv/transaction/prewrite.go
```diff
@@ -226,286 +227,368 @@ func (action actionPrewrite) handleSingleBatch(
...
-			}
+	if err := action.handleSingleBatchFailpoint(c, bo, batch); err != nil {
+		return err
+	}
+
+	handler := action.newSingleBatchPrewriteReqHandler(c, batch, bo)
+
+	for {
+		retryable, err := handler.sendReqAndCheck()
+		if !retryable {
```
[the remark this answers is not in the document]
cfzjywxk: +1, the implicit constraint `retryable == (err != nil)` would possibly raising the risk of mistakes in `sendReqAndCheck` function.

Another option, 
```
err := handler.sendReqAndCheck()
if err != nil {
    // Return error to the upper layer.
}
// Continue and retry.
```
and do not return error in `sendReqAndCheck` if it's considered retryable.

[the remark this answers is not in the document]
AndreMouche: there is one case that the err is nil and we should not retry:
When we encounter EpochNotMatch, we will perform doActionOnMutations immediately. Regardless of whether doActionOnMutations is successful or not, we should not attempt to retry at this time.

you06 (approved):

cfzjywxk (approved):

PR #1531, opened 2024-12-19
pick-6.5: Support adaptive update interval for low resolution ts (#1484)

ekexium:
Cherry pick #1484 and #1502 to tidb-6.5

### DIFF
--- a/oracle/oracles/pd.go
+++ b/oracle/oracles/pd.go
@@ -185,6 +185,7 @@ func (p *lastTSOPointer) compareAndSwap(old, new *lastTSO) bool {
 	return stdatomic.CompareAndSwapPointer(&p.p, unsafe.Pointer(old), unsafe.Pointer(new))
 }
 
+// PDOracleOptions is the options for creating a new pdOracle.
 type PDOracleOptions struct {
 	// The duration to update the last ts, i.e., the low resolution ts.
 	UpdateInterval time.Duration
--- a/oracle/oracles/pd.go
+++ b/oracle/oracles/pd.go
@@ -486,6 +486,8 @@ func (o *pdOracle) updateTS(ctx context.Context) {
 	ticker := time.NewTicker(currentInterval)
 	defer ticker.Stop()
 
+	// Note that as `doUpdate` updates last tick time while `nextUpdateInterval` may perform calculation depending on the
+	// last tick time, `doUpdate` should be called after finishing calculating the next interval.
 	doUpdate := func(now time.Time) {
 		// Update the timestamp for each txnScope
 		o.lastTSMap.Range(func(key, _ interface{}) bool {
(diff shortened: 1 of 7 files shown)


ekexium:
> Also pick #1502 in this PR?

I initially planned to cherry-pick these changes in separate PRs. After taking a look now, since it's only 11 lines, let's combine them into one.

you06 (approved):

oracle/oracles/pd.go
```diff
@@ -47,19 +48,111 @@ import (
 	"github.com/tikv/client-go/v2/metrics"
 	"github.com/tikv/client-go/v2/oracle"
 	pd "github.com/tikv/pd/client"
+	"go.uber.org/atomic"
```
MyonKeminta, line 51: Is there any reason to use the uber version rhe the original PR didn't? I think using the std one should be just fine🤔 Is this a mistake?

oracle/oracles/pd.go
```diff
@@ -47,19 +48,111 @@ import (
...
+)
+
 // pdOracle is an Oracle that uses a placement driver client as source.
 type pdOracle struct {
 	c pd.Client
 	// txn_scope (string) -> lastTSPointer (*lastTSOPointer)
 	lastTSMap sync.Map
 	quit      chan struct{}
+	// The configured interval to update the low resolution ts. Set by SetLowResolutionTimestampUpdateInterval.
+	// For TiDB, this is directly controlled by the system variable `tidb_low_resolution_tso_update_interval`.
```
MyonKeminta, line 123: You can consider picking from other cherrypicks to old branches, in which I've modified some contents. e.g. https://github.com/tikv/client-go/pull/1491

oracle/oracles/pd.go
ekexium: std atomic was introduced in go 1.19, while release-6.5 uses go 1.18.

MyonKeminta: 😰

oracle/oracles/pd.go
ekexium: That makes sense. I've picked #1491 as the last commit.

oracle/oracles/pd.go
```diff
@@ -486,8 +485,6 @@ func (o *pdOracle) updateTS(ctx context.Context) {
 	ticker := time.NewTicker(currentInterval)
 	defer ticker.Stop()
 
-	// Note that as `doUpdate` updates last tick time while `nextUpdateInterval` may perform calculation depending on the
-	// last tick time, `doUpdate` should be called after finishing calculating the next interval.
```
MyonKeminta: Looks this is part of #1502 and don't need to be removed

ekexium: Ah yes

MyonKeminta (approved):

PR #1532, opened 2024-12-19
*: Optimize executor runtime stats performance

crazycs520:
This PR tiny optimize `RegionRequestRuntimeStats` and `SnapshotRuntimeStats` by remove useless field and avoid create map.


internal/locate/region_request.go
```diff
@@ -141,30 +145,60 @@ type RequestErrorStats struct {
...
 		}
 		return
 	}
 	stat.Count++
-	stat.Consume += int64(d)
+	stat.Consume += d
+}
+
+// GetRPCCount returns the total rpc types count.
+func (r *RegionRequestRuntimeStats) GetRPCCount() int {
```
zyguan, line 184: How about to use `GetRPCStatsCount` (or sth else)? This name make me think it returns the total number of RPCs of all cmd types.

internal/locate/region_request.go
```diff
@@ -127,7 +127,11 @@ func (s *RegionRequestSender) String() string {
 
 // RegionRequestRuntimeStats records the runtime stats of send region requests.
 type RegionRequestRuntimeStats struct {
-	RPCStats map[tikvrpc.CmdType]*RPCRuntimeStats
+	// FirstRPCStats is the stats of first kinds of rpc request, since in most cases, only one kind of rpc request is sent at a time,
+	// this is to avoid allocating map memory.
+	FirstRPCStats RPCRuntimeStats
+	// OtherRPCStatsMap uses to record another types of RPC requests.
+	OtherRPCStatsMap map[tikvrpc.CmdType]*RPCRuntimeStats
```
zyguan, line 134: Just a suggestion, since one RegionRequestSender typically only send one (or a few) kind(s) of RPC, I guess we can just use a slice here to make the code clean and simple.

internal/locate/region_request.go
crazycs520: great, done.

internal/locate/region_request.go
crazycs520: great, done.

zyguan (approved):

.github/workflows/compatibility.yml
```diff
@@ -24,8 +24,8 @@ jobs:
     - name: Checkout TiDB
       uses: actions/checkout@v2
       with:
-        repository: JmPotato/tidb
-        ref: update_pd_client
+        repository: crazycs520/tidb
```
siddontang, line 27: why our own repo here? not offical pingcap/tidb? @cfzjywxk

crazycs520: This is to pass the CI test

crazycs520: After https://github.com/pingcap/tidb/pull/58420 merged, https://github.com/tikv/client-go/pull/1540 will recover ci reference to original tidb.

cfzjywxk: @siddontang 

The TiDB compatibility tests in `client-go` can be removed in the future. These tests make modifying `client-go` very cumbersome after the repository split. In practice, updating the `client-go` dependency in TiDB already serves as a compatibility test.

PR #1534, opened 2024-12-20
retry stale read as stale if leader is not accessible

mittalrishabh:
issue number : https://github.com/tikv/tikv/issues/17422
It is the extension of the original PR https://github.com/tikv/client-go/pull/1509. I have observed that in case of hot regions, leaders are inaccessible and stale reads are retried as follower_follower. I have added a change if leader is not reachable or epoch doesn't mach than retry stale read as stale instead of follower. 
Screenshot 2024-12-20 at 1 35 37 PM

### DIFF
--- a/internal/locate/replica_selector.go
+++ b/internal/locate/replica_selector.go
@@ -308,7 +308,7 @@ func (s *ReplicaSelectMixedStrategy) canSendReplicaRead(selector *replicaSelecto
 	replicas := selector.replicas
 	replica := replicas[s.leaderIdx]
 
-	if leader.attempts == 0 || replica.hasFlag(deadlineErrUsingConfTimeoutFlag) || replica.hasFlag(serverIsBusyFlag) {
+	if replica.attempts == 0 || replica.hasFlag(deadlineErrUsingConfTimeoutFlag) || replica.hasFlag(serverIsBusyFlag) {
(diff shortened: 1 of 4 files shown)


internal/locate/replica_selector.go
```diff
@@ -307,8 +307,10 @@ func (s *ReplicaSelectMixedStrategy) next(selector *replicaSelector) *replica {
 func (s *ReplicaSelectMixedStrategy) canSendReplicaRead(selector *replicaSelector) bool {
 	replicas := selector.replicas
 	replica := replicas[s.leaderIdx]
-	if replica.hasFlag(deadlineErrUsingConfTimeoutFlag) || replica.hasFlag(serverIsBusyFlag) {
-		// don't overwhelm the leader if it is busy
+
+	maxAttempt := 1
+	if !replica.isExhausted(maxAttempt, 0) || replica.hasFlag(deadlineErrUsingConfTimeoutFlag) || replica.hasFlag(serverIsBusyFlag) {
```
cfzjywxk, line 312: The `!replica.isExhausted(maxAttempt, 0)` is a little bit confusing, is it intended to check the `leader.attempts == 0` which means the leader is not attempted at all?

mittalrishabh: yes, i modified the code.

PR #1537, opened 2024-12-24
txn: store last commit transaction's info

you06:
Ref: pingcap/tidb#57165

Implement this comment: https://github.com/tikv/client-go/pull/1489#pullrequestreview-2457218539

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -120,6 +120,7 @@ type kvstore interface {
 	IsClose() bool
 	// Go run the function in a separate goroutine.
 	Go(f func()) error
+	storeCommitInfo
 }
 
 // twoPhaseCommitter executes a two-phase commit protocol.
@@ -343,7 +344,7 @@ type PlainMutations struct {
 	flags  []CommitterMutationFlags
 }
 
-// NewPlainMutations creates a PlainMutations object with sizeHint reserved.
+// NewPlainMutations creates a PlainMutations object with sizeHint reserved.G
 func NewPlainMutations(sizeHint int) PlainMutations {
 	return PlainMutations{
 		ops:    make([]kvrpcpb.Op, 0, sizeHint),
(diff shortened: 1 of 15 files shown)


util/commit_ts_verify.go
```diff
@@ -0,0 +1,86 @@
...
+type TSVerifier struct {
+	scope2commitInfo sync.Map
+}
+
+// NewTSVerifier creates a new TSVerifier.
+func NewTSVerifier() *TSVerifier {
+	return &TSVerifier{}
+}
+
+// StoreCommitInfo stores the commit information of a transaction.
```
cfzjywxk, line 56: How about using the same function name `SetLastCommitInfo` like in the upper layer.

util/commit_ts_verify.go
```diff
@@ -0,0 +1,86 @@
...
+func (t *TSVerifier) StoreCommitInfo(txnScope string, commitInfo *CommitInfo) {
+	if txnScope == "" {
+		txnScope = GlobalTxnScope
+	}
+	for {
+		old, loaded := t.scope2commitInfo.LoadOrStore(txnScope, commitInfo)
```
cfzjywxk, line 62: This is a bit costy in the performance crticial path, perhaps it could be simplified after `txnScope` is removed and there is just one cas operation required.

oracle/oracles/pd.go
```diff
@@ -228,8 +235,9 @@ func (o *pdOracle) GetAllTSOKeyspaceGroupMinTS(ctx context.Context) (uint64, err
 
 type tsFuture struct {
 	tso.TSFuture
-	o        *pdOracle
-	txnScope string
+	o          *pdOracle
+	txnScope   string
+	commitInfo *util.CommitInfo
```
cfzjywxk, line 240: Perhaps `lastCommitInfo` is better.

txnkv/transaction/2pc.go
```diff
@@ -1903,6 +1905,7 @@ func (c *twoPhaseCommitter) execute(ctx context.Context) (err error) {
 		}
 	}
 	atomic.StoreUint64(&c.commitTS, commitTS)
+	c.updateStoreCommitInfo()
```
cfzjywxk, line 1908: The transaction is not ensured to be committed successfully here, for example, if 2PC is used.

PR #1542, opened 2025-01-07
*: bump pd client

rleungx:

PR #1543, opened 2025-01-07
integrate circuitbreaker for region calls

Tema:
Issue Number: ref https://github.com/tikv/pd/issues/8678

Add a global circuit breaker to the context of each call to get region data from PD.  
The default circuit breaker is in the disabled state and has no effect. 
This PR also exposes a method to tweak circuit breaker settings which can be used from TiDB layer to enable and configure it.

### DIFF
--- a/config/client.go
+++ b/config/client.go
@@ -100,6 +100,8 @@ type TiKVClient struct {
 	// If a Region has not been accessed for more than the given duration (in seconds), it
 	// will be reloaded from the PD.
 	RegionCacheTTL uint `toml:"region-cache-ttl" json:"region-cache-ttl"`
+	// CircuitBreakerSettingsList is the config for default circuit breaker settings
+	CircuitBreakerSettingsList CircuitBreakerSettingsList `toml:"circuit-breaker" json:"circuit-breaker"`
 	// If a store has been up to the limit, it will return error for successive request to
 	// prevent the store occupying too much token in dispatching level.
 	StoreLimit int64 `toml:"store-limit" json:"store-limit"`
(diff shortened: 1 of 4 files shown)


rleungx:
I have updated the pd client version through https://github.com/tikv/client-go/pull/1542

MyonKeminta:
Please fix CI

internal/locate/region_cache.go
```diff
@@ -133,6 +134,20 @@ func nextTTL(ts int64) int64 {
 	return ts + regionCacheTTLSec + jitter
 }
 
+var pdRegionMetaCircuitBreaker = circuitbreaker.NewCircuitBreaker(
```
rleungx, line 137: Shall we move it to `config`? @MyonKeminta

rleungx:
> @rleungx I've rebased branch to bring your #1542 and added test through enforcing the CB in the context in mock PDClient. Not sure if we need more tests as it is challenging given all tests are using pdClient mock which does not have an interceptor.

IMO, we can add more integration tests later.

internal/locate/region_cache.go
MyonKeminta: Fine to me.
But if you are not going to add some fields to the config file, I think maybe it would be better to add another single file in the `internal/locate` directory for holding this variable.

MyonKeminta: Nevermind. I just feel it still makes me kind of incomfortable if it's put in the `config` package while nothing of it is configurable by config file. Now it looks fine to me.

config/client.go
```diff
@@ -149,6 +151,23 @@ type CoprocessorCache struct {
 	AdmissionMinProcessMs uint64 `toml:"admission-min-process-ms" json:"-"`
 }
 
+// CircuitBreakerSettings is the config for default circuit breaker settings excluding the error rate
+type CircuitBreakerSettings struct {
```
rleungx, line 155: How about referring to `config/retry/backoff.go`? We can use NewCircuitBreakerWithVars or something like that and only add a var for the error rate threshold. The other configs can be hard coded and we don't need to support changing them through the config file. @Tema @MyonKeminta @okJiang WDYT?

rleungx: How about propagating the error rate instead of the whole `Settings`? As for the other configs, as we discussed before, we can make them hard-coded temporarily. If we do need to change them, then we can make them configurable. Right now, it's better to hide them to reduce the complexity.

rleungx: LGTM, /cc @okJiang

okJiang: LGTM

internal/locate/region_cache.go
okJiang, line 1: Seems you forgot to update L2218 and L2283?

internal/locate/region_cache.go
```diff
@@ -133,6 +134,23 @@ func nextTTL(ts int64) int64 {
...
+		CoolDownInterval:     10,
+		HalfOpenSuccessCount: 1,
+	})
+
+func withPDCircuitBreaker(ctx context.Context) context.Context {
+	return circuitbreaker.WithCircuitBreaker(ctx, pdRegionMetaCircuitBreaker)
+}
+
+// ChangePdRegionMetaCircuitBreakerSettings changes circuit breaker changes for region metadata calls
+func ChangePdRegionMetaCircuitBreakerSettings(apply func(config *circuitbreaker.Settings)) {
```
rleungx, line 150:
```suggestion
func ChangePDRegionMetaCircuitBreakerSettings(apply func(config *circuitbreaker.Settings)) {
```

tikv/region.go
```diff
@@ -197,6 +198,11 @@ func SetRegionCacheTTLSec(t int64) {
 	locate.SetRegionCacheTTLSec(t)
 }
 
+// ChangePdRegionMetaCircuitBreakerSettings changes circuit breaker settings for region metadata calls
+func ChangePdRegionMetaCircuitBreakerSettings(apply func(config *circuitbreaker.Settings)) {
+	locate.ChangePdRegionMetaCircuitBreakerSettings(apply)
```
rleungx, line 203: ditto

rleungx (commented):
The rest LGTM

okJiang (approved):

rleungx (approved):

MyonKeminta (approved):

PR #1513, opened 2024-12-04
Try to validate read ts for all RPC requests

MyonKeminta:
Ref: https://github.com/pingcap/tidb/issues/57786

PR #1549, opened 2025-01-07
Try to validate read ts for all RPC requests (#1513)

you06:
Cherry pick #1513 and #1533 to `tidb-8.5`.


ekexium (approved):

oracle/oracles/pd.go
```diff
@@ -670,22 +673,34 @@ func (o *pdOracle) getCurrentTSForValidation(ctx context.Context, opt *oracle.Op
...
+
+	latestTSInfo, exists := o.getLastTSWithArrivalTS(opt.TxnScope)
+	// If we fail to get latestTSInfo or the readTS exceeds it, get a timestamp from PD to double-check.
 	// But we don't need to strictly fetch the latest TS. So if there are already concurrent calls to this function
 	// loading the latest TS, we can just reuse the same result to avoid too many concurrent GetTS calls.
-	if err != nil || readTS > latestTS {
+	if !exists || readTS > latestTSInfo.tso {
 		currentTS, err := o.getCurrentTSForValidation(ctx, opt)
 		if err != nil {
 			return errors.Errorf("fail to validate read timestamp: %v", err)
```
cfzjywxk, line 691: Maybe attach more information here as the read request would be rejected in this case?

MyonKeminta: The log didn't exist in master branch. I think it's better to add it to master first and then cherry-pick it if necessary.

MyonKeminta (approved):

oracle/oracles/pd.go
cfzjywxk: If this branch is reached, the user client may receive this error. Is this situation always unexpected? Could it lead to false-positive judgments or similar issues?

I understand that it is indeed unexpected because each PD timestamp (TS) will update the last TS first, Do I understand right?

PR #1551, opened 2025-01-07
Add txn write conflict metrics

### LINKED ISSUE #1550: Add transaction write conflict metrics
Currently, there is no metrics of transaction write conflict. When a transaction write conflict occurs, you can only view it in the log.
The write_conflict value of the grafana panel is always 0.
Image

ImSjt:
Issue Number: close #1550
Add transaction write conflict metrics.

### DIFF
--- a/txnkv/transaction/prewrite.go
+++ b/txnkv/transaction/prewrite.go
@@ -467,9 +467,6 @@ func (handler *prewrite1BatchReqHandler) extractKeyErrs(keyErrs []*kvrpcpb.KeyEr
 		// Extract lock from key error
 		lock, err1 := txnlock.ExtractLockFromKeyErr(keyErr)
 		if err1 != nil {
-			if tikverr.IsErrWriteConflict(err1) {
-				metrics.TiKVTxnWriteConflictCounter.Inc()
-			}
 			return nil, err1
 		}
 		if _, ok := logged[lock.TxnID]; !ok {
@@ -490,7 +487,6 @@ func (handler *prewrite1BatchReqHandler) extractKeyErrs(keyErrs []*kvrpcpb.KeyEr
 		// TiKV returns lock.TTL = 0, and we still need to resolve the lock.
 		if (lock.TxnID > handler.committer.startTS && !handler.committer.isPessimistic) ||
 			handler.committer.txn.prewriteEncounterLockPolicy == NoResolvePolicy {
-			metrics.TiKVTxnWriteConflictCounter.Inc()
 			return nil, tikverr.NewErrWriteConflictWithArgs(
 				handler.committer.startTS,
 				lock.TxnID,
--- a/error/error.go
+++ b/error/error.go
@@ -43,6 +43,7 @@ import (
 	"github.com/pingcap/log"
 	"github.com/pkg/errors"
 	"github.com/tikv/client-go/v2/internal/logutil"
+	"github.com/tikv/client-go/v2/metrics"
 	"github.com/tikv/client-go/v2/util"
 	"github.com/tikv/client-go/v2/util/redact"
 	"go.uber.org/zap"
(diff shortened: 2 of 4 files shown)


ekexium (commented):
LGTM. Please resolve the conflicts

ImSjt:
@ekexium PTAL again.

ekexium (commented):
Actually I think the metric is not related to lock_resolver. It's more like "WriteConflicts errrors generated from the client side".
Maybe we can remove `LockResolverCountWithWriteConflict` and make this one a new metric. What do you think?

ImSjt:
@ekexium It makes sense. I think we can add a metrics to count the error results of the **doActionOnMutations** function, so that we can clearly know the final result of the different actions in the commit phase. What do you think?
`
TikvCommitterErrorMetrics = prometheus.NewCounterVec(
    prometheus.CounterOpts{
      Namespace: namespace,
      Subsystem: subsystem,
      Name:      "committer_err_total",
      Help:      "Counter of committer errors.",
    }, []string{LblType, LblResult})

LblType: prewrite, commit, cleanup...
LblResult: error.Error()
`

ekexium:
Have you seen real needs for such metrics? Since metrics are not free, I suggest we either add just one metric for write conflicts or remove the existing one, unless we can justify their value.

ImSjt:
@ekexium PTAL again.

metrics/metrics.go
```diff
@@ -117,6 +117,7 @@ var (
 	TiKVLowResolutionTSOUpdateIntervalSecondsGauge prometheus.Gauge
 	TiKVStaleRegionFromPDCounter                   prometheus.Counter
 	TiKVPipelinedFlushThrottleSecondsHistogram     prometheus.Histogram
+	TiKvTxnWriteConflictCounter                    prometheus.Counter
```
ekexium, line 120:
```suggestion
	TiKVTxnWriteConflictCounter                    prometheus.Counter
```

txnkv/transaction/prewrite.go
```diff
@@ -467,6 +467,9 @@ func (handler *prewrite1BatchReqHandler) extractKeyErrs(keyErrs []*kvrpcpb.KeyEr
 		// Extract lock from key error
 		lock, err1 := txnlock.ExtractLockFromKeyErr(keyErr)
 		if err1 != nil {
+			if tikverr.IsErrWriteConflict(errors.Cause(err1)) {
```
ekexium, line 470: What's the purpose of `errors.Cause` here? Why not just `tikverr.IsErrWriteConflict(err1)`?

ImSjt: Since err1 might be 'withStack{ErrWriteConflict}', we need to use errors.Cause(err1) to get 'ErrWriteConflict'.

metrics/metrics.go
ImSjt: done

txnkv/transaction/prewrite.go
ekexium: I suppose `IsErrWriteConflict` handles that case as well and is a more common approach. What do you think?

ImSjt: Yes, I tried it and it works. PTAL again, thks~

zyguan (commented):
Can we just call `metrics.TiKVTxnWriteConflictCounter.Inc()` in `NewErrWriteConflictWithArgs`?

ImSjt:
> Can we just call `metrics.TiKVTxnWriteConflictCounter.Inc()` in `NewErrWriteConflictWithArgs`?

I think this might be a bit coupled, and in some scenarios, ErrWriteConflict is generated without going through NewErrWriteConflictWithArgs.

zyguan:
> > Can we just call `metrics.TiKVTxnWriteConflictCounter.Inc()` in `NewErrWriteConflictWithArgs`?
> 
> I think this might be a bit coupled, and in some scenarios, ErrWriteConflict is generated without going through NewErrWriteConflictWithArgs.

Besides `NewErrWriteConflictWithArgs`, `ExtractKeyErr` is another place where `ErrWriteConflict` gets generated. In this PR, only two calls of `ExtractLockFromKeyErr` (which calls `ExtractKeyErr` internally) are handled, and there are many other calls of `ExtractKeyErr`. May I know why do not increase the counter when `ErrWriteConflict` is generated from those calls?

zyguan:
> > > > Can we just call `metrics.TiKVTxnWriteConflictCounter.Inc()` in `NewErrWriteConflictWithArgs`?
> > > 
> > > 
> > > I think this might be a bit coupled, and in some scenarios, ErrWriteConflict is generated without going through NewErrWriteConflictWithArgs.
> > 
> > 
> > Besides `NewErrWriteConflictWithArgs`, `ExtractKeyErr` is another place where `ErrWriteConflict` gets generated. In this PR, only two calls of `ExtractLockFromKeyErr` (which calls `ExtractKeyErr` internally) are handled, and there are many other calls of `ExtractKeyErr`. May I know why do not increase the counter when `ErrWriteConflict` is generated from those calls?
> 
> I think write conflicts only occur in the prewrite1BatchReqHandler, actionPipelinedFlush, and actionPessimisticLock operations. I might have overlooked the actionPessimisticLock scenario. Perhaps we could add the NewErrWriteConflictWithConflict function and invoke metrics.TiKVTxnWriteConflictCounter.Inc() within it. All ErrWriteConflict would be exclusively generated through either NewErrWriteConflictWithArgs or NewErrWriteConflictWithConflict. What do you think?

I think so, it could be

```go
func NewErrWriteConflict(conflict *kvrpcpb.WriteConflict) *ErrWriteConflict {
	metrics.TiKVTxnWriteConflictCounter.Inc()
	return &ErrWriteConflict{WriteConflict: conflict}
}

func NewErrWriteConflictWithArgs(startTs, conflictTs, conflictCommitTs uint64, key []byte, reason kvrpcpb.WriteConflict_Reason) *ErrWriteConflict {
	return NewErrWriteConflict(&kvrpcpb.WriteConflict{
		StartTs:          startTs,
		ConflictTs:       conflictTs,
		Key:              key,
		ConflictCommitTs: conflictCommitTs,
		Reason:           reason,
	})
}
```

ImSjt:
@zyguan PTAL again, thks.

zyguan (approved):
LGTM

PR #1563, opened 2025-01-23
memdb: prevent iterator invalidation

ekexium:
ref pingcap/tidb#59153

To prevent potential misuse and iterator invalidation, modify the iterators provided by ART memdb as follows:

1. `Iter` and `IterReverse` now comes with an extra check: it is invalidated immediately by any write operation to the memdb after the creation of the iterator. Attempting to use such an invalidated iterator will result in a panic.
2. `SnapshotIter` and `SnapshotIterReverse` will be replaced by `BatchedSnapshotIter`.
2.1. `SnapshotIter` is different from `Iter` that it can be valid after write operations, but only becomes invalid if a write operation modifies the "snapshot".
2.2. We need to introduce `BatchedSnapshotIter` instead of directly modifying `SnapshotIter` because `SnapshotIter` maintains internal states and pointers. Consider a situation where a write operation causes changes to the internal data structure making the pointers invalid, while the snapshot should remain valid.
2.3. `SnapshotIter` and `SnapshotIterReverse` are not removed now for compatibility.

RBT is unchanged as it is no longer used.
Pipelined MemDB still doesn't support iterators as it was.

## Performance
Iterator microbenchmark
```
go test -run=^$ -bench=BenchmarkSnapshotIter -benchtime=3s
goos: linux
goarch: amd64
pkg: github.com/tikv/client-go/v2/internal/unionstore
cpu: AMD Ryzen 9 9900X 12-Core Processor            
BenchmarkSnapshotIter/RBT-SnapshotIter-24                     1783           1826587 ns/op             144 B/op               2 allocs/op
BenchmarkSnapshotIter/ART-SnapshotIter-24                     1870           1718771 ns/op             496 B/op              11 allocs/op
BenchmarkSnapshotIter/ART-BatchedSnapshotIter-24              1120           2964170 ns/op          417461 B/op             370 allocs/op
BenchmarkSnapshotIter/ART-ForEachInSnapshot-24                1771           1832084 ns/op             496 B/op              11 allocs/op
PASS
ok          github.com/tikv/client-go/v2/internal/unionstore        14.598s
```

TiDB union scan executor
BatchedSnapshotIter
```
go test -run=^$ -bench=BenchmarkUnionScanRead -benchtime=10s
    3999           3063035 ns/op          793197 B/op           17578 allocs/op
```
SnapshotIter
```
go test -run=^$ -bench=BenchmarkUnionScanRead -benchtime=10s
    3862           2990516 ns/op          386942 B/op           17434 allocs/op
```


internal/unionstore/memdb_art.go
```diff
@@ -165,3 +196,153 @@ func (db *artDBWithContext) SnapshotIterReverse(upper, lower []byte) Iterator {
...
+		}
+		snapshotIter = it.db.SnapshotIterReverse(searchUpper, it.lower)
+	} else {
+		searchLower := it.lower
+		if it.nextKey != nil {
+			searchLower = it.nextKey
+		}
+		snapshotIter = it.db.SnapshotIter(searchLower, it.upper)
+	}
+	defer snapshotIter.Close()
```
cfzjywxk, line 263: Perhaps we need to check if the newly created `snapshotIter` contains the same snapshot compared with the previous one, so the whole `snapshotBatchedIter` has a consistent snapshot semantics.

internal/unionstore/union_store.go
```diff
@@ -193,15 +198,45 @@ type MemBuffer interface {
...
+	// otherwise it uses the current checkpoint as snapshot.
+	//
+	// NOTE: returned kv-pairs are only valid during the iteration. If you want to use them after the iteration,
+	// you need to make a copy.
+	//
+	// The method is protected by a RWLock to prevent potential iterator invalidation, i.e.
+	// You cannot modify the MemBuffer during the iteration.
+	//
+	// Use it when you need to scan the whole range, otherwise consider using BatchedSnapshotIter.
+	ForEachInSnapshotRange(lower []byte, upper []byte, f func(k, v []byte) (stop bool, err error), reverse bool) error
```
you06, line 227: I see the `KvPair` struct, is it planned to be used here like `f func(pair KvPair)`?

internal/unionstore/art/art.go
```diff
@@ -52,6 +52,13 @@ type ART struct {
 	lastTraversedNode atomic.Uint64
 	hitCount          atomic.Uint64
 	missCount         atomic.Uint64
+
+	// The counter of every write operation, used to invalidate iterators that were created before the write operation.
+	SeqNo int
```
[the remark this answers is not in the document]
ekexium: I renamed it to `WriteSeqNo`, to be more consistent with `SnapshotSeqNo` since they have similar purposes and usages.

internal/unionstore/union_store.go
ekexium: I removed `KvPair` for performance reasons.

internal/unionstore/memdb_rbt.go
```diff
@@ -175,3 +201,12 @@ func (db *rbtDBWithContext) SnapshotIterReverse(upper, lower []byte) Iterator {
 func (db *rbtDBWithContext) SnapshotGetter() Getter {
 	return db.RBT.SnapshotGetter()
 }
+
+func (db *rbtDBWithContext) BatchedSnapshotIter(lower, upper []byte, reverse bool) Iterator {
+	// TODO: implement this
```
cfzjywxk, line 206: Comments should be removed?

internal/unionstore/memdb_test.go
```diff
@@ -1364,3 +1364,230 @@ func TestSnapshotReaderWithWrite(t *testing.T) {
...
+	check(newArtDBWithContext(), 3)
+	check(newArtDBWithContext(), 17)
+	check(newArtDBWithContext(), 64)
+
+	checkReverse(newArtDBWithContext(), 3)
+	checkReverse(newArtDBWithContext(), 17)
+	checkReverse(newArtDBWithContext(), 64)
+}
+
+func TestBatchedSnapshotIterEdgeCase(t *testing.T) {
```
cfzjywxk, line 1441: Do we have case to cover the memdb is changed bettween two `fillbatch` operations and errors shoul be reported?

internal/unionstore/memdb_art.go
```diff
@@ -165,3 +196,173 @@ func (db *artDBWithContext) SnapshotIterReverse(upper, lower []byte) Iterator {
...
+	snapshot MemDBCheckpoint
+}
+
+func (db *artDBWithContext) BatchedSnapshotIter(lower, upper []byte, reverse bool) Iterator {
+	if len(db.Stages()) == 0 {
+		logutil.BgLogger().Error("should not use BatchedSnapshotIter for a memdb without any staging buffer")
+	}
+	iter := &snapshotBatchedIter{
+		db:                    db,
+		snapshotTruncateSeqNo: db.SnapshotSeqNo,
```
cfzjywxk, line 225: Why not just using `snapshotSeqNo` but `snapshotTruncateSeqNo`?

internal/unionstore/memdb_art.go
```diff
@@ -165,3 +196,173 @@ func (db *artDBWithContext) SnapshotIterReverse(upper, lower []byte) Iterator {
...
+	if it.db.GetSnapshot() != it.snapshot {
+		return errors.Errorf(
+			"snapshot changed between batches, expected=%v, actual=%v",
+			it.snapshot,
+			it.db.GetSnapshot(),
+		)
+	}
+
+	it.db.RLock()
+	defer it.db.RUnlock()
```
cfzjywxk, line 255: There seems to be possible race condition bettwen the `GetSnapshot` check and `RLock`, it seems we need a CAS like semantics.

ekexium: I reconsidered this and it looks like we don't need to store and compare snapshots at all: `SnapshotSeqNo` is enough to guarantee the "snapshot" doesn't change because of its definition.

Though the race of `snapshotSeqNo` shall be considered another problem

internal/unionstore/art/art.go
```diff
@@ -52,6 +52,13 @@ type ART struct {
 	lastTraversedNode atomic.Uint64
 	hitCount          atomic.Uint64
 	missCount         atomic.Uint64
+
+	// The counter of every write operation, used to invalidate iterators that were created before the write operation.
+	WriteSeqNo int
```
cfzjywxk, line 57: The `WriteSeqNo` seems not to have the concurrency invariance like `SnapshotSeqNo` below, should atomic variable be used for it?

ekexium: The purpose of the sequence number is to tolerate interleaving access, but not concurrent access. Even if we change it to an atomic variable, concurrent read and write can still corrupt the iterator.
If there exists data race on this variable, there must be a bug on the caller side.
I've updated the comment to explain the rationale.

cfzjywxk: Trying to sort out the current issues:
1. memdb has concurrent usage scenarios, such as pessimistic lock setting key flags while membuf read is performing a memdb snapshot read.
2. The underlying implementation structure of memdb does not support concurrency and needs to be properly used by the upper layer with mutex protection. This is the issue being addressed by the PR at https://github.com/pingcap/tidb/pull/59219.
3. The situation where the memdb iterator is invalidated by interleaving writes, which is the purpose of introducing a mechanism like write sequence in this PR.

Implementation level:
1. `lastTraversedNode`, `hitCount`, and `missCount`, several internal states of ART, are implemented as atomic variables. These states are modified by write operations, and the WriteSeqNo below will also be modified by write operations, requiring the upper layer to handle concurrency correctly. 

This part of the code implementation is confusing, should we make the internal states of ART to be single-threaded, while allowing concurrent operations and protection to be handled at the upper layer of `MemDB`? So there is no need to consider data races or introduce atomic variables in the ART code.
/cc @you06

ekexium: One difference between `WriteSeqNo` and `lastTraversedNode` is that the latter one is updated during **read** operations. Concurrent read operations are allowed (by design?). So even if we acquire an Rlock of memdb for every read operation it still needs to be implemented by an atomic variable.

you06: If the traverse in snapshot-get do not cache the last-traversed node(maybe it's not worth cache for a snapshot-get), I think the `art.lastTraversedNode` can also be a non-atomic variable, but `hitCount` and `missCount` should be atomic.

https://github.com/tikv/client-go/blob/456c3af21c092c9b607c069906fc5cf19a69f061/internal/unionstore/art/art_snapshot.go#L79

internal/unionstore/art/art.go
```diff
@@ -498,7 +513,9 @@ func (t *ART) Release(h int) {
 	if h != len(t.stages) {
 		panic("cannot release staging buffer")
 	}
+	t.WriteSeqNo++
```
you06, line 516: Are we going to tolerance the interleaving usage between iterator and `Release`? It  won't affect the iterator I think.

internal/unionstore/art/art.go
```diff
@@ -115,6 +125,7 @@ func (t *ART) Set(key artKey, value []byte, ops ...kv.FlagsOp) error {
 		}
 	}
 
+	t.WriteSeqNo++
```
you06, line 128: Though there is no such usage in TiDB.
Will you consider add `SnapshotSeqNo` here if `len(t.stages) == 0`?

internal/unionstore/art/art_iterator.go
```diff
@@ -76,12 +80,41 @@ type Iterator struct {
 	currLeaf     *artLeaf
 	currAddr     arena.MemdbArenaAddr
 	endAddr      arena.MemdbArenaAddr
+
+	// only when seqNo == art.seqNo, the iterator is valid.
```
you06, line 84:
```suggestion
	// only when seqNo == art.WriteSeqNo, the iterator is valid.
```

internal/unionstore/art/art_iterator.go
```diff
@@ -19,6 +19,9 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/tikv/client-go/v2/internal/logutil"
+	"go.uber.org/zap"
```
you06, line 23: Remove blank line in imports

internal/unionstore/memdb_art.go
```diff
@@ -16,8 +16,13 @@ package unionstore
 
 import (
 	"context"
+	"fmt"
 	"sync"
 
+	"github.com/tikv/client-go/v2/internal/logutil"
+
+	"github.com/pingcap/errors"
+
```
you06, line 25: Remove blank lines.

you06 (commented):
rest LGTM

internal/unionstore/art/art.go
ekexium: Release stage[0] can change "the snapshot" so I suppose we cannot tolerate that. WDYT?

internal/unionstore/art/art.go
ekexium: It won't hurt. Let's do it

internal/unionstore/art/art.go
you06: Agree, even the non-snapshot iterator is not affected, prohibit this case looks good to me.

internal/unionstore/memdb_art.go
```diff
@@ -165,3 +196,161 @@ func (db *artDBWithContext) SnapshotIterReverse(upper, lower []byte) Iterator {
...
+func (it *snapshotBatchedIter) fillBatch() error {
+	if it.snapshotSeqNo != it.db.SnapshotSeqNo {
+		return errors.Errorf(
+			"invalid iter: snapshotSeqNo changed, iter's=%d, db's=%d",
+			it.snapshotSeqNo,
+			it.db.SnapshotSeqNo,
+		)
+	}
+
+	it.db.RLock()
```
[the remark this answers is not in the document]
you06: Oh I get you, `SnapshotSeqNo` is changed in very low frequency.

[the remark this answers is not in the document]
cfzjywxk: @ekexium 
Better to add some comments to explain the usage of `RLock` here, others may not be familiar with the upper `MemDB` and underlying `ART` implementation and usages.

internal/unionstore/memdb_test.go
ekexium: The "SnapshotChange" case covered the situation

cfzjywxk (approved):
LGTM

you06 (approved):

PR #1564, opened 2025-01-23
client: refactor batch client send loop

zyguan:
`batchConn.idleDetect` and `batchConn.closed` are checked in `fetchAllPendingRequests`, so, the conn should be closed or recycled when `a.reqBuilder.len() == 0` just after `fetchAllPendingRequests`, no need to `fetchMorePendingRequests`.

Also bump tidb used in integration tests.


cfzjywxk (approved):
LGTM

crazycs520 (approved):

PR #1566, opened 2025-02-03
retry stale read as stale if leader is not accessible

mittalrishabh:
issue number : https://github.com/tikv/tikv/issues/17422
It is the extension of the original PR https://github.com/tikv/client-go/pull/1509. I have observed that in case of hot regions, leaders are inaccessible and stale reads are retried as follower_follower. I have added a change if leader is not reachable or epoch doesn't mach than retry stale read as stale instead of follower.
github


mittalrishabh:
@cfzjywxk please review this PR as original PR was closed https://github.com/tikv/client-go/pull/1534

cfzjywxk (approved):

crazycs520 (approved):

PR #1572, opened 2025-02-11
(tidb-6.5) Do not set txn scope in snapshot

ekexium:
Ref https://github.com/pingcap/tidb/issues/59402

`TxnScope` should not be set to `ReadReplicaScope`.


cfzjywxk (approved):

crazycs520 (approved):

cfzjywxk:
@ekexium Need to bump up the client-go dependency.

PR #1587, opened 2025-02-19
Delete txn scope code

ekexium:
Part of pingcap/tidb#58838

Some changes that I'm not sure about:
1. Test `TestMinSafeTsFromMixed1` and `TestMinSafeTsFromMixed2`
2. `updateGlobalTxnScopeTSFromPD`

## Summary by CodeRabbit

* **Refactor**
  * Simplified transaction timestamp handling by removing scope-based differentiation. Timestamp acquisition and oracle operations now use a unified approach instead of per-scope management.
  * Streamlined commit protocols and timestamp retrieval to rely on default behavior rather than explicit scope configuration.

* **Tests**
  * Removed multi-datacenter-specific transaction test cases.
  * Updated integration tests to reflect simplified timestamp handling.


cfzjywxk:
@JmPotato PTAL

config/config.go
```diff
@@ -75,9 +75,10 @@ type Config struct {
 	OpenTracingEnable     bool
 	Path                  string
 	EnableForwarding      bool
-	TxnScope              string
-	EnableAsyncCommit     bool
-	Enable1PC             bool
+	// FIXME: rename
```
JmPotato, line 78: Can you provide more explanation about this comment?

oracle/oracle.go
```diff
@@ -41,8 +41,8 @@ import (
 )
 
 // Option represents available options for the oracle.Oracle.
+// TODO: remove this struct
 type Option struct {
-	TxnScope string
 }
```
JmPotato, line 46: Why not just remove it in this PR?

zyguan (commented):
Generally LGTM, any plan to deprecate (or rename) `oracle.GlobalTxnScope`?

oracle/oracle.go
ekexium: To minimize the change.. It has hundreds of usages. I prefer keeping this PR (relatively) small

oracle/oracles/pd.go
```diff
@@ -278,54 +276,38 @@ func (o *pdOracle) getMinTimestampInAllTSOGroup(ctx context.Context) (uint64, er
...
 		arrival: time.Now(),
 	}
-	lastTSInterface, ok := o.lastTSMap.Load(txnScope)
-	if !ok {
-		pointer := &atomic.Pointer[lastTSO]{}
-		pointer.Store(current)
-		// do not handle the stored case, because it only runs once.
-		lastTSInterface, _ = o.lastTSMap.LoadOrStore(txnScope, pointer)
+	if o.lastTS.Load() == nil {
+		o.lastTS.Store(current)
```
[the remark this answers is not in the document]
ekexium: Makes sense.

ekexium:
> any plan to deprecate (or rename) oracle.GlobalTxnScope?

Yes, I think it's the last step. After we deprecated txn-scope related code, we can rename those left to names like `ReadReplicaScope`

zyguan (approved):

cfzjywxk (approved):

you06:
You can ignore the leak goroutine to fix the test.

https://github.com/tikv/client-go/blob/469a7adf7ae88ce8ac69f49473f0b82324030f0b/integration_tests/main_test.go#L27-L31

```diff
	opts := []goleak.Option{
		goleak.IgnoreTopFunction("github.com/pingcap/goleveldb/leveldb.(*DB).mpoolDrain"),
		goleak.IgnoreTopFunction("github.com/tikv/client-go/v2/txnkv/transaction.keepAlive"), // TODO: fix ttlManager goroutine leak
		goleak.IgnoreTopFunction("github.com/tikv/client-go/v2/config/retry.(*Config).createBackoffFn.newBackoffFn.func2"),
+		goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start"),
	}
```

Issue #1586, opened 2025-02-19
Support async send requests


zyguan:
Currently, the KV client only provides synchronous API for sending requests. If a query needs to issue multiple requests concurrently, it inevitably has to create a new goroutine for each concurrent procedure. A typical scenario is batch-get, which groups keys into batches based on regions. When the number of batches exceeds one, a worker goroutine is created for each batch to send KV requests.

We aim to reduce the number of goroutines while still allowing requests to be issued concurrently as before. To achieve this, we need to introduce async API. In fact, the internal batch client processes requests and responses asynchronously through *send loop* and *recv loop*. It's not difficult to implement an async method for `RPCClient`, but rewriting upper layer in an async way require a significant amount of work. This issue tries to describe the brief work plan, it's also used to track related PRs.

Here is the related task breakdown

- Define and implement core interfaces (eg. executor, callback) that will be used in async API.
  - https://github.com/tikv/client-go/pull/1591
- Implement `SendRequestAsync` for `RPCClient` and its upper wrappers (eg. [collapse client](https://github.com/tikv/client-go/blob/master/internal/client/client_collapse.go), [interceptor client](https://github.com/tikv/client-go/blob/master/internal/client/client_interceptor.go)).
  - https://github.com/tikv/client-go/pull/1604
  - https://github.com/tikv/client-go/pull/1664
  - https://github.com/tikv/client-go/pull/1735
- Refactor `RegionRequestSender` to maximize code reuse when implementing the asynchronous interface.
  - https://github.com/tikv/client-go/pull/1565
- Implement `SendReqAsync` for `RegionRequestSender`.
  - https://github.com/tikv/client-go/pull/1618
- Refactor batch-get related methods in `KVSnapshot` and allow batch-get to execute asynchronously.
  - https://github.com/tikv/client-go/pull/1646

Bugfixes:
- https://github.com/tikv/client-go/pull/1747
- https://github.com/tikv/client-go/pull/1683
- https://github.com/tikv/client-go/pull/1711

PR #1591, opened 2025-02-25
util: define and implement core interfaces for async api

zyguan:
ref #1586

### DIFF
--- a/util/async/core.go
+++ b/util/async/core.go
@@ -1,0 +1,83 @@
+// Copyright 2025 TiKV Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package async
+
+import (
+	"sync"
+)
+
+// Pool is an interface for goroutine pool.
+type Pool interface {
+	// Go submits a function to the goroutine pool.
+	Go(f func())
+}
+
+// Executor is an interface that can append functions to be executed asynchronously.
+type Executor interface {
+	Pool
+	// Append adds functions to the executor. It should be safe to call Append concurrently.
+	Append(fs ...func())
+}
+
+// Callback defines a callback function that can be invoked immediately or later.
+type Callback[T any] interface {
+	// Executor returns the executor that the callback uses.
+	Executor() Executor
+	// Inject adds a deferred action that will be invoked before the callback.
+	Inject(g func(T, error) (T, error))
+	// Invoke invokes the callback immediately in current goroutine.
+	Invoke(val T, err error)
+	// Schedule schedules the callback to be invoked later, it's typically called in other goroutines.
+	Schedule(val T, err error)
+}
+
+// NewCallback creates a new callback function.
+func NewCallback[T any](e Executor, f func(T, error)) Callback[T] {
+	return &callback[T]{e: e, f: f}
+}
+
+type callback[T any] struct {
+	once sync.Once
+	e    Executor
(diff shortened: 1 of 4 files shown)


util/async/runloop.go
```diff
@@ -0,0 +1,158 @@
...
+// the other goroutine to wake it up later, or the context will be canceled finally to break the waiting. Exec should
+// only be called by one goroutine.
+func (l *RunLoop) Exec(ctx context.Context) (int, error) {
+	for {
+		l.lock.Lock()
+		if l.state != StateIdle {
+			l.lock.Unlock()
+			return 0, errors.New("runloop: already executing")
+		}
+		// assert l.state == stateIdle
```
cfzjywxk, line 107: Could be removed.

util/async/runloop.go
```diff
@@ -0,0 +1,158 @@
...
+				l.lock.Lock()
+				// move remaining running tasks to runnable
+				l.running = append(l.running[:0], l.running[i:]...)
+				l.running = append(l.running, l.runnable...)
+				l.running, l.runnable = l.runnable, l.running
+				l.state = StateIdle // running -> idle
+				l.lock.Unlock()
+				return count, ctx.Err()
+			default:
+				f()
```
cfzjywxk, line 145: How about passing the `ctx` to `f()` so it could be terminated? Timeout cancel may be needed.

util/async/runloop_test.go
```diff
@@ -0,0 +1,161 @@
...
+func TestGo(t *testing.T) {
+	l := NewRunLoop()
+	n := uint32(0)
+
+	// Go works when pool is nil (by default)
+	require.Nil(t, l.Pool)
+	l.Go(func() { atomic.StoreUint32(&n, 1) })
+	require.Eventually(t, func() bool { return atomic.LoadUint32(&n) == 1 }, time.Second, time.Millisecond)
+
+	// use a customized pool
```
cfzjywxk, line 36:
```suggestion
	// Use a customized pool
```

util/async/core.go
```diff
@@ -0,0 +1,83 @@
...
+// NewCallback creates a new callback function.
+func NewCallback[T any](e Executor, f func(T, error)) Callback[T] {
+	return &callback[T]{e: e, f: f}
+}
+
+type callback[T any] struct {
+	once sync.Once
+	e    Executor
+	f    func(T, error)
+	gs   []func(T, error) (T, error)
```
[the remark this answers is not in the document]
zyguan: `f` and `g` are common function names (in mathematics), and `gs` stands for a list of `g` (a common naming convention in FP).

util/async/runloop.go
zyguan: I leave this comment on purpose. It's not a runnable code in golang, just reminds us that the state should be idle in the following code.

util/async/runloop.go
zyguan: I prefer keeping it simple, if `f` here accepts a `ctx` as an arg, then `Callback` functions (`cb.f`, `cb.gs` and `call` related methods) should accept a `ctx` as well.

In fact, `ctx` is typically referenced in the `cb.f`. eg.
```go
func sendAll(ctx, reqs) {
  completed := 0
  e := NewExecutor(...)
  for req in reqs {
    cb := NewCallback(e, func(resp, err) {
      handleResponse(ctx, resp, err)
      completed++
    }
    cli.asyncSend(ctx, req, cb)
  }
  for completed < len(reqs) {
    e.Exec()
  }
}
```

cfzjywxk (approved):

ekexium (commented):
Generally LGTM. Just a few questions
1. What is the expected scope of a RunLoop? Is it one per TiDB instance?
2. Do we need to consider the behavior when shutting down? Do we need to drop or finish the pending tasks?

zyguan:
> 1. What is the expected scope of a RunLoop? Is it one per TiDB instance?

The scope shall be small, typically one per query/txn-cmd, [here] is an example.

> 2. Do we need to consider the behavior when shutting down? Do we need to drop or finish the pending tasks?

Yes, but it depends on the caller, RunLoop itself cannot perceive the business logic. Eg. when we send multiple requests, we can either collect all errors, or just return the first error (drop the rest).

ekexium (approved):

PR #1594, opened 2025-02-26
(tidb-7.5) Validate ts only for stale read

ekexium:
ref pingcap/tidb#59402
Only validate read ts for stale read and those who set `tidb_snapshot`


oracle/oracles/pd_test.go
```diff
@@ -237,56 +237,6 @@ func TestAdaptiveUpdateTSInterval(t *testing.T) {
 	assert.Equal(t, adaptiveUpdateTSIntervalStateNormal, o.adaptiveUpdateIntervalState.state)
 }
 
-func TestValidateReadTS(t *testing.T) {
```
MyonKeminta, line 240: Just noticed that this test seems can be added back

cfzjywxk (approved):

MyonKeminta (approved):

PR #1600, opened 2025-02-27
Add a retry when getting ts from PD for validating read ts

MyonKeminta:
This PR adds a retry to `ValidateReadTS` to avoid a false-positive when the checked ts comes from a different source other than the current pdOracle. This can happen in some non-TiDB caller (e.g. BR, lightning).

The `ValidateReadTS` function may use `getCurrentTSForValidation` to get a ts, which internally uses a singleflight. So it's possible that a later call to `ValidateReadTS` reuses an older on-the-way single flight and returns an older ts. If the ts to be checked comes from outside the current `pdOralce` (so that it's not cached by the low resolution ts), it may cause false positive.

The solution is to add a  retry: when `ValidateReadTS` check failed with a ts returned by `getCurrentTSForValidation`, try it one more time. So when the second call to `getCurrentTSForValidation` must uses a later-started singleflight, inwhich the `GetTS` RPC to PD is strictly later than the call. If the given ts is valid, it can be guaranteed by PD's TSO's monotonicity that the second ts returned by `getCurrentTSForValidation` must be greater than the given ts.

Ref: https://github.com/pingcap/tidb/issues/59758 , this PR can avoid this problem in master branch where the validation to read ts takes effect to all read.

### DIFF
--- a/oracle/oracles/pd.go
+++ b/oracle/oracles/pd.go
@@ -47,6 +47,7 @@ import (
 	"github.com/tikv/client-go/v2/internal/logutil"
 	"github.com/tikv/client-go/v2/metrics"
 	"github.com/tikv/client-go/v2/oracle"
+	"github.com/tikv/client-go/v2/util"
 	pd "github.com/tikv/pd/client"
 	"github.com/tikv/pd/client/clients/tso"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


cfzjywxk (approved):
LGTM for working around the BR/Lightning compatible issue

ekexium (approved):

PR #1612, opened 2025-03-13
support redact key in logs

tangenta:
Ref https://github.com/pingcap/tidb/issues/59279

This PR checks `RedactLogEnabled` in `"github.com/pingcap/errors"` to determine whether to redact the keys in logs.


ekexium (commented):
Can we also search and replace usages of `hex.EncodeToString`?

ekexium (approved):

tangenta:
Manual test:

```
2025/03/14 17:00:59.062 +08:00] [INFO] [split_region.go:85] ["split batch regions request"] ["split key count"=1] ["batch count"=1] ["first batch, region ID"=24] ["first split key"=?]
[2025/03/14 17:00:59.073 +08:00] [INFO] [split_region.go:187] ["batch split regions complete"] ["batch region ID"=24] ["first at"=?] ["first new region left"="{Id:134 StartKey:? EndKey:? RegionEpoch:{ConfVer:1 Version:67} Peers:[id:135 store_id:1 ] EncryptionMeta:<nil> IsInFlashback:false FlashbackStartTs:0}"] ["new region count"=1]
```

util/redact/redact.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// RedactLogEnabled defines whether the arguments of Error need to be redacted.
+var RedactLogEnabled atomic.String
+
+const (
+	RedactLogEnable  string = "ON"
+	RedactLogDisable string = "OFF"
+	RedactLogMarker  string = "MARKER"
+)
```
D3Hunter, line 17: this repo also include `pingcap/error` can we reuse them?

util/redact/redact.go
```diff
@@ -0,0 +1,76 @@
...
+// KeyBytes receives a key return omitted information if redact log enabled
+func KeyBytes(key []byte) []byte {
+	if NeedRedact() {
+		return []byte{'?'}
+	}
+	return ToUpperASCIIInplace(EncodeToString(key))
+}
+
+// String converts slice of bytes to string without copy.
+func String(b []byte) (s string) {
```
D3Hunter, line 42: maybe move them to `pingcap/errors`

util/redact/redact.go
tangenta: I think the global variable cannot be reused if the `pingcap/error` is different. Let me reuse the constants.

D3Hunter: do you mean there be 2 different global variables?

go will choose only one larger version of pingcap/errors` when compile, so it should be only one

tangenta: For example, let's define "→" as "depends on":

```
A → B
B → C (v1, with variable c1)
A → C (v2, removed c1, add variable c2)
```

Do you mean only v2 is used? How about the access to `c1`?

D3Hunter: see https://go.dev/ref/mod#minimal-version-selection

D3Hunter: depends on the usage, it might report error on compile in your case

util/redact/redact.go
tangenta: https://github.com/pingcap/errors/pull/75

util/redact/redact.go
tangenta: https://github.com/pingcap/errors/pull/75

you06 (commented):
There can be some error messages contain raw key like. `keyErr.String()` will cast the `KeyError` into string, which may contais key info.
 
https://github.com/tikv/client-go/blob/9e7da793ff00acb881a70cb8fd9e23520839607d/error/error.go#L336
https://github.com/tikv/client-go/blob/9e7da793ff00acb881a70cb8fd9e23520839607d/txnkv/txnsnapshot/snapshot.go#L813

A missing redact

https://github.com/tikv/client-go/blob/9e7da793ff00acb881a70cb8fd9e23520839607d/txnkv/transaction/2pc.go#L489

you06 (approved):
LGTM, need to update `integration_tests/go.mod`.

PR #1618, opened 2025-03-21
locate: implement SendReqAsync for RegionRequestSender

zyguan:
ref #1586


cfzjywxk:
@lcwangchao @tiancaiamao 
PTAL if you have time

internal/locate/region_request.go
```diff
@@ -1063,6 +1204,135 @@ func (s *sendReqState) send(bo *retry.Backoffer, req *tikvrpc.Request, timeout t
...
+}
+
+// toResponesExt converts the state to a ResponseExt .
+func (s *sendReqState) toResponesExt() (*tikvrpc.ResponseExt, error) {
+	if s.vars.err != nil {
+		return nil, s.vars.err
+	}
+	if s.vars.resp == nil {
+		return nil, errors.New("invalid state: response is nil")
+	}
```
cfzjywxk, line 1328: Should we also assert `s.vars.regionErr == nil` here?

zyguan: No, `toResponseExt` might be called with `.vars.regionErr` when the region error cannot be resolved by retry.

cfzjywxk (approved):
LGTM

you06 (approved):

PR #1619, opened 2025-03-24
enable ts validation for normal read

ekexium:
Enable the check on the master branch


cfzjywxk (approved):

zyguan (commented):
Should we introduce an "escape route" in case users encounter related issues after upgrading to the new version?

cfzjywxk:
> Should we introduce an "escape route" in case users encounter related issues after upgrading to the new version?

Currently, it is enabled experimentally on the master branch, and it may be reverted if any issues arise. 
Based on previous lessons, it is crucial to avoid false positive in production environments, as even with workarounds, they have already caused service unavailability.

The challendge is it's difficult to verify no false positive would happen.

zyguan:
> > Should we introduce an "escape route" in case users encounter related issues after upgrading to the new version?
> 
> Currently, it is enabled experimentally on the master branch, and it may be reverted if any issues arise. Based on previous lessons, it is crucial to avoid false positive in production environments, as even with workarounds, they have already caused service unavailability.
> 
> The challendge is it's difficult to verify no false positive would happen.

I known it's difficult to verify no false positive would happen. What if there is an unfound false positive which causes service unavailability in the customer‘s production environment?
- if there is no "escape route", we have to urgently release a version with the bug fix.
- if there is a switch to disable the validation, we only need to inform the customer to disable this feature.

IMO, an even better approach is to keep this feature disabled by default for upgraded clusters.

cfzjywxk:
Makes sense,  it is still needed to introduce the relevant configuration option or system variable, and as @zyguan  mentioned, it should only take effect for newly created clusters. Perhaps this is the safest way to enable the check.

oracle/oracles/pd.go
```diff
@@ -672,12 +672,6 @@ func (o *pdOracle) ValidateReadTS(ctx context.Context, readTS uint64, isStaleRea
 		return errors.Errorf("MaxInt64 <= readTS < MaxUint64, readTS=%v", readTS)
 	}
 
-	// only check stale reads and reads using `tidb_snapshot`
-	forTidbSnapshot := ctx.Value(ValidateReadTSForTidbSnapshot{}) != nil
```
you06, line 676: Shall we also remove the definition of `ValidateReadTSForTidbSnapshot`?

ekexium:
If we add a switch to this, what should be the behvaior when the switch is off? Skip all checks or still check stale reads?

oracle/oracles/pd.go
ekexium: I prefer not to remove it for now. We may need to check stale reads, e.g. after we add a switch to control checks for leader read.

you06:
> If we add a switch to this, what should be the behvaior when the switch is off? Skip all checks or still check stale reads?

I prefer skipping checks than logging the violations. Because the logs may be ignored, and too many logs can be another serious issue.

cfzjywxk:
> > If we add a switch to this, what should be the behvaior when the switch is off? Skip all checks or still check stale reads?
> 
> I prefer skipping checks than logging the violations. Because the logs may be ignored, and too many logs can be another serious issue.

The `stale read` check is already enabled in the LTS version with error logging, the behaviour of the new switch needs some considerations. Disabling for all is a more common approach.

zyguan (approved):

cfzjywxk (approved):

PR #1620, opened 2025-03-27
Refine pessimistic lock related metrics and stats

yibin87:
This PR mainly forcus to replace some existing perssimistic stats with lockCtx.Stats struct, and move the TiKVPessimisticLockKeysDuration metric to tidb side, since the final merged lockCtx.Stats will be ready in tidb side.
Using lockCtx.Stats to replace existing LockKeys_Duration stats have several advantages:

1. Won't miss any lock time when the lock is not successfully accquired at last.
2. Don't rely on the Waited flag to record lock time
3. No need to keep similar stats in lockCtx and lockCtx.Stats

### DIFF
--- a/kv/kv.go
+++ b/kv/kv.go
@@ -61,9 +61,6 @@ type LockCtx struct {
 	ForUpdateTS             uint64
 	lockWaitTime            *lockWaitTimeInMs
 	WaitStartTime           time.Time
-	PessimisticLockWaited   *int32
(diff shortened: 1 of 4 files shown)


zyguan (approved):

cfzjywxk (approved):

cfzjywxk:
@yibin87 We could continue with the TiDB PR

PR #1623, opened 2025-04-01
memdb: introduce snapshot interface

you06:
A common use case for memdb involves a write thread continuously writing to the memdb while multiple read threads simultaneously access it. Without mutex protection, this can cause data races. Since it's difficult to guarantee that every risky read operation calls `RLock` beforehand, we introduce the `Snapshot` interface.

A snapshot allows reading unstaged snapshot data within a staging. All reads through it are automatically mutex-protected, preventing developer from mistakes.

This interface simplifies the usage pattern: call `GetSnapshot` when creating a staging, and all subsequent reads can be performed using the retrieved snapshot. Before that, we commonly check [`IsDirty()`](https://github.com/tikv/client-go/blob/e3b8af59f428f01216cd8a60e1411632935ce4fa/internal/unionstore/union_store.go#L240-L241) when diciding whether reading from snapshot which can cause potential data race.


internal/unionstore/art/art_snapshot.go
```diff
@@ -31,8 +31,8 @@ func (t *ART) getSnapshot() arena.MemDBCheckpoint {
 }
 
 // SnapshotGetter returns a Getter for a snapshot of MemBuffer.
-func (t *ART) SnapshotGetter() *SnapGetter {
-	return &SnapGetter{
+func (t *ART) SnapshotGetter() *Snapshot {
```
cfzjywxk, line 34: Is it possible to keep only `Snapshot` function? As the upper callers should use the new `GetSnapshot` interface.

internal/unionstore/union_store.go
```diff
@@ -270,6 +249,8 @@ type MemBuffer interface {
 	FlushWait() error
 	// GetMetrics returns the metrics related to flushing
 	GetMetrics() Metrics
+	// GetSnapshot returns a snapshot of the MemBuffer.
+	GetSnapshot() MemBufferSnapshot
```
cfzjywxk, line 253: Better to comment this is the snapshot suppose to be used and it is thread safe. 

The commonents of `SnapshotGetter()` should also be updated to clarity the differences bettween `SnapshotGetter` and the new `GetSnapshot`, when to use which.

internal/unionstore/art/art_snapshot.go
cfzjywxk: Besides there are quite some `Snapshot` like definitions in the code.

you06: I remove the `SnapGetter` because it's a subset of `Snapshot`. Also make the `MemBuffer.SnapshotGetter` deprecated.

The current code doesn't have `SnapGetter` but there are still `uSnapshot` and `MemBufferSnapshot`, and some other `Snapshot` named structs for implementation. To make it more clear, I move the `SnapshotWithMutex` into an individual file.

cfzjywxk (approved):
LGTM

ekexium:
Is there a corresponding TiDB PR? Shall we review them together?

you06:
> Is there a corresponding TiDB PR? Shall we review them together?

Yes, see pingcap/tidb#59219

internal/unionstore/memdb_art.go
```diff
@@ -154,203 +153,49 @@ func (db *artDBWithContext) IterReverse(upper, lower []byte) (Iterator, error) {
...
-		}
-	}
-	return nil
+// SnapshotGetter implements the Getter interface, by wrapping GetSnapshot.
+func (db *artDBWithContext) SnapshotGetter() Getter {
+	return db.ART.GetSnapshot()
 }
 
 // SnapshotIter returns an Iterator for a snapshot of MemBuffer.
 func (db *artDBWithContext) SnapshotIter(lower, upper []byte) Iterator {
```
ekexium, line 162: Does this method still need to be public? 
Will we replace original calls to `SnapshotIter` with `GetSnapshot().ForEachInSnapshotRange()` and `GetSnapshot().BatchedSnapshotIter()`?

you06: This interface is no more needed for tidb, but leave it for keeping the interface compatible.

ekexium: Maybe add `Deprecated:` to mark them and avoid new usages?

internal/unionstore/union_store.go
```diff
@@ -270,6 +250,14 @@ type MemBuffer interface {
...
 	// GetMetrics returns the metrics related to flushing
 	GetMetrics() Metrics
+
+	// GetSnapshot returns a snapshot of the MemBuffer.
+	// The snapshot returned by this function is protected by an RWLock to ensure thread safety.
+	// And this snapshot can fully replace SnapshotGetter, SnapshotIter, and SnapshotIterReverse.
+	// Additionally, it provides two iteration methods: ForEachInSnapshotRange and BatchedSnapshotIter,
+	// which tolerate interleaving reads and writes for using them simply.
+	// The snapshot also verifies the snapshot sequence number to prevent reading from an invalid snapshot.
+	GetSnapshot() MemBufferSnapshot
```
ekexium, line 260: I suggest we define the `snapshot` here.

ekexium (commented):
The rest LGTM

internal/unionstore/union_store.go
you06: What do you mean by defining `snapshot` here?
The `MemBufferSnapshot` is defined in line 281 just after the `MemBuffer` interface.

cfzjywxk:
This has been lasted for some time, PTAL @you06 @ekexium

internal/unionstore/union_store.go
ekexium: I mean define the concept in comments to explain what snapshot is

you06: I added some explanations about snapshot, PTAL again.

ekexium (approved):
The rest LGTM

PR #1629, opened 2025-04-11
txn: fix resolver cache usage for async commit

cfzjywxk:
Ref: https://github.com/pingcap/tidb/issues/59494

1. Do not call `checkSecondaries` if the transaction status is determined for async commit transactions. 
2. Small refactor on the `TxnStatus` usages.
3. Do not use `ttl = 0` setting in test cases.

### DIFF
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -173,9 +173,9 @@ func (s TxnStatus) StatusCacheable() bool {
 	return s.IsStatusDetermined()
 }
 
-// StatusEqual checks whether the current status is equal to another status if the
+// HasSameDeterminedStatus checks whether the current status is equal to another status if the
 // transaction status is determined.
-func (s TxnStatus) StatusEqual(other TxnStatus) bool {
+func (s TxnStatus) HasSameDeterminedStatus(other TxnStatus) bool {
 	return (s.IsCommitted() && other.IsCommitted() && s.CommitTS() == other.CommitTS()) ||
 		(s.IsRolledBack() && other.IsRolledBack())
 }
@@ -234,7 +234,7 @@ func (lr *LockResolver) saveResolved(txnID uint64, status TxnStatus) {
 
 	if savedStatus, ok := lr.mu.resolved[txnID]; ok {
 		// The saved determined status should always equal to the new one.
-		if !(savedStatus.StatusEqual(status)) {
+		if !(savedStatus.HasSameDeterminedStatus(status)) {
 			logutil.BgLogger().Error("unexpected txn status saving to the cache, the existing status is not equal to the new one",
 				zap.Uint64("txnID", txnID),
 				zap.String("existing status", savedStatus.String()),
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -170,10 +170,7 @@ func (s TxnStatus) Action() kvrpcpb.Action { return s.action }
 //	 `action` equals `Action::NoAction` or `Action::LockNotExistRollback` or `Action::TtlExpireRollback`.
 //	 Refer to the tikv `CheckTxnStatus` handling logic for more information.
 func (s TxnStatus) StatusCacheable() bool {
-	if s.IsCommitted() || s.IsRolledBack() {
-		return true
-	}
-	return false
+	return s.IsStatusDetermined()
 }
 
 func (s TxnStatus) String() string {
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -166,7 +166,7 @@ func (s TxnStatus) Action() kvrpcpb.Action { return s.action }
 //		    resp.set_commit_version(commit_ts.into_inner())
 //		}
 //
-//	 So the transaction is regarded as committed if the commit_ts is not 0, and rollback if the
+//	 So the transaction is regarded as committed if the commit_ts is not 0, and rolled back if the
 //	 `action` equals `Action::NoAction` or `Action::LockNotExistRollback` or `Action::TtlExpireRollback`.
 //	 Refer to the tikv `CheckTxnStatus` handling logic for more information.
 func (s TxnStatus) StatusCacheable() bool {
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -225,7 +225,14 @@ func (lr *LockResolver) saveResolved(txnID uint64, status TxnStatus) {
 	lr.mu.Lock()
 	defer lr.mu.Unlock()
 
-	if _, ok := lr.mu.resolved[txnID]; ok {
+	if savedStatus, ok := lr.mu.resolved[txnID]; ok {
+		// The saved determined status should always equal to the new one.
+		if !(savedStatus.TTL() == status.TTL() && savedStatus.CommitTS() == status.CommitTS() && savedStatus.Action() == status.Action()) {
+			logutil.BgLogger().Error("unexpected txn status saving to the cache, the existing status is not equal to the new one",
+				zap.String("existing status", savedStatus.String()),
+				zap.String("new status", status.String()))
+			panic("unexpected txn status saved to cache with existing different entry")
+		}
 		return
 	}
 	lr.mu.resolved[txnID] = status
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -220,6 +220,11 @@ func NewLock(l *kvrpcpb.LockInfo) *Lock {
 }
 
 func (lr *LockResolver) saveResolved(txnID uint64, status TxnStatus) {
+	if !status.IsStatusDetermined() {
+		logutil.BgLogger().Error("unexpected undetermined status saved to cache",
+			zap.Uint64("txnID", txnID), zap.Stringer("status", status), zap.Stack("stack"))
+		panic("unexpected undetermined status saved to cache")
+	}
 	lr.mu.Lock()
 	defer lr.mu.Unlock()
 
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -229,6 +229,7 @@ func (lr *LockResolver) saveResolved(txnID uint64, status TxnStatus) {
 		// The saved determined status should always equal to the new one.
 		if !(savedStatus.TTL() == status.TTL() && savedStatus.CommitTS() == status.CommitTS() && savedStatus.Action() == status.Action()) {
 			logutil.BgLogger().Error("unexpected txn status saving to the cache, the existing status is not equal to the new one",
+				zap.Uint64("txnID", txnID),
 				zap.String("existing status", savedStatus.String()),
 				zap.String("new status", status.String()))
 			panic("unexpected txn status saved to cache with existing different entry")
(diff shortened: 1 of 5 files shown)


txnkv/txnlock/lock_resolver.go
```diff
@@ -815,11 +833,7 @@ func (lr *LockResolver) getTxnStatus(bo *retry.Backoffer, txnID uint64, primary
 		status.action = cmdResp.Action
 		status.primaryLock = cmdResp.LockInfo
 
-		if status.primaryLock != nil && status.primaryLock.UseAsyncCommit && !forceSyncCommit {
-			if !lr.store.GetOracle().IsExpired(txnID, cmdResp.LockTtl, &oracle.Option{TxnScope: oracle.GlobalTxnScope}) {
-				status.ttl = cmdResp.LockTtl
-			}
-		} else if cmdResp.LockTtl != 0 {
+		if cmdResp.LockTtl != 0 {
```
MyonKeminta, line 836: Can't remenber the purpose of this code. Could you explain this change?

MyonKeminta (commented):
Understanding this part of code has become much more difficult than I would imagine...

txnkv/txnlock/lock_resolver.go
```diff
@@ -113,17 +113,27 @@ func (lr *LockResolver) Close() {
...
-func (s TxnStatus) IsCommitted() bool { return s.ttl == 0 && s.commitTS > 0 }
+func (s TxnStatus) IsCommitted() bool { return s.commitTS > 0 }
 
 // IsRolledBack returns true if the txn's final status is rolled back.
-func (s TxnStatus) IsRolledBack() bool { return s.ttl == 0 && s.commitTS == 0 }
+func (s TxnStatus) IsRolledBack() bool {
+	return s.ttl == 0 && s.commitTS == 0 && (s.action == kvrpcpb.Action_NoAction ||
+		s.action == kvrpcpb.Action_LockNotExistRollback ||
+		s.action == kvrpcpb.Action_TTLExpireRollback)
+}
```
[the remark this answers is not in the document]
cfzjywxk: The `TxnStatus` is an abstraction that introduces some complexity. In the current implementation, `TxnStatus` is determined by the KV response from `CheckTxnStatus`. Whether a transaction is rolled back depends on the TTL and the specified action, while whether it's committed depends on whether the commit timestamp is greater than 0.

To make it clear two seperate functions `IsCommitted` and `IsRolledback` are abstarcated, according to the current `CheckTxnStatus` implementation.

```
TxnStatus::RolledBack => resp.set_action(Action::NoAction),
TxnStatus::TtlExpire => resp.set_action(Action::TtlExpireRollback),
TxnStatus::LockNotExist => resp.set_action(Action::LockNotExistRollback),
TxnStatus::Committed { commit_ts } => {
    resp.set_commit_version(commit_ts.into_inner())
}
```
Only from these four types of respones the transaction status is determined.

txnkv/txnlock/lock_resolver.go
```diff
@@ -1003,31 +1017,49 @@ func (lr *LockResolver) resolveAsyncResolveData(bo *retry.Backoffer, l *Lock, st
...
-	resolveData, err := lr.checkAllSecondaries(bo, l, &status)
-	if err != nil {
-		return TxnStatus{}, err
-	}
-	resolveData.keys = append(resolveData.keys, l.Primary)
+	var toResolveKeys [][]byte
+	if status.IsStatusDetermined() {
+		toResolveKeys = make([][]byte, 0, len(status.primaryLock.Secondaries)+1)
+		toResolveKeys = append(toResolveKeys, status.primaryLock.Secondaries...)
+		toResolveKeys = append(toResolveKeys, l.Primary)
```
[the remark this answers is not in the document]
cfzjywxk: Yes, I would add some comments. The performance impact should be minial as resolving async commit locks is not a common operation.

txnkv/txnlock/lock_resolver.go
cfzjywxk: The 
```
if status.primaryLock != nil && status.primaryLock.UseAsyncCommit && !forceSyncCommit {
			if !lr.store.GetOracle().IsExpired(txnID, cmdResp.LockTtl, &oracle.Option{TxnScope: oracle.GlobalTxnScope}) {
				status.ttl = cmdResp.LockTtl
			}
		}
```
logic is to change `TxnStatus.ttl` to zero, so the 
```
if status.ttl != 0
   return

if status.primaryLock != nil && status.primaryLock.usaAsyncCommit && !forceSyncCommit {
    resolveAsyncCommit
}
```
`resolveAsyncCommit` could be reached in the `resolve` function.

I’ve moved the TTL expiration check back into the `resolve` function. This helps avoid modifying the `TxnStatus` state deep inside the call stack—especially since setting the TTL to 0 can be confusing. It also ensures that the TTL field remains read-only throughout the process.

cfzjywxk: In actual code, all lock TTLs should be greater than 0. Whether a lock has expired should ideally be determined by the upper layer, rather than by modifying the internal state of `TxnStatus` IMO.

cfzjywxk:
@MyonKeminta @zyguan @you06 @ekexium 
PTAL, we may need to pick the fix to release versions.

txnkv/txnlock/lock_resolver.go
```diff
@@ -204,10 +217,23 @@ func NewLock(l *kvrpcpb.LockInfo) *Lock {
...
+		logutil.BgLogger().Error("unexpected undetermined status saved to cache",
+			zap.Uint64("txnID", txnID), zap.Stringer("status", status), zap.Stack("stack"))
+		panic("unexpected undetermined status saved to cache")
+	}
 	lr.mu.Lock()
 	defer lr.mu.Unlock()
 
-	if _, ok := lr.mu.resolved[txnID]; ok {
+	if savedStatus, ok := lr.mu.resolved[txnID]; ok {
+		// The saved determined status should always equal to the new one.
```
cfzjywxk, line 236: @MyonKeminta Please also take a look for the new check added here.

ekexium:
`TxnStatus` looks unnecessarily complicated and is hard to understand. Can we consider using different types to represent different status?

cfzjywxk:
> `TxnStatus` looks unnecessarily complicated and is hard to understand. Can we consider using different types to represent different status?

@ekexium 
Better to do the refactor on the master branch, for the release pick PR it could be risky to change the related logic too much.

txnkv/txnlock/lock_resolver.go
```diff
@@ -204,10 +217,23 @@ func NewLock(l *kvrpcpb.LockInfo) *Lock {
 }
 
 func (lr *LockResolver) saveResolved(txnID uint64, status TxnStatus) {
+	if !status.IsStatusDetermined() {
+		logutil.BgLogger().Error("unexpected undetermined status saved to cache",
```
ekexium, line 228: Can use a panic level log

cfzjywxk:
@MyonKeminta @zyguan @ekexium @you06 PTAL

txnkv/txnlock/lock_resolver.go
```diff
@@ -137,27 +147,30 @@ func (s TxnStatus) Action() kvrpcpb.Action { return s.action }
...
+// The mapping from `CheckTxnStatus` kv result to the tidb status:
+//
+//		TxnStatus::RolledBack => resp.set_action(Action::NoAction),
+//		TxnStatus::TtlExpire => resp.set_action(Action::TtlExpireRollback),
+//		TxnStatus::LockNotExist => resp.set_action(Action::LockNotExistRollback),
+//		TxnStatus::Committed { commit_ts } => {
+//		    resp.set_commit_version(commit_ts.into_inner())
+//		}
+//
+//	 So the transaction is regarded as committed if the commit_ts is not 0, and rollback if the
```
ekexium:
```suggestion
//	 So the transaction is regarded as committed if the commit_ts is not 0, and rolled back if the
```

ekexium (commented):
Seems the PR can fix the bug. But I'm still confused. 
Is the core problem an ambiguity in states? Like, 'committed' doesn't necessarily mean 'all locks (are being) resolved' but some part of code misunderstood this? 
In `addKeys` it doesn't append keys if `len(locks) != expected`. I don't understand why it is correct...

cfzjywxk:
> Seems the PR can fix the bug. But I'm still confused. Is the core problem an ambiguity in states? Like, 'committed' doesn't necessarily mean 'all locks (are being) resolved' but some part of code misunderstood this? In `addKeys` it doesn't append keys if `len(locks) != expected`. I don't understand why it is correct...

@ekexium Yes the core problem is the an ambiguity in async commit states processing, including:
1. The tikv return `commit_ts` only in  `check_secondary_keys` responses without other left locks, the kv-client does not know how many locks are left actually.
2. The kv-client should not continue to `check_secondary_keys` if the to be resolved async commit transaction status is determined, this is fixed by this PR. If so, combined with condition 1 there could be cases deadloop happens and the resolver continues to resolve locks but only the primary key lock is resolved.

txnkv/txnlock/lock_resolver.go
```diff
@@ -137,27 +147,37 @@ func (s TxnStatus) Action() kvrpcpb.Action { return s.action }
...
-	}
-	return false
+	return s.IsStatusDetermined()
+}
+
+// StatusEqual checks whether the current status is equal to another status if the
+// transaction status is determined.
+func (s TxnStatus) StatusEqual(other TxnStatus) bool {
+	return (s.IsCommitted() && other.IsCommitted() && s.CommitTS() == other.CommitTS()) ||
+		(s.IsRolledBack() && other.IsRolledBack())
```
MyonKeminta: Is it expected that undetermined states always considered not equal here?
If so, it looks more like a partial-equal, and it's possible that a `TxnStatus` does not equal to itself. Considering this, perhaps the name can be something like `HasSameDeterminedStatus`.

cfzjywxk: Make sense, already refactored, PTAL

txnkv/txnlock/lock_resolver.go
```diff
@@ -503,7 +538,7 @@ func (lr *LockResolver) resolveLocks(bo *retry.Backoffer, opts ResolveLocksOptio
 			cleanRegions = make(map[locate.RegionVerID]struct{})
 			cleanTxns[l.TxnID] = cleanRegions
 		}
-		if status.primaryLock != nil && status.primaryLock.UseAsyncCommit && !forceSyncCommit {
+		if expiredAsyncCommitLocks {
```
MyonKeminta, line 548: I find it very hard to understand what will be going on for some cases, such as not-expired async commit lock as primary with l being either pessimistic or optimistic secondary. Can there be any analysis or explaination about how it works for different cases in the comments?

cfzjywxk: Comments are added at L533, PTAL

txnkv/txnlock/lock_resolver.go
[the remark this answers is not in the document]
ekexium: Should `Action_TTLExpirePessimisticRollback` and `Action_LockNotExistDoNothing` be considered rolled back?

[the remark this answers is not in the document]
cfzjywxk: @ekexium 
`Action_TTLExpirePessimisticRollback` just pessimistic rollback the lock, 
`Action_LockNotExistDoNothing` does not write a rollback record, 

So the transaction status from these checking results is undetermined, should not be regarded as already rollback

Issue #1631, opened 2025-04-15
txn: collect more information when inconsistency or rollback happens


cfzjywxk:
For certain unexpected cases where data issues have been confirmed, more on-site information should be collected and printed in the logs to facilitate further investigation.

For example once this log is printed
```
"2PC failed commit key after primary key committed"
```

It means the transaction coordinator encounters an unexpected failure when committing the secondary keys **AFTER** the primary key has been successfully committed (at which point the transaction is already considered successfully committed), the atomicity of the transaction can no longer be guaranteed. 

In such cases, more detailed MVCC information about the failure commit key should be printed to help diagnosis, and then it could be known whether it is rolled back by some conflict transactions(with `start_ts`), if so it means there is critical bug in the transaction protocol. Or there is  data lost and the key is lost.

ytypy:
Is there any progress for this issue? we have encountered the same problem "The commit is returned successfully by async commit protocol, but the async commit process encunter errors like TxnLockNotFound", i think it might be a really critical bug.

If this issue has not been resolved yet, is there any way to avoid this problem?

PR #1636, opened 2025-04-29
*: update kvproto version

lcwangchao:
ref #1631 

update the kvproto version


cfzjywxk (approved):

zyguan (commented):
We may need to consider keyspace codec for newly added field.

zyguan (approved):

PR #1640, opened 2025-05-06
txn: provide more information in commit RPC / log mvcc debug info when commit failed for `TxnLockNotFound`

lcwangchao:
ref #1631

### DIFF
--- a/txnkv/transaction/commit.go
+++ b/txnkv/transaction/commit.go
@@ -73,10 +73,18 @@ func (action actionCommit) tiKVTxnRegionsNumHistogram() prometheus.Observer {
 
 func (action actionCommit) handleSingleBatch(c *twoPhaseCommitter, bo *retry.Backoffer, batch batchMutations) error {
 	keys := batch.mutations.GetKeys()
+	var commitRole kvrpcpb.CommitRole
+	if batch.isPrimary {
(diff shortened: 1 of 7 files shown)


txnkv/transaction/commit.go
```diff
@@ -212,13 +220,18 @@ func (action actionCommit) handleSingleBatch(c *twoPhaseCommitter, bo *retry.Bac
 					zap.Error(err),
 					zap.Uint64("txnStartTS", c.startTS),
 					zap.Uint64("commitTS", c.commitTS),
-					zap.Strings("keys", hexBatchKeys(keys)))
+					zap.Strings("keys", hexBatchKeys(keys)),
+					zap.Uint64("sessionID", c.sessionID),
+					zap.String("debugInfo", tikverr.ExtractDebugInfoStrFromKeyErr(keyErr)))
```
cfzjywxk, line 225: Also need to print the information for async commit path if secondary key commit fails.

you06 (commented):
Rest LGTM.

BTW, what about adding a counter metric, which is increased when critical errors(like `2PC failed commit key after primary key committed`) happen, so we can monitor and alert it.

you06 (approved):

PR #1642, opened 2025-05-16
txn: handle undetermined error in client go

### LINKED ISSUE #1641: handle `RegionError.UndeterminedResult` in client-go
`UndeterminedResult` is an new type error added to `RegionError`, see: https://github.com/pingcap/kvproto/pull/1310

When client get this error, it should return `ErrResultUndetermined` to the caller without retry.

The outside code in TiDB should process the `ErrResultUndetermined` because the txn is in an undetermined state currently. For example, TiDB should disconnect the current connection at this time.

lcwangchao:
close #1641

### DIFF
--- a/config/retry/backoff.go
+++ b/config/retry/backoff.go
@@ -420,19 +420,12 @@ func IsFakeRegionError(err *errorpb.Error) bool {
 	return err != nil && err.GetEpochNotMatch() != nil && len(err.GetEpochNotMatch().CurrentRegions) == 0
 }
 
-// MayBackoffOrFailFastForRegionError handles the region error for some cases:
-// - If the `regionError` should not retry, returns an error.
-// - If the `regionError` can retry immediately, it returns nil.
-// - If the `regionError` can retry but need to do backoff, it backoff first and returns nil.
-// - If the backoff in the above case fails (.i.e reaches the max backoff limit), it returns the error.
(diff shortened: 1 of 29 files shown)


config/retry/backoff.go
```diff
@@ -413,3 +414,34 @@ func (b *Backoffer) CheckKilled() error {
...
+// - If the `regionError` should not retry, returns an error.
+// - If the `regionError` can retry immediately, it returns nil.
+// - If the `regionError` can retry but need to do backoff, it backoff first and returns nil.
+// - If the backoff in the above case fails (.i.e reaches the max backoff limit), it returns the error.
+func MayBackoffOrFailFastForRegionError(regionErr *errorpb.Error, bo *Backoffer) error {
+	if regionErr == nil {
+		return nil
+	}
+
+	if regionErr.GetUndeterminedResult() != nil {
```
cfzjywxk, line 433: The undetermined error should be returned in following cases:
1. 2pc protocol is used, and the undetermined error is returned in the response of primary key commit request.
2. Async commit protocol is used, and  the undetermined error is returned in either responess of prewrite requests.

For other cases, the undetermined error could be handled in the same way as rpc timeout error, so we may not need to adapt `MayBackoffOrFailFastForRegionError` in other commands like `TxnHeartBeat`, `PessimsiticLock`.

ekexium (approved):
So many duplications 😢 . I hope we can refactor to deduplicate one day

cfzjywxk (approved):

PR #1644, opened 2025-05-20
txn: fix the implemention of undetermined error

### LINKED ISSUE #1641: handle `RegionError.UndeterminedResult` in client-go

lcwangchao:
close #1641

It is a fix for PR: #1642 to also handle `UndeterminedResult` in `RegionRequestSender` to not retry in it.

It also fixes the test case to add some failpoint in the right place.


cfzjywxk (approved):

internal/locate/region_request.go
```diff
@@ -1494,6 +1496,15 @@ func (s *RegionRequestSender) onRegionError(
 		s.recordRPCAccessInfo(req, ctx, regionErrorToLogging(regionErr, regionErrLabel))
 	}
 
+	if regionErr.GetUndeterminedResult() != nil {
+		logutil.Logger(bo.GetCtx()).Warn(
+			"tikv reports `UndeterminedResult`",
+			zap.Stringer("req", req),
```
ekexium, line 1502: Shall we consider log redaction here

ekexium (approved):

cfzjywxk (approved):

PR #1646, opened 2025-05-28
*: support async batch get

zyguan:
ref #1586


txnkv/txnsnapshot/snapshot_async.go
```diff
@@ -0,0 +1,299 @@
...
+	for i, batch1 := range batches {
+		var backoffer *retry.Backoffer
+		if i == 0 {
+			backoffer = forkedBo
+		} else {
+			backoffer = forkedBo.Clone()
+		}
+		batch := batch1
+		s.tryBatchGetSingleRegionUsingAsyncAPI(backoffer, batch, readTier, collectF, async.NewCallback(runloop, func(_ struct{}, e error) {
+			completed++
```
cfzjywxk, line 67: Is there a possibility that async requests are processed in parallel by multiple underlying workers, causing a race condition updateing the `completed` field?

config/config.go
```diff
@@ -99,6 +101,7 @@ func DefaultConfig() Config {
 		TxnScope:              "",
 		EnableAsyncCommit:     false,
 		Enable1PC:             false,
+		EnableAsyncBatchGet:   false,
```
cfzjywxk, line 104: What was the risk of enabling batch get by default that we discussed earlier?

txnkv/txnsnapshot/snapshot.go
```diff
@@ -323,6 +324,75 @@ func appendBatchKeysBySize(b []batchKeys, region locate.RegionVerID, keys [][]by
 	return b
 }
 
+type batchGetLockInfo struct {
+	lockedKeys [][]byte
+	locks      []*txnlock.Lock
+	keyErr     *kvrpcpb.KeyError
+}
+
+func collectBatchGetResponseData(
```
cfzjywxk, line 336: Seems `collectBatchGetResponseLockInfo` is more understandable.

txnkv/txnsnapshot/snapshot_async.go
zyguan: The callback is designed to executed in the same goroutine (where the RunLoop is running), however there is a possibility of misuse, eg. call cb.Invoke in another goroutine.

config/config.go
zyguan: It hasn't been fully tested and it's on the critical path, I think.

txnkv/txnsnapshot/snapshot.go
zyguan: It returns a `*batchGetLockInfo` but also collects kv pairs and exec details by the callback args.

txnkv/txnsnapshot/snapshot_async.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+		}))
+	}
+	for completed < len(batches) {
+		_, err = runloop.Exec(bo.GetCtx())
+		if err != nil {
+			break
+		}
+	}
+	if lastForkedBo == nil {
```
ekexium: Should it be `!=`?

zyguan: Yes, I made a mistake...

ekexium (commented):
`retryBatchGetSingleRegionAfterAsyncAPI` seems a bit complex to me.
How about separating concerns by extracting functions to handle region errors and lock errors respectively?
The rest part LGTM

config/config.go
cfzjywxk: We could make it enabled by default for the nightly version as discussed.

txnkv/txnsnapshot/snapshot_async.go
cfzjywxk: How about adding comments about it and adding defensive code or test to detect such usages if possible?

config/config.go
zyguan: It seems the `DefaultConfig` will be overwritten by [`GetTiKVConfig`](https://github.com/pingcap/tidb/blob/64a25c795b70cc8351dc98093172e08a15a8440a/pkg/config/config.go#L329), setting it to `true` here doesn't take effect in tidb actually. Should we set it to true anyway?

txnkv/txnsnapshot/snapshot_async.go
zyguan: Run tests with `-race` could help, I guess. The defensive code I can figure out relies on goroutine id, however it's not exported by go runtime (also not recommanded), although there are some hack methods to get it, I doubt whether it worth to do so.

config/config.go
cfzjywxk: Not needed if so.

cfzjywxk (approved):

ekexium (approved):

PR #1649, opened 2025-06-04
Fix stale read metrics

### LINKED ISSUE #1648: missing stale read metrics when the stale read is failed
From the KV errors, a lot of stale read requests are failed, but the correspond ops/traffic metrics are not tracked.

Image

Image

you06:
close #1648

Record the stale-read metrics correctly.

- Hit
image

- Miss
image
image

### DIFF
--- a/internal/client/client_collapse.go
+++ b/internal/client/client_collapse.go
@@ -40,6 +40,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/pkg/errors"
 	"github.com/tikv/client-go/v2/tikvrpc"
 	"github.com/tikv/client-go/v2/util/async"
@@ -136,6 +137,9 @@ func (r reqCollapse) collapse(ctx context.Context, key string, sf *singleflight.
 	addr string, req *tikvrpc.Request, timeout time.Duration) (resp *tikvrpc.Response, err error) {
 	// because the request may be used by other goroutines, copy the request to avoid data race.
 	copyReq := *req
+	if req.Type == tikvrpc.CmdResolveLock && req.Req != nil {
+		copyReq.Req = proto.Clone(req.ResolveLock())
+	}
 	rsC := sf.DoChan(key, func() (interface{}, error) {
 		return r.Client.SendRequest(context.Background(), addr, &copyReq, ReadTimeoutShort) // use resolveLock timeout.
 	})


internal/locate/region_request.go
```diff
@@ -571,6 +574,16 @@ func (s *RegionRequestSender) SendReqAsync(
...
+		if stmtExec := ctx.Value(util.ExecDetailsKey); stmtExec != nil {
+			execDetails := stmtExec.(*util.ExecDetails)
+			atomic.AddInt64(&execDetails.WaitKVRespDuration, int64(time.Since(startTime)))
+			collector := networkCollector{}
+			if state.invariants.staleRead {
+				collector.staleReadMetricsCollector = &staleReadMetricsCollector{}
+			}
+			collector.onReq(req, execDetails)
+			collector.onResp(req, resp, execDetails)
+		}
```
zyguan: Should we move it to [here](https://github.com/tikv/client-go/blob/af9bd3c5d2d5d2b5b7b0cc4e175d6b96877de9f7/internal/locate/region_request.go#L1268)?

you06: Yes we can.
What do you think about the stmt exec details in the context, may we track the traffic even we're not collecting the exec details?

zyguan: > What do you think about the stmt exec details in the context, may we track the traffic even we're not collecting the exec details?

I think it would be better to do so, maybe extract `staleReadMetricsCollector` from `networkCollector` or allow `networkCollector` methods accept a nil `ExecDetails`?

you06:
@cfzjywxk @zyguan PTAL again.

internal/locate/region_request.go
```diff
@@ -1114,6 +1129,17 @@ func (s *sendReqState) send() (canceled bool) {
...
+
+		if stmtExec := ctx.Value(util.ExecDetailsKey); stmtExec != nil {
+			execDetails := stmtExec.(*util.ExecDetails)
+			atomic.AddInt64(&execDetails.WaitKVRespDuration, int64(rpcDuration))
+			collector := networkCollector{
+				staleRead: s.invariants.staleRead,
+			}
+			collector.onReq(req, execDetails)
+			collector.onResp(req, s.vars.resp, execDetails)
+		}
```
zyguan, line 1137:
```go
var execDetails *util.ExecDetails
if val := ctx.Value(util.ExecDetailsKey); val != nil {
	execDetails = val.(*util.ExecDetails)
	atomic.AddInt64(&execDetails.WaitKVRespDuration, int64(rpcDuration))
}
collector := networkCollector{staleRead: s.invariants.staleRead}
collector.onReq(req, execDetails)
collector.onResp(req, resp, execDetails)
```

zyguan (approved):

cfzjywxk (approved):

PR #1651, opened 2025-06-05
metrics: observe tail latencies of batch send/recv

zyguan:
#1411 changes some histograms to summaries, thus we are unable to confirm if there are long tail operations.

image


cfzjywxk (approved):

metrics/metrics.go
```diff
@@ -346,15 +347,15 @@ func initMetrics(namespace, subsystem string, constLabels prometheus.Labels) {
...
 		prometheus.HistogramOpts{
 			Namespace:   namespace,
 			Subsystem:   subsystem,
 			Name:        "batch_send_tail_latency_seconds",
-			Buckets:     prometheus.ExponentialBuckets(0.005, 2, 10), // 5ms ~ 2.56s
+			Buckets:     prometheus.ExponentialBuckets(0.02, 2, 8), // 20ms ~ 2.56s
 			Help:        "batch send tail latency",
 			ConstLabels: constLabels,
-		})
+		}, []string{"store"})
```
you06: We now have [`LblStore`](https://github.com/tikv/client-go/blob/b7e019d31519e149a96a0f66341d5f039f2d9772/metrics/metrics.go#L129).

you06 (commented):
LGTM. The only concern I have is that the histogram with store labels could consume significant storage space for metrics.

I believe the tail latency metrics are useful. Would you consider avoiding the Cartesian product of metrics? (Currently impossible since TiDB instances are auto-recorded, but perhaps future metrics collectors could prune some dimensions when storing the metrics.)

zyguan:
> I believe the tail latency metrics are useful. Would you consider avoiding the Cartesian product of metrics? (Currently impossible since TiDB instances are auto-recorded, but perhaps future metrics collectors could prune some dimensions when storing the metrics.)

I have no idea of how to reduce the cartesian product of metrics currently : (
However, I think these metrics are relative lite comparing to `TiKVSendReqHistogram`, they have less buckets and does not have labels like `type`.

internal/client/client_batch.go
```diff
@@ -239,6 +239,11 @@ func newBatchCommandsBuilder(maxBatchSize uint) *batchCommandsBuilder {
 	}
 }
 
+const (
+	batchSendTailLatThreshold = 20 * time.Millisecond
```
ekexium, line 243: Is there a dependency between this line and the metric definition? Is it possible to remove the dependency?

ekexium (approved):
The rest LGTM

PR #1663, opened 2025-06-16
snapshot: set `ReplicaRead` to false when `ReplicaReadType` fallbacks to `ReplicaReadLeader`

you06:
ref pingcap/tidb#61745

The [`replicaReadAdjuster`](https://github.com/tikv/client-go/blob/96b6487fac411fea3b4e15170d419d7b8307d6f2/txnkv/txnsnapshot/snapshot.go#L147-L148) can choose leader and follower replica after a request is created.
When the request is created as a replica-read request but fallback to leader read by `replicaReadAdjuster`, we need to set the [`replica_read`](https://github.com/pingcap/kvproto/blob/dc99a8f6e3487e3a73013c2e7d7cf9d5cbb55c45/proto/kvrpcpb.proto#L861) to false as well. Unless it can perform a meaningless read-index in leader and harm the performance.

## Before this PR

image

Many read-index proposals.

image

## This PR

image

Few read-index proposals.

image

### DIFF
--- a/integration_tests/snapshot_test.go
+++ b/integration_tests/snapshot_test.go
@@ -447,6 +447,7 @@ func (s *testSnapshotSuite) TestReplicaReadAdjuster() {
 			// check the replica read type
 			fn := func(next interceptor.RPCInterceptorFunc) interceptor.RPCInterceptorFunc {
 				return func(target string, req *tikvrpc.Request) (*tikvrpc.Response, error) {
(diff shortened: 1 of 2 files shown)


glorv (approved):

you06:
@glorv @zyguan fixed the test, PTAL again.

zyguan (approved):

PR #1665, opened 2025-06-16
txnkv: prevent some actions from being interrupted by kill

zyguan:
Ignore the kill flag when trying to send CommitRequest, PessimisticRollbackRequest and BatchRollbackRequest so that the locks can be cleaned up properly.

Fix https://github.com/pingcap/tidb/issues/61454 .

### DIFF
--- a/integration_tests/2pc_test.go
+++ b/integration_tests/2pc_test.go
@@ -2707,6 +2707,34 @@ func (s *testCommitterSuite) TestKillSignal() {
 	s.ErrorContains(err, "query interrupted")
 }
 
+func (s *testCommitterSuite) TestRollbackOnKill() {
+	s.Run("PrewriteLock", func() {
+		var killed uint32 = 0
+		txn := s.begin()
+		txn.SetVars(kv.NewVariables(&killed))
+		err := txn.Set([]byte("k1"), []byte("v1"))
+		s.NoError(err)
+		committer, err := txn.NewCommitter(0)
+		s.NoError(err)
+		err = committer.PrewriteAllMutations(context.Background())
+		s.NoError(err)
+		atomic.StoreUint32(&killed, 2)
+		s.NoError(committer.CleanupMutations(context.Background()))
+	})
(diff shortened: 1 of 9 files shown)


txnkv/transaction/commit.go
```diff
@@ -247,6 +247,10 @@ func (action actionCommit) handleSingleBatch(c *twoPhaseCommitter, bo *retry.Bac
 	return nil
 }
 
+func (actionCommit) isInterruptable() bool {
+	return true
```
cfzjywxk: The background commit requests after the tranasction state is decided could be regarded as `Uninterruptable` to avoid leaving locks.

integration_tests/2pc_test.go
```diff
@@ -2707,6 +2707,34 @@ func (s *testCommitterSuite) TestKillSignal() {
...
+		txn := s.begin()
+		txn.SetVars(kv.NewVariables(&killed))
+		txn.SetPessimistic(true)
+		err := txn.LockKeys(context.Background(), kv.NewLockCtx(txn.StartTS(), kv.LockNoWait, time.Now()), []byte("k2"))
+		s.NoError(err)
+		atomic.StoreUint32(&killed, 2)
+		committer, err := txn.NewCommitter(0)
+		s.NoError(err)
+		s.NoError(committer.PessimisticRollbackMutations(context.Background(), committer.GetMutations()))
+	})
```
cfzjywxk, line 2735: The `Commit` requests could be covered here.

txnkv/transaction/2pc.go
```diff
@@ -75,6 +75,7 @@ const slowRequestThreshold = time.Minute
 type twoPhaseCommitAction interface {
 	handleSingleBatch(*twoPhaseCommitter, *retry.Backoffer, batchMutations) error
 	tiKVTxnRegionsNumHistogram() prometheus.Observer
+	isInterruptable() bool
```
you06: Maybe `isInterruptible`?

PR #1672, opened 2025-06-18
Bump gprc version to 1.73.0 - remove deprecated NewSharedBufferPool

tscolari:
Remove references to `grpc.NewSharedBufferPool()`, it was removed from the grpc package and causes build to fail - it's always on. (https://pkg.go.dev/google.golang.org/grpc/experimental#WithBufferPool),

This an update on https://github.com/tikv/client-go/pull/1526, which seems to be stale for 7 months and with conflicts.
Having the code referencing `grpc.NewSharedBufferPool` causes it to be impossible to use the client in any system that is up to date on the grpc library.


tscolari:
Thank you :bow:

Issue #1678, opened 2025-06-26
When "failed to decode region range key" error happens, region ranges should be logged


ekexium:
When "failed to decode region range key" error happens, region ranges should be logged. Currently only the key was logged, it is hard to find the erroneous region range keys.

gilbi123:
Hi! I’d like to work on this issue if it’s still open.

PR #1685, opened 2025-06-27
Log region ranges on "failed to decode region range key" error (#1678)

gilbi123:
This PR improves the error logging when the "failed to decode region range key" error occurs.
Previously, the error message lacked details about the region ranges involved, which made debugging difficult.

Now, when the decoding error happens, the relevant region key ranges (startKey and endKey) are logged in the error message, providing better context to understand and diagnose the issue.

Changes:

    Enhanced error messages in region cache functions to include region key ranges.

    Added detailed logging of region start and end keys when decode errors happen.

This addresses issue #1678.


ekexium:
Hi, remember to sign off your all commits, and remove unrelated files please.

ekexium:
How about adding key details in errors returned from `decodeBytes` so that callers don't have to worry much about logging?

PR #1696, opened 2025-07-03
region cache: add options for batch loading region cache

you06:
Add options for batch loading region cache. This PR doesn't change the function usecase and keeps compatibility.


zyguan (approved):

you06:
We don't want to make the user interface too complex. A new fix is #1698.

PR #1697, opened 2025-07-04
region_cache: fix refreshFullStoreList to avoid data race

guo-shaoge:


ekexium:
Current approach always discards other info that are not set in this method, is it safe or appropriate? How about protecting the settings of the fields with the store's mutex?

internal/locate/region_cache.go
```diff
@@ -782,14 +782,14 @@ func refreshFullStoreList(ctx context.Context, stores storeCache) {
 		if addr == "" {
 			continue
 		}
-		s := stores.getOrInsertDefault(store.GetId())
-		// TODO: maybe refactor this, together with other places initializing Store
+		s := newUninitializedStore(store.GetId())
```
zyguan: It resets `healthStatus`, @MyonKeminta PTAL.

PR #1699, opened 2025-07-07
pd client: add valid function for pd client config

lhy1024:


config/config.go
```diff
@@ -124,6 +124,14 @@ func DefaultPDClient() PDClient {
 	}
 }
 
+// Valid returns true if the configuration is valid.
+func (p *PDClient) Valid() error {
```
bufferflies, line 128: where is it used？

lhy1024: It will be used in TIDB to check whether this config is valid.

rleungx (approved):

gengliqi (approved):

PR #1686, opened 2025-06-30
region_cache: add ForceRefreshAllStores function

guo-shaoge:
related tidb pr: https://github.com/pingcap/tidb/pull/62085

PR #1700, opened 2025-07-07
client: split large files and rename unclear fields

zimulala:
This refactor focuses on improving the readability and structural clarity of the internal/client/client.go and client_batch.go files. Key changes include:

* Splitting large files and isolating batch-related logic into more focused structures.
  * Moved `batchConn` from` internal/client/client_batch.go` to `internal/client/conn_batch.go`
  * Moved `connArray` from `internal/client/client.go` to `internal/client/conn_pool.go`
  * Moved `monitoredConn` and connMonitor from internal/client/client.go` to `internal/client/conn_monitor.go`
* Renaming unclear struct fields, such as `connArray.v` → `conns`, and `RPCClient.conns` → `connPools`, for better semantic clarity.

  截屏2025-07-08 13 21 40


internal/client/conn_batch.go
```diff
@@ -0,0 +1,339 @@
...
+)
+
+type batchConn struct {
+	// An atomic flag indicates whether the batch is idle or not.
+	// 0 for busy, others for idle.
+	idle uint32
+
+	// batchCommandsCh used for batch commands.
+	batchCommandsCh        chan *batchCommandsEntry
+	batchCommandsConns     []*batchCommandsClient
```
MyonKeminta: This looks not proper to be renamed `...Conns`. It's not a real connection, and the word *client* comes from the naming of goproto. It's the client of the stream established by calling gRPC stream API.

MyonKeminta (commented):
I'm ok to the changes, but I think maybe we need @zyguan to give a convincing opinion about whether the refactor is proper

internal/client/conn_batch.go
zimulala: I think this `batchConn` is not particularly reasonable; it is embedded. I just want to say, with the `conns` field analogy, and then change `batchCommandsClients` to `batchCommandsConns`. Mostly, I didn't figure out what would be better to change `batchConn` to.
Let me just change the `batchCommandsClients` back.

internal/client/client.go
```diff
@@ -592,7 +292,7 @@ func (c *RPCClient) recycleIdleConnArray() {
 	var addrs []string
 	var vers []uint64
 	c.RLock()
-	for _, conn := range c.conns {
+	for _, conn := range c.connPools {
```
zyguan:
```suggestion
	for _, pool := range c.connPools {
```

internal/locate/region_cache.go
```diff
@@ -756,11 +756,6 @@ func NewRegionCache(pdClient pd.Client, opt ...RegionCacheOpt) *RegionCache {
 	return c
 }
 
-// ForceRefreshAllStores get all stores from PD and refresh store cache.
-func (c *RegionCache) ForceRefreshAllStores(ctx context.Context) {
```
zyguan, line 760: Is #1686 deprecated?

zimulala: Yes. @guo-shaoge added it. Recently, he found that it was not used, so he hoped that I could help delete it.

zimulala:
PTAL @zyguan

zyguan (approved):

PR #1712, opened 2025-07-10
region cache: limit key ranges sent to PD in one `BatchScanRegions` gRPC

### LINKED ISSUE #1704: The size of BatchScanRegionsRequest might exceed the default MaxRequestBytes
The size of BatchScanRegionsRequest might exceed the default MaxRequestBytes (150MiB) when there are a lot of key ranges waiting for scan. The related messages are shown as the following.
```
grpc:received message larger than max(297800997 vs 157810688)
pdrpc backoffer.maxsleep 20000ms is exceeded,errors:batchScanRegion from PD failed,range num:1059860,limit:128
```

you06:
close #1704

In some user scenario, millions of key ranges can be passed in one `BatchLocateKeyRanges` call, if we sent such number of key ranges to PD in one RPC, it can either exceeds the gRPC size limit or cause performance issues. Because we limit the number of response regions in one gRPC call, it's meaningless sending too many ranges.

This PR limit the number of key ranges to 16 times of region batch size to avoid unexpected error or network traffic/CPU waste.


internal/locate/region_cache.go
```diff
@@ -1350,7 +1350,16 @@ func (c *RegionCache) BatchLocateKeyRanges(bo *retry.Backoffer, keyRanges []kv.K
...
+		// If we send too many ranges to PD, it may exceed the size of grpc limitation or cause a timeout error.
+		// So we limit the number of ranges per batch to avoid this issue.
+		maxRangesPerBatch := 16 * defaultRegionsPerBatch
+		var toBeSentRanges []router.KeyRange
+		if len(uncachedRanges) > maxRangesPerBatch {
+			toBeSentRanges = uncachedRanges[:maxRangesPerBatch]
+		} else {
+			toBeSentRanges = uncachedRanges
+		}
+		regions, err := c.BatchLoadRegionsWithKeyRanges(bo, toBeSentRanges, defaultRegionsPerBatch, opts...)
```
MyonKeminta, line 1362: What will happen if existing invocation to `BatchLocateKeyRanges` is truncated? Will it call this again when processing the remaining ranges?
(If not, how about passing the limit by the `opts` parameter?)

you06: The regions won't be truncated by limit of the ranges in one RPC call.
This function returns after all the ranges are drained.

MyonKeminta: Ah, got it. I missed the outer `for`.

MyonKeminta (approved):

zyguan (approved):

PR #1717, opened 2025-07-17
metrcis: add replica read traffic metrics

you06:
ref pingcap/tidb#62740

Add some traffic metrics to monitor the effect of replica read, this helps user to make decision of the variable.

The metrics example:

image

### DIFF
--- a/internal/locate/metrics_collector.go
+++ b/internal/locate/metrics_collector.go
@@ -17,6 +17,7 @@ package locate
 import (
 	"sync/atomic"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/tikv/client-go/v2/kv"
 	"github.com/tikv/client-go/v2/metrics"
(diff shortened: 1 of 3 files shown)


internal/locate/metrics_collector.go
```diff
@@ -89,6 +92,29 @@ func (s *networkCollector) onResp(req *tikvrpc.Request, resp *tikvrpc.Response,
 	if s.staleRead {
 		s.onRespStaleRead(float64(size), isCrossZoneTraffic)
 	}
+
+	// replica read metrics
+	if isReadReq(req.Type) {
```
cfzjywxk, line 97: Is `retrying` request caused by region errors included in this counting?

you06: Yes, both region error req/resp and the retry req/resp will be counted.

cfzjywxk (approved):

PR #1720, opened 2025-07-22
Adjustment for complete support of keyspace level GC

MyonKeminta:
* Add some exported utility functions for getting lists of key prefixes for different codec. This makes users of client-go have the ability to distinguish different key ranges. Currently, this helps in implementing the GC worker for correctly running GC for the null keyspace, as it needs to exclude keyspaces running keyspace level GC.
* Set end key properly when sending the ScanLock request. In multi-keyspace environment (next gen), this avoids ScanLock return locks that does not belong to the current keyspace (though it should not be possible as data in different keyspaces should be split in different regions), and also helps tests.


cfzjywxk (approved):

internal/apicodec/codec.go
```diff
@@ -29,6 +29,25 @@ const (
 	NullspaceID = KeyspaceID(constants.NullKeyspaceID)
 )
 
+// CodecV2Prefixes returns a sorted list of prefixes that will be used by Codec V2.
+func CodecV2Prefixes() [][]byte {
+	result := [][]byte{
```
ekexium, line 34: Nitpick: it can be a global constant to avoid allocation every time

ekexium (approved):

ekexium (approved):

PR #1727, opened 2025-07-24
Make the txn safe point polling support falling back to old version of PD

MyonKeminta:


tikv/compatible_txn_safe_point_loader.go
```diff
@@ -0,0 +1,136 @@
...
+		endpoints: etcdEndpoints,
+		tlsConfig: tlsConfig,
+		codec:     codec,
+	}
+}
+
+func (l *compatibleTxnSafePointLoader) getEtcdCli() (*clientv3.Client, error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	cli := l.etcdCli.Load()
```
tiancaiamao, line 55: This is under the protect of mutex, we can just use `cli := l.etcdCli` and `l.etcdCli = cli` ?

tiancaiamao (approved):

tikv/compatible_txn_safe_point_loader.go
MyonKeminta: The first attempt to load it is in `loadTxnSafePoint` and is not protected by the mutex. In this case read outside the mutex can be parallel with the write in the mutex. It's still unsafe without memory barrier.

tikv/kv.go
```diff
@@ -165,16 +171,35 @@ func (s *KVStore) Go(f func()) error {
...
 	s.gcStateCacheMu.Lock()
 	defer s.gcStateCacheMu.Unlock()
 
 	s.gcStateCacheMu.lastCacheTime = now
 	s.gcStateCacheMu.cachedTxnSafePoint = txnSafePoint
 }
 
+func (s *KVStore) loadTxnSafePoint(ctx context.Context) (uint64, error) {
+	if s.gcStatesAPIUnavailable.Load() {
+		logutil.Logger(ctx).Warn("GC states API is not available, which is possibly caused by cluster version < 9.0. Txn safe point updating will be fallen back to direct etcd reading.")
```
cfzjywxk: Would this introduce a lot of logs if is called much times by the tools?

tikv/compatible_txn_safe_point_loader.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return 0, errors.WithStack(err)
+	}
+	if len(resp.Kvs) == 0 {
+		return 0, nil
+	}
+
+	str := string(resp.Kvs[0].Value)
+	if len(str) == 0 {
+		return 0, nil
```
cfzjywxk, line 124: Is this expected result without error?

MyonKeminta: Yes. Empty value is always not distinguished with not existing, and the not-existing case exists when txn safe point has never been written at the early time after cluster creation.
The same behavior also exist in `EtcdSafePointKV` and PD server internal.

tikv/kv.go
MyonKeminta: 🤔
How about changing it to Debug level and only print the warning once?

cfzjywxk: Do we have metrics or clues to verify the requests coming from new or fallback old API? Or do we need such diagnosing information? If we just print the warn log one time.

MyonKeminta: I'll try to add a metric

cfzjywxk (approved):

you06 (approved):

PR #1731, opened 2025-08-14
feat: check max_execution_time for pessimistic lock requests

ekexium:
ref pingcap/tidb#62960

Check `max_execution_time` for pessimistic lock requests.
The integration test will be in the TiDB PR.


txnkv/transaction/pessimistic.go
```diff
@@ -58,6 +58,47 @@ import (
...
+	return nil
+}
+
+// calculateEffectiveWaitTime calculates the effective timeout considering both lockWaitTime and max_execution_time.
+// Returns the minimum timeout that should be applied, or kv.LockAlwaysWait if no timeout constraints apply.
+func calculateEffectiveWaitTime(lockCtx *kv.LockCtx, lockWaitTime int64, lockWaitStartTime time.Time, now time.Time) int64 {
+	effectiveTimeout := lockWaitTime
+
+	// Consider lockWaitTime if set
+	if lockWaitTime > 0 && lockWaitTime != kv.LockAlwaysWait {
```
MyonKeminta, line 84: Should `lockWaitTime == LockNoWait` or `lockWaitTime <= 0` be checked first and return LockNoWait for that case?

txnkv/transaction/pessimistic.go
```diff
@@ -58,6 +58,47 @@ import (
 	"go.uber.org/zap"
 )
 
+// MaxExecTimeExceededSignal represents the signal value for max_execution_time exceeded errors.
+// This must match the value of MaxExecTimeExceeded in TiDB's sqlkiller package.
+const MaxExecTimeExceededSignal uint32 = 2
```
MyonKeminta, line 63: Should we consider put definitions of the signals together?

ekexium: I hope to. But these signals are defined inside TiDB. Do you have any good idea how to do this? 
One way is to have individual error types for each signal and translate them when propogating back to TiDB.

txnkv/transaction/pessimistic.go
ekexium: Done

txnkv/transaction/pessimistic.go
```diff
@@ -365,6 +407,11 @@ func (action actionPessimisticLock) handlePessimisticLockResponseNormalMode(
 	// If msBeforeTxnExpired is not zero, it means there are still locks blocking us acquiring
 	// the pessimistic lock. We should return acquire fail with nowait set or timeout error if necessary.
 	if resolveLockRes.TTL > 0 {
+		// Check max_execution_time deadline first
```
cfzjywxk, line 413: Better to have test cases to cover this path.

integration_tests/lock_test.go
```diff
@@ -1591,3 +1591,142 @@ func (s *testLockWithTiKVSuite) TestPessimisticRollbackWithRead() {
...
+
+	err = txn6.LockKeys(ctx, lockCtx6, k1)
+	s.Error(err)
+	// Should be immediate lock acquisition failure, not max execution time error
+	s.Equal(tikverr.ErrLockAcquireFailAndNoWaitSet.Error(), err.Error())
+
+	// Cleanup: Unlock k1 and test successful lock with max_execution_time
+	s.NoError(txn1.Rollback())
+
+	// Test case 6: Successful lock acquisition with max_execution_time set
```
cfzjywxk, line 1714: Is there a test case to verify after timeout the resolve lock should not be triggered?

ekexium: Good point. I missed that. Now fixed and test is added

PR #1733, opened 2025-08-14
txnkv: fix the issue that batch-get error might be ignored

zyguan:
In `asyncBatchGetByRegions`, the `err` set in async callback might be overwritten to nil by runloop execution, which leads to partial results get returned inappropriately.

### DIFF
--- a/integration_tests/split_test.go
+++ b/integration_tests/split_test.go
@@ -39,6 +39,7 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/keyspacepb"
 	"github.com/pingcap/kvproto/pkg/meta_storagepb"
 	"github.com/pingcap/kvproto/pkg/metapb"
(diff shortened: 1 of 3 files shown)


zyguan:
@cfzjywxk PTAL

txnkv/txnsnapshot/snapshot_async.go
```diff
@@ -75,8 +75,8 @@ func (s *KVSnapshot) asyncBatchGetByRegions(
 		}))
 	}
 	for completed < len(batches) {
-		_, err = runloop.Exec(bo.GetCtx())
-		if err != nil {
+		if _, e := runloop.Exec(bo.GetCtx()); e != nil {
+			err = errors.WithStack(e)
```
cfzjywxk, line 79: What if the `err` is already set to a non `nil` value in the callback? Is it still overwrittened by the `runloop.Exec` returning error?

integration_tests/split_test.go
```diff
@@ -172,6 +173,12 @@ func (s *testSplitSuite) TestBatchGetUsingAsyncAPI() {
 	s.Equal([]byte("a"), m[string([]byte{'a'})])
 	s.Equal([]byte("c"), m[string([]byte{'c'})])
 	s.NotContains(m, string([]byte{'b'}))
+
+	// inject an error on sending request.
+	failpoint.Enable("tikvclient/tikvStoreSendReqResult", `1*return("timeout")`)
```
ekexium, line 178: Should disable the failpoint after use

txnkv/txnsnapshot/snapshot_async.go
zyguan: If the `runloop.Exec` returns an error, it typically means the ctx is cancelled. So, I think it's ok to return such an error, here the semantics is "return the last error".

cfzjywxk (approved):

PR #1734, opened 2025-08-15
apicodec: do not decode empty key for codec v2

tangenta:
When next-gen is enabled in TiDB, many unit tests failed because `codecV2` cannot decode the empty key. These unit tests rely on the unistore, which doesn't have a assumption about the key existence of `ErrConflict`:

https://github.com/pingcap/tidb/blob/60e5a4b8b7087eef42e9478cbe2efa4c154876fe/pkg/store/mockstore/unistore/tikv/server.go#L267

This PR makes `codecV2` supports decoding the empty key. It is also compatible with `codecV1`:

```
func (c *codecV1) DecodeKey(key []byte) ([]byte, error) {
	return key, nil
}
```

### DIFF
--- a/util/intest/in_unittest.go
+++ b/util/intest/in_unittest.go
@@ -1,0 +1,20 @@
+// Copyright 2025 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


ekexium (commented):
How about adding a warn log here incase something wrong happens in production?

ekexium (approved):

util/intest/not_in_unittest.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2025 PingCAP, Inc.
```
iosmanthus:
```suggestion
// Copyright 2025 TiKV Authors
```

util/intest/in_unittest.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2025 PingCAP, Inc.
```
iosmanthus:
```suggestion
// Copyright 2025 TiKV Authors
```

PR #1735, opened 2025-08-18
client: support interceptor in SendRequestAsync

zyguan:
When topsql is enabled, tidb will init `KvExecCounter` for each stmt and inject a `RPCInterceptor` to collect some info. And since async-batch-get is enabled by default, it can lead to a lot of warnings. This PR simply adds support for RPC interceptor by trigger it upon receiving the response.

### DIFF
--- a/internal/client/client_interceptor.go
+++ b/internal/client/client_interceptor.go
@@ -19,7 +19,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/tikv/client-go/v2/internal/logutil"
 	"github.com/tikv/client-go/v2/internal/resourcecontrol"
 	"github.com/tikv/client-go/v2/tikvrpc"
 	"github.com/tikv/client-go/v2/tikvrpc/interceptor"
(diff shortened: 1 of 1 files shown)


zyguan:
@cfzjywxk PTAL

internal/client/client_interceptor.go
```diff
@@ -114,6 +113,14 @@ func (r interceptedClient) SendRequestAsync(ctx context.Context, addr string, re
 		}
 
 		cb.Inject(func(resp *tikvrpc.Response, err error) (*tikvrpc.Response, error) {
+			if ctxInterceptor := interceptor.GetRPCInterceptorFromCtx(ctx); ctxInterceptor != nil {
+				// TODO(zyguan): In async API, the interceptor is only triggered upon receiving the response. Maybe
```
cfzjywxk, line 117: Would this `TODO` block other key features like next-gen development or is it urgent requirement?

zyguan: No, AFAIK.

cfzjywxk (approved):

PR #1736, opened 2025-08-19
txnkv: backport some optimizations for batch get

zyguan:
Here is the benchmark result.

### DIFF
--- a/txnkv/txnsnapshot/snapshot.go
+++ b/txnkv/txnsnapshot/snapshot.go
@@ -39,6 +39,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"runtime"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


zyguan:
@cfzjywxk PTAL

cfzjywxk:
@zyguan 
Please reference a github issue(like https://github.com/pingcap/tidb/issues/57580) in the description so it could be triaged.

cfzjywxk:
@lcwangchao PTAL

txnkv/txnsnapshot/snapshot.go
```diff
@@ -352,12 +361,19 @@ func (s *KVSnapshot) batchGetKeysByRegions(bo *retry.Backoffer, keys [][]byte, r
...
+		if i == 0 {
+			backoffer = bo
+		} else {
+			backoffer = bo.Clone()
+		}
 		batch := batch1
 		go func() {
-			backoffer, cancel := bo.Fork()
-			defer cancel()
+			growStackForBatchGetWorker()
```
you06, line 376: Shalle the ballast memory be kept alive until the call of `batchGetSingleRegion` is finished? Do you belive the goroutine will allocated stack memory before the stack size is shrinked?

zyguan: Yes, these goroutines generally reach the ‘wait for response’ phase very quickly, by which time their stack has grown enough to avoid shrinking. Keeping the ballast memory alive requires the stack size twice as large.

you06 (approved):

PR #1555, opened 2025-01-14
*: reduce overhead of codec client

zyguan:
1. let RPCClient attach request context to avoid attaching multiple times at upper layers.
2. let Codec use a request pool to avoid mem allocs.

| Workload | Threads | QPS (this PR) | QPS (baseline) | Diff |
| -------- | ------: | ------------: | -------------: | ---: |
| pointget |      64 |      88480.65 |       87868.89 | 0.7% |
| batchget |      64 |      22442.88 |       22165.04 | 1.3% |

PR #1737, opened 2025-08-20
*: reduce overhead of kv client

zyguan:
pick #1555 & #1557


zyguan:
@cfzjywxk PTAL

cfzjywxk:
@lcwangchao PTAL

zyguan:
> Should we ensure the req is reset before returning to the pool, so that there won't be unexpected non-default value when someone `get` a req?

the `req` is always reset after `get` [here](https://github.com/tikv/client-go/pull/1737/files#diff-b45586756f34b57867e968124fdfc1f3cde2a695187d42c175689527c9a2c4dbR129).

ekexium (approved):

PR #1747, opened 2025-08-28
client: fix the bug of attach-context in async api

zyguan:
`req.ToBatchCommandsRequest()` should be called after `tikvrpc.AttachContext`.

### DIFF
--- a/internal/client/client_async.go
+++ b/internal/client/client_async.go
@@ -49,12 +49,6 @@ func (c *RPCClient) SendRequestAsync(ctx context.Context, addr string, req *tikv
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


internal/client/client_async.go
```diff
@@ -72,6 +66,12 @@ func (c *RPCClient) SendRequestAsync(ctx context.Context, addr string, req *tikv
 	}
 	tikvrpc.AttachContext(req, req.Context)
 
+	batchReq := req.ToBatchCommandsRequest()
```
cfzjywxk, line 70: Better to comment the order quirement here.

cfzjywxk (approved):

ekexium (approved):

PR #1748, opened 2025-08-28
region_cache: avoid to put a store with empty addr into cache

### LINKED ISSUE #1746: A store with empty addr might be inserted into cache unexpectedly
In `reResolve`, there is a chance that a store with empty addr can be put into the cache. 

https://github.com/tikv/client-go/blob/82ff38718248d09f7a7bc3bcfc12935a79dda8e3/internal/locate/store_cache.go#L471-L490

And since its state is `resolved` (then `initResolve` returns an empty addr without err because cache hits), it will be filtered out in `newRegion`.

https://github.com/tikv/client-go/blob/82ff38718248d09f7a7bc3bcfc12935a79dda8e3/internal/locate/region_cache.go#L352-L359

Then such an abnormal store info may remain in the cache for a long time (current logic makes it difficult to evict), which leads to issues like region unavailable.

zyguan:
fix #1746

### DIFF
--- a/internal/locate/store_cache.go
+++ b/internal/locate/store_cache.go
@@ -469,6 +469,9 @@ func (s *Store) reResolve(c storeCache, scheduler *bgRunner) (bool, error) {
 
 	storeType := tikvrpc.GetStoreTypeByMeta(store)
 	addr = store.GetAddress()
+	if addr == "" {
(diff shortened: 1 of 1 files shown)


zyguan:
@cfzjywxk @MyonKeminta PTAL

ekexium (approved):

PR #1751, opened 2025-09-04
metrics: clean up metrics when a store is removed from the cluster

zyguan:
fix pingcap/tidb#63245

### DIFF
--- a/internal/client/conn_pool.go
+++ b/internal/client/conn_pool.go
@@ -23,12 +23,9 @@ import (
 
 	grpc_opentracing "github.com/grpc-ecosystem/go-grpc-middleware/tracing/opentracing"
 	"github.com/pkg/errors"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/tikv/client-go/v2/config"
 	tikverr "github.com/tikv/client-go/v2/error"
-	"github.com/tikv/client-go/v2/metrics"
 	"github.com/tikv/client-go/v2/tikvrpc"
-	"github.com/tikv/client-go/v2/util"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/credentials"
@@ -55,12 +52,7 @@ type connPool struct {
 
 	monitor *connMonitor
 
-	metrics struct {
-		rpcLatHist        *rpcMetrics
-		rpcSrcLatSum      sync.Map
-		rpcNetLatExternal prometheus.Observer
-		rpcNetLatInternal prometheus.Observer
-	}
+	metrics atomic.Pointer[storeMetrics]
 }
 
 func newConnPool(maxSize uint, addr string, ver uint64, security config.Security,
@@ -74,9 +66,6 @@ func newConnPool(maxSize uint, addr string, ver uint64, security config.Security
 		dialTimeout:   dialTimeout,
 		monitor:       m,
 	}
-	a.metrics.rpcLatHist = deriveRPCMetrics(metrics.TiKVSendReqHistogram.MustCurryWith(prometheus.Labels{metrics.LblStore: addr}))
-	a.metrics.rpcNetLatExternal = metrics.TiKVRPCNetLatencyHistogram.WithLabelValues(addr, "false")
-	a.metrics.rpcNetLatInternal = metrics.TiKVRPCNetLatencyHistogram.WithLabelValues(addr, "true")
 	if err := a.Init(addr, security, idleNotify, enableBatch, eventListener, opts...); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


internal/client/conn_pool.go
```diff
@@ -221,35 +210,12 @@ func (a *connPool) Close() {
...
-				a.metrics.rpcNetLatInternal.Observe(lat.Seconds())
-			} else {
-				a.metrics.rpcNetLatExternal.Observe(lat.Seconds())
-			}
-		}
+	m := a.metrics.Load()
+	storeID := req.Context.GetPeer().GetStoreId()
+	if m == nil || m.storeID != storeID {
+		mm := newStoreMetrics(storeID)
+		a.metrics.CompareAndSwap(m, mm)
```
lcwangchao: Is it expected if `CompareAndSwap` returns false? Maybe we need some comments here

internal/client/conn_pool.go
```diff
@@ -221,35 +210,12 @@ func (a *connPool) Close() {
...
-			lat := latency - time.Duration(totalRpcWallTimeNs)
-			if internal {
-				a.metrics.rpcNetLatInternal.Observe(lat.Seconds())
-			} else {
-				a.metrics.rpcNetLatExternal.Observe(lat.Seconds())
-			}
-		}
+	m := a.metrics.Load()
+	storeID := req.Context.GetPeer().GetStoreId()
+	if m == nil || m.storeID != storeID {
```
MyonKeminta, line 215: Is there any actual case that `m.storeID != storeID` might happen?

internal/client/conn_pool.go
```diff
@@ -221,35 +210,12 @@ func (a *connPool) Close() {
...
+		mm := newStoreMetrics(storeID)
+		a.metrics.CompareAndSwap(m, mm)
+		m = mm
```
MyonKeminta: If we consider the possibility that `m.storeID != storeID`, should we check the result of the CAS operation and here, and then loop on fail?

internal/locate/store_cache.go
```diff
@@ -419,22 +451,7 @@ func (s *Store) initResolve(bo *retry.Backoffer, c storeCache) (addr string, err
...
-			return "", errors.Errorf("empty store(%d) address", s.storeID)
-		}
-		s.addr = addr
-		s.peerAddr = store.GetPeerAddress()
-		s.saddr = store.GetStatusAddress()
-		s.storeType = tikvrpc.GetStoreTypeByMeta(store)
-		s.labels = store.GetLabels()
-		// Shouldn't have other one changing its state concurrently, but we still use changeResolveStateTo for safety.
-		s.changeResolveStateTo(unresolved, resolved)
+		s.initByStoreMeta(store)
```
MyonKeminta: Check error?

ekexium:
**Bug:** if PD has already removed the store, `fetchStore` returns the sentinel "invalid store ID %d, not found". We hit this branch and return, so the deletion loop never runs and the stale metrics remain. `Store.reResolve` already treats the same sentinel as confirmation that the store is gone; please reuse `isStoreNotFoundError(err)` here so we can still proceed with cleanup.

internal/locate/store_cache.go
```diff
@@ -1098,3 +1117,90 @@ func (s *Store) resetReplicaFlowsStats(destType replicaFlowsType) {
...
+		u.nextCleanUpStore = metrics.FindNextStaleStoreID(metrics.TiKVStoreLivenessGauge, validStoreIDs)
+		// delay cleaning up this store for one cleanStoreMetricsInterval
+		return
+	} else {
+		u.nextCleanUpStore = 0
+	}
+
+	// confirm that the store is indeed stale from pd
+	store, err := u.stores.fetchStore(ctx, id)
+	if err != nil {
```
ekexium: Is it necessary to check `isStoreNotFoundError(err)` here?

internal/client/conn_pool.go
zyguan: Eg, let a tidb cluster deployed by tidb-operator has 5 tikvs and the last one's addr is tikv-4.xxx.xxx, if we scale in then out the cluster by 1 tikv replica, then the addr of the last tikv might be still tikv-4.xxx.xxx but the store id changed.

internal/client/conn_pool.go
zyguan: Actually we can use `Store` here, I also added some comments.

internal/client/conn_pool.go
zyguan: I’ve changed to atomic.Store and added some comments

internal/locate/store_cache.go
zyguan: Yes, good catch.

internal/locate/store_cache.go
zyguan: It would be better to check that to avoid some data residue. Added.

internal/client/conn_pool.go
```diff
@@ -221,35 +210,15 @@ func (a *connPool) Close() {
...
-		}
+	m := a.metrics.Load()
+	storeID := req.Context.GetPeer().GetStoreId()
+	if m == nil || m.storeID != storeID {
+		// The client selects a connPool by addr via RPCClient.getConnPool, so it's possible that the storeID of the
+		// selected connPool is not the same as the storeID in req.Context. We need to create a new storeMetrics for the
+		// new storeID. Note that connPool.metrics just works as a cache, the metric data is stored in corresponding
+		// MetricVec, so it's ok to overwrite it here.
+		m = newStoreMetrics(storeID)
+		a.metrics.Store(m)
```
MyonKeminta, line 221: As there is a `Store` instead of a CAS loop, is it guaranteed that this function is never called concurrently?

zyguan: No, it will be called concurrently, but the cache pointer (connPool.metrics) will be eventully updated to point to the right value. Note that for multiple objects created by `newStoreMetrics`, as long as their store IDs are the same, the underlying metrics data is same.

ekexium (approved):
Is this only for TiKV? Is it also needed for TiFlash?
The rest LGTM

PR #1753, opened 2025-09-05
txn: check constraints for all keys in nextgen build

you06:


cfzjywxk:
Possible future test cases fix in tidb repo PR /cc @ekexium @MyonKeminta

ekexium:
I'm hesitating about this because `DoConstraintCheck` was originally for the constraint_check_in_place_pessimistic feature. That means
1. It checks against start_ts, not for_update_ts. I think this is OK.
2. The write conflict it returns is `LazyUniquenessCheck`, which can be confusing.

you06:
> 2\. The write conflict it returns is `LazyUniquenessCheck`, which can be confusing.

Is it about the [reason of `WriteConflict`](https://github.com/pingcap/kvproto/blob/21e537898f9da31d8cb148f6b7099d6946a99784/proto/kvrpcpb.proto#L1052-L1059)? It's chosen by the storage engine, what about adding a new one for this case?

ekexium:
> Is it about the [reason of WriteConflict](https://github.com/pingcap/kvproto/blob/21e537898f9da31d8cb148f6b7099d6946a99784/proto/kvrpcpb.proto#L1052-L1059)?

Yes.

> It's chosen by the storage engine, what about adding a new one for this case?

They are both `DoConstraintCheck`. How can we tell them apart?

you06:
> They are both `DoConstraintCheck`. How can we tell them apart?

Aha you are right, I think client-go can tell it.

It's also strange to handle `PrewriteRequest_DO_CONSTRAINT_CHECK` as lazy uniqueness check.

My idea is that we can add a new conflict reason named `PessimisticLockSkippedConflict` and stop responsing `LazyUniquenessCheck`. The client-go replace the reason with `LazyUniquenessCheck` for those keys we want to do lazy uniqueness check.

PR #1762, opened 2025-09-24
metrics: optimize rpc related metrics

zyguan:
cherry pick #1556 & #1751

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -44,8 +44,7 @@ var (
 	TiKVTxnCmdHistogram                            *prometheus.HistogramVec
 	TiKVBackoffHistogram                           *prometheus.HistogramVec
 	TiKVSendReqHistogram                           *prometheus.HistogramVec
-	TiKVSendReqCounter                             *prometheus.CounterVec
-	TiKVSendReqTimeCounter                         *prometheus.CounterVec
+	TiKVSendReqBySourceSummary                     *prometheus.SummaryVec
 	TiKVRPCNetLatencyHistogram                     *prometheus.HistogramVec
 	TiKVLockResolverCounter                        *prometheus.CounterVec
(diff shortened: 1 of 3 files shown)


internal/locate/store_cache.go
```diff
@@ -386,6 +387,36 @@ func (s resolveState) String() string {
 	}
 }
 
+// initByStoreMeta initializes the store fields by the given store meta. It should be protected by `resolveMutex`.
+func (s *Store) initByStoreMeta(store *metapb.Store) error {
+	if store == nil || store.GetState() == metapb.StoreState_Tombstone {
+		// The store is a tombstone.
+		s.setResolveState(tombstone)
+		return nil
```
you06: Should we clean up `s.addr` while the store is a tombstone?

I see it will return `s.addr, nil` when this function returns no error.

zyguan: I believe `s.addr` is designed to be immutable, so to avoid data race, just let `initByStoreMeta` return the addr.

metrics/metrics.go
```diff
@@ -774,6 +792,7 @@ func initMetrics(namespace, subsystem string, constLabels prometheus.Labels) {
 			Help:        "Counter of flows under PreferLeader mode.",
 			ConstLabels: constLabels,
 		}, []string{LblType, LblStore})
+	storeMetrics = append(storeMetrics, TiKVPreferLeaderFlowsGauge)
```
you06, line 795: Some metrics have `"store"` label instead of `LblStore` are missing.

e.g.

https://github.com/tikv/client-go/blob/43ccdd0991500891a137abad681f677eede86916/metrics/metrics.go#L394-L401

you06: They are address instead of storeid actually. Ignore this comment.

you06 (approved):

ekexium (approved):

PR #1776, opened 2025-10-23
Trace events

ekexium:
ref pingcap/tidb#64008
Trace important events to boost problem investigation.
The main control logic is in the tidb part. The client-go part only offers
1. Probes in client-go
2. Passing the trace id to tikv

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -33,76 +33,81 @@ const (
 	CategoryKVRequest
 )
 
-// EventTracer is the interface for recording trace events.
-// This allows TiDB to inject its trace event implementation without creating a dependency.
-type EventTracer interface {
-	// TraceEvent records a trace event with the given category, name, and fields.
-	TraceEvent(ctx context.Context, category Category, name string, fields ...zap.Field)
-}
+// TraceEventFunc is the function signature for recording trace events.
+type TraceEventFunc func(ctx context.Context, category Category, name string, fields ...zap.Field)
 
-// CategoryChecker is an optional interface that EventTracer implementations can provide
-// to allow efficient category enablement checks before expensive event construction.
-type CategoryChecker interface {
-	// IsCategoryEnabled returns true if the specified category is currently enabled for tracing.
-	IsCategoryEnabled(category Category) bool
-}
+// IsCategoryEnabledFunc is the function signature for checking if a category is enabled.
+type IsCategoryEnabledFunc func(category Category) bool
 
-// TraceIDProvider is an optional interface that EventTracer implementations can provide
-// to enable trace ID extraction from context. This allows client-go to propagate trace IDs
-// to TiKV without knowing about the application's (e.g., TiDB's) context key implementation.
-type TraceIDProvider interface {
-	// ExtractTraceID extracts the trace ID from the context.
-	// Returns nil if no trace ID is present.
-	ExtractTraceID(ctx context.Context) []byte
-}
+// Default no-op implementations
+func noopTraceEvent(context.Context, Category, string, ...zap.Field) {}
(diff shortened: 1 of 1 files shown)


ekexium:
@tiancaiamao PTAL

trace/trace.go
```diff
@@ -0,0 +1,108 @@
...
+// IsCategoryEnabled checks if a category is enabled for tracing.
+// Returns true if the tracer supports category checking and the category is enabled,
+// or true if the tracer doesn't support checking (conservative default).
+func IsCategoryEnabled(category Category) bool {
+	tracer := globalTracer.Load().(EventTracer)
+	if checker, ok := tracer.(CategoryChecker); ok {
+		return checker.IsCategoryEnabled(category)
+	}
+	// If tracer doesn't implement CategoryChecker, assume enabled
+	return true
```
cfzjywxk: Is it risky in other usage scnearios like the tidb tools returning  `enabled` here?

ekexium: Non-TiDB users should not set the trace and will use the noop tracer.
But it's good to default to disabled for performance.

cfzjywxk (approved):

trace/trace.go
```diff
@@ -0,0 +1,108 @@
...
+	tracer := globalTracer.Load().(EventTracer)
+	tracer.TraceEvent(ctx, category, name, fields...)
+}
+
+// IsCategoryEnabled checks if a category is enabled for tracing.
+// Returns true if the tracer supports category checking and the category is enabled,
+// or true if the tracer doesn't support checking (conservative default).
+func IsCategoryEnabled(category Category) bool {
+	tracer := globalTracer.Load().(EventTracer)
+	if checker, ok := tracer.(CategoryChecker); ok {
```
tiancaiamao: How about just move `IsCategoryEnabled(category Category) bool` method to EventTracer interface?
When trace category is enabled, we can save a interface cast here.
In fact, we don't really the `CategoryChecker` interface outside the package.

trace/trace.go
```diff
@@ -0,0 +1,108 @@
...
+// TraceIDFromContext extracts the trace ID from the context using the global tracer.
+// Returns nil if the tracer doesn't implement TraceIDProvider or if no trace ID is present.
+func TraceIDFromContext(ctx context.Context) []byte {
+	if ctx == nil {
+		return nil
+	}
+	tracer := globalTracer.Load().(EventTracer)
+	if provider, ok := tracer.(TraceIDProvider); ok {
+		return provider.ExtractTraceID(ctx)
+	}
```
tiancaiamao: Maybe we will not have several Tracer implementation, so here it's trading performance for flexibility, which is not worth. Why not the tradition way, just a TraceKey definiotion and `TraceIDFromContext` is context.Value(traceKey) ...

trace/trace.go
```diff
@@ -0,0 +1,108 @@
...
+	// Returns nil if no trace ID is present.
+	ExtractTraceID(ctx context.Context) []byte
+}
+
+// noopTracer is a no-op implementation used when no tracer is set.
+type noopTracer struct{}
+
+func (noopTracer) TraceEvent(context.Context, Category, string, ...zap.Field) {}
+
+var globalTracer atomic.Value // stores EventTracer
```
tiancaiamao: There is a `type Pointer[T [any](https://pkg.go.dev/builtin#any)] struct`
https://pkg.go.dev/sync/atomic#Pointer

So this can be used ...

```
var globalTracer atomic.Pointer[EventTracer]
```

But if you unlike the Go generic, just keep the old way~~

ekexium:
Reimplemented the trace function and the category filter function. PTAL. @tiancaiamao

Issue #32, opened 2019-09-03
feature: Add options in 'put' and 'batchPut' for txn API and raw API


BeyongTheMemory:
The duplicated key can return error,  ignore, rewrite.
Just support rewrite now when duplicated key.

disksing:
Hi @BeyongTheMemory , for txn API, as tikv supports `Insert` which reports error when key already exists. I think we can fulfill your needs by using `Insert` instead of `Put` with some options. However, I will be busy with other things in the near future, and it may take some time to fix this. For the rawkv API, it cannot be supported before tikv implements CAS commands.

BeyongTheMemory:
Hi @disksing , sorry, I couldn't find  `Insert ` in txn API, just found `Set `

BeyongTheMemory:
BTW, Have any plan supported CAS commands in TiKV?

disksing:
@BeyongTheMemory No plan now.

PR #1779, opened 2025-10-30
upgrade grpc to v1.73

joechenrh:
**Why upgrade grpc**

Because we are using some expermential API that was already renamed. This will block those libraries that depend on this to use newer version of grpc. And we have introduce github.com/apache/arrow-go/v18 to TiDB repo, which requires google.golang.org/grpc@v1.73

**What has be changed**

- Change `WithRecvBufferPool` with `WithBufferPool`.
- Use old codec in grpc 1.63 to make `sharedBytes`([tipb #32](https://github.com/pingcap/tipb/pull/32), [kvproto #142](https://github.com/pingcap/kvproto/pull/142)) works.

**TiFlash test**

This part is to verify if the previous issue ([#9159](https://github.com/pingcap/tiflash/issues/9159)) still exists. You can check the result on tcms.

**sysbench test**

This part is to verify that there won't be performance degradation after upgrading grpc version.

```
# range scan
sysbench oltp_read_only \
  --db-driver=mysql --mysql-host=172.16.4.180 --mysql-port=32003 \
  --mysql-user=root --mysql-db=test \
  --tables=1 --table-size=100000000 \
  --threads=16 --time=1800 --report-interval=10 \
  --rand-type=uniform --range_size=4096 --range_selects=on \
  --point_selects=0 \
  --db-ps-mode=disable run
 ```

For range scan, the QPS/latency are almost same for both version. The memory for 1.73 is slightly reduced (4.09G->3.98G), while the gc rate increases (0.25->0.3), which is an expected result by introducing `sync.Pool` inside the grpc-go.

image

image

```
# point lookup
sysbench oltp_read_only \
  --db-driver=mysql --mysql-host=172.16.4.180 --mysql-port=32003 \
  --mysql-user=root --mysql-db=test \
  --tables=1 --table-size=100000000 \
  --threads=16 --time=1800 --report-interval=10 \
  --rand-type=uniform --range_size=4096 --range_selects=on \
  --db-ps-mode=disable run
 ```

For point lookup, both version show almost no difference in QPS/lantence/memory.

image

image

## Summary by CodeRabbit

* **Chores**
  * Upgraded gRPC and Protocol Buffer dependencies for improved stability and compatibility across releases.
  * Updated example modules to align with the dependency upgrades.

* **New Features**
  * Added a legacy protobuf codec to preserve compatibility with older gRPC/protobuf usage.

* **Performance**
  * Reworked buffer pooling to reduce memory overhead and improve throughput.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


joechenrh:
Previously, grpc was upgraded to 1.64 and reverted (by https://github.com/tikv/client-go/pull/1369) due to errors in TiFlash regression tests. So I keep using old API `DialContext` here. 

I've ran a test internally with this branch. The plan ID is 7972306, and you can check the result.

---

```
rpc error: code = DeadlineExceeded desc = received context error while waiting for new LB policy update: context deadline exceeded
rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 10.200.86.26:3930: i/o timeout
```

(Although I suspect that error is not related to grpc API change because it seems like we couldn't connect to `10.200.86.26:3930`. That is, the first mistake was the result, and the second mistake was the cause, as resolver didn't provide any address to LB.)

dveeden:
This would fix #1789, right?

internal/client/client_batch.go
```diff
@@ -61,11 +61,14 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
+	"google.golang.org/grpc/mem"
 	"google.golang.org/grpc/metadata"
 )
 
 // globalEncodedMsgDataPool is used to pool pre-encoded message data for batch commands.
-var globalEncodedMsgDataPool = grpc.NewSharedBufferPool()
+var globalEncodedMsgDataPool = mem.NewTieredBufferPool(
```
joechenrh, line 69: `NewSharedBufferPool` is replaced with `NewTieredBufferPool`.
- Previous size tier: 16B, 256B, 4KB, 64KB, 1MB
- New size tier: 256B, 4KB, 16KB, 32KB, 1MB

internal/client/conn_pool.go
```diff
@@ -141,8 +146,8 @@ func (a *connPool) Init(addr string, security config.Security, idleNotify *uint3
 				Timeout: cfg.TiKVClient.GetGrpcKeepAliveTimeout(),
 			}),
 		}, opts...)
-		if cfg.TiKVClient.GrpcSharedBufferPool {
-			opts = append(opts, experimental.WithRecvBufferPool(grpc.NewSharedBufferPool()))
+		if !cfg.TiKVClient.GrpcSharedBufferPool {
```
joechenrh, line 149: After grpc-go 1.66, the default pool was changed from noop pool to `defaultBufferPool`
https://github.com/grpc/grpc-go/blob/c7ec4d9ae3281bc57a8adce59b572e56965fb728/dialoptions.go#L705-L718
Here we just align with the old behavior: use noop pool by default.

internal/client/conn_pool.go
```diff
@@ -76,6 +77,7 @@ func (a *connPool) monitoredDial(ctx context.Context, connName, target string, o
 	conn = &monitoredConn{
 		Name: connName,
 	}
+	//nolint:SA1019
 	conn.ClientConn, err = grpc.DialContext(ctx, target, opts...)
```
joechenrh, line 81: Keep using the old (deprecated) API to avoid potential TiFlash problems.
Ref: https://github.com/pingcap/tiflash/issues/9159

go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	go.uber.org/goleak v1.2.0
 	go.uber.org/zap v1.26.0
 	golang.org/x/sync v0.12.0
-	google.golang.org/grpc v1.63.2
+	google.golang.org/grpc v1.73.0
```
bufferflies, line 33: What about other go.mod in examples?

joechenrh: I think it's OK to keep them, but we can upgrade them as well.

bufferflies: I think keep same is better

PR #1785, opened 2025-11-06
tracing: pass immediate logging down to kv request

ekexium:
ref https://github.com/pingcap/tidb/issues/64008

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -64,6 +64,7 @@ import (
 	"github.com/tikv/client-go/v2/kv"
 	"github.com/tikv/client-go/v2/metrics"
 	"github.com/tikv/client-go/v2/tikvrpc"
+	"github.com/tikv/client-go/v2/trace"
 	"github.com/tikv/client-go/v2/util"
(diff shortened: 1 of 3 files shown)


ekexium:
@tiancaiamao  PTAL

cfzjywxk (approved):

trace/flags.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Has checks if the given flag is set.
+func (f TraceControlFlags) Has(flag TraceControlFlags) bool {
+	return f&flag != 0
+}
+
+// Set returns a new flags value with the given flag set.
+func (f TraceControlFlags) Set(flag TraceControlFlags) TraceControlFlags {
+	return f | flag
+}
```
MyonKeminta: May it be confusing whether it modifies the original value?

tiancaiamao (approved):

trace/flags.go
ekexium: Renamed

MyonKeminta (approved):

PR #1801, opened 2025-11-20
trace: capture the callstack when region_request.SendReq() is called without trace ID

tiancaiamao:
For https://github.com/pingcap/tidb/pull/64569

### DIFF
--- a/internal/locate/region_request.go
+++ b/internal/locate/region_request.go
@@ -1015,6 +1015,17 @@ func (s *sendReqState) next() (done bool) {
 		}
 	}
 
+	if len(req.Context.TraceId) == 0 {
+		if trace.IsCategoryEnabled(trace.CategoryDevDebug) {
+			trace.TraceEvent(bo.GetCtx(), trace.CategoryDevDebug, "send request with trace_id missing",
+				zap.Stack("stack"))
+
+			// If trace ID is not set, it may indicate the caller is not using client-go properly.
+			// Print the stack trace to capture all those cases.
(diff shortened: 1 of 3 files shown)


internal/locate/region_request.go
```diff
@@ -1015,6 +1015,17 @@ func (s *sendReqState) next() (done bool) {
...
 	}
 
+	if len(req.Context.TraceId) == 0 {
+		if trace.IsCategoryEnabled(trace.CategoryDevDebug) {
+			trace.TraceEvent(bo.GetCtx(), trace.CategoryDevDebug, "send request with trace_id missing",
+				zap.Stack("stack"))
+
+			// If trace ID is not set, it may indicate the caller is not using client-go properly.
+			// Print the stack trace to capture all those cases.
+			trace.CheckFlightRecorderDumpTrigger(bo.GetCtx(), "dump_trigger.suspicious_event.dev_debug", "send_request_trace_id_missing")
```
ekexium, line 1025: This is a dependency of TiDB, which we should try our best to avoid I think.

One possible way I can think of to improve it is to make it something like `NullTraceIdHook` and let the adapter in TiDB translate it to `CheckFlightRecorderDumpTrigger("dump_trigger.suspicious_event.dev_debug", "send_request_trace_id_missing")`.

If there will be more dump triggers, a more flexible design is needed. Just like how we have client-go categories and its translation to tidb categories.

tiancaiamao: This is not a dependency of TiDB, it is a hook just as you described. @ekexium 
If TiDB set this function, then the hook is called, otherwise it noop function.

ekexium: It's not a code dependency but a knowledge dependency. If we don't go to TiDB code we cannot understand what "dump_trigger.suspicious_event.dev_debug" is.

PR #1810, opened 2025-12-01
txnkv: add SetCommitWaitUntilTSO and SetCommitWaitUntilTSOTimeout txn option

tiancaiamao:
close https://github.com/pingcap/tidb/issues/64797

### DIFF
--- a/integration_tests/option_test.go
+++ b/integration_tests/option_test.go
@@ -98,7 +98,7 @@ func (s *testOptionSuite) TestSetCommitWaitUntilTSOTimeout() {
 	txn, err := s.store.Begin()
 	s.NoError(err)
 	defer txn.Rollback()
-	s.Equal(time.Second, txn.GetMaxClockDriftInActiveActiveReplication())
+	s.Equal(time.Second, txn.GetCommitWaitUntilTSOTimeout())
 	txn.KVTxn.SetCommitWaitUntilTSOTimeout(2 * time.Second)
-	s.Equal(2*time.Second, txn.GetMaxClockDriftInActiveActiveReplication())
+	s.Equal(2*time.Second, txn.GetCommitWaitUntilTSOTimeout())
 }


txnkv/transaction/txn.go
```diff
@@ -1905,6 +1921,34 @@ func (txn *KVTxn) MemHookSet() bool {
...
+func (txn *KVTxn) GetTimestampForCommit(bo *retry.Backoffer, scope string) (uint64, error) {
+	ts, err := txn.store.GetTimestampWithRetry(bo, scope)
+	if err != nil {
+		return ts, err
+	}
+
+	for retryTimes := 0; ts <= txn.minCommitTS && retryTimes < 2; retryTimes++ {
+		interval := oracle.GetTimeFromTS(txn.minCommitTS).Sub(oracle.GetTimeFromTS(ts))
+		if interval > time.Second {
+			return 0, errors.Errorf("commit_ts(%d) << min_commit_ts(%d), interval more than 1 second", ts, txn.minCommitTS)
```
lcwangchao: Do we need another option `WaitMinCommitTSTimeout` to avoid some hardcoded logic? User may want to set the custom timeout value in the future.

txnkv/transaction/txn.go
```diff
@@ -211,6 +212,14 @@ type KVTxn struct {
...
 
 	prewriteEncounterLockPolicy PrewriteEncounterLockPolicy
+
+	// This minCommitTS does not has exactly same semantic with 2pc minCommitTSMgr.value.
+	// There might be suble subtle differs. (This minCommitTS value is not calculated, and is not
+	// even from the same cluster) It's only used by active-active replication feature currently.
+	// To obtain correctness without the burden of too much reasoning, it is simply implemented by
+	// getting current TSO and comparing the value with this minCommitTS. And sleep if the current
+	// tso is less than minCommitTS.
+	minCommitTS uint64
```
cfzjywxk: Better to avoid using `minCommitTS` directly which conflicts with the existing `minCommitTS` meaning in the transaction protocol.

Consider renaming it something like `ReplicationTS`.

tiancaiamao: Or `OriginCommitTS`, the txn is replicated from the other cluster and this is the original commit ts.
To guarantee the LWW (last write win) conflict resolve strategy, the commit ts for the txn must be greater than OriginCommitTS

txnkv/transaction/txn.go
```diff
@@ -211,6 +212,14 @@ type KVTxn struct {
...
 
 	prewriteEncounterLockPolicy PrewriteEncounterLockPolicy
+
+	// This originCommitTS does not has exactly same semantic with 2pc minCommitTSMgr.value.
+	// There might be suble subtle differs. (This originCommitTS value is not calculated, and is not
+	// even from the same cluster) It's only used by active-active replication feature currently.
+	// To obtain correctness without the burden of too much reasoning, it is simply implemented by
+	// getting current TSO and comparing the value with this originCommitTS. And sleep if the current
+	// tso is less than originCommitTS.
+	originCommitTS uint64
```
lcwangchao: or `CommitWaitUntilTSO` ? I think its better to decouple the semantic with active-active sync.

config/config.go
```diff
@@ -89,25 +90,28 @@ type Config struct {
 	// NOTE: The value of "TxnScope" is also the "zone" label, but since the feaute "txn-scope" is to be deprecated
 	//       (See: https://github.com/pingcap/tidb/issues/58838), we use a separate field to make its meaning clearer.
 	ZoneLabel string
+	// MaxClockDriftInActiveActiveReplication indicates the maximum clock drift allowed in active-active replication.
+	MaxClockDriftInActiveActiveReplication time.Duration
```
lcwangchao: We can make it a transaction-level config with some default value. In the future, if we want to introduce some session level variable, its easy to update this value.

lcwangchao: And same with minCommitTS, could we make a name that is decoupled from active-active sync?

txnkv/transaction/txn.go
```diff
@@ -1905,6 +1919,39 @@ func (txn *KVTxn) MemHookSet() bool {
...
+		}
+
+		if interval < time.Millisecond {
+			interval = time.Millisecond
+		}
+
+		if err == nil {
+			err = errors.Errorf("clock drift from the upstream cluster")
+		}
+		bo.BackoffWithCfgAndMaxSleep(retry.BoMaxTsNotSynced, int(interval.Milliseconds()), err)
```
lcwangchao: PD will update the TSO physical time every 40ms (not very sure, we can confirm with @bufferflies ). If so, when the interval is 1 millisecond, it may retry more than 2 times before to get a right TSO. If the max retry time is 2, in this case, the transaction will fail even if the TSO gap is small

tiancaiamao: No, not every 40ms, but every ms
I think you misunderstand it as how long to sync the ts to disk (etcd)

txnkv/transaction/test_probe.go
```diff
@@ -136,6 +136,10 @@ func (txn TxnProbe) GetAggressiveLockingPreviousKeysInfo() []AggressiveLockedKey
 	return keys
 }
 
+func (txn TxnProbe) GetMaxClockDriftInActiveActiveReplication() time.Duration {
```
lcwangchao:
```suggestion
func (txn TxnProbe) GetCommitWaitUntilTSOTimeout() time.Duration {
```

txnkv/transaction/txn.go
```diff
@@ -1905,6 +1927,47 @@ func (txn *KVTxn) MemHookSet() bool {
...
+	if err != nil {
+		return ts, err
+	}
+	if ts > txn.commitWaitUntilTSO {
+		return ts, nil
+	}
+
+	// slow path
+	interval := oracle.GetTimeFromTS(txn.commitWaitUntilTSO).Sub(oracle.GetTimeFromTS(ts))
+	if txn.commitWaitUntilTSOTimeout > 0 && interval > txn.commitWaitUntilTSOTimeout {
```
lcwangchao: We can remove the maxSleep before this line and compare the maxSleep and the intrval

PR #1814, opened 2025-12-03
trace event only when category is enabled

### LINKED ISSUE #1813: Some usages of `TraceEvent` are unconditional
All usages of `TraceEvent` should be conditioned by `IsCategoryEnabled` so they won't have overhead when disabled

ekexium:
close #1813


ekexium:
@tiancaiamao PTAL

PR #1635, opened 2025-04-22
tikv: disable health-feedback in next-gen

zyguan:
ref: https://github.com/pingcap/tidb/issues/60697

PR #1819, opened 2025-12-09
Enable health tick on next gen

MyonKeminta:
The tick in health status manager in client-go is disabled for next gen in #1635 . Currently, the related functionalities in TiKV were ported to next gen, so it can be enabled now.

By the way, it needs to be pointed out that the option `RegionCacheNoHealthTick` is for test purposes to remove the background works. It doesn't completely disable health status related features, instead, it may partially run and has risk to cause unexpected results.


cfzjywxk (approved):
LGTM

PR #1822, opened 2025-12-15
locate: clamp bucket boundaries to region boundaries

ekexium:
Clamp bucket boundaries that exceed region boundaries.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -1195,7 +1195,7 @@ func (l *KeyLocation) clampBucketToRegion(bucket *Bucket) *Bucket {
 	endKey := bucket.EndKey
 
 	// Clamp start: max(bucket.StartKey, region.StartKey)
-	if len(l.StartKey) > 0 && bytes.Compare(startKey, l.StartKey) < 0 {
+	if bytes.Compare(startKey, l.StartKey) < 0 {
 		startKey = l.StartKey
 	}
 
--- a/internal/locate/region_cache_test.go
+++ b/internal/locate/region_cache_test.go
@@ -2051,6 +2051,16 @@ func TestBucketClampingToRegion(t *testing.T) {
 			wantEnd:         []byte("m"), // Clamped to region end
 			shouldBeClamped: true,
 		},
+		{
+			name:            "infinity region - no clamping needed",
+			regionStart:     []byte{}, // Beginning of keyspace
+			regionEnd:       []byte{}, // End of keyspace (infinity)
+			bucketStart:     []byte("a"),
+			bucketEnd:       []byte("z"),
+			wantStart:       []byte("a"),
+			wantEnd:         []byte("z"),
+			shouldBeClamped: false,
+		},
 	}
 
 	for _, tt := range tests {


internal/locate/region_cache.go
```diff
@@ -1187,6 +1187,46 @@ func (b *Bucket) Contains(key []byte) bool {
...
 
+// clampBucketToRegion ensures bucket boundaries don't exceed region boundaries.
+// This is a defensive measure against stale bucket metadata that can have keys
+// outside the current region boundary.
+func (l *KeyLocation) clampBucketToRegion(bucket *Bucket) *Bucket {
+	startKey := bucket.StartKey
+	endKey := bucket.EndKey
+
+	// Clamp start: max(bucket.StartKey, region.StartKey)
+	if len(l.StartKey) > 0 && (len(startKey) == 0 || bytes.Compare(startKey, l.StartKey) < 0) {
```
MyonKeminta: Isn't the case completely included in `bytes.Compare(startKey, l.StartKey) < 0`?

internal/locate/region_cache.go
```diff
@@ -1187,6 +1187,46 @@ func (b *Bucket) Contains(key []byte) bool {
...
+		startKey = l.StartKey
+	}
+
+	// Clamp end: min(bucket.EndKey, region.EndKey)
+	// Note: empty endKey means infinity, so only clamp if both are non-empty
```
MyonKeminta: The comment looks not quite correct. You need to clamp if buckets' end key is empty but region's is not.

internal/locate/region_cache.go
```diff
@@ -1187,6 +1187,46 @@ func (b *Bucket) Contains(key []byte) bool {
...
+	if len(l.EndKey) > 0 && (len(endKey) == 0 || bytes.Compare(endKey, l.EndKey) > 0) {
+		endKey = l.EndKey
+	}
+
+	// Ensure clamped bucket is valid (start < end)
+	if len(endKey) > 0 && len(startKey) > 0 && bytes.Compare(startKey, endKey) >= 0 {
```
MyonKeminta: It looks the condition `len(startKey) > 0` is redundant?

internal/locate/region_cache_test.go
```diff
@@ -1918,6 +1918,124 @@ func (s *testRegionCacheSuite) TestLocateBucket() {
...
+		{
+			name:            "bucket completely before region - fallback to region",
+			regionStart:     []byte("m"),
+			regionEnd:       []byte("z"),
+			bucketStart:     []byte("a"),
+			bucketEnd:       []byte("f"), // Ends before region starts
+			wantStart:       []byte("m"), // Falls back to region boundaries
+			wantEnd:         []byte("z"),
+			shouldBeClamped: true,
+		},
```
MyonKeminta, line 2003: Looks there better be some cases where the bucket's keys has empty values, combined with the region's range keys either containing or not containing empty values.

internal/locate/region_cache.go
```diff
@@ -1195,18 +1195,18 @@ func (l *KeyLocation) clampBucketToRegion(bucket *Bucket) *Bucket {
 	endKey := bucket.EndKey
 
 	// Clamp start: max(bucket.StartKey, region.StartKey)
-	if len(l.StartKey) > 0 && (len(startKey) == 0 || bytes.Compare(startKey, l.StartKey) < 0) {
+	if len(l.StartKey) > 0 && bytes.Compare(startKey, l.StartKey) < 0 {
```
MyonKeminta: `len(l.StartKey) > 0` looks still unnecessary.

internal/locate/region_cache_test.go
```diff
@@ -1918,6 +1918,174 @@ func (s *testRegionCacheSuite) TestLocateBucket() {
 	}
 }
 
+// TestBucketClampingToRegion tests that bucket boundaries exceeding region
+// boundaries are clamped correctly.
+func TestBucketClampingToRegion(t *testing.T) {
```
cfzjywxk, line 1923: Should the infinity region be covered too?

cfzjywxk (approved):
LGTM

PR #1796, opened 2025-11-13
*: return commit timestamp for Get / BatchGet if needed

lcwangchao:
close #1795

- introduce options for `Get` and `BatchGet`
- introduce option `WithRequireCommitTS` to return commit timestamp
- Get / BatchGet returns `ValueEntry`, which contains the `CommitTS` field

PR #1831, opened 2025-12-26
*: support `WithReturnCommitTS` and `SetCommitWaitUntilTSO`

lcwangchao:
This PR is a cherry-pick for:

- https://github.com/tikv/client-go/pull/1796
- https://github.com/tikv/client-go/pull/1798
- https://github.com/tikv/client-go/pull/1810

- It adds a new option `WithReturnCommitTS` for `Get` and `BatchGet` to fetch the commit timestamp.
- It adds a new method `SetCommitWaitUntilTSO` which indicates the txn should wait for the TSO after the specified timestamp to get a commit ts large enough.
- It adds a new method `SetCommitWaitUntilTSOTimeout` which indicates the timeout for waiting the TSO to be greater than the specifiy ts.

### DIFF
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -1938,7 +1938,9 @@ func (txn *KVTxn) GetTimestampForCommit(bo *retry.Backoffer, scope string) (uint
 		return ts, nil
 	}
 
-	// slow path
+	// maxSleep is the maximum time we are allowed to wait for the expected commit TS.
+	// It is 1 second by default to avoid infinite blocking but can be overridden by commitWaitUntilTSOTimeout.
+	// If the TSO drift is larger than the maxSleep, return error directly.
 	maxSleep := 1000
 	if txn.commitWaitUntilTSOTimeout > 0 {
 		maxSleep = int(txn.commitWaitUntilTSOTimeout.Milliseconds())
--- a/txnkv/txnsnapshot/snapshot.go
+++ b/txnkv/txnsnapshot/snapshot.go
@@ -1129,6 +1129,8 @@ func (s *KVSnapshot) CleanCache(keys [][]byte) {
 	}
 }
 
+// getSnapCacheWithoutLock gets value from snapshot cache without locking.
+// NOTICE: this method should be called under read lock (s.mu).
 func (s *KVSnapshot) getSnapCacheWithoutLock(key []byte, returnCommitTS bool) (entry kv.ValueEntry, ok bool) {
 	if entry, ok = s.mu.cached[string(key)]; ok {
 		if !returnCommitTS {


.github/workflows/compatibility.yml
```diff
@@ -2,7 +2,7 @@ name: Compatibility Test
 
 on:
   push:
-    branches: [master]
+    branches: [master, "feature/*"]
```
cfzjywxk, line 5: Is `feature/*` still needed on master branch?

txnkv/transaction/txn.go
```diff
@@ -1905,6 +1927,47 @@ func (txn *KVTxn) MemHookSet() bool {
...
+	ts, err := txn.store.GetTimestampWithRetry(bo, scope)
+	if err != nil {
+		return ts, err
+	}
+	if ts > txn.commitWaitUntilTSO {
+		return ts, nil
+	}
+
+	// slow path
+	maxSleep := 1000
```
cfzjywxk, line 1944: Better to add comments to explain the choice of the constant value of `maxSleep` here.

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1079,9 +1122,31 @@ func (s *KVSnapshot) CleanCache(keys [][]byte) {
...
-		s.mu.cachedSize -= len(s.mu.cached[string(key)])
-		delete(s.mu.cached, string(key))
+		if oldVal, ok := s.mu.cached[string(key)]; ok {
+			s.mu.cachedSize -= oldVal.Size()
+			delete(s.mu.cached, string(key))
+		}
+	}
+}
+
+func (s *KVSnapshot) getSnapCacheWithoutLock(key []byte, returnCommitTS bool) (entry kv.ValueEntry, ok bool) {
```
cfzjywxk, line 1134: Need to add comments to remind it should be called under the protection of `s.mu`.

.github/workflows/compatibility.yml
lcwangchao: This change seems harmless; the next feature branch can benefit from it without any changes for this file.

cfzjywxk:
@tiancaiamao PTAL

integration_tests/go.mod
```diff
@@ -121,6 +174,7 @@ require (
 
 replace (
 	github.com/go-ldap/ldap/v3 => github.com/YangKeao/ldap/v3 v3.4.5-0.20230421065457-369a3bab1117
-	github.com/pingcap/tidb => github.com/lhy1024/tidb v1.1.0-beta.0.20251215110526-59e2c3c2eab3
+	github.com/pingcap/tidb => github.com/lcwangchao/tidb v1.1.0-beta.0.20251229025756-9746aadede81
```
tiancaiamao, line 177: So, client-go integration test is depending on fixed tidb commit, instead of latest?

tiancaiamao (approved):

integration_tests/go.mod
lcwangchao: This dependency is only used for unistore. But we should rely on a personal branch to make sure this integration can be built because the master TiDB does not fit the API change.

PR #1833, opened 2026-01-08
txn: support shared lock

you06:
ref tikv/tikv#19087

Support shared lock for client-go with some limitations:

- Shared lock cannot be PK (shared locks cannot being acquired when there is a PK choosed)
- Txn with shared lock cannot lock keys with aggressive mode
- Txn with shared lock cannot commit with 1pc/async commit protocol

## Summary by CodeRabbit

* **New Features**
  * Added shared-lock mode for pessimistic transactions enabling concurrent shared locks.

* **Behavior Changes**
  * Shared locks coexist with exclusive locks and affect prewrite/commit/1PC/async decisions, scanning, GC, resolution, and tracing.
  * Pessimistic shared locks require a primary key and cannot be upgraded to exclusive.

* **Metrics**
  * New label and observers to track shared-lock key usage.

* **Tests**
  * Extensive unit, integration, and end-to-end tests covering shared vs exclusive locking, blocking, resolution, GC/TTL, commits/rollbacks, and edge cases.

### DIFF
--- a/integration_tests/shared_lock_test.go
+++ b/integration_tests/shared_lock_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/stretchr/testify/suite"
+	"github.com/tikv/client-go/v2/config"
 	"github.com/tikv/client-go/v2/kv"
 	"github.com/tikv/client-go/v2/oracle"
 	"github.com/tikv/client-go/v2/tikv"
@@ -34,6 +35,10 @@ import (
 func TestSharedLock(t *testing.T) {
 	if !*withTiKV {
 		t.Skip("skipping TestSharedLock because with-tikv is not enabled")
+		return
+	}
+	if config.NextGen {
+		t.Skip("skipping TestSharedLock because next-gen doesn't support shared lock yet")
 		return
 	}
 	suite.Run(t, new(testSharedLockSuite))


txnkv/transaction/2pc.go
```diff
@@ -561,17 +571,22 @@ func (c *twoPhaseCommitter) initKeysAndMutations(ctx context.Context) error {
 	var err error
 	var assertionError error
 	toUpdatePrewriteOnly := make([][]byte, 0)
+	fflags := make([]kv.KeyFlags, 0, sizeHint)
+	hasValues := make([]bool, 0, sizeHint)
```
zyguan: Those two are used as log info only when no suitable primary key found, how about to build them as needed (in the `else if` branch) ?

txnkv/transaction/txn.go
zyguan, line 1: It seems [txn_test.go](https://github.com/zyguan/client-go/blob/0939dbe0e0a25e634a52a52db1322623db69d881/txnkv/transaction/txn_test.go) is removed by 36a8441ee6c5f33e34a41d0caae48c51a46bdc81 , may I known the reason?

you06: Good catch, I thought it was a conflict and removed it by mistake.
36a8441ee6c5f33e34a41d0caae48c51a46bdc81 reverted.

kv/keyflags.go
```diff
@@ -78,6 +78,12 @@ const (
 	// ends when the lock is acquired.
 	flagPreviousPresumeKNE
 
+	// flagKeyLockedInShareMode indicates the key is locked in share mode. The semantic of flagKeyLockedInShareMode is
+	// to indicate the lock mode. If this flag is set, then flagKeyLocked must also be set, indicating that the key has
+	// been locked in share mode. If only flagKeyLocked is set (and flagKeyLockedInShareMode is not), then it indicates
+	// that the key has been locked in exclusive mode.
+	flagKeyLockedInShareMode
```
cfzjywxk, line 85: Should this be a persistent flag? MemDB staging keeps only `AndPersistent()` flags when reverting values, would the 2PC convert the shared lock to exclusive lock type unexpectedly?

txnkv/transaction/test_util.go
```diff
@@ -0,0 +1,508 @@
...
+	"github.com/tikv/pd/client/clients/router"
+	"github.com/tikv/pd/client/clients/tso"
+	"github.com/tikv/pd/client/opt"
+	"github.com/tikv/pd/client/pkg/caller"
+	sd "github.com/tikv/pd/client/servicediscovery"
+)
+
+var _ oracle.Oracle = (*unimplementedOracle)(nil)
+
+type unimplementedOracle struct{}
```
cfzjywxk, line 46: For test purpose, should they be put into "*_test.go" file or test directory?

tikv/gc.go
```diff
@@ -288,9 +288,16 @@ func scanLocksInOneRegionWithRange(bo *retry.Backoffer, store Storage, startKey
 			return nil, loc, errors.Errorf("unexpected scanlock error: %s", locksResp)
 		}
 		locksInfo := locksResp.GetLocks()
-		locks = make([]*txnlock.Lock, len(locksInfo))
-		for i := range locksInfo {
-			locks[i] = txnlock.NewLock(locksInfo[i])
+		locks = make([]*txnlock.Lock, 0, len(locksInfo))
+		for _, lockInfo := range locksInfo {
+			if sharedLockInfos := lockInfo.GetSharedLockInfos(); sharedLockInfos != nil {
+				// expand shared lock into multiple locks, and drop the dummy wrapper lock.
```
cfzjywxk, line 294: After expanding, `len(locks)` no longer matches the server-side Limit, but `ResolveLocksForRange` uses
    `len(locks) < scanLimit` to decide region completion or advance. This can cause extra rescans and confusing.
If this is already considered, we may leave a TODO here.

integration_tests/shared_lock_test.go
```diff
@@ -0,0 +1,389 @@
...
+	suite.Run(t, new(testSharedLockSuite))
+}
+
+type testSharedLockSuite struct {
+	suite.Suite
+	store tikv.StoreProbe
+}
+
+func (s *testSharedLockSuite) SetupSuite() {
+	atomic.StoreUint64(&transaction.ManagedLockTTL, 3000) // 3s
```
cfzjywxk, line 53: They should be restored in `TearDownsuite`.

txnkv/transaction/test_util.go
you06: By now the `unimplementedOracle` is used in `txnkv/tranasaction/txn_test.go` only, but it's similar to those things in `test_probe.go`, providing the possibility of being imported by `*_test.go` files in other packages.

tikv/gc.go
you06: Better to limit the exact number of locks in TiKV side, I'll open a PR to compatible with the limit parameter of `ScanLock` RPC.

integration_tests/shared_lock_test.go
you06: I copied it from `integration_tests/2pc_test.go`, fixed the both.

tikv/gc.go
you06: See https://github.com/tikv/tikv/pull/19338

kv/keyflags.go
you06: Good catch, added `flagKeyLockedInShareMode` as a persistent flag.

> would the 2PC convert the shared lock to exclusive lock type unexpectedly?

No, the lock key process prevent this case strictly.

image

PR #1835, opened 2026-01-10
client: clean reference in priority queue

bufferflies:
ref #1834

### DIFF
--- a/internal/client/priority_queue.go
+++ b/internal/client/priority_queue.go
@@ -92,27 +92,32 @@ func (pq *PriorityQueue) pop() Item {
 	return e.(Item)
 }
 
-// Take returns the highest priority entries from the priority queue.
-// notice that the caller should ensure that call the clean method periodically to remove canceled entries.
-// to avoid the weak reference in the backing array.
-func (pq *PriorityQueue) Take(n int) ([]Item, func()) {
+// Take returns the highest priority entries from the priority queue and run the provided function on them.
+// ref https://github.com/tikv/client-go/issues/1834
+// Take retrieves and removes the first n items from the priority queue and passes them to the callback function.
+// If n is less than or equal to 0, the function returns without performing any operation.
+// If n is greater than or equal to the queue length, all items are removed and passed to the callback.
+// Otherwise, exactly n items are popped from the queue and passed to the callback.
+// The callback function fn is responsible for processing the items; it receives a slice of Items.
+// Ref issue: https://github.com/tikv/client-go/issues/1834
+func (pq *PriorityQueue) Take(n int, cb func([]Item)) {
 	if n <= 0 {
-		return nil, func() {}
+		return
 	}
 	if n >= pq.Len() {
 		ret := pq.ps
 		pq.ps = pq.ps[:0]
-		return ret, func() {
-			for i := range ret {
-				ret[i] = nil
-			}
+		cb(ret)
+		for i := range ret {
+			ret[i] = nil
 		}
+		return
 	} else {
 		ret := make([]Item, n)
 		for i := 0; i < n; i++ {
 			ret[i] = pq.pop()
 		}
-		return ret, func() {}
+		cb(ret)
 	}
 
 }
--- a/internal/client/priority_queue_test.go
+++ b/internal/client/priority_queue_test.go
@@ -15,7 +15,6 @@
 package client
 
 import (
-	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -45,27 +44,29 @@ func TestPriority(t *testing.T) {
 		re.Equal(uint64(5), aq.highestPriority())
 		aq.clean()
 		re.Equal(5, aq.Len())
+		arr := make([]Item, 0)
+		collect := func(items []Item) {
+			arr = arr[:0]
+			arr = append(arr, items...)
+		}
 
-		arr, fn := aq.Take(1)
+		aq.Take(1, collect)
 		re.Len(arr, 1)
 		re.Equal(uint64(5), arr[0].priority())
 		re.Equal(uint64(4), aq.highestPriority())
-		fn()
 
-		arr, fn = aq.Take(2)
+		aq.Take(2, collect)
 		re.Len(arr, 2)
 		re.Equal(uint64(4), arr[0].priority())
 		re.Equal(uint64(3), arr[1].priority())
 		re.Equal(uint64(2), aq.highestPriority())
-		fn()
 
-		arr, fn = aq.Take(5)
+		aq.Take(5, collect)
 		re.Len(arr, 2)
 		re.Equal(uint64(2), arr[0].priority())
 		re.Equal(uint64(1), arr[1].priority())
 		re.Equal(uint64(0), aq.highestPriority())
 		re.Equal(0, aq.Len())
-		fn()
 
 		aq.Push(&FakeItem{value: 1, pri: 1, canceled: true})
 		re.Equal(1, aq.Len())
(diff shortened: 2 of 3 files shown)


internal/client/priority_queue.go
```diff
@@ -127,14 +128,20 @@ func (pq *PriorityQueue) all() []Item {
 }
 
 // clean removes all canceled entries from the priority queue.
-func (pq *PriorityQueue) clean() {
+func (pq *PriorityQueue) Clean() {
```
JmPotato: Why make it exported?

internal/client/priority_queue_test.go
```diff
@@ -65,9 +66,67 @@ func TestPriority(t *testing.T) {
...
 }
+
+func TestPriorityQueuePopLeavesReferenceInBackingArray(t *testing.T) {
+	pq := NewPriorityQueue()
+	it := &FakeItem{pri: 1}
+	pq.Push(it)
+
+	_ = pq.pop()
+	if pq.Len() != 0 {
+		t.Fatalf("expected empty queue, got len=%d", pq.Len())
```
JmPotato: Use `re.Empty` directly?

internal/client/priority_queue_test.go
```diff
@@ -65,9 +66,67 @@ func TestPriority(t *testing.T) {
...
+	}
+
+	// Expand to full capacity to inspect the backing array.
+	backing := pq.ps[:cap(pq.ps)]
+	// If the removed slot is not cleared, it stays non-nil and retains the item.
```
JmPotato: Ditto. `re.Nil` should be more straightforward.

internal/client/priority_queue_test.go
```diff
@@ -65,9 +66,67 @@ func TestPriority(t *testing.T) {
...
+	item := pq.pop()
+	re.Len(pq.ps, 2)
+	re.NotNil(item)
+	re.False(checkReferences(), "expected no references in backing array yet")
+
+	// Take all items without clean, the references remain in the backing array.
+	_ = pq.Take(pq.Len())
+	re.Len(pq.ps, 0)
+	if pq.Len() != 0 {
+		t.Fatalf("expected empty queue, got len=%d", pq.Len())
```
JmPotato: Ditto.

internal/client/priority_queue.go
```diff
@@ -127,14 +128,20 @@ func (pq *PriorityQueue) all() []Item {
...
 	for i := 0; i < pq.Len(); {
 		if pq.ps[i].isCanceled() {
 			heap.Remove(&pq.ps, i)
 			continue
 		}
 		i++
 	}
+
+	// clean up the references in the backing array
+	cleanQueue := pq.ps[len(pq.ps):cap(pq.ps)]
```
rleungx: We need to ensure that the clean must be called after take. Will the hypothesis be too strong?

internal/client/priority_queue.go
```diff
@@ -127,14 +128,20 @@ func (pq *PriorityQueue) all() []Item {
...
+	for i := range cleanQueue {
+		cleanQueue[i] = nil
+	}
```
zyguan: Walking through `ps[len:cap]` for every `batchCommandsBuilder.reset` doesn’t look very efficient. How about to let `Take` returns a callback for free these items. eg.

```go
func (pq *PriorityQueue) Take(n int) ([]Item, func()) {
	if n <= 0 {
		return nil, func() {}
	}
	if n >= pq.Len() {
		ret := pq.ps
		pq.ps = pq.ps[:0]
		return ret, func() {
			for i := range ret {
				ret[i] = nil
			}
		}
	} else {
		ret := make([]Item, n)
		for i := 0; i < n; i++ {
			ret[i] = pq.pop()
		}
		return ret, func() {}
	}
}
```

Then the build process can be:
```go
	for (count < limit && b.entries.Len() > 0) || b.hasHighPriorityTask() {
		n := limit
		if limit == 0 {
			n = 1
		}
		reqs, free := b.entries.Take(int(n))
		if len(reqs) == 0 {
			break
		}
		build(reqs)
		free()
	}
```

bufferflies: good idea

bufferflies: done

internal/client/priority_queue.go
bufferflies: fix by callback function

internal/client/priority_queue_test.go
bufferflies: done

internal/client/priority_queue_test.go
bufferflies: done

internal/client/priority_queue_test.go
bufferflies: done

internal/client/priority_queue.go
bufferflies: done

internal/client/priority_queue.go
```diff
@@ -92,20 +93,26 @@ func (pq *PriorityQueue) pop() Item {
 }
 
 // Take returns the highest priority entries from the priority queue.
-func (pq *PriorityQueue) Take(n int) []Item {
+// notice that the caller should ensure that call the clean method periodically to remove canceled entries.
+// to avoid the weak reference in the backing array.
+func (pq *PriorityQueue) Take(n int) ([]Item, func()) {
```
lcwangchao: This method is a little weird for the release function. Could refactor it to:

```
func (pq *PriorityQueue) Take(n int, reserved []Item) []Item {
   ...
}
```

The caller should responsible for the reuse of the reserved slice and clear the internal refs after used.

Another way to implement it is:

```
func (pq *PriorityQueue) TakeAndProcess(n int, fn func([]Item)) {
   ...
}
```

internal/client/priority_queue.go
```diff
@@ -58,6 +58,7 @@ func (ps *prioritySlice) Pop() interface{} {
 	old := *ps
```
lcwangchao, line 58: Does `clean()` also need to set the nil for cancelled requests?

lcwangchao:
@bufferflies 

Please ref #1834 with this PR

internal/client/priority_queue.go
```diff
@@ -92,20 +93,26 @@ func (pq *PriorityQueue) pop() Item {
 }
 
 // Take returns the highest priority entries from the priority queue.
-func (pq *PriorityQueue) Take(n int) []Item {
+// notice that the caller should ensure that call the clean method periodically to remove canceled entries.
+// to avoid the weak reference in the backing array.
```
cfzjywxk: The "weak reference" comment is misleading, better explain more about the memory re-use mechanism in `Take` function and its backing-array retention/cleanup design.


Also link the related issue https://github.com/tikv/client-go/issues/1834 here in the comments to note the impact.

internal/client/priority_queue.go
cfzjywxk: Could you explain more specific? The signiture looks weird too IMO.

lcwangchao: - Take(n int, reserved []Item) []Item

Take n items and returns the copied items. If the input `cap(reserved) >= n`, copy the items to the reserved slice, otherwise alloc a new slice and copy the items to it. The caller can reuse `reserved` to reduce the frequency allocation overheads. It is simliar with some methods in TiDB like `EncodeBytes`:

https://github.com/pingcap/tidb/blob/69a1b2ce8ec7cb8e82722dd4a29bfd76350be919/pkg/util/codec/bytes.go#L50-L73

- TakeAndProcess(n int, fn func([]Item))

This is another approach. It means take n items and call `fn` to process them. In this way, the taken positions will be set to `nil` automatically in `TakeAndProcess` after `fn` is called. The caller does not to care about slice reuse, free memory, etc..

zyguan: IMO, take a `func([]Item)` as an arg is also error-prone. Callers may take a reference to the slice and do not know `TakeAndProcess` will set the slice items to nil. `TakeAndProcess(n int, fn func(Item))` looks better.

internal/client/priority_queue.go
bufferflies: No, /usr/local/go/src/container/heap/heap.go will call the pop function.

```
// Remove removes and returns the element at index i from the heap.
// The complexity is O(log n) where n = h.Len().
func Remove(h Interface, i int) any {
	n := h.Len() - 1
	if n != i {
		h.Swap(i, n)
		if !down(h, i, n) {
			up(h, i)
		}
	}
	return h.Pop()
}```

internal/client/priority_queue.go
bufferflies: has renamed as func (pq *PriorityQueue) Take(n int, cb func([]Item)), not return the clean function now

internal/client/priority_queue.go
bufferflies: has renamed as func (pq *PriorityQueue) Take(n int, cb func([]Item))

internal/client/priority_queue.go
bufferflies: I added the clean weak check in the test also?

PR #1837, opened 2026-01-12
client: break loop if entry is nil

bufferflies:
ref code:

```
reqs := b.entries.Take(int(n))
if len(reqs) == 0 {
	break
}
build(reqs)
```

## Summary by CodeRabbit

* **Refactor**
  * Improved internal batch processing to detect empty batches and exit earlier, reducing unnecessary work and improving performance while preserving public interfaces, behavior, and error handling.


rleungx:
@coderabbitai review

internal/client/client_batch.go
```diff
@@ -254,6 +255,7 @@ func (b *batchCommandsBuilder) buildWithLimit(limit int64, collect func(id uint6
 				batchReq.Requests = append(batchReq.Requests, e.req)
 			}
 			b.idAlloc++
+			exhaust = len(reqs) == 0
```
lcwangchao, line 258: exhaust should always be false? Because it is the loop to iter reqs

internal/client/client_batch.go
```diff
@@ -262,6 +264,9 @@ func (b *batchCommandsBuilder) buildWithLimit(limit int64, collect func(id uint6
 			n = 1
 		}
 		b.entries.Take(int(n), build)
+		if exhaust {
```
lcwangchao, line 267: Why we need exhaust?

internal/client/client_batch.go
bufferflies: yes, you're right, the b.entries.Len() > 0 can guard the loop will break

bufferflies:
no need,

Issue #1839, opened 2026-01-14
client/client_batch: Potential data race on shared pointers in  batchCommandsBuilder


AndreMouche:
There appears to be a potential data race risk related to pointer usage in client_batch.go.

Specifically, the following fields are accessed without any explicit synchronization or protection:
https://github.com/tikv/client-go/blob/01cdf70d1ad66f0c870410abb869ee2867c08da7/internal/client/client_batch.go#L115-L128 
At the same time, the code attempts to avoid concurrent execution in the following sections:
https://github.com/tikv/client-go/blob/01cdf70d1ad66f0c870410abb869ee2867c08da7/internal/client/client_batch.go#L208-L225 

https://github.com/tikv/client-go/blob/01cdf70d1ad66f0c870410abb869ee2867c08da7/internal/client/client_batch.go#L152C32-L152C46 

However, it seems that these checks do not fully prevent concurrent access to the underlying data structures.
In particular, the following shared fields appear to be vulnerable to concurrent read/write access:
``` 
entries    *PriorityQueue 
requests   []*tikvpb.BatchCommandsRequest_Request 
```

Since these pointers (and the data they reference) are mutated across different code paths without clear locking or ownership guarantees, a data race may occur under concurrent execution. 

## related PR：
 this PR(https://github.com/tikv/client-go/pull/1242) does not fully fix the issue reported in: https://github.com/pingcap/tidb/issues/51921 , We are still encountering a similar issue even after this PR was merged, which suggests that the underlying concurrency problem may remain unresolved.

PR #1841, opened 2026-01-14
client: pre-encode request before sending to batch-send-loop

zyguan:
This PR tries to prevent issues like #1839 by pre-encoding request.

The micro-bench results show that the overhead introduced by this PR is negligible. Besides, we can reduce the mem allocs of `AttachContext` after this PR.

workload | threads | pre-encode | baseline | diff
-- | -- | -- | -- | --
pointget | 64 | 83120.35 | 83940.81 | -0.98%
batchget | 64 | 24469.9 | 24629.87 | -0.65%
tblscan | 64 | 41522.92 | 41565.02 | -0.10%
idxscan | 64 | 40363.22 | 40328.56 | 0.09%
idxlookup | 64 | 16482.58 | 16497.08 | -0.09%

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for batch request encoding with improved validation of request states.

* **Chores**
  * Optimized batch request processing through object pooling to reduce memory allocations and improve efficiency.
  * Refined request lifecycle management with better cleanup and state tracking.

* **Tests**
  * Added comprehensive test coverage for batch encoding, pooling reuse, and concurrent operations to ensure reliability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/internal/client/client_batch.go
+++ b/internal/client/client_batch.go
@@ -65,11 +65,7 @@ import (
 )
 
 // globalEncodedMsgDataPool is used to pool pre-encoded message data for batch commands.
-var globalEncodedMsgDataPool grpc.SharedBufferPool
-
-func init() {
-	globalEncodedMsgDataPool = grpc.NewSharedBufferPool()
-}
+var globalEncodedMsgDataPool = grpc.NewSharedBufferPool()
 
 type encodedBatchCmd struct {
 	// implement isBatchCommandsRequest_Request_Cmd
@@ -81,6 +77,8 @@ type encodedBatchCmd struct {
 func (p *encodedBatchCmd) MarshalTo(data []byte) (int, error) {
 	if p.data == nil {
 		return 0, errors.New("message data has been released")
+	} else if len(data) < len(*p.data) {
+		return 0, errors.New("no enough space to marshal message")
 	}
 	n := copy(data, *p.data)
 	return n, nil
@@ -93,7 +91,14 @@ func (p *encodedBatchCmd) Size() int {
 	return len(*p.data)
 }
 
-// encodeRequestCmd encodes the `req.Cmd` into a `preparedBatchCmd` and updates the `req.Cmd` to it in place.
+// encodeRequestCmd encodes the `req.Cmd` into a `encodedBatchCmd` and updates the `req.Cmd` to it in place.
+//
+// SAFETY: This function is called just after `tikvrpc.Request.ToBatchCommandsRequest()` and before constructing
+// `batchCommandsEntry`. Note that the `req` argument is the output of `ToBatchCommandsRequest`, which is a new object
+// of `tikvpb.BatchCommandsRequest_Request`. So `req.Cmd` can be only modified inside the batch client by
+// `reuseRequestData` after the `req` has been sent. So there should be no data race on `req.Cmd` via public API
+// (`SendRequest` and `SendRequestAsync`). When writing unit tests, please make sure it's only called once for one `req`
+// (do NOT reuse `req`).
 func encodeRequestCmd(req *tikvpb.BatchCommandsRequest_Request) error {
 	if req.Cmd == nil {
 		// req.Cmd might be nil in unit tests.
--- a/internal/client/client_batch_test.go
+++ b/internal/client/client_batch_test.go
@@ -51,9 +51,9 @@ func TestEncodeRequestCmd_Basic(t *testing.T) {
 	err := encodeRequestCmd(batch1.Requests[0])
 	require.NoError(t, err)
 
-	// Verify req.Cmd is now a preparedBatchCmd
+	// Verify req.Cmd is now a encodedBatchCmd
 	_, ok := batch1.Requests[0].Cmd.(*encodedBatchCmd)
-	require.True(t, ok, "req.Cmd should be converted to *preparedBatchCmd")
+	require.True(t, ok, "req.Cmd should be converted to *encodedBatchCmd")
 
 	// Verify two batches have the same size after encoding
 	require.Equal(t, batch1.Size(), batch2.Size())


internal/client/client_batch.go
```diff
@@ -64,6 +64,74 @@ import (
...
+		return nil
+	}
+	if encoded, ok := req.Cmd.(*encodedBatchCmd); ok {
+		if encoded.data == nil {
+			return errors.New("request has already been encoded and sent")
+		}
+		return nil
+	}
+	data := globalEncodedMsgDataPool.Get(req.Cmd.Size())
+	n, err := req.Cmd.MarshalTo(data)
```
AndreMouche, line 114: is it possible that `req.Cmd` could being modified concurrently by other goroutines?
maybe we should add some comments for this function for example : 
// encodeRequestCmd encodes req.Cmd. The caller must ensure that req is not
// being modified concurrently by other goroutines.

zyguan: No, `encodeRequestCmd` is used in batch client internally. It's called just after `tikvrpc.Request.ToBatchCommandsRequest()` and before constructing `batchCommandsEntry`. Note that the `req` here is the output of `ToBatchCommandsRequest`, which is a new object of `tikvpb.BatchCommandsRequest_Request`. So `req.Cmd` can be only modified inside the batch client by `reuseRequestData` after the `req` has been sent. No data race on `req.Cmd` via public API (`SendRequest` and `SendRequestAsync`).

internal/client/client_async.go
```diff
@@ -82,6 +82,11 @@ func (c *RPCClient) SendRequestAsync(ctx context.Context, addr string, req *tikv
 		return
 	}
 
+	if err := encodeRequestCmd(batchReq); err != nil {
+		cb.Invoke(nil, err)
+		return
```
cfzjywxk, line 87: It can early-return via `cb.Invoke` before `entry.cb.Inject(...)` is installed at L105, would it be a problem that regionRPC.End()/spanRPC.Finish() won’t run on this path or should we consider move encoding after the inject setup?

internal/client/client_batch.go
```diff
@@ -64,6 +64,74 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+// globalEncodedMsgDataPool is used to pool pre-encoded message data for batch commands.
+var globalEncodedMsgDataPool grpc.SharedBufferPool
+
+func init() {
+	globalEncodedMsgDataPool = grpc.NewSharedBufferPool()
```
cfzjywxk: Is this a experimental API?

internal/client/client_batch.go
```diff
@@ -64,6 +64,74 @@ import (
...
+}
```
cfzjywxk: `var globalEncodedMsgDataPool = grpc.NewSharedBufferPool()` may help to avoid any init-order surprises.

internal/client/client_batch.go
```diff
@@ -64,6 +64,74 @@ import (
...
+
+type encodedBatchCmd struct {
```
cfzjywxk, line 70: How about implementing `isBatchCommandsRequest_Request_Cmd` directly to avoid embedding.

internal/client/client_batch.go
```diff
@@ -64,6 +64,74 @@ import (
...
+	tikvpb.BatchCommandsRequest_Request_Empty
+	// pre-encoded message data
+	data *[]byte
+}
+
+func (p *encodedBatchCmd) MarshalTo(data []byte) (int, error) {
+	if p.data == nil {
+		return 0, errors.New("message data has been released")
+	}
+	n := copy(data, *p.data)
```
cfzjywxk, line 83: Perhaps adding `asserting len(dst) >= len(src);` here .

internal/client/client_batch.go
```diff
@@ -64,6 +64,74 @@ import (
...
+}
+
+func (p *encodedBatchCmd) Size() int {
+	if p.data == nil {
+		return 0
+	}
+	return len(*p.data)
+}
+
+// encodeRequestCmd encodes the `req.Cmd` into a `preparedBatchCmd` and updates the `req.Cmd` to it in place.
```
cfzjywxk: Better to add short comment/doc not to prevent future re-use regressions.

internal/client/client_batch_test.go
```diff
@@ -0,0 +1,221 @@
...
+			Key: []byte("test-key"),
+		},
+	}
+	batch1 := &tikvpb.BatchCommandsRequest{Requests: []*tikvpb.BatchCommandsRequest_Request{{Cmd: cmd}}}
+	batch2 := &tikvpb.BatchCommandsRequest{Requests: []*tikvpb.BatchCommandsRequest_Request{{Cmd: cmd}}}
+
+	err := encodeRequestCmd(batch1.Requests[0])
+	require.NoError(t, err)
+
+	// Verify req.Cmd is now a preparedBatchCmd
```
cfzjywxk:
```suggestion
	// Verify req.Cmd is now a encodedBatchCmd
```

internal/client/client_async.go
zyguan: I think it's OK, if it early-returned, there is no RPC actually, besides, this function shouldn't return error in most of cases.

internal/client/client_batch.go
zyguan: Yes, and it has been moved to [mem](https://github.com/grpc/grpc-go/blob/master/mem/buffer_pool.go) package in new version.

internal/client/client_batch.go
zyguan: Unfortunately, you cannot implement an unexported interface method from a different package in golang.

zyguan: We may implement it in kvproto later.

PR #1871, opened 2026-02-05
client: support versioned coprocessor routing

wshwsh12:
- route version-aware cop to VersionedKv
- add versioned cop cmd types in tikvrpc

### DIFF
--- a/internal/apicodec/codec_v2.go
+++ b/internal/apicodec/codec_v2.go
@@ -295,6 +295,9 @@ func (c *codecV2) EncodeRequest(req *tikvrpc.Request) (*tikvrpc.Request, error) 
 		r := *req.SplitRegion()
 		r.SplitKeys = c.encodeKeys(r.SplitKeys)
 		req.Req = &r
+	case tikvrpc.CmdVersionedCop:
+		// TODO: Support VersionedCop in APIv2.
+		return nil, errors.New("codecV2 does not support VersionedCop request")
 	}
 
 	return req, nil


tikvrpc/tikvrpc.go
```diff
@@ -192,6 +195,8 @@ func (t CmdType) String() string {
 		return "CopStream"
 	case CmdBatchCop:
 		return "BatchCop"
+	case CmdVersionedCop:
```
ChangRui-Ryan, line 196: Suggest putting `CmdVersionedCop` right after `CmdCop` wherever it appears to keep consistency, except `CmdVersionedCop` is at the end of the `CmdType` enum.

ChangRui-Ryan:
lgtm

internal/client/client.go
```diff
@@ -371,12 +371,20 @@ func (c *RPCClient) sendRequest(ctx context.Context, addr string, req *tikvrpc.R
...
 
-	client := tikvpb.NewTikvClient(clientConn)
-
 	// Set metadata for request forwarding. Needn't forward DebugReq.
 	if req.ForwardedHost != "" {
 		ctx = metadata.AppendToOutgoingContext(ctx, forwardMetadataKey, req.ForwardedHost)
 	}
+
+	// Version-aware coprocessor requests should go through VersionedKv services.
+	if req.Type == tikvrpc.CmdVersionedCop {
```
cfzjywxk, line 380: `BatchReq` would be used by default at L354, I suppose this may not be reached, have you executed E2E test for this PR?

wshwsh12: "ToBatchCommandsRequest" can't converts CmdVersionedCop to BatchReq and return nil. So now CmdVersionedCop always reached this path. 
And I have run some sqls in my local and it works.

you06:
`isReadReq` needs to be updated.

https://github.com/tikv/client-go/blob/6e80a225f868f4447c82ed5bc74fcce720df3b23/internal/locate/replica_selector.go#L605-L613

internal/client/client.go
you06: Here are some performance risks. How often is this CmdVersionedCop called? If the frequency is high, I think it's necessary to modify kvproto to include it in the BatchReq and BatchResp.

wshwsh12: Currently, this feature does not have high performance requirements, and RPC requests are not the performance bottleneck. We can optimize it later if performance becomes an issue.”

wshwsh12:
> Should we handle `CmdVersionedCop` in codec v2 ?

My understanding is that only keyspace requires codec v2. Currently, VersionedCop has not been implemented for keyspace. When I need to implement VersionedCop for keyspace, I will add the corresponding support in client-go.

zyguan:
> My understanding is that only keyspace requires codec v2. Currently, VersionedCop has not been implemented for keyspace. When I need to implement VersionedCop for keyspace, I will add the corresponding support in client-go.

So, how about to add the following code to codec v2?
```go
	case tikvrpc.CmdVersionedCop:
		// TODO: ...
		return nil, errors.New("...")
```

you06 (approved):

PR #1769, opened 2025-10-02
retry on leader for region not available

mittalrishabh:

PR #1891, opened 2026-02-11
fast reload region on no-region-found error

mittalrishabh:
issue https://github.com/tikv/client-go/issues/1892

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache invalidation to mark stale entries for asynchronous reload, leading to more reliable and timely data refreshes.
* **Tests**
  * Updated tests to reflect the adjusted delayed-reload signaling and ensure correct reload behavior on subsequent access.

### DIFF
--- a/internal/locate/region_request_state_test.go
+++ b/internal/locate/region_request_state_test.go
@@ -543,7 +543,7 @@ func testStaleRead(s *testRegionCacheStaleReadSuite, r *RegionCacheTestCase, zon
 			return
 		}
 
-		s.Equal(*asyncReload, region.checkSyncFlags(needDelayedReloadPending | needDelayedReloadReady))
+		s.Equal(*asyncReload, region.checkSyncFlags(needDelayedReloadPending|needDelayedReloadReady))
 	}()
 
 	bo := retry.NewBackoffer(ctx, -1)


internal/locate/region_cache.go
```diff
@@ -2045,12 +2045,16 @@ func (c *RegionCache) BatchLoadRegionsFromKey(bo *retry.Backoffer, startKey []by
...
+// without invalidating it, so in-flight retries can still proceed.
+// It sets needDelayedReloadReady directly, skipping the GC promotion from
+// needDelayedReloadPending for faster reload.
 func (c *RegionCache) AsyncInvalidateCachedRegion(id RegionVerID) {
 	cachedRegion := c.GetCachedRegionWithRLock(id)
 	if cachedRegion == nil {
 		return
 	}
-	cachedRegion.setSyncFlags(needDelayedReloadPending)
+	cachedRegion.setSyncFlags(needDelayedReloadReady)
```
AndreMouche, line 2057: Great job! It sets needDelayedReloadReady (not needDelayedReloadPending) for two reasons:
1) Region split: The leader has already split and reported the new region to PD,but a follower may not have applied the split yet. TiDB can get RegionNotFoundwhen sending to that follower. We retry with the leader and mark the cache for reload. Using Ready ensures the next access reloads from PD immediately and gets the up-to-date region (range, version, peers). Using Pending would delay reload until the GC round runs (1s interval, and only ~50 regions per round), so we would keep using stale region info longer and risk more KeyNotInRegion or wrong  routing.
2) Data migrated away: When the region (or its replica) has actually been moved off this store, we need fresh peer list from PD so we stop sending to the old store. Ready triggers reload on the very next access; Pending would again wait for GC, prolonging failed attempts to the old store.

AndreMouche:
@cfzjywxk @MyonKeminta PTAL

internal/locate/region_cache.go
zyguan: <del>So why not use `needReloadOnAccess`? Set `needDelayedReloadReady` direclty is not a good idea.</del>

The delayed reload mechanism was firstly introduced by #843 , @you06 do you remember why we do reload in two phases (mark and reload)?

zyguan: Oh, I see `needReloadOnAccess` will force invalidate the old region.

zyguan: I guess the purpose of two-phase reloading is for batching.

Before this PR:
- Multiple `AsyncInvalidateCachedRegion` calls => all set `needDelayedReloadPending`
- GC scans (every cleanCacheInterval = 1 second) => batches them all to `needDelayedReloadReady`
- First access after GC => single batch reload (the rate of region reloads is limited by GC frequency)

After this PR:
- Multiple `AsyncInvalidateCachedRegion` calls => all set `needDelayedReloadReady` immediately
- Each access => triggers individual PD request (the region might be marked as `needDelayedReloadReady` repeatedly)

mittalrishabh: >>Oh, I see needReloadOnAccess will force invalidate the old region.

yes, we don't want to fail the retry

>>Each access => triggers individual PD request (the region might be marked as needDelayedReloadReady

region is NULL, we should try reloading the region in next access. It was doing InvalidateCachedRegion before i changed it to AsyncInvalidateCachedRegion in https://github.com/tikv/client-go/pull/1769/changes . So it doesn't regress the overall behavior now. Behavior was reverted back to what is was before this [change](https://github.com/tikv/client-go/pull/1769/changes)

you06 (approved):

you06:
Test fail in `internal/locate/region_request_state_test.go:L546`, to fix it:

```diff
-    s.Equal(*asyncReload, region.checkSyncFlags(needReloadOnAccess))
+    s.Equal(*asyncReload, region.checkSyncFlags(needDelayedReloadPending|needDelayedReloadReady))
```

PR #1911, opened 2026-03-12
tikv,txnkv: disable async commit and 1pc when tso-max-index > 1

lcwangchao:
ref: https://github.com/pingcap/tidb/issues/64281

### Problem Summary

In active-active mode, PD may allocate discontinuous TSOs when `tso-max-index > 1`. In that case, commit timestamps fetched from PD are still valid, but async commit and 1PC may derive `commit_ts` from `maxReadTS`, which can violate the TSO slot constraint and generate invalid commit timestamps.

### What changed and how does it work?

- read `tso-max-index` from PD when `KVStore` starts
- mark the store as not supporting async commit / 1PC when `tso-max-index > 1`
- gate transaction defaults, `SetEnableAsyncCommit`, `SetEnable1PC`, and committer `checkAsyncCommit` / `checkOnePC` with the store capability
- add tests covering PD config parsing, store capability initialization, and transaction-side gating

### Check List

- [x] Unit test
- [ ] Integration test

### Release note

```release-note
Disable async commit and 1PC automatically when PD `tso-max-index` is greater than 1.
```

## Summary by CodeRabbit

* **New Features**
  * App now detects and respects store/cluster capability for async commit and 1PC, enabling or disabling those paths automatically.

* **Bug Fixes**
  * Improved mock/client fallbacks to avoid nil-call issues during tests.

* **Tests**
  * Added coverage validating async-commit/1PC enablement behavior across configurations.

* **Chores**
  * CI workflows extended to include feature/* branches.
  * Updated dependency versions.

### DIFF
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -47,6 +47,7 @@ import (
 	"time"
 
 	"github.com/opentracing/opentracing-go"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/pkg/errors"
 	"github.com/tikv/client-go/v2/config"
@@ -316,14 +317,16 @@ func NewKVStore(
 			return nil, errors.WithStack(err)
 		}
 	}
+
+	if err = store.checkPDConfig(); err != nil {
+		return nil, err
+	}
+
 	store.clientMu.client = client.NewReqCollapse(client.NewInterceptedClient(tikvclient))
 	store.clientMu.client.SetEventListener(regionCache.GetClientEventListener())
 
 	store.lockResolver = txnlock.NewLockResolver(store)
 	loadOption(store, opt...)
-	if err = store.checkPDConfig(); err != nil {
-		return nil, err
-	}
 
 	store.wg.Add(2)
 	go store.runSafePointChecker()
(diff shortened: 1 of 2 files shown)


tikv/kv.go
```diff
@@ -315,6 +321,9 @@ func NewKVStore(
 
 	store.lockResolver = txnlock.NewLockResolver(store)
 	loadOption(store, opt...)
+	if err = store.checkPDConfig(); err != nil {
```
bufferflies: If the PD config changed and the client is restarted, can this client check this config again?

lcwangchao: I think it's a more complicated scene and not only related to "auto check PD config change". If `tso-max-index` changes from `1` to `2` (or a greater value), this means the clusters should change to an active-active sync mode, we should consider more, for example, how to keep two clusters beginning with the same snapshot before the changefeed is created. We still need some manual operations.

Currently, we can rolling upgrade TiDB manually to reload PD's config on the TiDB side for simplicity. We can have a further discussion about the actual requirement about how to upgrade exist cluster to active-active mode.

lcwangchao:
@coderabbitai review this PR

lcwangchao:
@coderabbitai please review again

tikv/kv.go
```diff
@@ -323,6 +332,83 @@ func NewKVStore(
...
+
+func (s *KVStore) checkPDConfig() error {
+	var cfg map[string]any
+	if _, err := util.EvalFailpoint("mockPDConfigToCheck"); err == nil {
+		cfg = mockedPDConfigToCheck
+	} else {
+		discovery := s.pdClient.GetServiceDiscovery()
+		if discovery == nil || len(discovery.GetServiceURLs()) == 0 {
+			// This only happens in the test environment, and we can skip the check in this case.
+			return nil
```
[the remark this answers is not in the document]
lcwangchao: I prefer not to reuse s.pdHttpClient because it can be nil in some scenarios (for example, during IMPORT INTO). This leads to different code paths that are not visible when TiDB starts.

Having a unified code path for checkPDConfig helps reduce such inconsistencies and makes potential errors easier to detect.

PR #1912, opened 2026-03-16
*: disable async commit in production to make sure the right commit ts of active-active replication

lcwangchao:
## Summary

ref: https://github.com/pingcap/tidb/issues/64281

This PR makes `client-go` safer for active-active replication in production, while keeping the existing async commit / 1PC test coverage intact.

1. Revert the PD dependency change in `go.mod` so the behavior stays compatible with TiDB/TiKV 8.5.
   This is aligned with the discussion in PD issue [#10427](https://github.com/tikv/pd/issues/10427), where PD is moving toward ensuring TSO uniqueness across clusters via suffix bits rather than the previous local/global TSO behavior.

2. Force disable async commit and 1PC in production.
   The reason is that in active-active scenarios, async commit and 1PC may derive commit TS from `maxReadTS + 1` instead of getting it directly from PD, which can violate the PD `tso-unique-index` constraint and potentially produce invalid commit TS behavior across clusters.

3. Keep async commit and 1PC available in test environments.
   Existing async commit / 1PC tests still need to verify those code paths, so this PR adds a test-only switch to re-enable them when building test stores.

4. Add coverage for the default behavior.
   This PR adds tests to verify that by default async commit and 1PC are disabled, even if a transaction explicitly calls `SetEnableAsyncCommit(true)` or `SetEnable1PC(true)`.

## Summary by CodeRabbit

* **Chores**
  * CI workflows now run on feature/* branches.
  * PD client dependency version updated across modules.

* **New Features**
  * Added a runtime toggle to disable active-active commit support for test scenarios and a probe to inspect one-phase commit behavior.

* **Tests**
  * New and extended integration tests to validate default commit protocol behavior and related scenarios.

### DIFF
--- a/integration_tests/2pc_test.go
+++ b/integration_tests/2pc_test.go
@@ -99,7 +99,7 @@ func (s *testCommitterSuite) SetupTest() {
 	pdCli := tikv.NewCodecPDClient(tikv.ModeTxn, pdClient)
 	spkv := tikv.NewMockSafePointKV()
 	store, err := tikv.NewKVStore("mocktikv-store", pdCli, spkv, client,
-		// To make sure some exists tests with async-commit or 1pc commit pass
+		// To make sure some existing tests with async-commit or 1pc commit pass
 		tikv.WithDisableActiveActiveCommitSupportForTest(),
 	)
 	store.EnableTxnLocalLatches(8096)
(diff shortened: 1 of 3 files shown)


lcwangchao:
@coderabbitai review this PR

lcwangchao:
@coderabbitai review this PR again

PR #1913, opened 2026-03-16
rawkv: Implement CompareAndDelete with integration and unit tests

healthwaite:
[#19382](https://github.com/tikv/tikv/pull/19382) extends the rawkv CompareAndSwap API to allow for atomic key deletions. It allows users to achieve atomic compare-and-delete semantics. This can be used in conjunction with compare-and-swap to safely insert/update/delete keys when using the rawkv API.

This PR adds support for compare and delete to the client. We do this by introducing a new method `CompareAndDelete` in `rawkv`. This method piggy-backs on the existing `CompareAndSwap` code path but sets the [delete](https://github.com/pingcap/kvproto/blob/e3b325474572ee1dcadd5c64a6a0b5719fc2eba2/proto/kvrpcpb.proto#L1556) field in the `RawCASRequest` proto message to signal we wish to delete the key.

* update `github.com/pingcap/kvproto` to the latest commit
* add integration tests
* add unit tests

## Release Notes

* **New Features**
  * Added atomic Compare-and-Delete operation for raw key-value operations, enabling conditional deletion based on expected values.

* **Tests**
  * Added comprehensive tests for Compare-and-Delete functionality.
  * Extended compare-and-swap tests with additional scenarios.

* **Chores**
  * Updated Go module dependencies to latest versions.

### DIFF
--- a/rawkv/rawkv.go
+++ b/rawkv/rawkv.go
@@ -697,7 +697,7 @@ func (c *Client) CompareAndSwap(ctx context.Context, key, previousValue, newValu
 // If the value retrieved is equal to previousValue, the key is deleted.
 // It returns the previous value and whether the key was successfully deleted.
 //
-//	previousValue  | key exists    | result
+//	previousValue   | key exists    | result
 //	----------------|---------------|----------------------------------
 //	nil             | no            | success, returns nil
 //	nil             | yes           | failure, returns existing value


healthwaite:
There is a versioning problem here. If a newer version of the client is talking to an older version of TiKV then this API will cause problems. The client will call `CompareAndDelete` which simply sends a `RawCASRequest`protobuf message with the `delete` field populated. If the server is old it doesn't know about the `delete` flag and will process the message as a normal compare-and-swap. This will result in the server _storing_ the defaulted `Value` for `RawCASRequest`, i.e. a nil. The client erroneously thinks they have deleted the key.

I'd like to know if anyone has any proposal for how to work around this. One idea is to mandate in the CompareAndDelete function documentation that tikv must be version x.y.z, but this is pretty brittle. Another option is to open a PR in TiKV and explicitly create a compare-and-delete API rather than piggy backing on compare-and-swap.

Issue #1892, opened 2026-02-11
region-not-found errors


mittalrishabh:
In 8.5 we are asynchronously invalidating the regions on region-not-found errors. And in 7.5 we were invalidating the region synchronously.
It is a regression caused by my [change](https://github.com/tikv/client-go/pull/1769) which added a retry on replica and delayed reload on region not found errors.
With this change, the stale region can remain in cache for an extended period leading to repeated RegionNotFound errors without triggering PD reloads. It is causing the retries on leader and increase in cross AZ traffic.,
Fix would be to set the flag to synchronously reload the region on next access.
Here trade off would be an extra RPC call to PD in the scenario where region split has occurred on the leader but the new region hasn’t yet been created on the replica (e.g., due to replica being down)(https://github.com/pingcap/tidb/issues/62932) (

PR #1916, opened 2026-03-17
hard-invalidate region on RegionNotFound instead of async reload

mittalrishabh:
issue https://github.com/tikv/client-go/issues/1892

The original [fix](https://github.com/tikv/client-go/pull/1891) does not work because in-flight requests continue to fail with region-not-found errors. This change hard-invalidates the region so that concurrent requests stop using the stale cache, and introduces a regionInvalidatedForRetry flag to allow the current request to retry on the leader before giving up.

Unit test

Deployed in prod to verify that it fixed `region-not-found `errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved region cache invalidation to be synchronous instead of asynchronous, ensuring more immediate consistency when region states change.
  * Enhanced retry logic to allow requests to attempt on the leader even after hard region invalidation, improving reliability in region-not-found scenarios.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -2045,18 +2045,6 @@ func (c *RegionCache) BatchLoadRegionsFromKey(bo *retry.Backoffer, startKey []by
 	return regions[len(regions)-1].EndKey(), nil
 }
 
-// AsyncInvalidateCachedRegion marks a cached region for reload on next access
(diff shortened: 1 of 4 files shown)


rpaik:
@you06 and @zyguan could you help with the review? 🙏

zyguan (approved):

you06 (approved):

PR #1926, opened 2026-03-30
ruv2: align bypass with ru v1

disksing:
## Summary
- bypass TiKV RU v2 updates only for requests that are already marked as bypassed by `RequestInfo`
- reuse the same bypass decision in unary, async, and coprocessor stream response paths
- add regression coverage for async requests, coprocessor streams, and the bypass decision itself

## Why
RU v2 bypass conditions had become broader than the existing resource control bypass rules by also skipping background requests. This change aligns the RU v2 bypass behavior with the existing request-info bypass semantics.

## Testing
- `go test ./internal/client -run 'TestBypassRUV2FollowsRequestInfoBypass|TestSendRequestAsyncUpdateTiKVRUV2'`
- `go test ./tikvrpc -run TestCopStreamResponseRecvBypassRUV2`
- `go test ./internal/resourcecontrol -run TestMakeRequestInfo`

## Summary by CodeRabbit

* **Bug Fixes**
  * Resource-usage V2 updates are now conditionally applied per request, preventing unintended updates for requests marked to bypass tracking.
  * Coprocessor streaming responses now honor the per-request bypass flag, avoiding resource-usage updates when bypass is set.

* **Tests**
  * Added tests covering bypass behavior and request-path handling to ensure correct tracking and non-tracking scenarios.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -451,7 +451,7 @@ func (c *RPCClient) getCopStreamResponse(ctx context.Context, client tikvpb.Tikv
 	copStream.Timeout = timeout
 	copStream.Cancel = cancel
 	copStream.Ctx = ctx
-	copStream.BypassRUV2 = resourcecontrol.MakeRequestInfo(req).Bypass()
+	copStream.Bypass = resourcecontrol.MakeRequestInfo(req).Bypass()
 	connPool.streamTimeout <- &copStream.Lease
 
 	// Read the first streaming response to get CopStreamResponse.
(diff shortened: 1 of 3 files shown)


internal/client/client.go
```diff
@@ -337,6 +338,7 @@ func (c *RPCClient) sendRequest(ctx context.Context, addr string, req *tikvrpc.R
 	}
 
 	start := time.Now()
+	bypassRUV2 := resourcecontrol.MakeRequestInfo(req).Bypass()
```
JmPotato: The `Bypass()` function name is quite generic, but the variable names here include the RUV2 concept. Could this be a bit confusing?

disksing: `Bypass` is used for both v1 and v1. and the variable name here is specificlly used for ruv2.

disksing: I will switch to use just `bypass` for variable name, look more clear.

PR #1940, opened 2026-04-10
txnkv: reduce mem / goroutine costs for resolve lock tasks

### LINKED ISSUE #1937: LockResolver: unbounded goroutines and error accumulation cause OOM under sustained lock contention
## Summary

Under sustained lock contention (many locks across many regions, TiKV returning "server is busy"), `LockResolver` spawns unbounded concurrent goroutines that retry indefinitely, each accumulating error objects in `Backoffer.errors`. This creates a positive feedback loop — more retries cause more TiKV pressure, more "server is busy", more retries — eventually leading to TiDB OOM.

The core issue is that **there is no concurrency limit on lock resolution goroutines and no circuit breaker** to stop amplifying TiKV overload.

## Evidence

Profiling data from a TiDB pod during memory growth (OOM at ~15 GB RSS).

### Heap: 8.9 GB live, 95% in lock resolution backoff

```
      flat  flat%        cum   cum%
 5822.21MB 65.32%  5898.77MB 66.18%  fmt.Sprintf
 1365.81MB 15.32%  1365.81MB 15.32%  github.com/pkg/errors.callers
  346.16MB  3.88%     7513MB 84.29%  retry.(*Backoffer).BackoffWithCfgAndMaxSleep
```

5.8 GB in error strings + 1.4 GB in stack traces, both from [backoff.go L184](https://github.com/tikv/client-go/blob/master/config/retry/backoff.go#L184). 95% cumulative through `LockResolver.resolveLock`. 42% through `onServerIsBusy` -> `backoffOnNoCandidate` (all replicas busy, nowhere to send).

### CPU: 40% lock resolution, 20% protobuf serialization, 14% GC

20% of CPU wasted in [`RPCContext.String()`](https://github.com/tikv/client-go/blob/master/internal/locate/region_cache.go#L843) serializing `*metapb.Region` into error strings nobody reads. 14% on GC under memory pressure.

### Trace: 2,725 hours of goroutine-time blocked

85% of goroutine-time sleeping in backoff timers. 6:1 ratio of backoff sleep to actual RPC work. Massive concurrent goroutines confirmed by `.func2.1` suffix (anonymous closures in `resolveLocks`).

### RSS vs heap (15 GB vs 8.9 GB)

Expected: GOGC=100 allows heap up to 2x live set. 14% CPU on GC can't keep up, so dead objects accumulate.

## Root Cause

Four compounding issues:

**1. No concurrency limit on lock resolution goroutines.** `resolveLocks` spawns unbounded `go func()` at [L619](https://github.com/tikv/client-go/blob/master/txnkv/txnlock/lock_resolver.go#L619), [L1074](https://github.com/tikv/client-go/blob/master/txnkv/txnlock/lock_resolver.go#L1074), [L1129](https://github.com/tikv/client-go/blob/master/txnkv/txnlock/lock_resolver.go#L1129), [L1168](https://github.com/tikv/client-go/blob/master/txnkv/txnlock/lock_resolver.go#L1168). Each retries up to 40s ([asyncResolveLockMaxBackoff](https://github.com/tikv/client-go/blob/master/txnkv/txnlock/lock_resolver.go#L49)).

**2. Unbounded `Backoffer.errors` slice.** Every retry appends `errors.Errorf(...)` at [backoff.go L184](https://github.com/tikv/client-go/blob/master/config/retry/backoff.go#L184) — never truncated, only last 3 are ever read ([L156](https://github.com/tikv/client-go/blob/master/config/retry/backoff.go#L156)). `pkg/errors.Errorf` captures a full stack trace each time (1.4 GB). `Fork()`/`Clone()` copy the entire slice to children ([L267](https://github.com/tikv/client-go/blob/master/config/retry/backoff.go#L267), [L286](https://github.com/tikv/client-go/blob/master/config/retry/backoff.go#L286)).

**3. Protobuf serialization in error messages.** `errors.Errorf("server is busy, ctx: %v", ctx)` at [replica_selector.go L571](https://github.com/tikv/client-go/blob/master/internal/locate/replica_selector.go#L571) and ~20 other sites in [region_request.go](https://github.com/tikv/client-go/blob/master/internal/locate/region_request.go) and [lock_resolver.go](https://github.com/tikv/client-go/blob/master/txnkv/txnlock/lock_resolver.go) call `proto.CompactTextString()` on region metadata, producing large strings stored in the unbounded error slice.

**4. Feedback loop.** Unbounded goroutines -> TiKV overloaded -> "server is busy" -> more retries -> more memory -> GC pressure (14% CPU) -> slower -> more locks pile up -> OOM.

## Suggested Fixes

**Fix 1 (critical): Add a global semaphore to `LockResolver`** bounding concurrent resolution goroutines (e.g., 128). When full, async resolutions are shed (`select/default`), sync ones block. This breaks the avalanche by capping TiKV request rate and bounding memory.

**Fix 2: Cap `Backoffer.errors` to ~16 entries** and switch to `fmt.Errorf` (no stack traces). Stop copying the slice in `Fork()`/`Clone()`.

**Fix 3: Replace `%v` on `RPCContext`/`regionErr` with scalar fields** (region ID, addr) in all backoff error messages. Eliminates 20% CPU from proto serialization.

**Fix 4 (longer-term): Circuit breaker** — track "server is busy" ratio in `LockResolver`, pause spawning when overloaded.

| Fix | Effect |
|-----|--------|
| 1. Global semaphore | **Breaks the avalanche.** Caps memory and TiKV request rate. |
| 2. Cap errors + fmt.Errorf | Bounds per-goroutine memory. Eliminates 1.4 GB stack traces. |
| 3. Stop proto serialization | Eliminates 20% CPU waste. |
| 4. Circuit breaker | Self-healing under sustained pressure. |

lcwangchao:
close #1937.

Under sustained lock contention, `LockResolver` could spawn too many async resolve goroutines and keep accumulating heavy backoff errors. That amplified TiKV overload, increased memory use, and could eventually lead to OOM.

## What is changed and how does it work?
- Add a global semaphore for async lock resolution so background resolve tasks have a bounded concurrency.
- Record lock-resolver async running tasks and fallback counts with metrics.
- Change `Backoffer` error recording from unbounded growth to a fixed-size ring buffer that only keeps recent backoff errors.
- Avoid generating stack traces and heavyweight RPC-context strings for backoff-only errors, so retries are much cheaper in memory and CPU.
- Add and complete tests covering bounded async resolve behavior, backoff error retention, and lighter backoff error formatting.
- Related metrics are already added：
    - A new metric: `lock_resolver_async_running_tasks` to show how many async tasks are running.
    - Two new labels for lock resolver counter: `async_resolve_async_commit_fallback` and `read_async_resolve_fallback` . If the rate is not 0, this indicates the semaphore is used up.

## Tests
- `go test ./config/retry ./internal/locate ./txnkv/txnlock -count=1`

## Summary by CodeRabbit

* **New Features**
  * Richer backoff/retry error text that includes detailed request and region context.
  * Concurrency-limited async lock/region resolution with automatic synchronous fallback when capacity is exhausted.

* **Improvements**
  * Bounded, timestamped recent-error buffer for clearer diagnostics and consistent reporting.
  * Backoff context strings now handle nil contexts safely and produce more consistent formatting.

* **Monitoring**
  * New metrics tracking async lock-resolution tasks and fallback occurrences.

* **Tests**
  * Added unit tests validating error formatting, backoff error retention, and async-resolve concurrency behavior.

### DIFF
--- a/config/retry/backoff.go
+++ b/config/retry/backoff.go
@@ -46,15 +46,26 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/pingcap/kvproto/pkg/errorpb"
-	"github.com/pingcap/log"
 	"github.com/pkg/errors"
 	tikverr "github.com/tikv/client-go/v2/error"
 	"github.com/tikv/client-go/v2/internal/logutil"
 	"github.com/tikv/client-go/v2/kv"
 	"github.com/tikv/client-go/v2/util"
 	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 )
+
+// MaxRecordBackoffErrCount is the maximum number of backoff errors to be recorded in the log.
+const MaxRecordBackoffErrCount = 3
+
+type backoffErr struct {
+	Reason string
+	Time   time.Time
+}
+
+// Error implements the error interface for backoffErr.
+func (e backoffErr) Error() string {
+	return fmt.Sprintf("%s at %s", e.Reason, e.Time.Format(time.RFC3339Nano))
+}
 
 // Backoffer is a utility for retrying queries.
 type Backoffer struct {
@@ -68,7 +79,14 @@ type Backoffer struct {
 	vars *kv.Variables
 	noop bool
 
-	errors         []error
+	// errors is a fixed-length array to record the backoff errors,
+	// and it records the last maximum MaxRecordBackoffErrCount errors.
+	// If errorsNum <= MaxRecordBackoffErrCount, only the errors[:errorsNum] are valid.
+	// Otherwise, if errorsNum > MaxRecordBackoffErrCount, errors can be regarded as a circular array,
+	// and errors[errorsNum % MaxRecordBackoffErrCount] is the first error located
+	errors    [MaxRecordBackoffErrCount]backoffErr
+	errorsNum int
+
 	configs        []*Config
 	backoffSleepMS map[string]int
 	backoffTimes   map[string]int
@@ -130,6 +148,26 @@ func (b *Backoffer) BackoffWithMaxSleepTxnLockFast(maxSleepMs int, err error) er
 	return b.BackoffWithCfgAndMaxSleep(cfg, maxSleepMs, err)
 }
 
+func (b *Backoffer) appendErr(err error) {
+	b.errors[b.errorsNum%MaxRecordBackoffErrCount] = backoffErr{
+		Reason: err.Error(),
+		Time:   time.Now(),
+	}
+	b.errorsNum++
+}
+
+func (b *Backoffer) latestErrors() []backoffErr {
+	if b.errorsNum <= len(b.errors) {
+		return b.errors[:b.errorsNum]
+	}
+	errs := make([]backoffErr, len(b.errors))
+	firstIndex := b.errorsNum % MaxRecordBackoffErrCount
+	for i := 0; i < len(b.errors); i++ {
+		errs[i] = b.errors[(firstIndex+i)%len(b.errors)]
+	}
+	return errs
+}
+
 // BackoffWithCfgAndMaxSleep sleeps a while base on the Config and records the error message
 // and never sleep more than maxSleepMs for each sleep.
 func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err error) error {
@@ -153,11 +191,8 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 	if b.maxSleep > 0 && maxTimeExceeded {
 		longestSleepCfg, longestSleepTime := b.longestSleepCfg()
 		errMsg := fmt.Sprintf("%s backoffer.maxSleep %dms is exceeded, errors:", cfg.String(), b.maxSleep)
-		for i, err := range b.errors {
-			// Print only last 3 errors for non-DEBUG log levels.
-			if log.GetLevel() == zapcore.DebugLevel || i >= len(b.errors)-3 {
-				errMsg += "\n" + err.Error()
-			}
+		for _, err := range b.latestErrors() {
+			errMsg += "\n" + err.Error()
 		}
 		var backoffDetail bytes.Buffer
 		totalTimes := 0
@@ -181,7 +216,7 @@ func (b *Backoffer) BackoffWithCfgAndMaxSleep(cfg *Config, maxSleepMs int, err e
 		// Use the backoff type that contributes most to the timeout to generate a MySQL error.
 		return errors.WithStack(returnedErr)
 	}
-	b.errors = append(b.errors, errors.Errorf("%s at %s", err.Error(), time.Now().Format(time.RFC3339Nano)))
+	b.appendErr(err)
 	b.configs = append(b.configs, cfg)
 
 	// Lazy initialize.
@@ -264,7 +299,8 @@ func (b *Backoffer) Clone() *Backoffer {
 		totalSleep:     b.totalSleep,
 		excludedSleep:  b.excludedSleep,
 		vars:           b.vars,
-		errors:         append([]error{}, b.errors...),
+		errors:         b.errors,
+		errorsNum:      b.errorsNum,
 		configs:        append([]*Config{}, b.configs...),
 		backoffSleepMS: copyMapWithoutRecursive(b.backoffSleepMS),
 		backoffTimes:   copyMapWithoutRecursive(b.backoffTimes),
@@ -283,7 +319,8 @@ func (b *Backoffer) Fork() (*Backoffer, context.CancelFunc) {
 		maxSleep:       b.maxSleep,
 		totalSleep:     b.totalSleep,
 		excludedSleep:  b.excludedSleep,
-		errors:         append([]error{}, b.errors...),
+		errors:         b.errors,
+		errorsNum:      b.errorsNum,
 		configs:        append([]*Config{}, b.configs...),
 		backoffSleepMS: copyMapWithoutRecursive(b.backoffSleepMS),
 		backoffTimes:   copyMapWithoutRecursive(b.backoffTimes),
@@ -303,6 +340,7 @@ func (b *Backoffer) UpdateUsingForked(forked *Backoffer) {
 			b.totalSleep = forked.totalSleep
 			b.excludedSleep = forked.excludedSleep
 			b.errors = forked.errors
+			b.errorsNum = forked.errorsNum
 			b.backoffSleepMS = forked.backoffSleepMS
 			b.backoffTimes = forked.backoffTimes
 			break
@@ -363,7 +401,7 @@ func (b *Backoffer) GetBackoffSleepMS() map[string]int {
 
 // ErrorsNum returns the number of errors.
 func (b *Backoffer) ErrorsNum() int {
-	return len(b.errors)
+	return b.errorsNum
 }
 
 // Reset resets the sleep state of the backoffer, so that following backoff
--- a/config/retry/backoff.go
+++ b/config/retry/backoff.go
@@ -156,6 +156,10 @@ func (b *Backoffer) appendErr(err error) {
 	b.errorsNum++
 }
 
+// latestErrors is called when the backoff time exceeds the max sleep time,
+// and it returns the latest N (N <= MaxRecordBackoffErrCount) backoff errors for logging.
+// It should only be called when the backoff time exceeds the max sleep time to reduce the performance overhead,
+// because it may involve some array copy when the number of errors exceeds MaxRecordBackoffErrCount.
 func (b *Backoffer) latestErrors() []backoffErr {
 	if b.errorsNum <= len(b.errors) {
 		return b.errors[:b.errorsNum]
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -841,12 +841,32 @@ type RPCContext struct {
 }
 
 func (c *RPCContext) String() string {
+	if c == nil {
+		return "<nil>"
+	}
 	var runStoreType string
 	if c.Store != nil {
 		runStoreType = c.Store.StoreType().Name()
 	}
 	res := fmt.Sprintf("region ID: %d, meta: %s, peer: %s, addr: %s, idx: %d, reqStoreType: %s, runStoreType: %s",
 		c.Region.GetID(), c.Meta, c.Peer, c.Addr, c.AccessIdx, c.AccessMode, runStoreType)
+	if c.ProxyStore != nil {
+		res += fmt.Sprintf(", proxy store id: %d, proxy addr: %s", c.ProxyStore.storeID, c.ProxyStore.GetAddr())
+	}
+	return res
+}
+
+// BackoffInfoString returns the string used for backoff error logs
+func (c *RPCContext) BackoffInfoString() string {
+	if c == nil {
+		return "<nil>"
+	}
+	var runStoreType string
+	if c.Store != nil {
+		runStoreType = c.Store.StoreType().Name()
+	}
+	res := fmt.Sprintf("region: %s, peerID: %d, storeID: %d, addr: %s, idx: %d, reqStoreType: %s, runStoreType: %s",
+		c.Region.String(), c.Peer.GetId(), c.Peer.GetStoreId(), c.Addr, c.AccessIdx, c.AccessMode, runStoreType)
 	if c.ProxyStore != nil {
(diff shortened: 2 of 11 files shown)


lcwangchao:
@coderabbitai please review it again

internal/locate/region_cache.go
```diff
@@ -853,6 +856,23 @@ func (c *RPCContext) String() string {
 	return res
 }
 
+// BackoffInfoString returns the string used for backoff error logs
+func (c *RPCContext) BackoffInfoString() string {
```
cfzjywxk: This seems to be more like `RPCContext.String()` , why naming it with `BackoffInfo`?

txnkv/txnlock/lock_resolver.go
```diff
@@ -1126,14 +1192,19 @@ func (lr *LockResolver) resolveAsyncCommitLock(bo *retry.Backoffer, l *Lock, sta
 	logutil.BgLogger().Info("resolve async commit locks", zap.Uint64("startTS", l.TxnID), zap.Uint64("commitTS", status.commitTS), zap.Stringer("TxnStatus", status))
 	if asyncResolveAll {
 		asyncBo := retry.NewBackoffer(lr.asyncResolveCtx, asyncResolveLockMaxBackoff)
-		go func() {
+		asyncResolveAll = lr.tryAsyncResolve(func() {
```
cfzjywxk: Better use a different variable for "did we actually async resolve all". The current implementation is a bit confusing for L1207

txnkv/txnlock/lock_resolver.go
```diff
@@ -680,6 +689,63 @@ func (lr *LockResolver) resolveLocks(bo *retry.Backoffer, opts ResolveLocksOptio
 	}, nil
 }
 
+// AsyncResolveLockSemaphoreLimit is the max concurrency of async resolve lock.
+// It is set to a very large value to make sure it won't introduce some performance drawback in most cases,
+// and for some extreme cases it can also provide some protection to avoid OOM
+// by too many async resolve lock goroutines.
+const AsyncResolveLockSemaphoreLimit = 10000
```
cfzjywxk, line 718: Do we need a http api to change it in case of some extrem situations?

config/retry/backoff.go
```diff
@@ -130,6 +148,26 @@ func (b *Backoffer) BackoffWithMaxSleepTxnLockFast(maxSleepMs int, err error) er
...
 
+func (b *Backoffer) appendErr(err error) {
+	b.errors[b.errorsNum%MaxRecordBackoffErrCount] = backoffErr{
+		Reason: err.Error(),
+		Time:   time.Now(),
+	}
+	b.errorsNum++
+}
+
+func (b *Backoffer) latestErrors() []backoffErr {
```
cfzjywxk, line 163: Better to add comments that it should only be called on the `max-sleep-exceeded` path, and it should never be called in a per-backoff like path.

txnkv/txnlock/lock_resolver_test.go
```diff
@@ -48,3 +53,115 @@ func TestLockResolverCache(t *testing.T) {
...
+
+	var enter atomic.Int32
+	var exit atomic.Int32
+	waitCounter := func(c *atomic.Int32, n int) {
+		assert.Eventually(t, func() bool {
+			return c.Load() == int32(n)
+		}, 10*time.Second, 10*time.Millisecond)
+		if c == &exit {
+			// wait the goroutine exit completely and related metrics updated.
+			time.Sleep(5 * time.Millisecond)
```
cfzjywxk: This seems to be flaky on a busy CI runner.

txnkv/txnlock/lock_resolver.go
```diff
@@ -680,6 +689,63 @@ func (lr *LockResolver) resolveLocks(bo *retry.Backoffer, opts ResolveLocksOptio
...
+	ch := make(chan struct{}, limit)
+	for i := 0; i < limit; i++ {
+		select {
+		case ch <- struct{}{}:
+		default:
+			panic("asyncResolveLockSemaphore channel should never be full")
+		}
+	}
+	return ch
+}
```
cfzjywxk: Seems
```
func createAsyncResolveLockSemaphore(limit int) chan struct{} { 
      ch := make(chan struct{}, limit)
      for i := 0; i < limit; i++ {                                                        
          ch <- struct{}{}
      }                                                                                   
      return ch                                                   
 }
```
is enough

txnkv/txnlock/lock_resolver.go
lcwangchao: Seems client-go does not expose this kind of API.

txnkv/txnlock/lock_resolver.go
```diff
@@ -613,19 +617,24 @@ func (lr *LockResolver) resolveLocks(bo *retry.Backoffer, opts ResolveLocksOptio
...
 					}
-				}()
-			} else {
+				}, metrics.LockResolverAsyncRunningTasksForReadResolve)
+				if !asyncResolve {
+					metrics.LockResolverCountWithReadAsyncResolveFallback.Inc()
+				}
+			}
+			if !asyncResolve {
 				err = lr.resolveLock(bo, l, status, lite, cleanRegions)
```
zyguan, line 640: When the semaphore is saturated, it will fallback to synchronous cleanup, which may lead to spike of read latency and cleanup error returns compared to the previous best-effort background cleanup. Is this an acceptable tradeoff for the issue it tries to fix? Would it be better to just drop the async read cleanup when the semaphore is saturated?

txnkv/txnlock/lock_resolver.go
```diff
@@ -1124,16 +1186,22 @@ func (lr *LockResolver) resolveAsyncCommitLock(bo *retry.Backoffer, l *Lock, sta
 		return status, nil
 	}
 	logutil.BgLogger().Info("resolve async commit locks", zap.Uint64("startTS", l.TxnID), zap.Uint64("commitTS", status.commitTS), zap.Stringer("TxnStatus", status))
+	doSyncResolve := false
 	if asyncResolveAll {
 		asyncBo := retry.NewBackoffer(lr.asyncResolveCtx, asyncResolveLockMaxBackoff)
-		go func() {
+		doSyncResolve = lr.tryAsyncResolve(func() {
 			err := lr.resolveAsyncResolveData(asyncBo, l, status, toResolveKeys)
```
zyguan, line 1245: In [`resolveLocks`](https://github.com/tikv/client-go/pull/1940/changes#diff-13c8df1fbf35abf2fd05615e715d5a5cbd79b30ae0839b7d67ca135e7b5e5004R627), one async `resolveLock` call consumes one token from the semaphore, however, here mutliple concurrent `resolveRegionLocks` calls only consume one token. Is this an intentional design?

lcwangchao: I think this is fine. What we need right now is simply a bound to prevent unbounded forking of background goroutines. The exact limit is not very important, so we set it to a relatively large value, 10000, for now. Although resolveRegionLocks consumes only one token and this could still allow, for example, 50000 background goroutines to run, that is acceptable because there is still an upper bound.

Also, a large number of locks to resolve is likely not caused by the number of transactions, since the number of in-flight transactions is bounded by the connection count. It is more likely caused by the number of keys in a single large transaction. Therefore, the number of background tasks needed to resolve an entire async transaction should not be too large.

txnkv/txnlock/lock_resolver.go
lcwangchao: Yeah, I think that’s another possible approach, since reads can still succeed even if lock resolution has not finished yet.

However, if the “committed locks” or “resolved locks” are ignored on the TiKV side, they could lead to a long-running dead loop, which would also be dangerous.

ekexium (commented):
Overall, LGTM. I suggest adding necessary observability enhancements alongside this fix, since it could lead to slow queries that users care about. There are already metrics. Consider if logs, alerts, and fields in execution details are needed

lcwangchao:
> Overall, LGTM. I suggest adding necessary observability enhancements alongside this fix, since it could lead to slow queries that users care about. There are already metrics. Consider if logs, alerts, and fields in execution details are needed

Related metrics are already added：

- A new metric: `lock_resolver_async_running_tasks` to show how many async tasks are running.
- Two new labels for lock resolver counter: `async_resolve_async_commit_fallback` and `read_async_resolve_fallback` . If the rate is not 0, this indicates the semaphore is used up.

ekexium:
> > Overall, LGTM. I suggest adding necessary observability enhancements alongside this fix, since it could lead to slow queries that users care about. There are already metrics. Consider if logs, alerts, and fields in execution details are needed
> 
> Related metrics are already added：
> 
> * A new metric: `lock_resolver_async_running_tasks` to show how many async tasks are running.
> * Two new labels for lock resolver counter: `async_resolve_async_commit_fallback` and `read_async_resolve_fallback` . If the rate is not 0, this indicates the semaphore is used up.

I see. My concern is when someone meets a slow query caused by this change, the first approach is usually to check slow logs and TiDB logs. Few would know to check these metrics

lcwangchao:
> > > Overall, LGTM. I suggest adding necessary observability enhancements alongside this fix, since it could lead to slow queries that users care about. There are already metrics. Consider if logs, alerts, and fields in execution details are needed
> > 
> > 
> > Related metrics are already added：
> > 
> > * A new metric: `lock_resolver_async_running_tasks` to show how many async tasks are running.
> > * Two new labels for lock resolver counter: `async_resolve_async_commit_fallback` and `read_async_resolve_fallback` . If the rate is not 0, this indicates the semaphore is used up.
> 
> I see. My concern is when someone meets a slow query caused by this change, the first approach is usually to check slow logs and TiDB logs. Few would know to check these metrics

Updated in https://github.com/tikv/client-go/pull/1940/changes/ec3e385f6d7d568e6842d69f595c0349eaa60e1f , PTAL again. A warn log printed if the Semaphore used up

txnkv/txnlock/lock_resolver.go
zyguan: It's a bit weird that the semaphore only prevents unbounded goroutines, however, it does not impose a strict numerical limit, meaning the actual number of goroutines may exceed the semaphore size. I'm wondering if it might be better to introduce a goroutine pool with a dynamic size, subject to a strict upper limit. It would not only strictly limit the number of goroutines but also reuse goroutines for those short-lived tasks, thereby reducing the overhead associated with frequently creating new goroutines.

lcwangchao:
@coderabbitai please review this PR again

lcwangchao:
@coderabbitai review this PR again

lcwangchao:
@coderabbitai review again

txnkv/txnlock/lock_resolver_test.go
```diff
@@ -48,3 +52,107 @@ func TestLockResolverCache(t *testing.T) {
...
+	require.True(t, tryAsync())
+	waitSemaphoreSizeWithCheck(5)
+
+	// more async task will be rejected
+	require.False(t, tryAsync())
+	require.False(t, tryAsync())
+	require.False(t, tryAsync())
+	waitSemaphoreSizeWithCheck(5)
+	time.Sleep(time.Millisecond)
+	checkMetricVal(5)
```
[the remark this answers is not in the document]
lcwangchao: This is to test the metric should still be correct and that pending async tasks do not cause metric change.

txnkv/txnlock/lock_resolver.go
lcwangchao: PTAL：https://github.com/tikv/client-go/pull/1940/changes/da1a660e0e861d0eeabfffafb1ce696fac0c9869 which introduced 2 changes:

- use `gp.Pool` as a goroutine pool
- Other async tasks also need to acquire a permit to run background

txnkv/txnlock/lock_resolver.go
```diff
@@ -110,6 +115,7 @@ func NewLockResolver(store storage) *LockResolver {
 // Close cancels all background goroutines.
 func (lr *LockResolver) Close() {
 	lr.asyncResolveCancel()
+	lr.asyncResolvePool.Close()
```
cfzjywxk, line 118: Is the `func (lr *LockResolver) Close()` ensured to be called? What would happen if the close is not called in products using client-go as a lib?

lcwangchao: For this PR specifically, if `Close()` is missed, the incremental impact is bounded: in-flight async resolve tasks may continue until they finish, and the idle `gp.Pool` workers recycle after 10s (`gp.New(..., 10*time.Second)`). So this change should not introduce an unbounded permanent goroutine leak by itself, but `Close()` is still the proper shutdown path to proactively cancel background async resolve work.

cfzjywxk (approved):
LGTM

zyguan (approved):

PR #1947, opened 2026-04-21
resourcecontrol, tikvrpc: add PredictedReadBytes hint for RC paging pre-charge

### LINKED ISSUE #1953: resource control: add PredictedReadBytes hint transport for RC paging pre-charge
### Problem

Under RC (resource control), paging coprocessor RPCs are today billed only at response settlement. A misbehaving resource group can burst large reads before its token bucket reacts, because no RU is reserved up front.

To enable a pre-charge mechanism in the PD controller, an upstream caller (TiDB cop iterator) needs to attach a read-bytes prediction to each paging RPC. client-go has no transport for such a hint today: `tikvrpc.Request` has no field for it, and `internal/resourcecontrol.RequestInfo` exposes nothing the controller can consume.

### Proposal

Thread an optional caller-supplied read-bytes estimate from `tikvrpc.Request` into the resource-control layer so the PD controller can read it via a duck-typed optional interface.

- **`tikvrpc.Request`**: new `PredictedReadBytes uint64` field. Client-go-internal hint, **not** carried in the wire proto — consumed before the RPC is sent; TiKV never observes it.
- **`internal/resourcecontrol.RequestInfo`**: new `predictedReadBytes` field populated from `tikvrpc.Request.PredictedReadBytes` in `MakeRequestInfo`, exposed via a `PredictedReadBytes()` getter.

Together these let the PD resource-group controller pre-charge RU before the RPC goes out (see the related PD issue). When the hint is absent or zero, behavior is unchanged — the request is billed by actual read bytes at settlement.

### Related issues

- TiDB caller: pingcap/tidb#68085
- PD controller: tikv/pd#10612

### Non-Goals

- No change to the wire protocol.
- No change to billing semantics — this issue only adds a transport channel for an optional hint; the controller decides what to do with it.

YuhaoZhang00:
Issue Number: close #1953

Adds an optional caller-supplied read-bytes estimate that the resource-control layer forwards to PD's controller as a pre-charge basis for paging-style coprocessor requests.

- `tikvrpc.Request`: new `PredictedReadBytes uint64` field. This is a client-go-internal hint, not carried in the wire proto; it is consumed before the RPC is sent and TiKV never observes it.
- `internal/resourcecontrol.RequestInfo`: new `predictedReadBytes` and `isCop` fields populated in `MakeRequestInfo`, exposed through `PredictedReadBytes()` and `IsCop()`.

Together these let an upstream caller (TiDB cop iterator, driven by a per-logical-scan EMA) attach a read-bytes prediction to a paging RPC. PD's resource-group controller consumes this hint via the matching `RequestInfo` methods and pre-charges RU before the RPC goes out. When the hint is zero, behavior is unchanged and the request is billed from actual read bytes at settlement.

This PR also adds targeted unit coverage for:

- `PredictedReadBytes()` propagation from `tikvrpc.Request` to `resourcecontrol.RequestInfo`.
- `IsCop()` classification for `CmdCop` / `CmdCopStream` versus other read requests.
- Existing resource-control interceptor behavior on transport failures where no response is returned. These are regression tests only; this PR does not change the interceptor runtime settlement path.

### PD client sync

This PR also bumps `github.com/tikv/pd/client` to a merged PD master version. That brings in PD client interface changes from other PD PRs, so client-go's mocks and integration-test helpers are updated in the same PR to keep the upgraded dependency buildable:

- tikv/pd#10883 (`8ed2b3fb8e3c`, `client: fix resource group config watch recovery`): meta-storage/resource-manager `Watch` now returns `chan *metastorage.WatchResponse` so compact-revision information can be propagated.
- tikv/pd#10911 (`17e9771e4b9e`, `keyspace: load keyspace by ID through gRPC`): the PD client keyspace interface now includes `LoadKeyspaceByID(ctx, id uint32)`.

`integration_tests/go.mod` temporarily replaces `github.com/pingcap/tidb` and `github.com/pingcap/tidb/pkg/parser` with an adapted fork because the current upstream TiDB dependency still uses the old PD client mock signatures. After the TiDB-side PR lands, a follow-up client-go PR should remove that replace and bump back to the official TiDB master pseudo-version.

### Related PRs

- PD controller side: tikv/pd#10611
- TiDB caller side: pingcap/tidb#67941

### Check List

- Unit test

Validation:

```sh
env GOCACHE=/tmp/client-go-gocache go test ./internal/resourcecontrol ./internal/client ./tikvrpc ./util -count=1
cd integration_tests && env GOCACHE=/tmp/client-go-gocache go test ./... -run '^$' -count=0
git diff --check HEAD
```

```release-note
None
```


internal/resourcecontrol/resource_control.go
```diff
@@ -87,11 +90,12 @@ func MakeRequestInfo(req *tikvrpc.Request) *RequestInfo {
...
-			storeID:     storeID,
-			bypass:      bypass,
-			requestSize: uint64(req.GetSize()),
-			accessType:  toPDAccessLocationType(req.AccessLocation),
+			writeBytes:         -1,
+			storeID:            storeID,
+			bypass:             bypass,
+			requestSize:        uint64(req.GetSize()),
+			accessType:         toPDAccessLocationType(req.AccessLocation),
+			predictedReadBytes: req.PredictedReadBytes,
```
rleungx, line 103: With PD pre-charge enabled, this hint needs a settlement path even when the RPC fails before returning a response. Today `OnResponseWait` is only called when `resp != nil`, so a transport error/timeout after pre-charge can leave the predicted read RU charged without refund. Can we add a no-response settlement/refund path or a test that documents this behavior?

YuhaoZhang00: Addressed by adding calling `resourceControlInterceptor.OnRequestCancel()` which does refund when `resp == nil`

Commit 54a00f471548bac0d270cd7c32a2c3c860f74fdd

internal/client/client_interceptor.go
```diff
@@ -73,14 +73,21 @@ func (r interceptedClient) SendRequest(ctx context.Context, addr string, req *ti
...
+				return nil, err
+			}
+			if ruDetails != nil {
+				ruDetails.Update(consumption, waitDuration)
+			}
+		} else {
+			// Transport-level failure produced no response — the settlement
+			// path will not run. Cancel the pre-charge so the predicted RU
+			// debited in OnRequestWait is refunded to the limiter.
+			resourceControlInterceptor.OnRequestCancel(ctx, resourceGroupName, reqInfo)
```
rleungx: On the no-response path we refund the pre-charge in the PD controller, but `RUDetails` has already been updated with the `OnRequestWait` consumption above. Because `OnRequestCancel` does not return a refund delta and this branch does not apply a negative update, TiDB runtime RU stats will still include the predicted/base RU from the failed attempt. If the request is retried and later succeeds, the failed attempt can be counted as phantom RU in runtime stats even though controller tokens/consumption were rolled back. The async cancel path below has the same issue. Could we either have `OnRequestCancel` return the refund delta and update `RUDetails`, or keep the request-side consumption here and subtract it on cancel?

YuhaoZhang00: Fixed in ea180e1: the no-response cancel path now rolls back the request-side RUDetails RRU/WRU that was added after OnRequestWait, with sync/async regression tests covering the phantom RU case.

YuhaoZhang00: Follow up issue: Current cancellation semantics are tied to “no settlement response available”, not to “no TiKV work happened”. This path intentionally refunds the request-side precharge because PD has no reliable execution signal there. The tradeoff is that work already done by TiKV but lost with the response can be under-accounted. Fixing that precisely would require client-go/TiKV to propagate partial execution stats on error/no-response paths; PD cannot infer it from the current API.

rleungx (approved):

JmPotato (approved):

cfzjywxk (approved):

PR #1948, opened 2026-04-21
tikv: keep txn safe point updater interval fixed

wfxr:
## Summary

Remove the previous 1s quick retry behavior and keeping the txn safe point updater interval fixed at `pollTxnSafePointInterval` even after load failures to avoid potentially adding unnecessary load to PD.

## Test Plan

- `go test ./tikv -run '^$' -count=1`
- `git diff --check -- tikv/kv.go tikv/safepoint.go tikv/kv_test.go`

### DIFF
--- a/tikv/kv.go
+++ b/tikv/kv.go
@@ -428,17 +428,17 @@ func (s *KVStore) runTxnSafePointUpdater() {
 	if _, e := util.EvalFailpoint("noBuiltInTxnSafePointUpdater"); e == nil {
 		return
 	}
-	d := pollTxnSafePointInterval
 	for {
 		select {
-		case now := <-time.After(d):
+		case now := <-time.After(pollTxnSafePointInterval):
 			txnSafePoint, err := s.loadTxnSafePoint(context.Background())
-			if err == nil {
-				s.UpdateTxnSafePointCache(txnSafePoint, now)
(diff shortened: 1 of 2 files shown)


rleungx (approved):

MyonKeminta (approved):

tikv/kv.go
```diff
@@ -428,17 +428,17 @@ func (s *KVStore) runTxnSafePointUpdater() {
...
 			txnSafePoint, err := s.loadTxnSafePoint(context.Background())
-			if err == nil {
-				s.UpdateTxnSafePointCache(txnSafePoint, now)
-				d = pollTxnSafePointInterval
-			} else {
-				d = pollTxnSafePointQuickRepeatInterval
+			if err != nil {
+				// Intentionally ignore the error here. loadTxnSafePoint already logs it,
+				// and the updater retries on the next fixed poll interval to avoid
+				// potentially adding unnecessary load to PD.
```
rleungx, line 438: Do we need a log here?

wfxr: I think it's not necessary since `loadTxnSafePoint` already logs it?

https://github.com/tikv/client-go/blob/f6464e7cfcd41acac6035469f7f51556045067be/tikv/kv.go#L204-L213

PR #1952, opened 2026-04-27
txnkv: always fire commit callback for zero session id

windtalker:
Currently, the commit callback is called only if
* session_id is set, and session_id > 0
* session_id is not set

https://github.com/tikv/client-go/blob/2d8a6e6ec9f9593f0aba81d8f36c1b085a7d6347/txnkv/transaction/txn.go#L761-L763
And
https://github.com/tikv/client-go/blob/2d8a6e6ec9f9593f0aba81d8f36c1b085a7d6347/txnkv/transaction/txn.go#L782-L784

For all internal session, the session_id is 0, so internal sql can not get last_txn_info

There wasn't really a strong reason for the original restriction. It looks like it got added just as an offhand comment during code review:
https://github.com/pingcap/tidb/pull/19949#discussion_r486954059

This pr remove this restriction because in materialized view refresh, it is executed in internal session, while we want to get the commit_ts.

### DIFF
--- a/integration_tests/2pc_test.go
+++ b/integration_tests/2pc_test.go
@@ -39,6 +39,7 @@ package tikv_test
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	stderrs "errors"
 	"fmt"
 	"math"
@@ -63,6 +64,7 @@ import (
 	"github.com/tikv/client-go/v2/txnkv"
 	"github.com/tikv/client-go/v2/txnkv/transaction"
 	"github.com/tikv/client-go/v2/txnkv/txnlock"
+	"github.com/tikv/client-go/v2/util"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


wfxr:
This change looks reasonable to me. Could we add a test for the pessimistic transaction path with `SessionID=0`? MV refresh uses `BEGIN PESSIMISTIC`, and pessimistic transactions bypass local latches, so covering that path explicitly would make the test match the motivating case better.

lcwangchao (approved):

windtalker:
> This change looks reasonable to me. Could we add a test for the pessimistic transaction path with `SessionID=0`? MV refresh uses `BEGIN PESSIMISTIC`, and pessimistic transactions bypass local latches, so covering that path explicitly would make the test match the motivating case better.

I've added these mv refresh test in tidb repo

wfxr (approved):

cfzjywxk (approved):

gengliqi (approved):

PR #1956, opened 2026-05-07
txn: reject unsupported shared lock transaction paths

### LINKED ISSUE #1955: Reject unsupported shared lock transaction paths
## Enhancement

TiDB shared-lock work needs client-go to reject unsupported transaction modes before requests are sent to TiKV / next-gen storage. This is the client-go tracking issue for TiDB issue https://github.com/pingcap/tidb/issues/66154.

## Motivation

Shared lock transactions have compatibility constraints that should be enforced in client-go as well as in the storage engine:

- Shared lock transactions should not use async commit.
- Shared lock transactions should not use 1PC.
- Shared lock keys should not be selected as the transaction primary key.
- Shared locks should not be used in pipelined transactions.
- Shared lock transactions should not be used in aggressive/fair locking mode.

Rejecting these paths in client-go gives both original TiDB and next-gen TiDB/TiDB-X a consistent early failure path instead of relying only on downstream storage errors.

## Proposed change

Add client-side guards in the transaction committer and lock path, with focused regression tests for the unsupported combinations above.

## Related

- TiDB tracking issue: https://github.com/pingcap/tidb/issues/66154

cfzjywxk:
Issue Number: close #1955

Problem Summary:

Shared lock transactions should not enter commit or lock paths that are incompatible with shared locks. TiDB issue: https://github.com/pingcap/tidb/issues/66154

What is changed:

- Reject a shared-lock key if it is selected or pre-set as the transaction primary key.
- Keep async commit and 1PC disabled for transactions that contain shared locks.
- Reject shared-lock requests in aggressive/fair locking mode.
- Reject shared-lock use in pipelined transactions and pipelined flush.
- Add focused regression tests for the unsupported combinations.

Tests:

- `go test ./txnkv/transaction -run "TestLockKeys|TestSharedLockCommitterIncompatibilities"`
- `go test ./txnkv/transaction`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted certain lock configurations from using commit optimizations.
  * Prevented specific lock types from being designated as transaction primary keys.
  * Added validation to block lock mode conflicts in pipelined and aggressive locking contexts.

* **Tests**
  * Extended test coverage for transaction lock configuration constraints.

### DIFF
--- a/txnkv/transaction/2pc.go
+++ b/txnkv/transaction/2pc.go
@@ -567,6 +567,7 @@ func (c *twoPhaseCommitter) initKeysAndMutations(ctx context.Context) error {
 	c.mutations = newMemBufferMutations(sizeHint, memBuf)
 	c.isPessimistic = txn.IsPessimistic()
 	filter := txn.kvFilter
+	primaryKeyIsSharedLock := false
 
 	var err error
 	var assertionError error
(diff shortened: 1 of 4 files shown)


cfzjywxk:
@wfxr PTAL

you06 (approved):

zyguan (approved):

PR #1958, opened 2026-05-10
region_cache: return nil reason from GetTiFlashRPCContext in 8.5

gengliqi:
Return a reason when GetTiFlashRPCContext returns nil to help callers know why the TiFlash RPC context is unavailable.

### DIFF
--- a/internal/locate/region_cache.go
+++ b/internal/locate/region_cache.go
@@ -81,6 +81,26 @@ const (
 // LabelFilter returns false means label doesn't match, and will ignore this store.
 type LabelFilter = func(labels []*metapb.StoreLabel) bool
 
+// TiFlashRPCContextUnavailableReason is the reason why GetTiFlashRPCContext returns a nil RPCContext.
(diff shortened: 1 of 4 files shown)


internal/locate/region_cache.go
```diff
@@ -1035,6 +1069,7 @@ func (c *RegionCache) GetTiFlashRPCContext(bo *retry.Backoffer, id RegionVerID,
 		storeFailEpoch := atomic.LoadUint32(&store.epoch)
 		if storeFailEpoch != regionStore.storeEpochs[storeIdx] {
 			cachedRegion.invalidate(Other)
+			epochMismatchCount++
 			logutil.Logger(bo.GetCtx()).Info("invalidate current region, because others failed on same store",
 				zap.Uint64("region", id.GetID()),
 				zap.String("store", store.addr))
```
JaySon-Huang, line 1075: Also log down the region ver/confVer?

PR #1973, opened 2026-05-21
txnkv: enable TiKV-side async resolve region lock for read path

### LINKED ISSUE #1963: Track client-go support for async region ResolveLock signals
## Background

This tracks client-go-side support for async region-level ResolveLock signals. It is related to TiDB tracking issue: https://github.com/pingcap/tidb/issues/68352.

When TiDB/client-go encounters locks from large transactions, synchronous region-level ResolveLock may be expensive because TiKV can need to scan lock CF for the whole region. Repeated foreground requests may therefore trigger overlapping region-level resolve work.

## Scope

The intended direction is to keep foreground correctness on the existing synchronous paths, such as transaction status checks and keyed lock resolve when the current request must be unblocked. For region-level cleanup that the foreground request does not need to wait for, client-go can send a best-effort async resolve signal to TiKV instead.

On the client-go side, this likely includes deciding where async signals should be emitted, keeping async signals separate from synchronous ResolveLock requests because their response semantics differ, and gating the behavior behind configuration so it can be enabled gradually.

lcwangchao:
close #1963

## Summary by CodeRabbit

* **New Features**
  * Optional TiKV-side asynchronous lock resolution behind a feature gate; read paths default to async when results aren't required and behavior can be toggled per resolver.

* **Tests**
  * New integration and unit tests validating TiKV-side async resolve behavior, request patterns, and resolver metrics.

* **Chores**
  * Bumped kvproto dependency version across examples and tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tikv/client-go/pull/1973?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -1562,8 +1562,3 @@ func (lr *LockResolver) resolvePessimisticLock(bo *retry.Backoffer, l *Lock, pes
 		return nil
 	}
 }
-
-// SetEnableTiKVSideAsyncResolve sets whether to enable TiKV-side async resolve.
-func (lr *LockResolver) SetEnableTiKVSideAsyncResolve(enabled bool) {
-	lr.enableTiKVSideAsyncResolve.Store(enabled)
-}
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -1428,6 +1428,9 @@ func (lr *LockResolver) resolveLock(bo *retry.Backoffer, l *Lock, status TxnStat
 			metrics.LockResolverCountWithResolveLockLite.Inc()
 			lreq.Keys = [][]byte{l.Key}
 		} else if !resultRequired && lr.enableTiKVSideAsyncResolve.Load() {
+			if intest.InTest && !config.NextGen {
+				panic("TiKV side async resolve should only be enabled in next-gen cluster")
+			}
 			lreq.IsAsync = true
 		}
 		req := tikvrpc.NewRequest(tikvrpc.CmdResolveLock, lreq, kvrpcpb.Context{
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -87,6 +87,9 @@ type LockResolver struct {
 	// The Cancel function is to cancel these goroutines for passing goleak test.
 	asyncResolveCtx    context.Context
 	asyncResolveCancel func()
+
+	// enableTiKVSideAsyncResolve indicates whether to enable TiKV side async resolve.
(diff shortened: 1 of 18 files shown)


cfzjywxk (approved):
LGTM

txnkv/txnlock/lock_resolver.go
```diff
@@ -1420,6 +1427,8 @@ func (lr *LockResolver) resolveLock(bo *retry.Backoffer, l *Lock, status TxnStat
 			// prevent from scanning the whole region in this case.
 			metrics.LockResolverCountWithResolveLockLite.Inc()
 			lreq.Keys = [][]byte{l.Key}
+		} else if !resultRequired && lr.enableTiKVSideAsyncResolve.Load() {
```
ekexium, line 1430: I suggest checking nextgen here to avoid misuse.

txnkv/txnlock/test_probe.go
```diff
@@ -54,7 +54,7 @@ func (l LockResolverProbe) ResolveAsyncCommitLock(bo *retry.Backoffer, lock *Loc
 
 // ResolveLock resolves single lock.
 func (l LockResolverProbe) ResolveLock(bo *retry.Backoffer, lock *Lock) error {
-	return l.resolveLock(bo, lock, TxnStatus{}, false, make(map[locate.RegionVerID]struct{}))
+	return l.resolveLock(bo, lock, TxnStatus{}, false, true, make(map[locate.RegionVerID]struct{}))
```
ekexium, line 57: Why is this set to true?

txnkv/txnlock/lock_resolver.go
lcwangchao: The current runtime gate is intentionally centralized in NewLockResolver: enableTiKVSideAsyncResolve is initialized from config.NextGen and the resolve path only consumes that flag.

I added a test-only assertion here instead of duplicating config.NextGen in the runtime condition. This keeps enableTiKVSideAsyncResolve as the single source of truth, while still catching accidental misuse in non-NextGen tests. If non-NextGen supports this async mode in the future, the assertion will fail and remind us to remove/update this assumption.

https://github.com/tikv/client-go/pull/1973/changes/21149a03c6c717d37b272a07614fbc1adce0c360

txnkv/txnlock/test_probe.go
lcwangchao: This is intentional to keep the test helper’s previous behavior. Before this PR, LockResolverProbe.ResolveLock was synchronous: nil meant the resolve request completed. Passing resultRequired=true preserves that contract; false could use TiKV-side async resolve in NextGen and make tests less deterministic.

ekexium (approved):

PR #1981, opened 2026-06-02
txnkv: batch lite lock resolves by region

lcwangchao:
ref https://github.com/pingcap/tidb/issues/68352

## What changed

- Batch lite lock resolves by transaction before grouping keys by region.
- Keep the single-key lite resolve path unchanged while resolving multi-key lite locks with per-region key lists.
- Allow read-path lite cleanup to run through the async resolve pool.
- Increase the default `resolve-lock-lite-threshold` from 16 to 512.
- Add integration coverage for both synchronous and read async batch-lite resolve paths.

## Why

The old lite resolve path could send one `ResolveLock` request per lock. With the new batched lite resolve path, increasing the threshold lets more small and medium transactions avoid whole-region resolve without regressing to a long serial sequence of per-lock resolve requests.

## Validation

- `go test ./config`
- `go test . -run 'TestLock/TestBatchLiteResolveLocksByRegion' -count=1`
- `go test . -run 'TestLock/TestResolveLocksForRead|TestBatchLiteResolveLocksByRegion' -count=1`
- `go test ./txnkv/txnlock`
- `git diff --check -- integration_tests/lock_test.go txnkv/txnlock/test_probe.go txnkv/txnlock/lock_resolver.go config/client.go`

## Summary by CodeRabbit

* **Configuration Changes**
  * TiKV client default lock-resolution threshold increased (16 → 256), altering lite-vs-region resolve decision

* **Lock Resolution Improvements**
  * Per-transaction lite-resolution batching, better region grouping, and refined async vs sync dispatch with request metadata preserved

* **Tests**
  * New integration test for lite resolve batching; test helpers added to control and report async resolve tasks

* **Chores**
  * Bumped indirect dependencies in example modules to newer revisions

### DIFF
--- a/txnkv/txnlock/lock_resolver.go
+++ b/txnkv/txnlock/lock_resolver.go
@@ -1357,7 +1357,11 @@ func (lr *LockResolver) checkAllSecondaries(bo *retry.Backoffer, l *Lock, status
 
 func (lr *LockResolver) newAsyncResolveBackoffer(originalBo *retry.Backoffer) *retry.Backoffer {
 	// Async cleanup must not inherit the caller's cancellation, but it should
-	// keep request metadata such as RequestSource for observability.
+	// keep RequestSource for observability.
+	// This intentionally preserves the existing for-read async resolve behavior:
+	// resource group attribution is not copied to the detached cleanup context here.
+	// Whether background lock cleanup should be charged to the triggering request's resource group is a separate
+	// policy decision.
 	asyncCtx := context.WithValue(lr.asyncResolveCtx, util.RequestSourceKey, originalBo.GetCtx().Value(util.RequestSourceKey))
 	return retry.NewBackoffer(asyncCtx, asyncResolveLockMaxBackoff)
 }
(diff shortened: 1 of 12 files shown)


txnkv/txnlock/lock_resolver.go
```diff
@@ -1314,6 +1355,92 @@ func (lr *LockResolver) checkAllSecondaries(bo *retry.Backoffer, l *Lock, status
...
+		logutil.BgLogger().Info("failed to batch resolve lock lite asynchronously",
+			zap.String("lock", l.String()),
+			zap.Strings("keys", keysForLog),
+			zap.Uint64("commitTS", status.CommitTS()),
+			zap.Bool("fallbackSync", fallbackSync),
+			zap.Error(err),
+		)
+	}
+
+	asyncResolveOrNot := func(f func(*retry.Backoffer) error) error {
```
cfzjywxk: The `asyncResolveOrNot` naming is a bit confusing, the `async or not` is not actually returned.

txnkv/txnlock/lock_resolver.go
```diff
@@ -1314,6 +1355,92 @@ func (lr *LockResolver) checkAllSecondaries(bo *retry.Backoffer, l *Lock, status
...
+			if !bytes.Equal(l.Key, keys[0]) {
+				panic(fmt.Sprintf("the only key in keys should be the same as l.Key, but got %v and %v", l.Key, keys[0]))
+			}
+		}
+		return asyncResolveOrNot(func(asyncBo *retry.Backoffer) error {
+			return lr.resolveLock(asyncBo, l, status, true, true, nil)
+		})
+	}
+
+	regions, _, err := lr.store.GetRegionCache().GroupKeysByRegion(bo, keys, nil)
```
cfzjywxk: The old path did key location inside async resolveLock with a detached async backoffer, the behaviour is changed after batching logic, a region-cache/location/backoff error during grouping can return from `ResolveLocksForRead`.

Is it expected change?

txnkv/txnlock/lock_resolver.go
```diff
@@ -1314,6 +1355,92 @@ func (lr *LockResolver) checkAllSecondaries(bo *retry.Backoffer, l *Lock, status
 	return &shared, nil
 }
 
+func (lr *LockResolver) newAsyncResolveBackoffer(originalBo *retry.Backoffer) *retry.Backoffer {
```
cfzjywxk, line 1358: The resouce group souce would be lost in resolving, but it seems to be a pre-exsiting issue unrelated to this PR, perhapse we could add some comments about it.

txnkv/txnlock/lock_resolver.go
lcwangchao: renamed to inplaceOrScheduleAsync

txnkv/txnlock/lock_resolver.go
lcwangchao: I've moved GroupKeysByRegion to async path when forRead is true

txnkv/txnlock/lock_resolver.go
lcwangchao: comment added to this method.

txnkv/txnlock/lock_resolver.go
```diff
@@ -1314,6 +1355,105 @@ func (lr *LockResolver) checkAllSecondaries(bo *retry.Backoffer, l *Lock, status
...
+				metrics.LockResolverCountWithReadAsyncResolveFallback.Inc()
+			}
+		}
+
+		// isAsync is false means either it is not for read or the async resolve
+		// task is rejected. For reads, the lock status is already known, so the
+		// physical cleanup is best-effort even when it runs inline.
+		if !isAsync {
+			if err = f(bo); err != nil && forRead {
+				err = nil
```
cfzjywxk, line 1410: Why not propagate the error to `ResolveLocksForRead` next? Is this intended and do we have test coverage for this path?

lcwangchao: Yes, it is intended, a new test is here to cover it: https://github.com/tikv/client-go/pull/1981/changes/cca53206301ad53a8cf3716bdd956084b8287720

When the asyncPool is full, the forRead resolve will back to inplace-mode, but I think it's better to keep the same behavior when failed when background goroutine

cfzjywxk (approved):
LGTM

ekexium (approved):

PR #1994, opened 2026-06-08
apicodec: complete API v2 key-bearing command coverage

### LINKED ISSUE #1993: apicodec: incomplete API v2 encode/decode handling for key-bearing commands
## Summary

`internal/apicodec` does not consistently cover all key-bearing requests and responses under API v2.

As a result, some commands can miss the expected API v2 key encode/decode step, which can lead to incorrect key handling when keyspace prefixes are involved.

## Why this matters

This is not only a theoretical cleanup issue.

https://github.com/tikv/client-go/pull/1986 fixed one real bug we hit in a PoC. That PR addressed an actual API v2 codec gap and showed that the current coverage is incomplete in practice.

## Problem statement

API v2 handling in `apicodec` is correct for many common paths, but it is not uniformly applied to every command that carries keys or key ranges.

That leaves room for bugs such as:

- prefixed keys leaking past decode paths
- raw keys being sent without the expected API v2 encoding
- similar issues surfacing again in less frequently used paths

## Expected behavior

All key-bearing commands in API v2 should have complete and consistent encode/decode handling in `apicodec`.

This issue is intended to track the problem at a high level. The follow-up PR will describe the concrete fixes and command-by-command details.

wfxr:
## Summary

Closes #1993

This PR completes the missing API v2 encode/decode coverage for key-bearing commands in `internal/apicodec`.

`apicodec` already handled many common API v2 paths correctly, but coverage was still partial. Some commands could bypass request-side key encoding or response-side key decoding, which leaves room for key-format mismatches once keyspace prefixes are involved. That kind of mismatch is a correctness issue rather than a cleanup issue.

This work follows up on https://github.com/tikv/client-go/pull/1986, which fixed a real bug we hit in a PoC. That fix addressed one concrete API v2 codec gap in lock error handling. This PR extends the same principle to the remaining key-bearing command paths so that API v2 handling is complete and consistent at the codec layer.

## Changes

This PR is split into three logical parts.

### 1. Cover the hot paths used by flush and buffered reads

The first part adds API v2 handling for:

- `CmdFlush`
  - encode request `Mutations`
  - encode request `PrimaryKey`
  - decode response `RegionError`
  - decode response `Errors`
- `CmdBufferBatchGet`
  - encode request `Keys`
  - decode response `RegionError`
  - decode response `Error`
  - decode response `Pairs`

These paths are already used by pipelined DML and buffered snapshot reads, so leaving them outside API v2 coverage means the client can still send or consume keys in inconsistent formats.

### 2. Cover the remaining key-bearing command paths in `apicodec`

The second part adds API v2 handling for the remaining commands in this area that still carry keys or key ranges:

- `CmdCheckLockObserver`
- `CmdLockWaitInfo`
- `CmdMvccGetByStartTs`
- `CmdFlashbackToVersion`
- `CmdPrepareFlashbackToVersion`

These commands are not all on the same hot path as flush and buffered reads, but they still cross the API v2 boundary with keys, key ranges, lock information, or region errors. Leaving them partially covered means the client can still send raw keys without the expected API v2 encoding or return storage-encoded keys without decoding them first.

### 3. Add repository guidance for future API v2 codec work

This PR also adds `AGENTS.md` to document repository-specific guidance for future development and review.

In particular, the new guidance makes two expectations explicit:

- `client-go` is shared infrastructure for both TiKV and CSE, and it is consumed by multiple TiDB ecosystem components such as TiDB, TiCDC, BR, and others. Changes in this repository therefore need to be coded and reviewed with downstream compatibility and stability in mind, rather than only validating a single TiDB-facing path.
- API v2 key-bearing commands should be handled symmetrically in `apicodec`. If a request or response carries keys, key ranges, lock information, or region errors with encoded keys, the codec needs complete and consistent handling in both directions. Partial coverage makes it too easy to introduce encode/decode mismatches and correctness bugs.

## Tests

```bash
go test ./internal/apicodec
```

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive repository guidelines on development, testing, coding style, API expectations, and contribution workflow.

* **Improvements**
  * Extended API v2 codec to handle additional command types and improve decoding of key-related fields and region/lock errors, yielding more accurate error reporting for affected operations.

* **Tests**
  * Expanded test coverage to validate request encoding and response decoding across the newly supported command scenarios.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,48 @@
+# Repository Guidelines
+
+## Project Structure & Module Organization
+
+This repository is the Go module `github.com/tikv/client-go/v2`, a Go client library for TiKV and CSE. Public packages live at the repository root, including `tikv`, `rawkv`, `txnkv`, `kv`, `tikvrpc`, `oracle`, `config`, `metrics`, `trace`, and `error`. Implementation-only packages live under `internal/`, such as `internal/client`, `internal/locate`, `internal/apicodec`, `internal/kvrpc`, `internal/mockstore`, and `internal/unionstore`.
+
+Tests are colocated with packages in `*_test.go` files. Broader integration coverage is under `integration_tests/`, with a separate `go.mod`; raw KV integration tests live in `integration_tests/raw/`. Runnable examples are under `examples/`, and several example directories are independent Go modules.
+
+## Build, Test, and Development Commands
+
+- `go test ./...`: run the main module unit tests locally.
+- `go test --tags=intest ./...`: run the full unit-test set used by CI.
+- `go test -race --tags=intest ./...`: run CI-equivalent race tests.
+- `go test ./internal/apicodec`: run targeted API codec tests; use this after changing API v2 request or response key handling.
+- `go generate ./...`: regenerate generated files, then check `git diff` for unexpected output.
+- `golangci-lint run`: run repository linting and formatting checks from `.golangci.yml`.
+- `cd integration_tests && gotestsum --format short-verbose -- ./...`: run local integration tests from the integration module.
+- `cd integration_tests && go test --with-tikv`: run integration tests against local `pd-server` and `tikv-server` binaries after starting them as described in `README.md`.
+
+## Coding Style & Naming Conventions
+
+Use Go 1.25.x and standard Go formatting. Run `gofmt`/`goimports` through `golangci-lint`, or format touched files before submitting. Keep package names short, lowercase, and consistent with existing directories. Exported identifiers must have clear Go doc comments when they are part of the public client API.
+
+Prefer existing error, retry, backoff, logging, metrics, and failpoint helpers instead of introducing parallel mechanisms. Keep public API changes small and deliberate because this module is consumed by multiple TiDB ecosystem components, including TiDB, TiCDC, BR, go-ycsb, and other projects. Design and review for compatibility and stability across these consumers, not only for TiDB.
+
+## API v2 Codec Guidelines
+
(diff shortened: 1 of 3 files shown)


cfzjywxk (approved):

AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+- `go test -race --tags=intest ./...`: run CI-equivalent race tests.
+- `go test ./internal/apicodec`: run targeted API codec tests; use this after changing API v2 request or response key handling.
+- `go generate ./...`: regenerate generated files, then check `git diff` for unexpected output.
+- `golangci-lint run`: run repository linting and formatting checks from `.golangci.yml`.
+- `cd integration_tests && gotestsum --format short-verbose -- ./...`: run local integration tests from the integration module.
+- `cd integration_tests && go test --with-tikv`: run integration tests against local `pd-server` and `tikv-server` binaries after starting them as described in `README.md`.
+
+## Coding Style & Naming Conventions
+
+Use Go 1.25.x and standard Go formatting. Run `gofmt`/`goimports` through `golangci-lint`, or format touched files before submitting. Keep package names short, lowercase, and consistent with existing directories. Exported identifiers must have clear Go doc comments when they are part of the public client API.
```
ekexium, line 22: Should not limit to specific go verions.

AGENTS.md
```diff
@@ -0,0 +1,48 @@
+# Repository Guidelines
+
+## Project Structure & Module Organization
```
ekexium, line 3: These info can easily found by agents. I don't think we need to describe them in the prompt.

ekexium (approved):
LGTM overall. One nit: I’d prefer to make AGENTS.md more concise, keeping only repo-specific guidance such as the API v2 codec rules

zyguan (approved):

AGENTS.md
wfxr: Makes sense, but the PR has already been automatically merged. I’ll fold this into my next client-go PR (rather than opening a separate one just for it).

AGENTS.md
wfxr: Same as above

PR #1998, opened 2026-06-09
transaction: Support file based transaction

pingyu:
### Changes

- Cherry pick "file based transaction" PRs to master:
  - #1119
  - https://github.com/tikv/client-go/pulls?q=is%3Apr++is%3Amerged+author%3Apingyu+txn-file+

- Integration test in next PR: https://github.com/tikv/client-go/pull/2038

## Summary by CodeRabbit

* **New Features**
  * Added file-based transaction processing with chunked uploads, concurrent writes, retries, lock resolution, and region-aware batching.
  * Added configuration options for chunk sizing, writer concurrency, RU discounting, and request-source filtering.
  * Added an option to disable file-based transactions and use standard 2PC.
* **Bug Fixes**
  * Improved split-region handling by resolving split-key locks before continuing.
* **Observability**
  * Added Prometheus metrics for transaction-file requests, bytes, mutation sizes, and durations.
* **Tests**
  * Expanded coverage for sizing, retries, region changes, configuration validation, and request selection.


txnkv/transaction/2pc.go
```diff
@@ -337,6 +340,38 @@ type CommitterMutations interface {
...
+				break
+			}
+
+			pos++
+			if !isInRange(pos) {
+				break
+			}
+		}
+		return firstDataKey, true
+	}
```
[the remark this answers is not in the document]
pingyu: The `MutationsHasDataInRange` is expected to return `true` as long as the mutations has data in the range.
The `firstDataKey` is used to build sample key, and can be empty.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1312 @@
...
+		if r.err != nil {
+			if returnEarly {
+				return regionErrChunks, r.err
+			} else if err == nil {
+				err = r.err
+			}
+		} else if r.regionErrSlice != nil {
+			regionErrChunks.appendSlice(r.regionErrSlice)
+		}
+	}
```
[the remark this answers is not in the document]
pingyu: `rateLimiter.GetToken(exitCh)` return `false` only when `exitCh` is closed, so the `executeTxnFileSlice` must reach fast fail to returned.

It should not be deadlock.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1312 @@
...
+		loc, err := c.LocateKey(bo, startKey)
+		if err != nil {
+			logutil.Logger(bo.GetCtx()).Error("locate key failed", zap.Error(err), zap.String("startKey", redact.Key(startKey)))
+			return nil, nil, errors.Wrap(err, "locate key failed")
+		}
+		firstKey, ok := MutationsHasDataInRange(
+			mutations,
+			util.GetMaxStartKey(r.smallest, loc.StartKey),
+			util.GetMinEndKey(exclusiveBiggest, loc.EndKey),
+		)
```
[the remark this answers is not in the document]
pingyu: Fixed.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1312 @@
...
+		if resp.Resp == nil {
+			return nil, errors.WithStack(tikverr.ErrBodyMissing)
+		}
+		prewriteResp := resp.Resp.(*kvrpcpb.PrewriteResponse)
+		regionErr := prewriteResp.GetRegionError()
+		if regionErr != nil {
+			// For other region error and the fake region error, backoff because
+			// there's something wrong.
+			// For the real EpochNotMatch error, don't backoff.
+			if regionErr.GetEpochNotMatch() == nil || locate.IsFakeRegionError(regionErr) {
```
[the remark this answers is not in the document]
pingyu: Fixed.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1312 @@
...
+	if discountRatio > 0.0 && discountRatio < 1.0 {
+		writeBytes = int64(float64(writeBytes) * discountRatio)
+	}
+
+	reqInfo := resourcecontrol.NewRequestInfo(
+		writeBytes,
+		region.GetLeaderStoreID(),
+		replicaNumber,
+		false,
+	)
```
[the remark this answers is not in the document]
pingyu: Fixed.

txnkv/transaction/txn.go
```diff
@@ -787,6 +790,10 @@ func (txn *KVTxn) GetScope() string {
 	return txn.scope
 }
 
+func (txn *KVTxn) DisableTxnFile() {
+	txn.disableTxnFile = true
+}
```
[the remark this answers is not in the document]
pingyu: Fixed.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1312 @@
...
+			logutil.Logger(ctx).Warn("service error", zap.String("http status", resp.Status), zap.String("body", bodyStr))
+			err = bo.Backoff(retry.BoTiKVServerBusy, fmt.Errorf("service error, http status %s", resp.Status))
+			if err != nil {
+				return nil, errors.WithStack(err)
+			}
+			continue
+		}
+		data, err := io.ReadAll(resp.Body)
+		return data, errors.WithStack(err)
+	}
```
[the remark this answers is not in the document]
pingyu: Fixed.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1312 @@
...
+
+			if inflightChunks >= concurrency {
+				r := <-resultCh
+				if r.err != nil {
+					logutil.Logger(bo.GetCtx()).Error(buildChunkErrMsg, zap.Error(err))
+					return errors.Wrap(r.err, buildChunkErrMsg)
+				}
+				results = append(results, r)
+				inflightChunks -= 1
+			}
```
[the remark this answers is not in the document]
pingyu: Fixed.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1312 @@
...
+		})
+		if errGo != nil {
+			c.store.WaitGroup().Done()
+			logutil.Logger(bo.GetCtx()).Warn("txn file: create goroutine failed",
+				zap.Uint64("startTS", c.startTS),
+				zap.Stringer("action", action),
+				zap.Error(errGo))
+		}
+		return nil
```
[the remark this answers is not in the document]
pingyu: Fixed.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1326 @@
...
+
+	prewriteBo := retry.NewBackofferWithVars(ctx, int(PrewriteMaxBackoff.Load()), c.txn.vars)
+	err = c.executeTxnFileAction(prewriteBo, c.txnFileCtx.slice, txnFilePrewriteAction{})
+	stepDone("prewrite")
+	if err != nil {
+		return
+	}
+
+	commitBo := retry.NewBackofferWithVars(ctx, int(CommitMaxBackoff), c.txn.vars)
+	c.commitTS, err = c.store.GetTimestampWithRetry(commitBo, c.txn.GetScope())
```
wfxr: Should the txn-file path use the same commitTS preparation/check path as normal 2PC here?

Normal 2PC gets the commitTS through `txn.GetTimestampForCommit(...)`, then checks schema validity, `MaxTxnTimeUse`, and `commitTSUpperBoundCheck` before committing. This path calls `store.GetTimestampWithRetry(...)` directly, and the `CommitTsExpired` retry path does the same, so it seems to bypass `SetCommitWaitUntilTSO`, schema lease validation, and commitTS upper-bound constraints.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1326 @@
...
+	}
+
+	err = c.preSplitTxnFileRegions(buildBo)
+	stepDone("pre-split")
+	if err != nil {
+		return
+	}
+
+	prewriteBo := retry.NewBackofferWithVars(ctx, int(PrewriteMaxBackoff.Load()), c.txn.vars)
+	err = c.executeTxnFileAction(prewriteBo, c.txnFileCtx.slice, txnFilePrewriteAction{})
```
wfxr, line 789: Should txn-file either handle binlog the same way as normal 2PC, or fall back when `c.shouldWriteBinlog()` is true?

The normal 2PC path calls `binlog.Prewrite(...)` before KV prewrite and later calls `binlog.Commit(...)` or `binlog.Skip(...)`. I don't see equivalent handling in `executeTxnFile`, and `useTxnFile` does not appear to exclude transactions with a binlog executor. This may cause a large txn-file transaction to commit to TiKV without producing the corresponding binlog.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1326 @@
...
+	})
+	markTxnFileRetryRequest(req, bo)
+	sender := locate.NewRegionRequestSender(c.store.GetRegionCache(), c.store.GetTiKVClient(), c.store.GetOracle())
+	for {
+		resp, _, err := sender.SendReq(bo, req, batch.region.Region, client.ReadTimeoutMedium)
+		if batch.isPrimary && sender.GetRPCError() != nil {
+			c.setUndeterminedErr(errors.WithStack(sender.GetRPCError()))
+		}
+		// Unexpected error occurs, return it.
+		if err != nil {
```
wfxr, line 552: Should this return `ErrResultUndetermined` for primary txn-file commit RPC errors, matching normal 2PC?

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1326 @@
...
+		ResourceGroupTag:       c.resourceGroupTag,
+		DiskFullOpt:            c.diskFullOpt,
+		TxnSource:              c.txnSource,
+		MaxExecutionDurationMs: uint64(client.ReadTimeoutMedium.Milliseconds()),
+		RequestSource:          c.txn.GetRequestSource(),
+		ResourceControlContext: &kvrpcpb.ResourceControlContext{
+			ResourceGroupName: c.resourceGroupName,
+		},
+	})
+	markTxnFileRetryRequest(req, bo)
```
wfxr, line 355: Is the txn-file path expected to preserve dynamic resource-group tags from `resourceGroupTagger`?

The normal prewrite/commit/rollback paths call `c.resourceGroupTagger(req)` when `c.resourceGroupTag == nil && c.resourceGroupTagger != nil`. I don't see the same handling for txn-file prewrite, commit, or rollback requests, so a transaction may lose its dynamic resource-group tag once it crosses the txn-file threshold.

metrics/metrics.go
```diff
@@ -1055,6 +1060,44 @@ func initMetrics(namespace, subsystem string, constLabels prometheus.Labels) {
 			ConstLabels: constLabels,
 		}, []string{LblResult})
 
+	TiKVTxnFileRequestCounter = prometheus.NewCounterVec(
```
wfxr, line 1064: Looks like the new txn-file metrics may be missing from `RegisterMetrics()`?

config/client.go
```diff
@@ -121,6 +121,21 @@ type TiKVClient struct {
 
 	// RUV2 is the RU v2 TiKV-side weights used to calculate TiKV RU values from ExecDetailsV2.RuV2.
 	RUV2 RUV2TiKVConfig `toml:"ru-v2" json:"ru-v2"`
+
+	// TxnChunkWriterAddr is the address of the txn chunk writer for file-based txn.
+	TxnChunkWriterAddr string `toml:"txn-chunk-writer-addr" json:"txn-chunk-writer-addr"`
+	// TxnChunkWriterConcurrency is the concurrency to request the txn chunk writer for file-based txn.
+	TxnChunkWriterConcurrency uint `toml:"txn-chunk-writer-concurrency" json:"txn-chunk-writer-concurrency"`
+	// TxnChunkMaxSize is the maximum size of a txn chunk of file-based txn.
+	TxnChunkMaxSize uint64 `toml:"txn-chunk-max-size" json:"txn-chunk-max-size"`
```
wfxr, line 132: Maybe we should validate the new txn-file config values before accepting them, especially `TxnChunkMaxSize > 0`? If this is set to `0`, the txn-file path can divide by zero when calculating chunk counts or parallelism, so rejecting or normalizing it in `Valid()` would make the failure mode clearer.

util/misc.go
```diff
@@ -175,3 +176,23 @@ func None[T interface{}]() Option[T] {
...
+		return lhs
+	}
+	if len(lhs) == 0 {
+		return rhs
+	}
+	if bytes.Compare(lhs, rhs) < 0 {
+		return lhs
+	}
+	return rhs
+}
```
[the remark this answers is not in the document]
pingyu: Comment addressed. PTAL.
https://github.com/tikv/client-go/pull/1998/commits/ace38d1b7b008ba0065aaf432148315b2e5227b5

txnkv/transaction/txn_file_test.go
```diff
@@ -0,0 +1,805 @@
...
+	pd := &mockPDClient{}
+	regionCache := locate.NewTestRegionCache()
+	regionCache.SetPDClient(pd)
+	defer regionCache.Close()
+
+	chunkWriter := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		require.Equal(t, http.MethodPost, r.Method)
+		_, err := w.Write([]byte(`{"chunk_id":1}`))
+		require.NoError(t, err)
+	}))
```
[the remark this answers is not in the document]
pingyu: Comment addressed. PTAL.
https://github.com/tikv/client-go/pull/1998/commits/4f653f2847e64c0b7b1036d1496f1ac5a5204544

txnkv/transaction/txn_file.go
pingyu: Fixed in 44fa21205d27ee447dc95e6ba9626897762c9854.

txnkv/transaction/txn_file.go
pingyu: Fixed in d50071dc597f8d61582ec4da8206fefbee418576.

As the txn file feature is support in TiDBCloud only, simply disable txn file when "should write binlog".

txnkv/transaction/txn_file.go
pingyu: Fixed in 26bd55b00e3ba8920ef6857ba1d587b290c5570e.

txnkv/transaction/txn_file.go
pingyu: Fixed in 70cee6ea7398774946af3af0d836ddc8d209a7ea & b15c470e282c77f9184463744642936e671fb0d1.

metrics/metrics.go
pingyu: Fixed in dd46c190e4cdd1f0b19f11955c35ac05edf0010a.

config/client.go
pingyu: Fixed in 06541e7b8cacc3aea9352e170b42fe0f34d5d29e.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1422 @@
...
+	asyncExecuteSecondaries() bool
+	String() string
+}
+
+type txnFilePrewriteAction struct{}
+
+var _ txnFileAction = (*txnFilePrewriteAction)(nil)
+
+func (a txnFilePrewriteAction) executeBatch(c *twoPhaseCommitter, bo *retry.Backoffer, batch chunkBatch) (*tikvrpc.Response, error) {
+	primaryLock := c.txnFileCtx.slice.chunkRanges[0].smallest
```
wfxr: **P2: Should not use the minimum key of the first chunk as the transaction primary**

PrimaryLock cannot always be derived from the first chunk. initKeysAndMutations skips CheckNotExists/SharedLock when choosing c.primaryKey, so a smaller prewrite-only mutation can precede the real primary. In that case this either fails with primary out of first batch, or makes the write locks point to a key that has no lock. Please use c.primary() and explicitly select the batch containing it; the pos == 0 assumption in MutationsHasDataInRange needs the same treatment.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1422 @@
...
+	minMutationSize := c.txn.vars.TxnFileMinMutationSize
+	if minMutationSize == 0 {
+		minMutationSize = conf.TiKVClient.TxnFileMinMutationSize
+	}
+	if c.txn.isInternal() {
+		// Relax the requirement for internal requests.
+		minMutationSize = minMutationSize / 2
+	}
+
+	if c.txn.isPessimistic ||
```
wfxr, line 1192: **P1: Should exclude pipelined transactions from the txn-file path**

Commit intentionally skips initKeysAndMutations for a pipelined transaction, so committer.mutations remains nil. If this gate succeeds, buildTxnFiles dereferences it at mutations.GetKey(0) and panics. More importantly, mutations already flushed by pipelined DML are no longer available to serialize into txn files. Please exclude c.txn.isPipelined here unless the txn-file path is extended to handle flushed mutations explicitly.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1422 @@
...
+		var locks []*txnlock.Lock
+		for _, keyErr := range keyErrs {
+			// Check already exists error
+			if alreadyExist := keyErr.GetAlreadyExist(); alreadyExist != nil {
+				e := &tikverr.ErrKeyExist{AlreadyExist: alreadyExist}
+				return nil, c.extractKeyExistsErr(e)
+			}
+
+			// Extract lock from key error
+			lock, err1 := txnlock.ExtractLockFromKeyErr(keyErr)
```
wfxr: **P1: Should expand shared-lock wrappers before resolving locks**

ExtractLockFromKeyErr returns the shared-lock wrapper when SharedLockInfos is present, but ResolveLocksWithOpts explicitly rejects Lock.IsShared() and requires callers to resolve the inner locks one by one. As a result, txn-file prewrite fails whenever it encounters a shared lock. Please mirror or factor the normal prewrite extraction logic here; the same issue exists in handleSplitRegionKeyErrors, and this path should also preserve NoResolvePolicy.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1422 @@
...
+
+	defer func() {
+		// Always clean up all written keys if the txn does not commit.
+		c.mu.RLock()
+		committed := c.mu.committed
+		undetermined := c.mu.undeterminedErr != nil
+		c.mu.RUnlock()
+		if !committed && !undetermined {
+			if c.txnFileCtx.slice.Len() > 0 {
+				err1 := c.executeTxnFileAction(retry.NewBackofferWithVars(ctx, int(CommitMaxBackoff), c.txn.vars), c.txnFileCtx.slice, txnFileRollbackAction{})
```
wfxr: **P2: Should not reuse the commit context for rollback cleanup**

If prewrite succeeds and the caller then cancels or times out the commit, this context is already canceled, so the rollback RPC cannot be sent and the locks remain until TTL cleanup. Normal 2PC derives cleanup from c.store.Ctx() for this reason. Please use an independent store-scoped cleanup context here, preferably through the store pool.

config/client.go
```diff
@@ -246,6 +267,22 @@ func (config *TiKVClient) Valid() error {
...
 		return fmt.Errorf("grpc-keepalive-timeout should be at least 0.05, but got %f", config.GrpcKeepAliveTimeout)
 	}
+	return validateTxnFileConfig(config)
+}
+
+func validateTxnFileConfig(config *TiKVClient) error {
+	if config.TxnChunkMaxSize == 0 {
+		return fmt.Errorf("txn-chunk-max-size should be greater than 0")
+	}
+	if config.TxnChunkMaxSize > math.MaxInt {
```
wfxr, line 283: **P2: Should reject chunk sizes that make the parallelism limit zero**

This validation still accepts TxnChunkMaxSize values greater than MaxTxnChunkSizeInParallel (4 GiB). For such a valid value, 4 GiB / TxnChunkMaxSize becomes 0 and NewRateLimit(0) deadlocks the multi-secondary path. Please reject values above the parallel budget and defensively clamp the calculated rate limit to at least 1.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1422 @@
...
+			zap.Error(err),
+			zap.String("requestSource", c.txn.GetRequestSource()),
+			zap.Stringers("steps", steps))
+	}()
+
+	logutil.Logger(ctx).Debug("execute txn file", zap.Uint64("startTS", c.startTS))
+
+	buildBo := retry.NewBackofferWithVars(ctx, int(BuildTxnFileMaxBackoff.Load()), c.txn.vars)
+
+	rcInterceptor := client.ResourceControlInterceptor.Load()
```
wfxr: **P2: Should honor the normal resource-control enable and bypass checks**

Loading ResourceControlInterceptor directly bypasses ResourceControlSwitch and the normal empty-group, background-request, and RequestInfo.Bypass checks. DisableResourceControl leaves the interceptor pointer installed, so txn-file transactions can still be charged—or fail while looking up an empty resource group—when normal RPCs would skip resource control. Please reuse or factor the same selection logic used by getResourceControlInfo before applying the aggregate txn-file charge.

txnkv/transaction/txn_file.go
wfxr: applyTxnFilePrewriteResourceGroupTagger skips the tagger entirely when sampleKeys is empty. This loses more than the key-derived table/row label: TiDB's ResourceGroupTagBuilder still encodes the keyspace, SQL digest, and plan digest when the key is empty. The current test explicitly expects the tag to be absent for such a batch, so this remains different from normal 2PC. Please invoke the tagger even when no representative key is available; only the key-derived fields need to be omitted.

config/client.go
```diff
@@ -246,6 +267,22 @@ func (config *TiKVClient) Valid() error {
 	if config.GetGrpcKeepAliveTimeout() < time.Millisecond*50 {
 		return fmt.Errorf("grpc-keepalive-timeout should be at least 0.05, but got %f", config.GrpcKeepAliveTimeout)
 	}
+	return validateTxnFileConfig(config)
```
wfxr, line 272: **P2: Could txn-file validation be skipped while the feature is disabled?**

`TiKVClient.Valid` now validates `TxnChunkMaxSize` and `TxnChunkWriterConcurrency` even when `TxnChunkWriterAddr` is empty, which currently means txn-file is disabled. Is this intended?

An existing caller that constructs `TiKVClient` directly, or decodes an old configuration into a zero-valued struct, can now fail validation without using txn-file at all. Callers starting from `DefaultTiKVClient` are unaffected because these fields already have defaults, but client-go consumers are not necessarily required to do that.

Would it make sense to validate these txn-file-only fields only when `TxnChunkWriterAddr` is configured, or normalize the zero values before validation? A feature-off compatibility test using an old-style configuration would also help preserve this behavior.

tikv/split_region.go
```diff
@@ -179,6 +180,20 @@ func (s *KVStore) batchSendSingleRegion(bo *Backoffer, batch kvrpc.Batch, scatte
 	}
 
 	spResp := resp.Resp.(*kvrpcpb.SplitRegionResponse)
+
+	keyErrs := spResp.GetErrors()
```
wfxr: **P1: Is the lock-aware SplitRegions behavior intended for all callers?**

This handling is added to the generic `KVStore.SplitRegions` path, so it also affects existing external callers and normal 2PC's best-effort pre-split, even when txn-file is not used.

Previously, a split request that encountered a locked key returned promptly and normal 2PC continued with prewrite. With this change, the same request can resolve locks, sleep for an active lock's TTL, and retry within the split backoff budget. It can also return an error for a shared-lock wrapper because `ResolveLocksWithOpts` does not accept shared locks directly.

Was this intended as a general behavior change for `SplitRegions`, or is it only needed by txn-file pre-split? If it is txn-file-specific, could we keep the lock-aware retry in that path? If it is intended as a generic fix, separating it from the feature and adding compatibility coverage for active, expired, and shared locks may make the behavior change easier to evaluate.

txnkv/transaction/txn_file.go
wfxr, line 1192: Could we also exclude `c.hasSharedLocks` here? `isPessimistic` covers the normal shared-lock path today, but `hasSharedLocks` is the post-init capability signal and is already used to exclude async commit and 1PC. Since txn-file prewrite currently does not support `Op_SharedLock`, relying on the transaction mode alone makes this boundary implicit. An eligibility test with `hasSharedLocks` set would also document that these transactions always fall back to normal 2PC.

txnkv/transaction/txn_file.go
wfxr: One more detail when mirroring the normal prewrite path: the newer-TS/write-conflict check needs to use each inner lock’s `TxnID`, and `RecordResolvingLocks`, `UpdateResolvingLocks`, and `ResolveLocksWithOpts` should all receive the expanded inner locks rather than the wrapper.

Could we also add a multi-holder test with one expired holder and one active holder, verifying that cleanup removes only the expired holder while the active holder remains and keeps the prewrite waiting? Reusing the normal prewrite resolver options, including `PessimisticRegionResolve`, would also keep the cleanup behavior and RPC profile aligned.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1422 @@
...
+var _ txnFileAction = (*txnFilePrewriteAction)(nil)
+
+func (a txnFilePrewriteAction) executeBatch(c *twoPhaseCommitter, bo *retry.Backoffer, batch chunkBatch) (*tikvrpc.Response, error) {
+	primaryLock := c.txnFileCtx.slice.chunkRanges[0].smallest
+	req := tikvrpc.NewRequest(tikvrpc.CmdPrewrite, &kvrpcpb.PrewriteRequest{
+		StartVersion:   c.startTS,
+		PrimaryLock:    primaryLock,
+		LockTtl:        c.lockTTL,
+		MaxCommitTs:    c.maxCommitTS,
+		AssertionLevel: kvrpcpb.AssertionLevel_Off,
```
wfxr, line 339: **P1: Preserve strict mutation assertions in the txn-file path, or fall back to normal 2PC**

Normal prewrite copies `IsAssertExists` / `IsAssertNotExist` into each `Mutation.Assertion` and sends `c.txn.assertionLevel`. The txn-file path instead serializes only key/op/value into the chunks and unconditionally sends `AssertionLevel_Off`; `useTxnFile` also does not exclude transactions that have strict assertions. The server cannot reconstruct these assertions from the chunk format.

This is observable for optimistic TiDB DML: update and delete mutations carry `AssertExist`, and NextGen defaults `tidb_txn_assertion_level` to `STRICT`. Once such a transaction crosses the txn-file size threshold, an invariant violation that normal 2PC would return as `ErrAssertionFailed` can be prewritten and committed without that check, provided there is no other MVCC error. `Op_Insert` does not cover this case because update/delete remain `Put`/`Del` with an independent assertion.

Please either extend the txn-chunk protocol and server implementation to preserve `AssertExist` / `AssertNotExist` together with the request assertion level, or conservatively fall back to normal 2PC when `c.txn.assertionLevel != AssertionLevel_Off` and any mutation has an Exist/NotExist assertion. Please also add a real txn-file integration test that intentionally violates `AssertExist` or `AssertNotExist` and expects `ErrAssertionFailed`.

txnkv/transaction/txn_file.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/8b878c4428756fa26dd618d3bd16796973e1c0fc

txnkv/transaction/txn_file.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/b30dcc8b5417dbcc22650f6dcc015db03110cd0e.

The change also remove the "Always return primary key if it's in the range". I check server side codes, and find that the primary is not necessary. The keys are used to check transaction committed/rollback on commit/rollback. So `isOpForWrite" should be enough.

txnkv/transaction/txn_file.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/83ca9736ed8c549f5eb600f2a9de122ab5f4c627

txnkv/transaction/txn_file.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/88763ffa550bcd353e9746996a16edded0185134

txnkv/transaction/txn_file.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/d82de8faab1b256caa8ce541470ac5f59b6dd4b9

config/client.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/ebab9a900a9a0b5090bcf662c46d75c410600ef0

txnkv/transaction/txn_file.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/fdca77b0d23eb70895e1ec6a4fd0e0c1c3f4fbe2

config/client.go
pingyu: Fixed by https://github.com/tikv/client-go/pull/1998/changes/b8b8131ab918c766e5c1ec5fd34442e550e48b52

tikv/split_region.go
pingyu: Fixed by 0d123892765b316a5d7a6fc60f26335b43ae9d8c

txnkv/transaction/txn_file.go
pingyu: Fixed by 9b6b05ce6247cedfc358902ac4b566f61ab3c195

txnkv/transaction/txn_file.go
pingyu: Fixed by f0a60e91690de616fce7b5808a6985642c693b52

wfxr: The fallback in `f0a60e9` looks good. Could we also add the integration coverage mentioned above? An integration test with txn-file enabled that violates `AssertExist` or `AssertNotExist` and verifies `Commit` returns `ErrAssertionFailed` would help confirm the fallback works end to end. Ideally, it could also check that no txn chunk is written.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1477 @@
...
+			logutil.Logger(ctx).Warn("txn file commit result undetermined",
+				zap.Error(err),
+				zap.NamedError("rpcErr", undeterminedErr),
+				zap.Uint64("txnStartTS", c.startTS))
+			err = errors.WithStack(tikverr.ErrResultUndetermined)
+		}
+		return
+	}
+
+	err = c.afterExecuteTxnFile(rcInterceptor, reqInfo, ruDetails)
```
wfxr: **P2: Do not return a fallible custom resource-accounting error after the primary has committed**

When resource control is enabled for a non-empty, non-background, non-bypassed resource group, `executeTxnFileAction` has already received a definitive primary-commit success and set `c.mu.committed` before this call. A custom interceptor installed through `tikv.SetResourceControlInterceptor` is allowed to return an error from `OnResponse`; propagating that error reports the transaction as failed even though its data is committed, which can make the caller retry an already committed transaction. The current built-in PD controller returns `nil` from `OnResponse`, so the default implementation is not affected today, but the public custom-interceptor path can still produce this contradictory outcome. Please preserve the successful commit result—for example, log the post-commit accounting failure instead of returning it—and add a test where `OnResponse` returns an error after a successful primary commit.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1477 @@
...
+		},
+	})
+	c.applyTxnFileResourceGroupTagger(req)
+	markTxnFileRetryRequest(req, bo)
+	sender := locate.NewRegionRequestSender(c.store.GetRegionCache(), c.store.GetTiKVClient(), c.store.GetOracle())
+	resp, _, err1 := sender.SendReq(bo, req, batch.region.Region, client.ReadTimeoutShort)
+	if err1 != nil {
+		return nil, err1
+	}
+	return resp, nil
```
wfxr, line 670: **P2: Propagate key errors from the primary rollback**

This returns success for every non-transport response, while the primary path never calls `extractKeyError`; only the secondary/common batch path does. Consequently, a `BatchRollbackResponse.Error` from the primary region is silently treated as successful cleanup. This masks a failed primary cleanup and, depending on the returned key error, transaction locks may remain until TTL-based cleanup; some errors, such as an already-committed transaction, do not imply a remaining lock. Please validate the rollback response here—or handle `extractKeyError` in the common primary path—and add primary-rollback key-error coverage.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1477 @@
...
+func getHTTPClient() (*http.Client, error) {
+	once.Do(func() {
+		var (
+			cfg       = config.GetGlobalConfig()
+			timeout   = time.Duration(BuildTxnFileMaxBackoff.Load()) * time.Millisecond
+			tlsConfig *tls.Config
+		)
+
+		scheme = "http://"
+		transport := &http.Transport{
```
wfxr, line 1373: **P2: Release the txn-file uploader idle HTTP connections**

This process-global custom transport has `IdleConnTimeout == 0`, so it imposes no client-side idle expiry, and it is never passed to `CloseIdleConnections`. Consequently, txn-file sockets can survive `KVStore.Close` until the peer or process closes them. The retention is bounded by `MaxIdleConnsPerHost` (20 for a stable single writer address) and `MaxIdleConns` (100 globally across hosts), but it still violates the client lifecycle and can retain stale descriptors across stores or writer-address changes. #2038 already adds the required close hook; please either pull that lifecycle fix into this PR or make the merge/release dependency explicit. A bounded `IdleConnTimeout` would also provide a useful fallback.

txnkv/transaction/txn_file.go
```diff
@@ -0,0 +1,1477 @@
...
+		}
+
+		secondaryBo := retry.NewBackofferWithVars(c.store.Ctx(), CommitSecondaryMaxBackoff, c.txn.vars)
+		if c.store.IsClose() {
+			logutil.Logger(bo.GetCtx()).Warn("the store is closed",
+				zap.Uint64("startTS", c.startTS), zap.Uint64("commitTS", c.commitTS),
+				zap.Stringer("action", action))
+			return nil
+		}
+		errGo := c.txn.spawnWithStorePool(func() {
```
wfxr, line 1062: **P2: Release MemDB values before committing**

Unlike regular 2PC, the txn-file path does not call DiscardValues after the values have been uploaded. The asynchronous secondary closure retains the committer, and therefore the entire MemDB, until it finishes. Since txn-file transactions are large and secondary retries may take minutes, concurrent commits can significantly increase memory usage or cause OOM.

Please discard the values before committing the primary, once they are no longer needed. The keys and metadata required by secondary retries will remain available.

txnkv/transaction/txn_file.go
pingyu: Integration coverage added by fd466357bd10185a4f749fef2379cdd04fd3dc3f.

txnkv/transaction/txn_file.go
pingyu: Fixed in 0bb1ee0485b42c900ce54d65285336ee6c4591ca

txnkv/transaction/txn_file.go
pingyu: Fixed in 8cf5a023b53d759615261cbb8c396b9de2204d14

txnkv/transaction/txn_file.go
pingyu: Fixed in 98432af4047e114f62bf7e735e6a6d6000f8a789

txnkv/transaction/txn_file.go
pingyu: Fixed in 18258918aba9f5fea72d995db360f806a4ff7635

wfxr (approved):

cfzjywxk (approved):

PR #2006, opened 2026-06-16
fix(tikvrpc): attach context to lock wait info requests

### LINKED ISSUE #2005: CmdLockWaitInfo requests do not carry keyspace context in API v2
## Problem

`CmdLockWaitInfo` requests are missing from the `tikvrpc` request-context attachment list.

For normal KV requests, `AttachContext` attaches the region and keyspace context to the underlying protobuf request before it is sent to TiKV. `CmdLockWaitInfo` also needs this context because it is used to read lock-wait entries from TiKV's wait table.

In API v2 / next-gen TiKV, the keyspace context is required to read the wait-table entries for the caller's keyspace. Without it, callers such as TiDB's `INFORMATION_SCHEMA.DATA_LOCK_WAITS` can miss lock waits that are present in storage.

## Impact

`DATA_LOCK_WAITS` can be incomplete when it relies on TiKV wait-table data.

This is not limited to shared locks. The same issue can affect ordinary pessimistic lock waits when the result must come from the storage wait table instead of TiDB's local resolving-lock fallback.

## Reproduction

This was reproduced with TiDB RealTiKV tests on next-gen TiKV by forcing `DATA_LOCK_WAITS` to read from the storage wait table and skipping TiDB's local lock-resolver fallback.

The query should return the waiting transaction for the current keyspace, but it returns no row when `CmdLockWaitInfo` is sent without request context.

## Expected behavior

`CmdLockWaitInfo` should carry the same region and keyspace context as other KV requests that go through `AttachContext`.

## Related context

`CmdLockWaitInfo` response key decoding is already covered by `internal/apicodec` tests. The missing part is request-side context attachment in `tikvrpc`.

wfxr:
## Summary

Fixes #2005

`CmdLockWaitInfo` is used by callers such as TiDB's `INFORMATION_SCHEMA.DATA_LOCK_WAITS` to read TiKV wait-table entries. In API v2 / next-gen TiKV, those requests need the same region and keyspace context as other KV requests.

Without the request context, the wait-table query can miss waits in the caller's keyspace.

This PR makes `CmdLockWaitInfo` participate in `tikvrpc.AttachContext` and adds regression coverage at the `tikvrpc` boundary.

## Changes

- Add `LockWaitInfo` to the generated `tikvrpc` command list that supports request-context attachment.
- Regenerate `tikvrpc/cmds_generated.go` so `CmdLockWaitInfo`:
  - is treated as a valid request type for context attachment
  - attaches context to `GetLockWaitInfoRequest`
  - follows the same reattach / copy-on-write behavior as other KV requests
- Add unit coverage for:
  - first-time context attachment
  - repeated context attachment on an existing request
  - `CmdLockWaitInfo` alongside a baseline `CmdGet` request

## Why

`DATA_LOCK_WAITS` needs complete wait-table visibility in API v2 / next-gen deployments. The request-side keyspace context determines which keyspace TiKV should read wait entries from.

Response key decoding for `CmdLockWaitInfo` is already covered in `internal/apicodec`; this PR complements that by fixing and covering the request-side context path.

## Tests

```bash
go test ./tikvrpc -run '^TestAttachContextSetsRequestContext$' -count=1
ok  	github.com/tikv/client-go/v2/tikvrpc	0.008s
```

## Summary by CodeRabbit

* **New Features**
  * Added support for lock wait information requests in command processing.

* **Tests**
  * Expanded test coverage for context attachment across multiple request types.

### DIFF
--- a/tikvrpc/cmds_generated.go
+++ b/tikvrpc/cmds_generated.go
@@ -269,6 +269,15 @@ func patchCmdCtx(req *Request, cmd CmdType, ctx *kvrpcpb.Context) bool {
 			req.Req = &cmd
 		}
 		req.rev++
+	case CmdLockWaitInfo:
+		if req.rev == 0 {
+			req.LockWaitInfo().Context = ctx
+		} else {
+			cmd := *req.LockWaitInfo()
+			cmd.Context = ctx
+			req.Req = &cmd
+		}
+		req.rev++
 	case CmdCop:
 		if req.rev == 0 {
 			req.Cop().Context = ctx
@@ -442,6 +451,8 @@ func isValidReqType(cmd CmdType) bool {
 	case CmdRemoveLockObserver:
 		return true
 	case CmdPhysicalScanLock:
+		return true
+	case CmdLockWaitInfo:
 		return true
 	case CmdCop:
 		return true
--- a/tikvrpc/gen.sh
+++ b/tikvrpc/gen.sh
@@ -56,6 +56,7 @@ cmds=(
   CheckLockObserver
   RemoveLockObserver
   PhysicalScanLock
+  LockWaitInfo
   Cop
   BatchCop
   MvccGetByKey
--- a/tikvrpc/tikvrpc_test.go
+++ b/tikvrpc/tikvrpc_test.go
@@ -121,6 +121,58 @@ func TestDefaultRequestOrigin(t *testing.T) {
 	require.Equal(t, kvrpcpb.RequestOrigin_RequestOriginTiDB, req.GetRequestOrigin())
 }
 
+func TestAttachContextSetsRequestContext(t *testing.T) {
+	rpcCtx := kvrpcpb.Context{
+		RegionId:     123,
+		ApiVersion:   kvrpcpb.APIVersion_V2,
+		KeyspaceId:   456,
+		KeyspaceName: "test-keyspace",
+	}
+	nextRPCtx := kvrpcpb.Context{
+		RegionId:     789,
+		ApiVersion:   kvrpcpb.APIVersion_V2,
+		KeyspaceId:   101112,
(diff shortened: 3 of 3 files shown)


cfzjywxk (approved):

ekexium (commented):
The fix itself LGTM. But I'm wondering can we avoid such mistakes next time? It looks really easy to miss modifying the script.

wfxr:
> The fix itself LGTM. But I'm wondering can we avoid such mistakes next time? It looks really easy to miss modifying the script.

@ekexium I checked the current structure again. Under the current design, completely preventing this class of mistake at the source is hard.

CI already covers one important case: it runs `go generate ./... && git diff --exit-code`, so script/generated-output drift should be caught. But that does not catch the semantic omission where a command is already supported by `tikvrpc`, its protobuf request has a `Context` field, but the command is simply missing from the generator list.

The more thorough fixes would require reducing or removing this manual command list. For example:

- Use a Go generator with AST/type-checking to discover `Request` getters whose return types have a `Context` field.
- Implement a generic `patchCmdCtx` path with reflection, doing copy-on-write on `req.Req` and setting the `Context` field when present.
- Introduce a command registry as the single source of truth, where each command declares its `CmdType`, request accessor, whether it needs context attachment, whether it supports batch, etc.

But these approaches have non-trivial cons:

- An AST/type-checked generator would be more complex than the current shell generator and would still need explicit rules for exceptions.
- A reflection-based `AttachContext` would run on the RPC path. Even with caching, it is more complex and less explicit than the current generated type switch, and we should be careful about performance and maintainability there.
- A command registry would be cleaner long-term, but migrating to it has to handle existing special cases such as `CmdCopStream` sharing the `CmdCop` request type, plus MPP/store-level/debug requests that do not follow the same region-request semantics.

ekexium (approved):

PR #2014, opened 2026-07-01
txnkv: support shared lock upgrades

wfxr:
## Summary

This PR adds opt-in support for upgrading a pessimistic shared lock to an exclusive lock through `LockCtx.AllowSharedLockUpgrade`.

TiDB needs this capability for foreign-key checking with `tidb_foreign_key_check_in_shared_lock` enabled. In that flow, a transaction can first acquire a shared lock while checking a parent row and later update the same row. Supporting an in-place upgrade preserves the transaction's lock ownership across those operations.

The PR also defines the transaction contract for upgrade failures. A server-confirmed `SharedLockLost` or `LockUpgradeConflict` means the transaction cannot safely continue, so client-go preserves the typed error and makes the transaction rollback-only. Later lock and commit attempts return the original error, while explicit rollback remains available to release pessimistic locks. This defensive state protects callers that accidentally swallow the original upgrade error.

Generic RPC and response errors are intentionally non-fatal. A missing response alone does not reveal whether the remote lock stayed shared, became exclusive, or was removed, so it is not sufficient evidence to poison the transaction or report an undetermined commit result. The caller can retry the upgrade, and the existing pessimistic-lock ownership checks prevent a write from committing without the required exclusive lock.

Related to pingcap/tidb#68815.

## Dependency

The protocol support and typed shared-lock upgrade errors have merged in pingcap/kvproto#1495. Both the main module and `integration_tests` use the official `github.com/pingcap/kvproto` version `v0.0.0-20260820034643-9327469bb3ce`.

## Testing

Local validation includes the main transaction suite, focused race coverage for the rollback-only state, dependency verification, and integration-test compilation with workspace overrides disabled:

```bash
GOWORK=off go mod verify
GOWORK=off go test --tags=intest ./...
go test -race --tags=intest ./txnkv/transaction -run '^TestSharedLockUpgrade$' -count=1

cd integration_tests
GOWORK=off go mod verify
GOWORK=off go test -run '^$' ./...
```

The matching TiDB next-gen test also exercises a real CSE `SharedLockLost` response end to end and verifies that TiDB exits the SQL transaction and releases its remaining pessimistic locks.


txnkv/transaction/txn.go
```diff
@@ -1506,28 +1747,41 @@ func (txn *KVTxn) lockKeys(ctx context.Context, lockCtx *tikv.LockCtx, fn func()
...
 			}
 		}
 	}
+	if len(upgradeKeys) > 0 {
+		if len(keys) > 0 {
+			keys = deduplicateKeys(keys)
+		}
+		upgradeKeys = deduplicateKeys(upgradeKeys)
+		return txn.lockKeysWithSharedLockUpgrade(ctx, lockCtx, keys, upgradeKeys)
+	}
```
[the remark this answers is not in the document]
wfxr: Fixed by rejecting shared-lock upgrades when `ForUpdateTS == 0`, before sending any RPC or changing the local lock flags. The zero-TS path is only valid for legacy local bookkeeping; locally promoting an existing shared pessimistic lock would make the local state inconsistent with TiKV.

txnkv/transaction/txn.go
```diff
@@ -1384,6 +1393,225 @@ func (txn *KVTxn) LockKeysFunc(ctx context.Context, lockCtx *tikv.LockCtx, fn fu
...
+			if errors.As(err, &sharedLockLost) {
+				txn.committer.setFatalTxnErr(err)
+				return 0, err
+			}
+			if isLockUpgradeResultUndetermined(err) {
+				txn.committer.setUndeterminedErr(err)
+				return 0, errors.WithStack(tikverr.ErrResultUndetermined)
+			}
+			return 0, err
+		}
```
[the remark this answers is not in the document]
wfxr: Shared-lock upgrades are not registered as waiters in CSE's deadlock detector; concurrent upgrades are reported as `LockUpgradeConflict` instead. This error is intentionally non-retryable and makes the transaction rollback-only, so the losing upgrader releases its shared lock and allows the other upgrader to proceed.

txnkv/transaction/txn.go
```diff
@@ -799,6 +799,12 @@ func (txn *KVTxn) Commit(ctx context.Context) error {
 	if !txn.valid {
 		return tikverr.ErrInvalidTxn
 	}
+	if txn.committer != nil {
+		undeterminedErr, fatalTxnErr := txn.committer.getTxnStateErrs()
+		if undeterminedErr == nil && fatalTxnErr != nil {
+			return fatalTxnErr
+		}
+	}
```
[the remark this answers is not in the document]
wfxr: This check needs to remain before `defer txn.close()`. `txn.close()` only invalidates the transaction, while pessimistic locks and the TTL manager are cleaned up by `Rollback()`. Closing here would make the subsequent rollback return `ErrInvalidTxn` and prevent it from immediately releasing the losing upgrader's shared lock.

txnkv/transaction/txn.go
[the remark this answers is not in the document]
zyguan: Should we add a defensive `Warn` log when an `ErrDeadlock` reaches the shared-lock upgrade path? Under the current protocol this should be unreachable, since upgrade waits are not registered with the deadlock detector and concurrent upgraders use `LockUpgradeConflict`. If it does occur unexpectedly, a warning would make a server/client protocol mismatch or future behavior change visible.

txnkv/transaction/txn.go
```diff
@@ -1409,6 +1418,225 @@ func (txn *KVTxn) LockKeysFunc(ctx context.Context, lockCtx *tikv.LockCtx, fn fu
...
+	var deadlock *tikverr.ErrDeadlock
+	if errors.As(err, &deadlock) {
+		return false
+	}
+	var lockUpgradeConflict *tikverr.ErrLockUpgradeConflict
+	if errors.As(err, &lockUpgradeConflict) {
+		return false
+	}
+	var assertionFailed *tikverr.ErrAssertionFailed
+	return !errors.As(err, &assertionFailed)
```
cfzjywxk: Should we return an explicit undermined error type instead of `!errors.As(err, &assertionFailed)`? Or if is just a RPC unrechable like error?

zyguan (approved):

txnkv/transaction/txn.go
wfxr: I looked into this further. The negative filter is too broad, but generic RPC/body errors do not need to become `ErrResultUndetermined` either. The uncertain state is only whether the same transaction holds a shared or exclusive lock: if the upgrade was not applied, it can be retried; if applied but the response was lost, the stronger remote lock can be reconciled by retry/prewrite.

In a13502e, I removed the classifier and return generic errors as-is, while keeping `ErrSharedLockLost` and `ErrLockUpgradeConflict` fatal. Tests cover both unapplied failures and successful upgrades with lost responses. Does this reasoning and handling look right to you?

cfzjywxk: The naming and definition of `fatalTxnErr` are also too broad and unclear, try to avoid such naming convention. 

If this is only encountered in shared lock scenarios, we could consider abstracting it into a method like `sharedLockAbort(err error)` to determine the specific shared-lock-related error types that should abort the transaction.

wfxr: A pure `sharedLockAbort(err) bool` helper can centralize the type checks, but it is not enough to preserve the current behavior. It only classifies the error for the current `LockKeys` call. If later `LockKeys` or `Commit` calls must keep rejecting the transaction with the original error while still allowing an explicit `Rollback`, the abort state has to be retained somewhere.

There are two options:

1. Keep the state in client-go, but narrow it to `sharedLockAbortErr`. A helper such as `recordSharedLockAbort(err)` would recognize only `ErrSharedLockLost` and `ErrLockUpgradeConflict`, store the first matching error, and leave generic upgrade errors unchanged.
2. Remove the abort state from client-go entirely and only return the typed errors. TiDB already rolls back on `ErrSharedLockLost`, and converts `ErrLockUpgradeConflict` to a non-retryable deadlock which also rolls back the transaction.

I personally prefer the second one. Shared-lock upgrade is currently consumed through TiDB, and the policy of aborting the SQL transaction is already implemented and tested there. Keeping another persistent abort state in client-go would duplicate that responsibility and introduce extra transaction state solely for this path. The trade-off is that standalone client-go callers would need to handle these typed errors by rolling back explicitly.

Does this ownership model look reasonable to you, or would you prefer client-go to retain and enforce the shared-lock abort state?

cfzjywxk: I don't see the two responsibilities as mutually exclusive. TiDB should consume the typed errors and perform the E2E action, while client-go should retain a transaction-scoped rollback-only state to prevent the same `KVTxn` from continuing after such an error.

I'd suggest combining the two options instead of choosing one:
1. Return the exact typed errors (`ErrSharedLockLost`/`ErrLockUpgradeConflict`) as the contract, and let TiDB own the abort policy, similar to how `ErrResultUndetermined` is handled by closing the connection on the TiDB side. As a E2E feature it is reasonable for TiDB to follow the feature design.
2. Still keep a defensive state in client-go to reject further `LockKeys`/`Commit` calls, in case the caller swallows the error by mistake. But narrow it down: name it something like `sharedLockAbortErr`, only record these two specific error types, and return the stored original typed error on rejection instead of introducing a generic `fatalTxnErr` concept.

Also, this state is transaction-scoped, so it should live on `KVTxn` rather than `twoPhaseCommitter`. The committer is lazily created (the current `getTxnStateErr` needs a nil guard because of this), and `asyncPessimisticRollback` clones bare committers, so the committer is more like a per-execution engine. `undeterminedErr` stays on the committer since it's commit-attempt state.

Rollback should remain allowed to release the pessimistic locks. It would be good to have tests verifying that subsequent `LockKeys`/`Commit` calls return the same typed error without sending RPCs, and that rollback still works.

wfxr: Thank you for the detailed analysis and thoughtful suggestion.

Implemented this direction in 1069594, PTAL. `KVTxn` now owns a first-error-wins `sharedLockAbortErr`, protected by a dedicated mutex, and only records `ErrSharedLockLost` or `ErrLockUpgradeConflict`. Later `LockKeys`, `LockKeysFunc`, `LockKeysWithWaitTime`, and `Commit` return the original typed error before issuing PD/TiKV RPCs, while `Rollback` remains available. Generic upgrade errors remain non-fatal, and `ErrResultUndetermined` keeps its existing precedence and close behavior.

The tests cover no-RPC rejection including the TSO path, rollback cleanup, first-error-wins/race safety, and undetermined-error precedence. We also added a TiDB RealTiKV/CSE test for the `SharedLockLost` rollback path.

cfzjywxk (approved):

Issue #2018, opened 2026-07-13
tikvrpc: potential memory leak in stream RPC lease tracking


windtalker:
## Bug Report

Stream RPC leases can be retained indefinitely in `tikvrpc.CheckStreamTimeoutLoop` when a stream is created but not closed on some error/cancellation paths.

### What happened

In a TiDB workload with heavy internal maintenance queries against TiFlash/MPP, heap/goref analysis showed a large amount of memory retained from:

```text
tikvrpc.CheckStreamTimeoutLoop.array
  -> []*tikvrpc.Lease
```

The retained objects were stale stream leases rather than active MPP coordinator goroutines. The goroutine dump did not show corresponding active MPP receive goroutines, which suggests some leases were left in the timeout loop after the owning stream had already stopped being consumed.

### Why this can happen

For stream RPC responses, `Recv()` temporarily sets `Lease.deadline` before reading from the underlying gRPC stream, then sets it back to `0` after `Recv()` returns.

`CheckStreamTimeoutLoop.keepOnlyActive` treats `deadline == 0` as active/idle and keeps the lease:

```go
if deadline == 0 || deadline > now {
    array[idx] = array[i]
    idx++
}
```

This means a stream lease can remain in the loop forever if the stream is not explicitly closed after `Recv()` returns.

One concrete path is the first `Recv()` in the stream setup path:

```go
first, err = copStream.Recv()
if err != nil {
    if errors.Cause(err) != io.EOF {
        return nil, errors.WithStack(err)
    }
}
```

The stream has already been registered into `connArray.streamTimeout`, but this non-EOF error path returns without calling `Close()`. Because `Recv()` has already reset `deadline` to `0`, the lease can be kept by the timeout loop indefinitely.

The same pattern exists for CopStream, BatchCopStream, and MPPStream setup.

### Impact

Long-running workloads with many stream RPCs can accumulate stale `Lease` objects in `CheckStreamTimeoutLoop.array`, causing TiDB/client-go memory growth.

This is easier to hit when queries use MPP/TiFlash fanout and are frequently cancelled or short-circuited, for example queries with small `LIMIT` where not all established streams are fully consumed.

### Expected behavior

Stream leases should be removed from `CheckStreamTimeoutLoop` when the stream is no longer usable or should no longer be consumed.

[...]

PR #2019, opened 2026-07-13
tikvrpc: avoid leaking stream RPC leases

windtalker:
Issue Number: ref #2018

Problem Summary:

Stream RPC leases can be retained indefinitely in `tikvrpc.CheckStreamTimeoutLoop` if a stream has been registered into `connArray.streamTimeout` but is not closed on some error/cancellation paths. One concrete case is the first `Recv()` error path for CopStream, BatchCopStream, and MPPStream: `Recv()` resets the lease deadline to `0`, then the setup path returns a non-EOF error without closing the stream. Since `keepOnlyActive` treats `deadline == 0` as active/idle, the stale lease can remain in the loop.

What is changed:

- Close CopStream, BatchCopStream, and MPPStream when the first `Recv()` returns a non-EOF error after the lease has been registered.
- Add an optional absolute hard lifetime to `tikvrpc.Lease`, so the timeout loop can cancel and remove selected stream leases even when `deadline == 0`.
- Enable the hard lifetime only for TiDB materialized-view maintenance stream RPCs, identified by request source or context source `internal_mv_maintain`.

Why this approach:

- Normal stream timeout behavior is unchanged when hard timeout is not set.
- The hard timeout is scoped to MV maintenance stream RPCs instead of all user queries.
- The hard timeout is an absolute lifetime guardrail rather than an idle timeout: selected MV maintenance streams can be canceled after the hard TTL even if a `Recv` is currently active.
- The first-Recv error path is closed immediately, so it does not rely on the hard timeout guardrail.

Tests:

- `go test ./tikvrpc -run '^(TestKeepOnlyActiveWithHardDeadline|TestKeepOnlyActiveHardDeadlineOverridesRecvDeadline)$'`
- `go test ./internal/client -run '^(TestStreamHardTTLForMVMaintenance|TestStreamFirstRecvErrorClosesLease)$'`

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -94,6 +94,11 @@ const (
 // forwardMetadataKey is the key of gRPC metadata which represents a forwarded request.
 const forwardMetadataKey = "tikv-forwarded-host"
 
+const (
+	mvMaintenanceRequestSource = "internal_mv_maintain"
+	mvMaintenanceStreamHardTTL = time.Hour
+)
+
 // Client is a client that sends RPC.
 // It should not be used after calling Close().
 type Client interface {
@@ -686,6 +691,7 @@ func (c *RPCClient) getCopStreamResponse(ctx context.Context, client tikvpb.Tikv
 	copStream := resp.Resp.(*tikvrpc.CopStreamResponse)
 	copStream.Timeout = timeout
 	copStream.Lease.Cancel = cancel
+	setStreamHardTimeout(ctx, req, &copStream.Lease)
 	connArray.streamTimeout <- &copStream.Lease
 
 	// Read the first streaming response to get CopStreamResponse.
@@ -695,6 +701,7 @@ func (c *RPCClient) getCopStreamResponse(ctx context.Context, client tikvpb.Tikv
 	first, err = copStream.Recv()
 	if err != nil {
 		if errors.Cause(err) != io.EOF {
+			copStream.Close()
 			return nil, errors.WithStack(err)
 		}
 		logutil.BgLogger().Debug("copstream returns nothing for the request.")
@@ -721,6 +728,7 @@ func (c *RPCClient) getBatchCopStreamResponse(ctx context.Context, client tikvpb
 	copStream := resp.Resp.(*tikvrpc.BatchCopStreamResponse)
 	copStream.Timeout = timeout
 	copStream.Lease.Cancel = cancel
+	setStreamHardTimeout(ctx, req, &copStream.Lease)
 	connArray.streamTimeout <- &copStream.Lease
 
 	// Read the first streaming response to get CopStreamResponse.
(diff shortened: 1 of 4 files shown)


tikvrpc/tikvrpc.go
```diff
@@ -1295,8 +1295,18 @@ func CallDebugRPC(ctx context.Context, client debugpb.DebugClient, req *Request)
 
 // Lease is used to implement grpc stream timeout.
 type Lease struct {
-	Cancel   context.CancelFunc
-	deadline int64 // A time.UnixNano value, if time.Now().UnixNano() > deadline, cancel() would be called.
+	Cancel       context.CancelFunc
+	deadline     int64 // A time.UnixNano value, if time.Now().UnixNano() > deadline, cancel() would be called.
+	hardDeadline int64 // A UnixNano value for force-closing long-lived idle streams. 0 means disabled.
```
lcwangchao: Could you please to explain the difference between deadline and hardDeadline？

lcwangchao (approved):

tikvrpc/tikvrpc.go
windtalker: Done

tikvrpc/tikvrpc.go
wfxr, line 1430: Would it make sense to clear the compacted-away tail before returning here? `array[:idx]` only changes the slice length, so the removed `*Lease` values may still remain in the backing array retained by `CheckStreamTimeoutLoop` and potentially keep the enclosing stream responses reachable.

Something like:

```go
clear(array[idx:])
return array[:idx]
```

internal/client/client.go
```diff
@@ -94,6 +94,11 @@ const (
 // forwardMetadataKey is the key of gRPC metadata which represents a forwarded request.
 const forwardMetadataKey = "tikv-forwarded-host"
 
+const (
+	mvMaintenanceRequestSource = "internal_mv_maintain"
+	mvMaintenanceStreamHardTTL = time.Hour
```
wfxr, line 99: Is the one-hour limit intended to be an absolute stream lifetime rather than an idle timeout? Since `hardDeadline` is independent of the current `Recv` state, an MV maintenance stream would be canceled after one hour even if it is still active. Just wondering whether all such operations are guaranteed to finish within this limit.

It would also be useful to cover the case where the regular `Recv` deadline is still active but the hard deadline has expired, to make the intended behavior explicit.

wfxr (commented):
It looks like this is an absolute lifetime limit rather than an idle timeout. Would it make sense to adjust the PR description accordingly?

internal/client/client.go
windtalker: Yes, it is an absolute timeout

wfxr (approved):

ekexium (approved):

zyguan (approved):

PR #2024, opened 2026-07-20
*: update kvproto to apiv3

disksing:

The kvproto APIV3 branch moves keyspace-related protobuf fields into oneofs, so the current client code no longer compiles against the latest definitions.

Issue: pingcap/kvproto#1506 (introduce namespace-scoped keyspace identity for API V3); the protos are added by pingcap/kvproto#1458.

### What changed?

- Update kvproto to fd32127adca58438f6b05c869951182bbdd6041c in the root, integration-test, and example modules.
- Update the PD client dependency to the latest master revision `ba23e721a` (2026-08-04) in the root, integration-test, and example modules.
- Adapt API codec, region location, TiKV RPC, and tests to generated oneof getters and wrappers.
- Cache the keyspace oneof wrapper in the codec: the keyspace ID never changes per codec, so reuse one `*Context_KeyspaceId` instead of allocating a new wrapper on every encoded V1/V2 request. The isolated `setAPICtx` path goes back to 0 B/0 allocs per call (was 4 B/1 alloc).
- Make the lifecycle-hook tests non-flaky: `Test2PCLifecycleHooks`, `Test2PCCleanupLifecycleHooks` and `TestAsyncCommitLifecycleHooks` asserted `reachedPost == false` immediately after `Commit` returned, assuming the background (secondary commit / cleanup) goroutine had not finished yet. With the mock store that goroutine can finish before `Commit` returns, so the assertion is racy and `integration-local-race` failed intermittently. The fragile assertion is removed; the tests still verify that `Pre` runs and that `Post` eventually runs.

### Tests

- `go test -tags=intest ./...`
- `go test -race ./...` in `integration_tests/`

## Summary by CodeRabbit

* **Compatibility**
  * Updated keyspace-aware request and response handling for newer protocol formats across MPP, compaction, transaction, and RPC flows.
  * Improved consistency when reading keyspace identifiers in garbage collection and related operations.
* **Maintenance**
  * Refreshed example and integration components to align with newer protocol and client revisions.
  * Updated supporting test coverage to reflect the latest keyspace metadata structures.


disksing:
Addressed the CodeRabbit nit in 901d142 by using the expected/actual argument order. This update also moves integration_tests to the APIV3-compatible TiDB revision and fixes the remaining KeyspaceMeta getter; the full integration test module now compiles.

disksing:
CI follow-up:

- `integration-local-race` was failing on `TestCommitter/Test2PCLifecycleHooks` (and the same pattern in `Test2PCCleanupLifecycleHooks` and `TestAsyncCommitLifecycleHooks`). The assertion `s.Equal(reachedPost.Load(), false)` right after `Commit` returns is racy: with the mock store the background (secondary commit / cleanup) goroutine can finish before `Commit` returns, so the `Post` hook may already have run. Removed the fragile assertion and kept the checks that `Pre` ran and that `Post` eventually runs.

- Updated `github.com/tikv/pd/client` to the latest master revision `ba23e721a` (2026-08-04) in the root, integration-test, and example modules.

- The `pull-integration-test-nextgen` prow job still fails because the pinned base image ships Go 1.25.10 while go.mod requires 1.25.12 (and `GOTOOLCHAIN=local` is set). Fixed in PingCAP-QE/ci#4857 (bump base image to `v2026.7.12-6-gc8582dde-go1.25`, Go 1.25.12); once that is merged the job should be re-run.

internal/apicodec/codec.go
```diff
@@ -113,7 +114,7 @@ func DecodeKey(encoded []byte, version kvrpcpb.APIVersion) ([]byte, []byte, erro
 
 func setAPICtx(c Codec, r *tikvrpc.Request) {
 	r.ApiVersion = c.GetAPIVersion()
-	r.KeyspaceId = uint32(c.GetKeyspaceID())
+	r.Keyspace = &kvrpcpb.Context_KeyspaceId{KeyspaceId: uint32(c.GetKeyspaceID())}
```
rleungx: Every encoded V1/V2 request reaches this assignment, and the new oneof wrapper escapes to the heap. On the exact PR head the isolated path changes from 0 B/0 allocs to 4 B/1 alloc per call versus the base, adding allocation and GC pressure to the default RPC hot path.

disksing: Fixed in 49d3f90: the keyspace ID of a codec never changes, so the oneof wrapper is now cached in the codec (`GetKeyspaceOneof`) and reused by every encoded request instead of being rebuilt in `setAPICtx`.

Benchmarked the isolated path (`BenchmarkSetAPICtxV1`):
- before: 4 B/op, 1 allocs/op
- after: 0 B/op, 0 allocs/op (back to the pre-oneof baseline)

internal/apicodec/codec.go
```diff
@@ -69,6 +70,9 @@ type Codec interface {
 	GetKeyspace() []byte
 	// GetKeyspaceID return the keyspace id of the codec.
 	GetKeyspaceID() KeyspaceID
+	// GetKeyspaceOneof returns the cached keyspace oneof wrapper of the codec.
+	// The returned wrapper is shared by all encoded requests and must not be modified.
+	GetKeyspaceOneof() *kvrpcpb.Context_KeyspaceId
```
rleungx: `tikv.Codec` is a public alias accepted by `tikv.WithCodec`, so adding this method for the allocation optimization creates a separate source-compatibility break: external codec implementations that compile against the preceding PR head fail on this head with `missing method GetKeyspaceOneof`, even if they do not use the protobuf fields migrated to oneofs.

disksing: Good catch, thanks! Fixed in 0a7a5ac: the public `Codec` interface is unchanged. The internal codecs now implement an unexported `keyspaceOneofProvider`, and `setAPICtx` uses a type assertion to reuse the cached wrapper when available, falling back to building a fresh one otherwise. External codec implementations keep compiling and behave as before; the internal hot path stays at 0 B/0 allocs.

cfzjywxk:
@disksing 
Please create an related github issue and reference it to the original design/RFC

internal/apicodec/codec_v2.go
```diff
@@ -59,12 +59,14 @@ type codecV2 struct {
...
 	keyspaceMeta *keyspacepb.KeyspaceMeta
+	// keyspaceID is the cached keyspace oneof wrapper shared by all encoded requests.
+	keyspaceID *kvrpcpb.Context_KeyspaceId
 }
 
 // NewCodecV2 returns a codec that can be used to encode/decode
 // keys and requests to and from APIv2 format.
 func NewCodecV2(mode Mode, keyspaceMeta *keyspacepb.KeyspaceMeta) (Codec, error) {
-	keyspaceID := keyspaceMeta.Id
+	keyspaceID := keyspaceMeta.GetId()
```
rleungx, line 79: `KeyspaceMeta.GetId()` returns 0 when the new oneof contains `KeyspaceIdentity`, while API V3 defines zero as invalid and requires the complete identity. Because this constructor is exported as `tikv.NewCodecV2`, passing a V3 meta now succeeds as a default-keyspace codec and encodes keys and request context for keyspace 0 rather than rejecting the unsupported identity arm.

disksing:
@cfzjywxk Done - the PR description now references the design issue: pingcap/kvproto#1506 (introduce namespace-scoped keyspace identity for API V3). The protos themselves were added in pingcap/kvproto#1458.

internal/apicodec/codec_v2.go
disksing: Fixed in afe0924: `NewCodecV2` now rejects a meta whose keyspace oneof holds `KeyspaceIdentity` (API V3) with an explicit error, instead of silently building a default-keyspace (ID 0) codec. Added `TestNewCodecV2RejectsKeyspaceIdentity` to cover it.

internal/apicodec/codec_v2.go
```diff
@@ -59,12 +59,21 @@ type codecV2 struct {
...
 	keyspaceMeta *keyspacepb.KeyspaceMeta
+	// keyspaceID is the cached keyspace oneof wrapper shared by all encoded requests.
+	keyspaceID *kvrpcpb.Context_KeyspaceId
 }
 
 // NewCodecV2 returns a codec that can be used to encode/decode
 // keys and requests to and from APIv2 format.
 func NewCodecV2(mode Mode, keyspaceMeta *keyspacepb.KeyspaceMeta) (Codec, error) {
-	keyspaceID := keyspaceMeta.Id
+	if keyspaceMeta.GetKeyspaceIdentity() != nil {
```
rleungx, line 72: `GetKeyspaceIdentity()` and `GetId()` are both nil-safe, so `tikv.NewCodecV2(mode, nil)` now succeeds as keyspace 0 whereas the pre-PR code failed immediately. A missing PD metadata response or caller bug can therefore be turned into a default-keyspace request instead of a visible failure.

internal/locate/pd_codec.go
```diff
@@ -87,7 +87,7 @@ func GetKeyspaceID(client pd.Client, name string) (uint32, error) {
 	if meta.State != keyspacepb.KeyspaceState_ENABLED {
 		return 0, errors.Errorf("keyspace %s not enabled", name)
 	}
-	return meta.Id, nil
+	return meta.GetId(), nil
```
rleungx, line 96: `GetId()` returns 0 when the oneof holds `KeyspaceIdentity`, so this helper reports the default keyspace ID without an error for V3 metadata. A caller can therefore route an identity-scoped keyspace as keyspace 0, the same silent fallback rejected by `NewCodecV2`.

ekexium (commented):
LGTM after fixing existing comments

internal/apicodec/codec_v2.go
disksing: Fixed in 86c6e6e: `NewCodecV2` now returns an error for a nil meta (the pre-PR code read `keyspaceMeta.Id` directly and failed immediately, which the nil-safe getters had silently lost). Added `TestNewCodecV2RejectsNilMeta`.

internal/locate/pd_codec.go
disksing: Fixed in 86c6e6e: `GetKeyspaceID` now rejects a meta whose keyspace oneof holds `KeyspaceIdentity` with an explicit error instead of returning 0. Added `TestGetKeyspaceIDRejectsV3Identity`.

disksing:
Updated `github.com/tikv/pd/client` to `afa43111d` (client: recognize v2 keyspace-level GC, tikv/pd#11100) in the root, integration-test, and example modules.

disksing:
@ekexium All of rleungx's review comments have been addressed and replied to in the threads:
- oneof wrapper allocation (codec.go) → fixed in 0a7a5ac (0 B/0 allocs, public interface unchanged via unexported `keyspaceOneofProvider`)
- V3 keyspace identity rejected in `NewCodecV2` → fixed in afe0924
- nil meta rejected in `NewCodecV2` → fixed in 86c6e6e
- V3 keyspace identity rejected in `GetKeyspaceID` → fixed in 86c6e6e

CI is green (all GitHub Actions jobs + prow nextgen). Could you take another look / approve?

rleungx (approved):

pingyu (approved):

disksing:
@pingyu Thanks for the approval. The PR now has the required approvals and `lgtm`, but the `/hold` is keeping Tide blocked. Is there anything remaining to address? If not, please run `/hold cancel` when ready so Tide can proceed.

cfzjywxk (approved):

PR #2025, opened 2026-07-21
txnkv: add readPoolTaskDetails to SnapshotRuntimeStats

lcwangchao:

Issue Number: ref https://github.com/tidbcloud/cloud-storage-engine/issues/5480

Problem Summary:

TiKV/CSE can now return optional request-level read-pool scheduling and execution metrics in `ExecDetailsV2.read_pool_task_details`. client-go currently drops these details, so TiDB and other downstream consumers cannot include read-pool queueing, wakeup, dispatch/poll, and CPU/wall-time breakdowns in snapshot runtime diagnostics.

- Bump kvproto to the revision that adds `PoolTaskDetails` to `ExecDetailsV2`.
- Add `util.PoolTaskDetails` to aggregate reported task samples, including:
  - task, poll, and dispatch counts;
  - task wall time, queue wait, wake wait, and poll CPU/wall time;
  - total, minimum, and maximum values;
  - fair-queue waited-task-slice statistics.
- Merge `read_pool_task_details` from responses into `SnapshotRuntimeStats`.
- Support cloning, merging, empty checks, and compact string formatting for the aggregate.
- Expose `GetReadPoolTaskDetails()` as a defensive copy so callers cannot mutate the stored runtime stats.
- Keep the field lazy and backward-compatible: responses from older servers do not allocate or print empty read-pool statistics.
- Refresh root, integration-test, and example module metadata for the new kvproto revision.

The aggregation operates on reported response samples because the protobuf currently carries no task identity or delta information.

### Check List

- [x] Unit tests
  - `go test ./util ./txnkv/txnsnapshot`
  - `go test --tags=intest ./util ./txnkv/txnsnapshot`
- [x] Backward compatibility
  - The protobuf field is additive and optional.
  - Existing behavior is unchanged when `read_pool_task_details` is absent.

### Related changes

- https://github.com/pingcap/kvproto/pull/1502
- https://github.com/tidbcloud/cloud-storage-engine/pull/5646

## Summary by CodeRabbit

* **New Features**
  * Enhanced snapshot runtime statistics to include read-pool task details (poll/dispatch and related timing/wait metrics).
  * Added aggregation support so read-pool details can be merged, cloned, formatted, and retrieved from runtime stats.
* **Maintenance**
  * Bumped Go toolchain and refreshed several indirect module versions across examples and integration.
  * Updated the primary kv-proto dependency revision.
* **Tests**
  * Added unit and integration test coverage for read-pool metric aggregation, formatting, merge/clone behavior, and empty-state handling.

### DIFF
--- a/examples/gcworker/go.mod
+++ b/examples/gcworker/go.mod
@@ -1,6 +1,6 @@
 module gcworker
 
-go 1.25.8
+go 1.25.10
 
 require github.com/tikv/client-go/v2 v2.0.0
 
@@ -22,7 +22,7 @@ require (
 	github.com/opentracing/opentracing-go v1.2.0 // indirect
 	github.com/pingcap/errors v0.11.5-0.20241219054535-6b8c588c3122 // indirect
 	github.com/pingcap/failpoint v0.0.0-20240528011301-b51a646c7c86 // indirect
-	github.com/pingcap/kvproto v0.0.0-20260601035955-b2b3bb492278 // indirect
+	github.com/pingcap/kvproto v0.0.0-20260721064811-683dad8fa368 // indirect
 	github.com/pingcap/log v1.1.1-0.20221110025148-ca232912c9f3 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/prometheus/client_golang v1.20.5 // indirect
@@ -31,7 +31,7 @@ require (
 	github.com/prometheus/procfs v0.15.1 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/tiancaiamao/gp v0.0.0-20221230034425-4025bc8a4d4a // indirect
-	github.com/tikv/pd/client v0.0.0-20260601035915-3ef6a3b10c84 // indirect
+	github.com/tikv/pd/client v0.0.0-20260708075407-4e05b9d2c2d3 // indirect
 	github.com/twmb/murmur3 v1.1.3 // indirect
 	go.etcd.io/etcd/api/v3 v3.5.10 // indirect
 	go.etcd.io/etcd/client/pkg/v3 v3.5.10 // indirect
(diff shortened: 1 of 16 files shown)


examples/gcworker/go.mod
```diff
@@ -1,6 +1,6 @@
 module gcworker
 
-go 1.25.8
+go 1.25.10
```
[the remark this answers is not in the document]
lcwangchao: This is only to keep the submodule at the same version as the root module. The upgrade should be in another PR.

util/execdetails.go
```diff
@@ -457,6 +458,320 @@ func getUnit(d time.Duration) time.Duration {
...
+		return
+	}
+	hadPollSamples := d.PollCount > 0
+	hadQueueWaitSamples := d.TotalQueueWaitTime > 0
+	hadWakeWaitSamples := d.TotalWakeWaitTime > 0
+	hadFairQueueSamples := d.FairQueueSampleCount > 0
+	hadTaskWallTimeSamples := d.TaskWallTimeSampleCount > 0
+	hadTasks := d.TaskCount > 0
+
+	d.TaskCount++
```
cfzjywxk, line 535: The `PoolTaskDetails` is a cumulative snapshot of one physical read-pool task, but `batch_get_command` can attach successive snapshots of the same shared task to multiple child responses.

Incrementing the `TaskCount` and summing each snapshot here therefore double-counts task/poll/time totals and may attribute earlier children’s work to later requests.

lcwangchao: This aggregation is intended for request-latency diagnostics rather than de-duplicated physical task accounting. Each response contributes one task observation. For a child response from batch_get_command, the cumulative snapshot intentionally includes work performed before that child becomes visible, because that work contributes to its observed latency. Totals are sums across reported observations and are primarily used to calculate averages; overlapping or shared physical read-pool tasks are not de-duplicated. We will clarify this behavior in the field and merge documentation.

We could alternatively report statistics based on distinct physical tasks by de-duplicating shared execution on the TiKV side. However, doing so would require substantially more changes, such as introducing task identity and snapshot semantics across kvproto, TiKV, and client-go, and retaining that identity during aggregation. We need to evaluate whether the additional complexity is justified. In practice, a single SQL statement only encounters this distinction in a limited set of cases, such as multiple UNION branches that each perform a point get.

cfzjywxk (approved):

zyguan (approved):

PR #2026, opened 2026-07-22
locate: add per-attempt request limiter hook

### LINKED ISSUE #2035: Add a per-attempt request limiter hook after client-go selects the target store
# Add a per-attempt request limiter hook after client-go selects the target store

## Background

TiDB needs to limit concurrent coprocessor requests per statement and per TiKV store. The target store is selected inside client-go and may change when a region request is retried, so a caller cannot accurately limit the actual target store before invoking client-go.

The limiter token must cover exactly one RPC attempt. It must be acquired after client-go selects the target store, released exactly once when that attempt finishes, and released before a retry acquires another token.

## Proposed API

Add an optional callback to `tikvrpc.Request`:

```go
type RequestAttemptLimiterFunc func(ctx context.Context, storeID uint64) (release func(), err error)

type Request struct {
    // ...
    RequestAttemptLimiter RequestAttemptLimiterFunc
}
```

For every synchronous or asynchronous RPC attempt, including retries, client-go calls the callback with the actual target store ID after selecting the store and before starting the RPC. If the callback returns a release function and no error, client-go calls the release function exactly once after the attempt finishes.

When the callback is nil, existing behavior is unchanged.

## Error, cancellation, and timing semantics

- A limiter error terminates the request without sending an RPC.
- Implementations must return promptly when the request context is canceled.
- The limiter wait follows the request context and is not bounded by the per-attempt RPC timeout.
- The per-attempt RPC timeout and RPC runtime statistics start after the limiter allows the attempt.
- If the callback returns both a release function and an error, client-go calls the release function immediately because no RPC will be sent.
- If acquiring the existing store-limit token fails after the limiter returns a release function, client-go releases the limiter token.

## Compatibility

This is an additive client API. Existing callers are unaffected unless they set `RequestAttemptLimiter`.

## Related

- https://github.com/tikv/client-go/pull/2026
- https://github.com/pingcap/tidb/pull/69360

solotzg:
## Problem

Issue Number: close #2035

TiDB needs to limit concurrent coprocessor requests per statement and per TiKV store. The target store is selected inside client-go and may change during region retries, so acquiring a limiter token before calling client-go cannot accurately limit the actual target store.

The limiter token should cover exactly one RPC attempt and must be released before another attempt acquires a token.

## What changed

- Add `tikvrpc.RequestAttemptLimiterFunc` and `Request.RequestAttemptLimiter`.
- Invoke the limiter after selecting the actual target store and before every RPC attempt, including retries.
- Pass the target store ID rather than the proxy store ID.
- Invoke the returned release function exactly once after the attempt finishes.
- Release the token when store-token acquisition fails or when the limiter returns both a release function and an error.
- Run a potentially blocking limiter asynchronously in `SendReqAsync`.
- Exclude limiter wait time from asynchronous RPC runtime statistics.
- Define timeout semantics: the limiter follows the request context, and the per-attempt RPC timeout starts after the limiter allows the attempt.

When the callback is nil, existing request behavior is unchanged.

## Error and cancellation semantics

- A limiter error terminates the request without sending an RPC.
- Limiter implementations must return promptly when the request context is canceled.
- Limiter wait is not bounded by the per-attempt RPC timeout.

## Tests

Added coverage for:

- synchronous and asynchronous limiting;
- limiter rejection and context cancellation;
- release on store-token failure;
- release exactly once per attempt;
- selecting the new store after a retry;
- releasing the previous attempt before the retry acquires a token;
- excluding limiter wait from asynchronous RPC runtime statistics.

Tests run:

- `GOCACHE=/tmp/client-go-go-cache go test ./internal/locate -run 'TestRegionRequestToSingleStore/TestRequestAttemptLimiter|TestRegionRequestToThreeStores/TestRequestAttemptLimiterUsesRetryStore' -count=10`
- `GOCACHE=/tmp/client-go-go-cache go test -race ./internal/locate -run 'TestRegionRequestToSingleStore/TestRequestAttemptLimiter|TestRegionRequestToThreeStores/TestRequestAttemptLimiterUsesRetryStore' -count=1`

## Compatibility

This is an additive client API. Existing callers are unaffected unless they set `RequestAttemptLimiter`.

The associated TiDB change installs the callback only for TiKV coprocessor requests.

## Related

- https://github.com/pingcap/tidb/pull/69360


internal/locate/region_request.go
```diff
@@ -907,6 +936,22 @@ type sendReqState struct {
 	invariants reqInvariants
 }
 
+func (s *sendReqState) acquireRequestAttemptAdmission() (release func(), err error) {
```
qw4990: Could you add some comments for this function?

tikvrpc/tikvrpc.go
```diff
@@ -239,6 +239,14 @@ func (t CmdType) String() string {
 	return "Unknown"
 }
 
+// RequestAttemptAdmissionFunc waits until an RPC attempt to storeID is admitted.
+// When this function returns a non-nil release function and a nil error,
+// client-go calls release exactly once after that attempt finishes.
+// The context follows the request lifecycle; the per-attempt RPC timeout starts
+// after admission succeeds and does not bound admission wait.
+// Implementations should return promptly when ctx is canceled.
+type RequestAttemptAdmissionFunc func(ctx context.Context, storeID uint64) (release func(), err error)
```
cfzjywxk: Better to use another name instead of `admission control` which is used in the new RC mechanism in cse tikv. Also avoiding name like `resource control`.

How about somehting like `planningConcurrencyLimiter` or `plannerConcurrencyLimiter`.

internal/locate/region_request.go
solotzg: done

tikvrpc/tikvrpc.go
solotzg: Renamed the public API to `RequestAttemptLimiterFunc` and `RequestAttemptLimiter`, and renamed the internal helper to `acquireRequestAttemptToken`.

I avoided `planner`/`planning` in the name because the hook runs in client-go after the target store is selected and is not limited to SQL planning use cases. The TiDB implementation, tests, and design document have been updated accordingly.

qw4990 (approved):

tikvrpc/tikvrpc.go
```diff
@@ -272,6 +280,9 @@ type Request struct {
 	// controller, which decides whether the request type is eligible for
 	// paging pre-charge; non-cop hints may be ignored.
 	PredictedReadBytes uint64
+	// RequestAttemptLimiter, when non-nil, is called after client-go selects
+	// the actual target store and before every RPC attempt, including retries.
+	RequestAttemptLimiter RequestAttemptLimiterFunc
```
cfzjywxk, line 295: What is the scope of this limiter?

If it is statement-level, we need to consider whether statement-level retries should be included.

If it is region-request-level, we also need to clarify which layer it covers, since there are multiple layers of retries here:

* Region-miss-related region errors that require relocation; this layer will be retried above the region request sender.
* Best-effort retries inside the region request sender.

We need to define this clearly upfront to ensure the limiter’s lifecycle covers the corresponding scope and comment about it.

solotzg: `RequestAttemptLimiter` is a per-physical-RPC-attempt hook. client-go invokes it for the initial attempt and retries handled inside `RegionRequestSender`; client-go owns only the acquire/release interval.

The limiter’s lifecycle and scope are caller-owned. For the TiDB use case, every request created or rebuilt within one statement must attach a callback that captures the same statement-owned, per-store limiter. This includes higher-level region-miss retries that rebuild a request; client-go does not propagate the hook automatically.

internal/locate/region_request.go
```diff
@@ -547,59 +547,88 @@ func (s *RegionRequestSender) SendReqAsync(
...
+
+		sendToAddr := state.vars.rpcCtx.Addr
+		if state.vars.rpcCtx.ProxyStore == nil {
+			req.ForwardedHost = ""
+		} else {
+			req.ForwardedHost = state.vars.rpcCtx.Addr
+			sendToAddr = state.vars.rpcCtx.ProxyAddr
+		}
+
+		rpcStart := time.Now()
```
cfzjywxk: We should preserve the existing timing baseline when `RequestAttemptLimiter` is nil, as moving `rpcStart` here changes runtime stats, `WaitKVRespDuration`, slow-score calculation, and replica retry accounting for normal async requests.

tikvrpc/tikvrpc.go
cfzjywxk: Better to add to the comments above the `RequestAttemptLimiter`.

internal/locate/region_request.go
solotzg: Fixed

tikvrpc/tikvrpc.go
solotzg: Done

internal/locate/region_request.go
```diff
@@ -547,59 +547,94 @@ func (s *RegionRequestSender) SendReqAsync(
...
+					if retryTimes := state.vars.sendTimes - 1; retryTimes > 0 && retryTimes%100 == 0 {
+						logutil.Logger(bo.GetCtx()).Warn("retry", zap.Uint64("region", regionID.GetID()), zap.Int("times", retryTimes))
+					}
+				}
+				cb.Schedule(state.toResponseExt())
+			})
+		}))
+	}
+	if runLimiterAsync {
+		cb.Executor().Go(sendFirstAttempt)
```
cfzjywxk, line 638: When the limiter is set, `sendFirstAttempt `may park for an arbitrarily long time inside `RequestAttemptLimiter`. While routing it through `cb.Executor().Go` means that when the executor is backed by a recycling pool.

In this case, there could be degradation that `gp.Pool` overflows soon and does not take effect.


Non-blocking for current tidb as the async path would not be used by coprocessor by default.
Better to comment the risk here, it could be a problem if the async request optimization is integrated into the coprocessor processing in the future. 

/cc @zyguan

internal/locate/region_request.go
```diff
@@ -547,59 +547,94 @@ func (s *RegionRequestSender) SendReqAsync(
...
+
+		s.client.SendRequestAsync(ctx, sendToAddr, req, async.NewCallback(cb.Executor(), func(resp *tikvrpc.Response, err error) {
+			state.vars.sendTimes++
+			canceled := err != nil && hookCtx.Err() != nil && errors.Cause(hookCtx.Err()) == context.Canceled
+			var execDetails *util.ExecDetails
+			if val := ctx.Value(util.ExecDetailsKey); val != nil {
+				execDetails = val.(*util.ExecDetails)
+			}
+			if state.handleAsyncResponse(rpcStart, canceled, resp, err, execDetails, cancels...) {
+				cb.Invoke(state.toResponseExt())
```
zyguan: `cb.Invoke` is designed to be called within the run-loop goroutine, I think we need to use `finishBeforeSend` here.

internal/locate/region_request.go
solotzg: RequestAttemptLimiter may block, and running it through cb.Executor().Go can occupy workers in a recycling pool. Under high fan-out, this may cause the pool to overflow and weaken its concurrency bound.  
This is not exercised by the current coprocessor path, but I added a comment at the scheduling site to document the risk for future async coprocessor integrations.

This is not exercised by the current coprocessor path, but I added a comment at the scheduling site to document the risk for future async coprocessor integrations.

internal/locate/region_request.go
solotzg: Fixed. The terminal response path now uses finishBeforeSend instead of cb.Invoke.  
With an attempt limiter, sendFirstAttempt runs outside the run loop, so finishBeforeSend resolves to cb.Schedule and preserves the callback execution context. I also added a test with a client that invokes its async callback synchronously.

zyguan (approved):

cfzjywxk (approved):

Issue #2012, opened 2026-06-24
Follower read target selection should consult PD for store recovery, not only TiKV's self-reported slow score


AndreMouche:
## Background

When a TiKV store hits a transient problem (e.g. a faulty/jittery disk), PD marks it as a slow store and evicts its leaders. Follower stale-read traffic, however, can still be routed to that store while it is recovering.

## Current Behavior

When client-go selects the target replica for a follower read, it relies on the slow score that the TiKV store reports about itself to decide whether the store has recovered and is eligible to serve reads again.

The problem is that a recovered self-reported slow score does not guarantee the store is actually ready to take traffic:

- The signal is local to the store and can be noisy/flapping 
- It is not coordinated with PD's view of the store's recovery / leader-eviction state.

As a result, follower reads can be sent back to a not-yet-recovered store too early,  hotspotting it.

## Expected Behavior

Before routing a follower (stale) read to a store that was recently slow/recovering, client-go should consult PD to confirm the store has fully recovered (e.g. that PD no longer considers it a slow store / has stopped evicting its leaders), rather than relying solely on the store's self-reported slow score.

## Proposed Direction

- Treat PD's store recovery state as the authoritative signal for "is this store ready to serve reads again," and gate follower-read target selection on it.
- Keep the TiKV self-reported slow score as a fast local hint, but require agreement with PD's view before sending traffic back to a recently-slow store.
- Consider a hysteresis / cool-down so a flapping slow score cannot repeatedly re-add a store as a follower-read target.

## Related
- Related PD behavior: tikv/pd#9384 — Recovery time is not expected in slow store scheduler (https://github.com/tikv/pd/issues/9384) (leaders transferred back prematurely when recovery time < jitter duration).

LykxSassinator:
For addressing this issue, we have made a proposal as the following shows:

## Design summary

The key point is to treat PD as the source of truth for `recovery-time`.

Today, `client-go` only uses TiKV-side health feedback (`slow_score`) to decide whether a replica can serve stale-read traffic. This is insufficient for the slow-store recovery case, because PD may still keep the
store in `evicted_as_slow_store = true` during the recovery window, even after TiKV-side slow score has already returned to normal.

So the design should be:

- `client-go` keeps **two independent signals** for each store:
- TiKV-side health status
- PD-side slow-store eviction status
- stale-read replica routing must allow a store only when:
- TiKV-side health is normal, and
- PD-side `evicted_as_slow_store` is false
[...]

PR #2027, opened 2026-07-23
locate: honor PD slow-store recovery state for replica reads

LykxSassinator:

Replica selection currently relies on client-go's local health observations. A store can recover locally before PD's slow-store recovery-time expires, causing follower/learner/stale reads to be routed to a store that PD still evicts as slow.

Related issue: ref #2012

## What is changed?

- Consume `GetStoreResponse.scheduling_state` introduced by kvproto #1503 and exposed by PD #11041.
- Use the optional `pd.RPCClientExt` capability to preserve source compatibility for existing `pd.Client` implementations and mocks.
- Keep PD-owned scheduling state separate from `StoreHealthStatus`, which only records local and TiKV-reported health observations.
- Gate every non-leader replica candidate on both local health and PD scheduling state, covering follower, learner, stale reads, and leader-read replica fallback.
- Use an epoch-based pending confirmation gate so an older PD response cannot re-admit a store after a newer local health transition.
- Query the PD leader for state-aware `GetStore` refreshes because Router intentionally does not provide dynamic scheduling state.
- Update codec/interceptor wrappers and test mocks for the optional full-response API.

## Compatibility

- Clients connected to an older PD or a `pd.Client` implementation without `RPCClientExt` retain the existing replica-selection behavior.
- The branch temporarily depends on the kvproto #1503 and PD #11041 branches until those changes are merged.

## Tests

- `go test ./internal/locate`
- `go test ./... -run '^$'`
- `cd integration_tests && go test ./...`


LykxSassinator:
I considered adding `StoreSchedulingState` directly to `metapb.Store` so client-go could reuse `pd.Client.GetStore` and retain the Router-first query path. I do not recommend doing that in the current proposal.

`EvictedAsSlowStore` is not regular Store metadata. It is a dynamic, PD-leader-owned scheduling decision whose lifetime is governed by PD's recovery-time policy. The current PD implementation intentionally keeps it separate:

- `Store` is populated from `store.GetMeta()`.
- `SchedulingState` is populated from `store.EvictedAsSlowStore()`.
- Router serves `GetStore` from `BasicCluster`, which does not retain this dynamic scheduling state and therefore returns it as unknown.

Moving the field into `metapb.Store` would remove the `RPCClientExt` plumbing, but would not safely remove the PD leader query:

1. A Router response would still carry an absent or stale state unless PD first introduced a state-synchronization path from the PD leader to Router.
2. Treating an absent Router field as `false` could re-admit a store that PD still evicts; treating it as unknown requires a fallback query to the PD leader.
3. Router-first followed by a leader fallback adds an RPC without reducing the required leader query for the current Router implementation.
4. `metapb.Store` is broadly reused by store metadata, registration, and watch paths. Publishing a runtime scheduling decision there expands its propagation and freshness contract substantially.
5. `metapb.proto` cannot directly reference `pdpb.StoreSchedulingState` because `pdpb.proto` already depends on `metapb.proto`; the type would also need to move into the generic metadata package.

The response-level field keeps the contract explicit: it is PD scheduling state for this `GetStore` response, and an unset value means `unknown`, not `false`.

If reducing PD leader traffic becomes a goal, the follow-up should be to replicate a versioned/freshness-aware routing-state snapshot to Router. Client-go could then accept a fresh Router response and fall back to the PD leader only when the state is absent or stale. That achieves the desired offload without conflating dynamic scheduling policy with Store metadata.

PR #2029, opened 2026-07-23
resourcecontrol: settle batched cop tasks' execution details

### LINKED ISSUE #2043: resourcecontrol: batched coprocessor responses undercount read bytes and KV CPU
## Problem

`MakeResponseInfo` accounts only for the top-level coprocessor response.

A store-batched response also contains nested `batch_responses`, each carrying a logical task's data and execution details. These children are not separate RPC responses, so their read bytes and KV CPU are never settled independently.

For example, if the top-level response reports 80 read bytes and a nested response reports 20, resource control currently records only 80 instead of 100.

## Expected behavior

Resource-control settlement should include every nested response's:

- Scan bytes, falling back to data size when scan details are unavailable.
- KV CPU time.
- Existing legacy or NextGen scan-byte accounting rules.

Non-batched responses should remain unchanged.

## Impact

Resource-control-enabled workloads using store batching can be charged fewer RUs than the work performed. This can weaken resource-group limits and RU-based runaway detection.

Query correctness is not affected.

## History

- November 2022: kvproto and TiKV added batched coprocessor task support.
- December 2022: TiDB added store-batched coprocessor requests.
- January 2023: client-go added resource-control response accounting, which only inspected top-level execution details.
- February 2023: TiDB enabled store batching by default.

The accounting gap may therefore have affected resource-control-enabled batched workloads since early 2023.

0xPoe:

Issue Number: Close #2043

`MakeResponseInfo` ignores execution details from nested batched coprocessor responses, causing resource control to undercount read bytes and KV CPU.

- Include scan bytes and KV CPU from every nested batch response.
- Fall back to response data size when scan details are unavailable.
- Preserve the existing legacy and NextGen scan-byte accounting rules.

There is no public API or wire-format change.

### Tests

- `go test ./internal/resourcecontrol -count=1`
- `go test -tags nextgen ./internal/resourcecontrol -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource usage reporting for batched coprocessor responses.
  * Read-byte metrics now accurately include embedded batch results, with appropriate fallbacks when scan details are unavailable.
  * KV CPU time reporting now includes processing time from all batched tasks.
  * Preserved existing read-byte behavior for legacy scan responses.
  * Improved accuracy of performance metrics across both newer and legacy scan processing modes.

### DIFF
--- a/internal/resourcecontrol/resource_control.go
+++ b/internal/resourcecontrol/resource_control.go
@@ -203,15 +203,29 @@ func MakeResponseInfo(resp *tikvrpc.Response) *ResponseInfo {
 	}
 	// Parse the response to extract the info.
 	var (
-		readBytes uint64
-		detailsV2 *kvrpcpb.ExecDetailsV2
-		details   *kvrpcpb.ExecDetails
+		readBytes        uint64
+		detailsV2        *kvrpcpb.ExecDetailsV2
+		details          *kvrpcpb.ExecDetails
+		batchedReadBytes uint64
+		batchedKVCPU     time.Duration
 	)
 	switch r := resp.Resp.(type) {
 	case *coprocessor.Response:
 		detailsV2 = r.GetExecDetailsV2()
 		details = r.GetExecDetails()
 		readBytes = uint64(r.Data.Size())
+		// A batched coprocessor request answers several tasks in one response.
+		// Each nested response carries execution details that are not included
+		// in the top-level details, so account for every nested task.
+		for _, batchResp := range r.GetBatchResponses() {
+			batchDetailsV2 := batchResp.GetExecDetailsV2()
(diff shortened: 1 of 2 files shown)


0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback)

- [x] Code compiles successfully

- [x] Unit tests added

- [x] No AI-generated elegant nonsense in PR.

- [x] Comments added where necessary

- [x] PR title and description updated

- [x] Documentation PR created (or confirmed not needed)

- [x] PR size is reasonable


/cc @cfzjywxk @time-and-fate

JmPotato (approved):

cfzjywxk (approved):

PR #2034, opened 2026-07-28
util/async: avoid blocking Append on Exec cancellation

### LINKED ISSUE #2033: util/async: cancellation race can block RunLoop.Append and batchRecvLoop indefinitely
## Summary

There is a cancellation race between `RunLoop.Exec(ctx)` and `RunLoop.Append()` that can leave `Append` blocked forever while sending on the unbuffered `ready` channel.

When `Append` is called by `async.Callback.Schedule` from `batchCommandsClient.batchRecvLoop`, the receive loop cannot finish dispatching or retiring that response. Later responses on the same batch-command stream are then head-of-line blocked even though the gRPC connection and server are healthy.

This is reproducible on current `master` at `61ecd7c114166c08e747d252fb9ae592e54bb599`.

## Root cause

[`RunLoop.Append`](https://github.com/tikv/client-go/blob/61ecd7c114166c08e747d252fb9ae592e54bb599/util/async/runloop.go#L74-L91) does the following:

1. Acquire `RunLoop.lock`.
2. Append the callback.
3. If the state is `StateWaiting`, change it to `StateIdle` and set `notify=true`.
4. Release the lock.
5. Send on the unbuffered `ready` channel.

At the same time, [`RunLoop.Exec`](https://github.com/tikv/client-go/blob/61ecd7c114166c08e747d252fb9ae592e54bb599/util/async/runloop.go#L100-L127) waits for either `ready` or `ctx.Done()`.

The problematic interleaving is:

1. `Exec(ctx)` enters `StateWaiting`.
2. The context is canceled while `Append` is starting.
3. `Append` observes `StateWaiting`, changes the state to `StateIdle`, and decides to notify `Exec`.
4. `Exec` selects `ctx.Done()`, resets the state, and returns.
5. `Append` executes `l.ready <- struct{}{}` after the only receiver has exited.
6. The unbuffered send blocks permanently.

[`Callback.Schedule`](https://github.com/tikv/client-go/blob/61ecd7c114166c08e747d252fb9ae592e54bb599/util/async/core.go#L74-L76) calls `Append` from inside `sync.Once.Do`, so the `sync.Once` operation also remains in progress.

For an asynchronous batch request, [`batchCommandsEntry.response`](https://github.com/tikv/client-go/blob/61ecd7c114166c08e747d252fb9ae592e54bb599/internal/client/client_batch.go#L203-L209) calls `Callback.Schedule`. If that blocks, [`batchRecvLoop`](https://github.com/tikv/client-go/blob/61ecd7c114166c08e747d252fb9ae592e54bb599/internal/client/client_batch.go#L1202-L1321) cannot delete/retire the request or continue processing later responses on that stream.

One observed blocked stack had this shape:

```text
batchCommandsClient.batchRecvLoop
  -> batchCommandsEntry.response
  -> callback.Schedule
  -> sync.Once.Do
  -> RunLoop.Append
  -> chan send
```

Other goroutines attempting to complete the same callback can block on the `sync.Once` mutex. Request counters then show a growing difference between tracked and retired requests while the gRPC connections remain `READY`.

When `EnableAsyncBatchGet` is enabled, the multi-region batch-get path can reach this code through:

```text
KVSnapshot.BatchGetWithTier
  -> batchGetKeysByRegions
  -> asyncBatchGetByRegions
  -> SendRequestAsync
  -> batchRecvLoop
  -> Callback.Schedule
  -> RunLoop.Append
```

## Minimal reproduction

The following test uses the package lock only to force the otherwise timing-dependent interleaving. It detects the bug when `Append` remains blocked after the canceled `Exec` has returned.

<details>
<summary>Reproduction test for util/async/runloop_test.go</summary>

```go
func TestCanceledExecCanStrandAppend(t *testing.T) {
	oldProcs := runtime.GOMAXPROCS(1)
	defer runtime.GOMAXPROCS(oldProcs)

	for attempt := 0; attempt < 100; attempt++ {
		loop := NewRunLoop()
		ctx, cancel := context.WithCancel(context.Background())
		execDone := make(chan error, 1)
		go func() {
			_, err := loop.Exec(ctx)
			execDone <- err
		}()

		deadline := time.Now().Add(time.Second)
		for loop.State() != StateWaiting {
			if time.Now().After(deadline) {
				t.Fatal("run loop did not enter waiting state")
			}
			runtime.Gosched()
		}

		// Queue Append on the mutex before waking Exec through cancellation.
		loop.lock.Lock()
		appendDone := make(chan struct{})
		go func() {
			loop.Append(func() {})
			close(appendDone)
		}()
		runtime.Gosched()
		cancel()
		runtime.Gosched()
		loop.lock.Unlock()

		select {
		case err := <-execDone:
			if err == nil {
				t.Fatal("expected canceled Exec to return an error")
			}
		case <-time.After(time.Second):
			t.Fatal("Exec did not return after cancellation")
		}

		select {
		case <-appendDone:
			// Mutex wake-up ordering may select Exec first; retry.
		case <-time.After(20 * time.Millisecond):
			t.Logf("reproduced stranded RunLoop.Append on attempt %d", attempt+1)
			return
		}
	}

	t.Fatal("did not reproduce the cancellation/notification interleaving")
}
```

Required imports:

```go
import (
	"context"
	"runtime"
	"testing"
	"time"
)
```

</details>

Run:

```bash
go test ./util/async -run '^TestCanceledExecCanStrandAppend$' -count=20 -v
```

On `darwin/arm64` with Go `1.25.10`, all 20 runs reproduced the stranded `Append` on the first attempt:

```text
reproduced stranded RunLoop.Append on attempt 1
PASS
```

This is a positive reproducer: it passes when the blocking bug is observed. A permanent regression test should invert the final assertion and require `Append` to complete.

## Expected behavior

Cancellation of the `RunLoop` executor must not cause a concurrent or late callback submission to block indefinitely. A late response should be safely executed, rejected, or discarded without stopping the batch receive loop.

Requests following a canceled asynchronous request on the same batch-command stream should continue to receive and retire their responses.

## Actual behavior

`RunLoop.Append` can block forever on the unbuffered wake-up send. When this happens in `batchRecvLoop`, one callback causes stream-wide head-of-line blocking:

- the response being dispatched is not retired;
- later responses on that stream are not processed;
- tracked-but-not-retired requests accumulate;
- unrelated client requests sharing the stream can hang or time out.

## Suggested regression coverage

In addition to the `util/async` interleaving test, an `internal/client` regression test could:

1. Start a mock batch-command server.
2. Issue an asynchronous request using a `RunLoop` callback.
3. Cancel its context concurrently with the server response.
4. Send an unrelated synchronous request through the same batch stream.
5. Verify that the second request completes and both request entries are retired.

zyguan:
Fix #2033

`RunLoop.Append` could block indefinitely when racing with cancellation of a waiting `RunLoop.Exec`.

`Append` changes the state from `StateWaiting` to `StateIdle` under the mutex, then sends a notification through the
unbuffered `ready` channel after releasing the mutex. If `Exec` selected `ctx.Done()` and returned before receiving that
notification, the `Append` caller was left blocked. When invoked from an async batch response callback, this could stop
the batch receive loop from retiring the request and processing later responses on the same stream.

Changes:
- Make the cancellation path distinguish between a run loop that is still waiting and one for which an `Append` has
  already committed to sending a notification.
- Consume the committed notification before returning from `Exec`, ensuring the notifying `Append` can complete.
- Document the notification handshake and its single-`Exec` safety invariant.
- Add regression coverage for the cancellation race, queued-task preservation, and subsequent `Exec` behavior.
- Make the concurrent `Exec` test deterministic by synchronizing on the first executor entering `StateRunning`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved run loop behavior when execution is cancelled while another task is being added.
  * Prevented potential hangs during concurrent task scheduling.
  * Ensured the run loop remains usable for subsequent executions after cancellation.

* **Tests**
  * Added coverage for cancellation and concurrent task execution scenarios.

### DIFF
--- a/util/async/runloop.go
+++ b/util/async/runloop.go
@@ -37,7 +37,11 @@ type RunLoop struct {
 	ready    chan struct{}
 	runnable []func()
 	running  []func()
-	state    State
+
+	// state and ready form a notification handshake. Under Exec's single-goroutine contract, while Exec is waiting,
+	// Append is the only concurrent operation that may change StateWaiting to StateIdle. Exactly the Append that performs
+	// the transition sends one notification on ready after releasing lock.
+	state State
 }
 
 // NewRunLoop creates a new run-loop.
@@ -114,8 +118,17 @@ func (l *RunLoop) Exec(ctx context.Context) (int, error) {
 				continue
 			case <-ctx.Done():
 				l.lock.Lock()
-				l.state = StateIdle // waiting -> idle
-				l.lock.Unlock()
+				if l.state == StateWaiting {
+					// No Append has committed to sending a notification.
+					l.state = StateIdle // waiting -> idle
+					l.lock.Unlock()
+				} else {
+					// SAFETY: Under Exec's single-goroutine contract, Append is the only concurrent operation that can
+					// change StateWaiting to StateIdle. Exactly one Append performed that transition and must send one
+					// notification after releasing lock. Consume it before returning so the sender cannot be stranded.
+					l.lock.Unlock()
+					<-l.ready
+				}
 				return 0, ctx.Err()
 			}
 		} else {
--- a/util/async/runloop_test.go
+++ b/util/async/runloop_test.go
@@ -141,24 +141,89 @@ func TestExecCancelWhileWaiting(t *testing.T) {
 	require.Equal(t, 0, n)
 }
 
-func TestExecConcurrent(t *testing.T) {
-	defaultMaxProcs := runtime.GOMAXPROCS(0)
+func TestExecCancelWhileAppendNotifying(t *testing.T) {
+	defaultMaxProcs := runtime.GOMAXPROCS(1)
 	defer runtime.GOMAXPROCS(defaultMaxProcs)
-	runtime.GOMAXPROCS(1)
-	l := NewRunLoop()
-	l.Append(func() {
-		time.Sleep(time.Millisecond)
-	})
-	done := make(chan struct{})
-	go func() {
+
+	for attempt := 0; attempt < 100; attempt++ {
+		l := NewRunLoop()
+		called := false
+		ctx, cancel := context.WithCancel(context.Background())
+		execDone := make(chan error, 1)
+		go func() {
+			_, err := l.Exec(ctx)
+			execDone <- err
+		}()
+
+		require.Eventually(t, func() bool {
+			return l.State() == StateWaiting
+		}, time.Second, time.Millisecond, "run loop did not enter waiting state")
+
+		// Queue Append on the mutex before waking Exec through cancellation. If Append acquires the mutex first, it
+		// changes the state to idle and then tries to notify an Exec that has already selected ctx.Done().
+		l.lock.Lock()
+		appendDone := make(chan struct{})
+		go func() {
+			l.Append(func() {
+				called = true
+			})
+			close(appendDone)
+		}()
+		runtime.Gosched()
+		cancel()
+		runtime.Gosched()
+		l.lock.Unlock()
+
+		select {
+		case err := <-execDone:
+			require.ErrorIs(t, err, context.Canceled)
+		case <-time.After(time.Second):
+			t.Fatal("Exec did not return after cancellation")
+		}
+
+		select {
+		case <-appendDone:
+		case <-time.After(time.Second):
+			t.Fatalf("Append remained blocked after Exec cancellation on attempt %d", attempt+1)
+		}
+
+		require.Equal(t, StateIdle, l.State())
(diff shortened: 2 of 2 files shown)


zyguan:
@D3Hunter @cfzjywxk @lcwangchao PTAL

cfzjywxk (approved):

zyguan:
CI failed because #2048 introduced a flacky test `TestSnapshotRuntimeStatsAsyncBatchGetMultipleRegionsCoverage`. It has been fixed in #2074 , so we can merge #2074 first.

lcwangchao (approved):

PR #2039, opened 2026-08-03
tikv: support CES keyspace-level transaction safepoint

pingyu:

Essential v1 keyspaces advertise keyspace-level GC with `safe_point_version=v2`, while client-go only recognizes the native `gc_management_type=keyspace_level` metadata. The compatibility transaction safepoint loader therefore reads the unified GC key instead of the keyspace-scoped key.

- Route both native and Essential v1 keyspace-level GC metadata to `/keyspaces/tidb/{id}/tidb/store/gcworker/saved_safe_point`.
- Keep all other keyspaces on the unified GC key.
- Add focused coverage for nil, native, Essential v1, and unified metadata.

## Stack dependencies

- Draft and stacked on #2024 for API-v3 keyspace metadata compatibility.
- Uses the PD predicate introduced by tikv/pd#11100.

## Tests

- `go test ./tikv -run ^TestCompatibleTxnSafePointPath$ -count=1 -v`
- `gopls check tikv/compatible_txn_safe_point_loader.go tikv/compatible_txn_safe_point_loader_test.go`

The full `go test ./tikv -count=1` currently hits the same mockstore teardown nil-pointer panic in `KVStore.updateSafeTS` on the clean #2024 stacked base; the focused changed behavior passes.


pingyu:
Superseded by #2040, which implements the CES predicate locally and avoids the API-v3 dependency blast while retaining tikv/pd#11100 as the canonical PD helper.

PR #2040, opened 2026-08-03
tikv: support CSE keyspace-level transaction safepoint

pingyu:

Related PD issue: tikv/pd#11108

Related downstream issue: pingcap/ticdc#5785

Essential v1 keyspaces advertise keyspace-level GC with `safe_point_version=v2`, while client-go only recognizes the native `gc_management_type=keyspace_level` metadata. The compatibility transaction safepoint loader therefore reads the unified GC key instead of the keyspace-scoped key.

- Add `tikv.IsCSEKeyspaceLevelGC` locally so current client-go can recognize CSE metadata without pulling the API-v3 PD dependency graph.
- Route both native and CSE keyspace-level GC metadata to `/keyspaces/tidb/{id}/tidb/store/gcworker/saved_safe_point`.
- Keep all other keyspaces on the unified GC key.
- Document that the local helper should delegate to `pd.IsCSEKeyspaceLevelGC` after API-v3 support is merged into client-go.

The canonical PD helper remains in tikv/pd#11100. This PR intentionally has no module dependency changes.

## Tests

- `go test ./tikv -run "^(TestIsCSEKeyspaceLevelGC|TestCompatibleTxnSafePointPath)$" -count=1 -v`
- `gopls check tikv/compatible_txn_safe_point_loader.go tikv/compatible_txn_safe_point_loader_test.go`

### DIFF
--- a/tikv/compatible_txn_safe_point_loader.go
+++ b/tikv/compatible_txn_safe_point_loader.go
@@ -10,6 +10,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/pingcap/kvproto/pkg/keyspacepb"
 	"github.com/pkg/errors"
 	"github.com/tikv/client-go/v2/internal/apicodec"
 	"github.com/tikv/client-go/v2/util"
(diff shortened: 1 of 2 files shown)


pingyu:
Superseded by tikv/pd#11100 commit 31378cb, which extends pd.IsKeyspaceUsingKeyspaceLevelGC to recognize safe_point_version=v2. The API v3 base in #2024 already calls that predicate in compatibleTxnSafePointLoader, so no client-go-specific OR/helper is needed. Verified #2024 unchanged against the reviewed PD client with: go test ./tikv -run TestPD11100RecognizesV2TxnSafePointPath -count=1.

PR #2041, opened 2026-08-04
internal/locate: probe follower once when the cached leader keeps rejecting leader reads with ServerIsBusy(0)

### LINKED ISSUE #2028: region-cache: leader-read keeps retrying same store on `ServerIsBusy` without `EstimatedWaitMs` (stale leader cache)
## Description
For leader reads, when TiKV returns `ServerIsBusy` without `EstimatedWaitMs` (e.g. unified read pool full / `MaxPendingTasksExceeded`), replicaSelectorV2 does not switch to other replicas and does not invalidate the region cache. It only marks the store slow, backs off (`BoTiKVServerBusy`), and retries the same cached leader.
This becomes a severe availability gap when the cached leader store is `half-dead`:
- Raft / PD has already transferred leaders away from that store
- The old requests on this store were stuck at `async_snapshot` forever(https://github.com/tikv/tikv/issues/18491) which will lead to increasing running tasks(never done and will not timeout)
- When the new request comes, the store can still answer `gRPC` and return `ServerIsBusy` before Raft-level `NotLeader` checks
- TiDB keeps hammering the stale leader until the store becomes unreachable
https://github.com/tikv/client-go/blob/4d50d6744f0c276a0814e44504fd23debd1a9658/internal/locate/replica_selector.go#L483-L526

metrics also showed: after leader count on the store dropped to 0, unified read-pool running tasks stayed high with ~0 CPU, while gRPC was still alive — consistent with returning `ServerIsBusy` instead of `NotLeader`.

## Expected behavior

When a leader-read hits ServerIsBusy (especially with EstimatedWaitMs == 0) repeatedly on the cached leader, client-go should do one of:
- Invalidate / refresh region cache (so a later attempt can discover the new leader), or
- Fail over to other replicas (follower read / reload leader from PD), instead of backoff+retrying the same peer only.

## Impact
Leader eviction / slow-store protection on the PD/Raft side cannot fully protect TiDB latency if the half-dead store keeps answering ServerIsBusy and the client never updates its leader cache.

## workaround:
   stop the `half-dead` tikv instance

ekexium:
Close tikv/client-go#2028

## Problem

When a TiKV store becomes half-dead (e.g. unified read pool wedged, tikv/tikv#18491),
leader reads are rejected at the read-pool entrance with `ServerIsBusy` and
`EstimatedWaitMs == 0`, before the request reaches the raft layer. Even after PD has
moved the leader away, the store keeps answering gRPC with `ServerIsBusy(0)` instead of
`NotLeader`. The client only marks the store slow and backs off, retrying the same
cached leader indefinitely — a production incident lasted ~25 minutes and ended only
when the store became fully unreachable.

## Fix

"Sleep first, probe once":

- The 1st `ServerIsBusy(0)` on the cached leader behaves exactly as today (mark slow +
  backoff + retry leader) — transient busyness costs nothing.
- On the 2nd `ServerIsBusy(0)` from the same cached leader within one selector,
  mark the leader replica with a new dedicated `suspectNotLeaderFlag` (at most once
  per selector; the count restarts only when the cached leader changes).
- The next attempt skips the flagged leader, so the existing mixed strategy picks a
  follower with unchanged request semantics (`req.ReplicaRead` stays false). A follower
  can only reject a leader read with `NotLeader` + leader hint — never serve it — so
  there is no stale-read risk.
- The hint drives the existing `onNotLeader` → `updateLeader` path, healing the shared
  region cache for all subsequent requests. `onUpdateLeader` clears the flag, so a
  misjudgment (the store is still the leader) costs exactly one rejected RPC and the
  request reverts to today's behavior.
- If the followers yield no hint (election in progress, or followers also
  busy/exhausted), the selector restores the cached leader when the mixed strategy
  finds no candidate, instead of invalidating the region: it falls back to the plain
  backoff-and-retry loop, keeping essentially today's retry shape in an all-busy
  meltdown and adding no PD traffic. Region invalidation / PD reload still happen only
  through the pre-existing paths (e.g. after the leader exhausts its attempts).

No read/write semantics change, no extra PD traffic on the busy path, no new state
machines/timers/configs. Bounded extra cost: the probe fires at most once per
selector — when it succeeds (or is a misjudgment) it costs one rejected RPC; in the
worst case (all followers also busy) each follower is contacted once before the leader
is restored. Applies to reads and writes alike — a write rejected by a follower gets
the same NotLeader treatment.

## What it does NOT cover

The window where the wedged store is still the leader (before PD eviction): the hint
points back to the same store and behavior degrades to status quo; that window is
covered by PD slow-store eviction. Complementary to the server-side fix
tikv/tikv#19932 (return NotLeader at the rejection gate): that one heals old clients
on new TiKV, this one heals old TiKV on new clients.

The forwarding (proxy) path is out of scope: `EnableForwarding` defaults to false, and
this change is designed and tested for the default selection path only — its
interaction with `ReplicaSelectLeaderWithProxyStrategy` is not covered.

A compound failure — wedged old leader AND all followers also rejecting with
ServerIsBusy(0) AND PD has already moved the leader — is intentionally not cured
eagerly: the probe cannot obtain a hint, and invalidating/reloading would tax PD in
the (much more common) case where the cache is actually correct. The request falls
back to today's behavior; the cure arrives once any follower can answer again.

## Tests

`TestReplicaSelectorLeaderBusyProbe`: no probe on a single busy; probe on the 2nd busy
with leader-read semantics kept; cache healed via hint and a new selector goes
straight to the new leader; misjudgment restores the leader with no further probing;
no-hint goes through the region-scheduling backoff path; stale/follower/leaderOnly
reads never probe; a single-replica region (no probe target) restores the leader
without invalidating; the busy count restarts when the cached leader changes.

Two existing access-path tests were updated: their expectations encoded the old
behavior this PR removes (leader hammered until exhausted before trying a follower).
The all-busy meltdown case now ends with the leader restored — twelve accesses ending
in success with the region cache intact, essentially the same retry shape as before
this PR, with only one early fruitless probe of each follower.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request routing when a cached leader repeatedly reports that it is busy.
  - Temporarily probes an available follower and preserves leader-read behavior where appropriate.
  - Automatically recovers from stale or incorrect leader information.
  - Prevents unnecessary probing for stale reads, follower-only requests, leader-only requests, and single-replica configurations.
  - Improves handling of transient busy responses while avoiding false leader suspicions.

### DIFF
--- a/internal/locate/replica_selector.go
+++ b/internal/locate/replica_selector.go
@@ -40,6 +40,11 @@ type replicaSelector struct {
 	proxy                     *replica
 	attempts                  int
 	regionInvalidatedForRetry bool // set when region is hard-invalidated but this selector should still retry on leader
+	// leaderBusyCount counts the ServerIsBusy(0) errors received on the cached leader, and
+	// leaderBusyProbed records whether a suspect-not-leader probe has been fired. Both are
+	// per-selector states used to work around tikv/client-go#2028, see onServerIsBusy.
+	leaderBusyCount  int
+	leaderBusyProbed bool
 }
 
 // disableReadFeaturesForNextGen disables replica-read and stale-read feature
@@ -244,7 +249,10 @@ type ReplicaSelectLeaderStrategy struct {
 
 func (s ReplicaSelectLeaderStrategy) next(replicas []*replica) *replica {
 	leader := replicas[s.leaderIdx]
-	if isLeaderCandidate(leader) {
+	// Skip a leader replica flagged with suspectNotLeaderFlag, so that the selector falls
+	// back to the mixed strategy and probes a follower to trigger a NotLeader reply which
+	// heals the region cache (tikv/client-go#2028).
+	if isLeaderCandidate(leader) && !leader.hasFlag(suspectNotLeaderFlag) {
 		return leader
 	}
 	return nil
@@ -583,6 +591,25 @@ func (s *replicaSelector) onServerIsBusy(
 		} else {
 			// Mark the server is busy (the next incoming READs could be redirected to expected followers.)
 			ctx.Store.healthStatus.markAlreadySlow()
+			// Workaround for tikv/client-go#2028: if the store's read pool is wedged, leader
+			// reads are rejected with ServerIsBusy(0) at the pool entrance, so the request
+			// never reaches the raft layer and no NotLeader error is returned even if PD has
+			// already moved the leader away. Retrying the cached leader then hammers the
+			// half-dead store indefinitely. After 2 consecutive such rejections on the cached
+			// leader, mark it suspect-not-leader so that the next attempt probes a follower
+			// with the leader read (req.ReplicaRead is kept unchanged). The follower replies
+			// NotLeader with the real leader hint, which heals the shared region cache via
+			// onNotLeader/updateLeader. Probe at most once per selector; if the store is
+			// still the leader, the hint points back to it, onUpdateLeader clears the flag,
+			// and the only cost is one rejected RPC.
+			if s.replicaReadType == kv.ReplicaReadLeader && !s.isStaleRead && !s.option.leaderOnly &&
+				s.target != nil && s.target.peer.Id == s.region.GetLeaderPeerID() && !s.leaderBusyProbed {
+				s.leaderBusyCount++
+				if s.leaderBusyCount >= 2 {
+					s.target.addFlag(suspectNotLeaderFlag)
+					s.leaderBusyProbed = true
+				}
+			}
 		}
 	}
 	backoffErr := newBackoffErrWithRPCContext("server is busy", ctx)
--- a/internal/locate/replica_selector.go
+++ b/internal/locate/replica_selector.go
@@ -609,8 +609,9 @@ func (s *replicaSelector) onServerIsBusy(
 			// reads are rejected with ServerIsBusy(0) at the pool entrance, so the request
 			// never reaches the raft layer and no NotLeader error is returned even if PD has
 			// already moved the leader away. Retrying the cached leader then hammers the
-			// half-dead store indefinitely. After 2 consecutive such rejections on the cached
-			// leader, mark it suspect-not-leader so that the next attempt probes a follower
+			// half-dead store indefinitely. After 2 such rejections from the same cached
+			// leader within this selector (the count restarts only when the cached leader
+			// changes), mark it suspect-not-leader so that the next attempt probes a follower
 			// with the leader read (req.ReplicaRead is kept unchanged). The follower replies
 			// NotLeader with the real leader hint, which heals the shared region cache via
 			// onNotLeader/updateLeader. Probe at most once per selector; if the store is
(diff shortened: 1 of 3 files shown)


internal/locate/replica_selector.go
```diff
@@ -583,6 +605,34 @@ func (s *replicaSelector) onServerIsBusy(
...
+				// The count belongs to a specific cached leader: whenever the cached leader
+				// changes (e.g. switched by a NotLeader hint), restart the count for the new
+				// leader so that it won't be marked after inheriting the old leader's count.
+				leaderPeerID := s.region.GetLeaderPeerID()
+				if s.leaderBusyPeerID != leaderPeerID {
+					s.leaderBusyPeerID = leaderPeerID
+					s.leaderBusyCount = 0
+				}
+				s.leaderBusyCount++
+				if s.leaderBusyCount >= 2 {
```
cfzjywxk: What is the impact for case the leader is busy for real? Backoff would be skipped for some times?

ekexium: It won't be skipped. The sequence would be `leader: busy -> leader: busy -> follower: not leader(leader is the original one) -> leader...` The test case at line 2585 covered such situation

cfzjywxk (approved):

internal/locate/replica_selector.go
```diff
@@ -583,6 +605,34 @@ func (s *replicaSelector) onServerIsBusy(
...
+			// NotLeader with the real leader hint, which heals the shared region cache via
+			// onNotLeader/updateLeader. Probe at most once per selector; if the store is
+			// still the leader, the hint points back to it, onUpdateLeader clears the flag,
+			// and the only cost is one rejected RPC.
+			if s.replicaReadType == kv.ReplicaReadLeader && !s.isStaleRead && !s.option.leaderOnly &&
+				s.target != nil && s.target.peer.Id == s.region.GetLeaderPeerID() && !s.leaderBusyProbed {
+				// The count belongs to a specific cached leader: whenever the cached leader
+				// changes (e.g. switched by a NotLeader hint), restart the count for the new
+				// leader so that it won't be marked after inheriting the old leader's count.
+				leaderPeerID := s.region.GetLeaderPeerID()
```
zyguan: Should we use `leaderPeerID := s.region.GetLeaderPeerID()` here to avoid double read of `region.GetLeaderPeerID()`?

internal/locate/replica_selector.go
```diff
@@ -583,6 +609,34 @@ func (s *replicaSelector) onServerIsBusy(
 		} else {
 			// Mark the server is busy (the next incoming READs could be redirected to expected followers.)
 			ctx.Store.healthStatus.markAlreadySlow()
+			// Workaround for tikv/client-go#2028: if the store's read pool is wedged, leader
```
AndreMouche, line 612: `ServerIsBusy.estimated_wait_ms = 0` seems must not be interpreted as “the store is healthy” or “expected wait time is zero.” in this case, could we add more comments for `ServerIsBusy.estimated_wait_ms = 0`

AndreMouche (approved):
LGTM

PR #2042, opened 2026-08-04
retry: delegate killed checks to signal handler

YangKeao:
## Problem

TiDB needs to run richer kill-signal handling at interruptible TiKV retry checkpoints so it can cooperatively detect a disconnected client while a statement is blocked in TiKV. The existing client-go code only reads the legacy `Killed` flag, and the 2PC path duplicates that check.

This is needed by [pingcap/tidb#68682](https://github.com/pingcap/tidb/issues/68682).

Downstream integration: [pingcap/tidb#70343](https://github.com/pingcap/tidb/pull/70343).

## What changed

- Add the optional `kv.KillSignalHandler` interface to `kv.Variables`.
- Let `Backoffer.CheckKilled` delegate to the handler when configured, while preserving the existing `Killed` fallback for compatibility.
- Reuse `Backoffer.CheckKilled` for interruptible 2PC actions instead of duplicating the atomic kill-flag check.
- Add unit coverage for handler precedence, legacy fallback, and nil variables.

## Compatibility

This is an additive API change. Existing callers that only populate `Variables.Killed` keep the previous behavior. TiDB can opt in by installing its SQL kill-signal handler.

## Tests

- `go test ./config/retry ./txnkv/transaction`

## Summary by CodeRabbit

* **New Features**
  * Added configurable handling for kill signals during interruptible operations.
  * Kill-signal handlers can now provide custom interruption errors.
  * Existing kill-state behavior remains supported as a fallback.

* **Bug Fixes**
  * Improved cancellation checks before processing transaction batches.
  * Correctly reports query interruptions caused by kill signals.
  * Avoids unnecessary checks when kill-related settings are unavailable.
  * Improved reliability when kill-signal handling changes during concurrent operations.

### DIFF
--- a/config/retry/backoff.go
+++ b/config/retry/backoff.go
@@ -447,7 +447,7 @@ func (b *Backoffer) CheckKilled() error {
 	if b.vars == nil {
 		return nil
 	}
-	if handler := b.vars.LoadKillSignalHandler(); handler != nil {
+	if handler := b.vars.KillSignalHandler; handler != nil {
 		return handler.HandleSignal()
 	}
 	if b.vars.Killed != nil {
--- a/config/retry/backoff.go
+++ b/config/retry/backoff.go
@@ -447,9 +447,6 @@ func (b *Backoffer) CheckKilled() error {
 	if b.vars == nil {
 		return nil
 	}
-	if handler := b.vars.KillSignalHandler; handler != nil {
-		return handler.HandleSignal()
-	}
 	if b.vars.Killed != nil {
 		killed := atomic.LoadUint32(b.vars.Killed)
 		if killed != 0 {
@@ -459,6 +456,9 @@ func (b *Backoffer) CheckKilled() error {
 			)
 			return errors.WithStack(tikverr.ErrQueryInterruptedWithSignal{Signal: killed})
 		}
+	}
+	if b.vars.KillSignalHandler != nil {
+		return b.vars.KillSignalHandler.HandleSignal()
 	}
 	return nil
 }
--- a/config/retry/backoff.go
+++ b/config/retry/backoff.go
@@ -444,7 +444,13 @@ func (b *Backoffer) longestSleepCfg() (*Config, int) {
 }
 
 func (b *Backoffer) CheckKilled() error {
-	if b.vars != nil && b.vars.Killed != nil {
+	if b.vars == nil {
+		return nil
+	}
+	if handler := b.vars.LoadKillSignalHandler(); handler != nil {
+		return handler.HandleSignal()
+	}
+	if b.vars.Killed != nil {
 		killed := atomic.LoadUint32(b.vars.Killed)
 		if killed != 0 {
 			logutil.BgLogger().Info(
--- a/kv/variables.go
+++ b/kv/variables.go
@@ -34,6 +34,15 @@
 
 package kv
 
+import "sync/atomic"
+
+// KillSignalHandler handles kill signals at interruptible KV request checkpoints.
+// Implementations must be safe for concurrent use.
+type KillSignalHandler interface {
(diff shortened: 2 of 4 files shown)


kv/variables.go
```diff
@@ -56,6 +65,30 @@ type Variables struct {
...
+
+	// killSignalHandler takes precedence over Killed when it is set. It is stored
+	// atomically because Variables can be shared by concurrent backoffers.
+	killSignalHandler atomic.Pointer[KillSignalHandler]
+}
+
+// SetKillSignalHandler atomically sets the handler used at interruptible KV request
+// checkpoints. Passing nil clears the handler.
+func (v *Variables) SetKillSignalHandler(handler interface{ HandleSignal() error }) {
+	if handler == nil {
```
ekexium: Nit: we can also check if it's a typed nil

kv/variables.go
```diff
@@ -56,6 +65,30 @@ type Variables struct {
 	// TxnFileMinMutationSize is the minimum size of mutations to use file-based txn.
 	// When its value is 0, use the config of "txn-file-min-mutation-size".
 	TxnFileMinMutationSize uint64
+
+	// killSignalHandler takes precedence over Killed when it is set. It is stored
+	// atomically because Variables can be shared by concurrent backoffers.
+	killSignalHandler atomic.Pointer[KillSignalHandler]
```
ekexium: AI review:
  > Do we need to replace the handler concurrently with CheckKilled? Backoffers already share the same *Variables, so an initialization-only interface field would still let them concurrently use the same handler safely.
  >
  > If there is no runtime replacement path, please make SetKillSignalHandler lifecycle-only instead of embedding atomic.Pointer, which makes the public Variables non-copyable and forces downstream field-by-field cloning. If concurrent replacement is required, please document the use case and provide a client-go-owned clone API.

ekexium (commented):
The rest LGTM

ekexium (approved):

kv/variables.go
```diff
@@ -56,6 +65,40 @@ type Variables struct {
...
+	}
+	v.killSignalHandler = handler
+}
+
+// LoadKillSignalHandler returns the handler used at interruptible KV request checkpoints.
+func (v *Variables) LoadKillSignalHandler() KillSignalHandler {
+	return v.killSignalHandler
+}
+
+func isNilKillSignalHandler(handler KillSignalHandler) bool {
```
lcwangchao: Why does it need to introduce `isNilKillSignalHandler` instead of `handler == nil` ? Does this have any special case to make us to have to do this?

YangKeao: No. For the current use case `== nil` is enough. But I received a previews comment:
image

And I think it would be better than not :facepalm:. As this branch will not be called very frequently, it'll not introduce high performance influence.

kv/variables.go
YangKeao: After discuss with @lcwangchao, as we don't pass any typed nil here yet, it's too redudant to handle the case. I'll remove the typed nil check based on reflect.

config/retry/backoff.go
```diff
@@ -444,7 +444,13 @@ func (b *Backoffer) longestSleepCfg() (*Config, int) {
 }
 
 func (b *Backoffer) CheckKilled() error {
-	if b.vars != nil && b.vars.Killed != nil {
+	if b.vars == nil {
+		return nil
+	}
+	if handler := b.vars.LoadKillSignalHandler(); handler != nil {
```
lcwangchao: Do we need to aslo check `b.vars.Killed` here ? Is there any case we need to ignore `b.vars.Killed` ?

YangKeao: No. I will move it to the end! :+1:

lcwangchao (approved):

PR #2048, opened 2026-08-24
txnkv: expose point response runtime stats

### LINKED ISSUE #2057: txnkv: expose structured response stats for point reads
## Motivation

Coprocessor requests already carry `ScanDetailV2` through `ExecDetails.ScanDetail`, allowing TiDB to aggregate and consume scan details through its coprocessor runtime statistics.

Point Get and Batch Point Get bypass that coprocessor path. Their statistics are collected in `SnapshotRuntimeStats`, but upper layers currently cannot obtain:

- structured point-read scan details;
- whether scan details cover every response;
- the logical response payload size.

Without coverage information, a zero scan detail cannot be distinguished from missing execution details. This prevents upper layers from using point-read statistics reliably for resource accounting and observability.

## Proposal

Expose point-read response statistics from `SnapshotRuntimeStats`, containing:

- the aggregated scan detail;
- logical response payload bytes;
- whether scan-detail collection is complete.

Payload bytes use logical KV contents:

- Get: returned value bytes;
- BatchGet and BufferBatchGet: key and value bytes of successful pairs;
- misses and errors: zero bytes.

A recognized point-read response without `ScanDetailV2` marks scan-detail coverage as incomplete.

This information should be collected in client-go because it owns response decoding, retries, and point-read statistics aggregation.

winoros:
Close https://github.com/tikv/client-go/issues/2057
## Problem

`SnapshotRuntimeStats` aggregates point-read scan detail, but callers cannot tell whether every recognized response contributed scan detail, nor inspect the logical payload returned by Get, BatchGet, and BufferBatchGet. In particular, aggregate zero values alone cannot distinguish a real zero from missing response data.

## What changed

- Add `PointReadScanDetail` and `PointResponseStats`, exposed through `txnkv`, and add `SnapshotRuntimeStats.GetPointResponseStats()`.
- Track recognized response bodies, scan-detail records, payload records, and logical successful-response payload bytes independently so callers can check response-level completeness.
- Account for Get, BatchGet, and BufferBatchGet responses after transport and region handling, including retry, empty-response, and key-error semantics; fail closed on conversion or arithmetic overflow.
- Keep clone, merge, and accessor paths concurrency-safe, with unit and integration coverage for retry/error paths, async multi-region BatchGet, and pipelined BufferBatchGet.

## Compatibility

This is an additive public API change. It does not change existing exported signatures, wire protocols, or module dependencies.

## Tests

```text
go test -count=1 -race --tags=intest ./...
go test -count=1 --tags=intest ./txnkv/...
go vet ./txnkv/...
go generate ./...

cd integration_tests
go test -count=1 -race -run 'TestSnapshot/TestSnapshotRuntimeStats(GetAndBatchGetCoverage|ExcludesEpochNotMatch|PointGetLockRetryCoverage|AsyncBatchGetMultipleRegionsCoverage)$|TestPipelinedMemDB/TestBufferBatchGetPointResponseStats$' .

env GOLANGCI_LINT_CACHE=/tmp/client-go-golangci-cache \
  go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.8.0 \
  run --new-from-rev=origin/master ./txnkv/...

git diff --check origin/master
```

## Summary by CodeRabbit

* **New Features**
  * Added point-read response statistics for Get, BatchGet, and buffered batch-get operations.
  * Added tracking for scan details, payload sizes, response coverage, and completeness.
  * Extended statistics collection across asynchronous requests, retries, lock handling, and multi-region operations.

* **Bug Fixes**
  * Improved runtime-statistics accuracy when response details are missing or requests are retried.

* **Tests**
  * Added comprehensive coverage for recording, merging, invalidation, payload metrics, and response-statistics reporting.

### DIFF
--- a/txnkv/txnsnapshot/snapshot.go
+++ b/txnkv/txnsnapshot/snapshot.go
@@ -401,7 +401,11 @@ type batchGetLockInfo struct {
 func collectBatchGetResponseData(
 	resp *tikvrpc.Response,
 	onKvPair func([]byte, kv.ValueEntry),
-	onResponse func(*kvrpcpb.ExecDetailsV2, uint64, bool),
+	// onResponse receives the execution details and logical payload bytes.
+	onResponse func(
+		details *kvrpcpb.ExecDetailsV2,
+		payloadBytes uint64,
+	),
 ) (*batchGetLockInfo, error) {
 	if resp.Resp == nil {
 		return nil, errors.WithStack(tikverr.ErrBodyMissing)
@@ -424,10 +428,10 @@ func collectBatchGetResponseData(
 		return nil, errors.Errorf("unknown response %T", v)
 	}
 	if onResponse != nil {
-		payloadBytes, payloadValid := batchGetResponsePayloadBytes(data.keyErr, pairs)
+		payloadBytes := batchGetResponsePayloadBytes(data.keyErr, pairs)
 		// Record every recognized response, including missing ExecDetailsV2 and
 		// key errors, so missing detail remains visible across successful retries.
-		onResponse(details, payloadBytes, payloadValid)
+		onResponse(details, payloadBytes)
 	}
 	if data.keyErr != nil {
 		// If a response-level error happens, skip reading pairs.
@@ -466,25 +470,18 @@ func collectBatchGetResponseData(
 	return data, nil
 }
 
-func batchGetResponsePayloadBytes(responseError *kvrpcpb.KeyError, pairs []*kvrpcpb.KvPair) (uint64, bool) {
+func batchGetResponsePayloadBytes(responseError *kvrpcpb.KeyError, pairs []*kvrpcpb.KvPair) uint64 {
 	if responseError != nil {
-		return 0, true
+		return 0
 	}
 	var payloadBytes uint64
 	for _, pair := range pairs {
 		if pair.GetError() != nil {
 			continue
 		}
-		pairBytes, ok := checkedAddUint64(uint64(len(pair.GetKey())), uint64(len(pair.GetValue())))
-		if !ok {
-			return 0, false
-		}
-		payloadBytes, ok = checkedAddUint64(payloadBytes, pairBytes)
-		if !ok {
-			return 0, false
-		}
+		payloadBytes += uint64(len(pair.GetKey())) + uint64(len(pair.GetValue()))
 	}
-	return payloadBytes, true
+	return payloadBytes
 }
 
 //go:noinline
@@ -908,7 +905,7 @@ func (s *KVSnapshot) get(ctx context.Context, bo *retry.Backoffer, k []byte, opt
 			if cmdGetResp.GetError() != nil {
 				payloadBytes = 0
 			}
-			onResponse(cmdGetResp.ExecDetailsV2, payloadBytes, true)
+			onResponse(cmdGetResp.ExecDetailsV2, payloadBytes)
 		}
 		if cmdGetResp.ExecDetailsV2 != nil {
 			readKeys := len(cmdGetResp.Value)
@@ -978,7 +975,7 @@ func (s *KVSnapshot) mergeExecDetail(detail *kvrpcpb.ExecDetailsV2) {
 	if detail == nil || s.mu.stats == nil {
 		return
 	}
-	s.mu.stats.pointResponseStats.recordScanDetail(detail.ScanDetailV2)
+	s.mu.stats.scanDetail.MergeFromScanDetailV2(detail.GetScanDetailV2())
 	s.mu.stats.timeDetail.MergeFromTimeDetail(detail.TimeDetailV2, detail.TimeDetail)
 	if details := detail.GetReadPoolTaskDetails(); details != nil {
 		if s.mu.stats.readPoolTaskDetails == nil {
@@ -988,7 +985,7 @@ func (s *KVSnapshot) mergeExecDetail(detail *kvrpcpb.ExecDetailsV2) {
 	}
 }
 
-func (s *KVSnapshot) pointResponseRecorder() func(*kvrpcpb.ExecDetailsV2, uint64, bool) {
+func (s *KVSnapshot) pointResponseRecorder() func(*kvrpcpb.ExecDetailsV2, uint64) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	if s.mu.stats == nil {
@@ -997,13 +994,18 @@ func (s *KVSnapshot) pointResponseRecorder() func(*kvrpcpb.ExecDetailsV2, uint64
 	return s.mergePointResponse
 }
 
-func (s *KVSnapshot) mergePointResponse(detail *kvrpcpb.ExecDetailsV2, payloadBytes uint64, payloadValid bool) {
+func (s *KVSnapshot) mergePointResponse(detail *kvrpcpb.ExecDetailsV2, payloadBytes uint64) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.mu.stats == nil {
 		return
 	}
-	s.mu.stats.pointResponseStats.recordResponse(detail.GetScanDetailV2(), payloadBytes, payloadValid)
+	scanDetail := detail.GetScanDetailV2()
+	s.mu.stats.pointResponseExtra.recordResponse(
+		scanDetail,
+		payloadBytes,
+	)
+	s.mu.stats.scanDetail.MergeFromScanDetailV2(scanDetail)
 	if detail == nil {
 		return
 	}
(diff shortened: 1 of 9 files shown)


txnkv/txnsnapshot/snapshot.go
```diff
@@ -1281,21 +1342,88 @@ func (s *KVSnapshot) SetPipelined(ts uint64) {
...
+		stats.CompletedResponses > 0 &&
+		stats.PayloadRecords == stats.CompletedResponses
+}
+
 // SnapshotRuntimeStats records the runtime stats of snapshot.
 type SnapshotRuntimeStats struct {
 	rpcStats            *locate.RegionRequestRuntimeStats
 	backoffSleepMS      map[string]int
 	backoffTimes        map[string]int
+	pointResponseMu     sync.RWMutex
```
qw4990: Can we merge `pointResponseMu` into `pointResponseStats`?

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1281,21 +1342,88 @@ func (s *KVSnapshot) SetPipelined(ts uint64) {
...
+	// BatchGet and BufferBatchGet contribute key plus value bytes for successful
+	// pairs. Misses, response errors, and pair errors contribute zero; commit
+	// timestamps, error text, protobuf fields, and transport framing are excluded.
+	PayloadBytes uint64
+	// ScanDetailRecords counts recognized responses carrying ScanDetailV2.
+	ScanDetailRecords uint64
+	// PayloadRecords counts recognized responses whose payload was accounted for.
+	PayloadRecords uint64
+	// CompletedResponses counts all recognized response bodies.
+	CompletedResponses uint64
```
qw4990: Could we remove these fields to simplify the implementation...

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1281,21 +1342,88 @@ func (s *KVSnapshot) SetPipelined(ts uint64) {
...
+}
+
 // SnapshotRuntimeStats records the runtime stats of snapshot.
 type SnapshotRuntimeStats struct {
 	rpcStats            *locate.RegionRequestRuntimeStats
 	backoffSleepMS      map[string]int
 	backoffTimes        map[string]int
+	pointResponseMu     sync.RWMutex
 	scanDetail          util.ScanDetail
+	pointResponseStats  PointResponseStats
```
qw4990: Could we also put `PointResponseStats` into the `util` package?

winoros:
I tested the current PR head (`bcde13d7ed1c811899d59fe360c9095cb03a98fe`) as a consumer from TiDB master at `e0cfb27df3212f1112bddba141aee3d6f743b3fb`.

The current `util.PointResponseStats` / `GetPointResponseStats()` API was sufficient for the TiDB statement-RU consumer. The consumer handles Point Get and Batch Point Get at the statement terminal: complete scan/payload evidence contributes RU inputs; registered local completion and recognized misses contribute zero remote work; missing, incomplete, invalid, or overflowing evidence fails closed.

Correctness validation on the TiDB side passed, including:

- Point Get and Batch Point Get executor lifecycle tests through EOF/Close.
- Complete response evidence, recognized misses, local completion, missing scan detail, invalid data, nil embedded stats, and missing providers.
- Focused failpoint-enabled tests, `make bazel_prepare`, `make lint`, formatting, and diff checks.

I also measured the combined overhead of this PR's client-go collection plus the TiDB statement-RU consumer.

Go session benchmarks (`-benchtime=1s -count=5 -benchmem -cpu=1`):

- Point Get median: 18,080 -> 18,232 ns/op (+0.84%); 276 -> 277 allocs/op.
- Batch Point Get median: 17,404 -> 17,777 ns/op (+2.14%); allocations remained at 272/op.
- The focused TiDB terminal evidence helper measured 45.48 ns/op for a registered no-response snapshot and 49.42 ns/op for a complete response, both with zero allocations.

TiUP + sysbench `oltp_point_select.lua` A/B:

- One million rows, 16 threads, prepared statements (`--db-ps-mode=auto`), uniform key distribution.
- Same tagged nightly PD/TiKV data for both binaries; each side used prewarm, one unrecorded 30-second warmup, then three 30-second runs.
- Baseline QPS: 84,858.72 / 85,859.68 / 85,821.71 (mean 85,513.37; median 85,821.71; p95 0.26 ms).
- PR + TiDB consumer QPS: 83,989.49 / 84,687.68 / 84,347.34 (mean 84,341.50; median 84,347.34; p95 0.27 ms).
- Mean throughput delta: -1.37%; median delta: -1.72%; all six runs had zero errors.

These measurements consistently put the observed local overhead in the low single digits. The sysbench runs used baseline-first order on a single-host playground, so the percentages should be treated as local observations rather than a statistically controlled production-capacity claim.

txnkv/txnsnapshot/snapshot.go
```diff
@@ -401,7 +401,7 @@ type batchGetLockInfo struct {
 func collectBatchGetResponseData(
 	resp *tikvrpc.Response,
 	onKvPair func([]byte, kv.ValueEntry),
-	onDetails func(*kvrpcpb.ExecDetailsV2),
+	onResponse func(*kvrpcpb.ExecDetailsV2, uint64, bool),
```
lcwangchao: Could you add some comments to explain the arguments in `onResponse '? Expecially the `uint64` and `bool`

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1281,12 +1341,63 @@ func (s *KVSnapshot) SetPipelined(ts uint64) {
...
+// PointReadScanDetail is the storage work reported by point-read responses.
+type PointReadScanDetail = util.PointReadScanDetail
+
+// PointResponseStats is a value snapshot of point-read response statistics.
+type PointResponseStats = util.PointResponseStats
+
+// pointResponseStats keeps the diagnostic scan detail and point-response
+// snapshot under one lock. Only the data, never the lock, is copied or merged.
+type pointResponseStats struct {
+	mu         sync.RWMutex
```
lcwangchao: Could you clarify why `pointResponseStats` needs its own mutex? Giving it independent synchronization makes its ownership and concurrency boundaries less clear, and makes it difficult for callers to determine whether they still need to hold the outer lock when accessing it.
A clearer approach would be to keep `pointResponseStats` lock-free and let its owner handle synchronization. For example, when it belongs to `KVSnapshot`, all mutations should be performed while holding `KVSnapshot.mu`. If the stats need to be exposed externally, return a clone instead of an internal pointer.
Is there a concrete use case that cannot be protected by the owner’s lock and therefore requires `pointResponseStats` itself to support concurrent access?

txnkv/txnsnapshot/snapshot.go
```diff
@@ -1281,12 +1341,63 @@ func (s *KVSnapshot) SetPipelined(ts uint64) {
...
+	s.stats.Merge(stats)
+}
+
 // SnapshotRuntimeStats records the runtime stats of snapshot.
 type SnapshotRuntimeStats struct {
 	rpcStats            *locate.RegionRequestRuntimeStats
 	backoffSleepMS      map[string]int
 	backoffTimes        map[string]int
-	scanDetail          util.ScanDetail
+	pointResponseStats  pointResponseStats
```
lcwangchao: It is confusing for pointResponseStats to replace ScanDetail, which is changed to a subfield of `pointResponseStats`, since ScanDetail is also needed when no point get is executed. Could we keep them as separate fields of SnapshotRuntimeStats?

util/point_response_stats.go
```diff
@@ -0,0 +1,144 @@
...
+	ProcessedKeys int64
+	// ProcessedKeysSize is the total size of the processed user keys and values.
+	ProcessedKeysSize int64
+}
+
+// PointResponseStats is a value snapshot of Get, BatchGet, and BufferBatchGet
+// response statistics. Its zero value is valid but has no response coverage.
+// Use RecordResponse and Merge to populate both the values and coverage state.
+// The value may be copied; callers must synchronize concurrent reads and writes.
+type PointResponseStats struct {
```
lcwangchao, line 33: Could we avoid storing PointResponseStats directly? SnapshotRuntimeStats could keep the existing ScanDetail together with a private pointReadResponseExtraStats containing only the additional payload, coverage, and validity state. GetPointResponseStats() could then construct and return a PointResponseStats value by projecting the relevant fields from ScanDetail. This would preserve the narrower public API without storing and aggregating the same scan counters twice.

txnkv/txnsnapshot/snapshot.go
```diff
@@ -946,7 +975,40 @@ func (s *KVSnapshot) mergeExecDetail(detail *kvrpcpb.ExecDetailsV2) {
 	if detail == nil || s.mu.stats == nil {
 		return
 	}
-	s.mu.stats.scanDetail.MergeFromScanDetailV2(detail.ScanDetailV2)
+	s.mu.stats.scanDetail.MergeFromScanDetailV2(detail.GetScanDetailV2())
```
lcwangchao, line 984: line 978 - 984 seems to be a duplicate code with mergePointResponse 1008-1017, could we reduce this because the further developer may need to update both of these two methods when ExecDetailsV2 is extended

lcwangchao: I found mergeExecDetail is never used in the production, clould we remove it?

lcwangchao (approved):

0xPoe (approved):
Thanks! :shipit:

cfzjywxk:
@winoros 
Please reference an related github issue,  clearly explain what requirement or specific scenario this change is intended to address.

If there is any expected significant performance regression, please inform @MimeLyc and @kennedy8312 in advance.

winoros:
> @winoros Please reference an related github issue, clearly explain what requirement or specific scenario this change is intended to address.
> 
> If there is any expected significant performance regression, please inform @MimeLyc and @kennedy8312 in advance.

Issue updated. And I have pasted the current benchmark results—no significant regression.

cfzjywxk (approved):

PR #2050, opened 2026-08-26
apicodec: decode StoreBatchCop response keys

0xPoe:

Issue Number: ref https://github.com/pingcap/tidb/issues/70665

Some API v2 response fields remained encoded after `DecodeResponse`. Reusing those keys can add the keyspace prefix twice during lock resolution, while caching them can mix encoded bucket boundaries with decoded user keys.

Decode API v2 keys in:

- `StoreBatchCop` child region errors and lock information
- `BucketVersionNotMatch.Keys`
- `SplitRegionResponse.Errors`

`BucketVersionNotMatch.Keys` is decoded before bucket metadata reaches the region cache. If a key has malformed mem-comparable encoding, `DecodeResponse` returns a decode error instead of exposing it to normal `BucketVersionNotMatch` handling.

API v1 is unchanged.

### Tests

- `go test ./internal/apicodec -count=1`
- `go test ./internal/locate -count=1`

### DIFF
--- a/internal/apicodec/codec_v2.go
+++ b/internal/apicodec/codec_v2.go
@@ -665,6 +665,16 @@ func (c *codecV2) DecodeResponse(req *tikvrpc.Request, resp *tikvrpc.Response) (
 		if err != nil {
 			return nil, err
 		}
+		for _, batchResp := range r.BatchResponses {
+			batchResp.RegionError, err = c.decodeRegionError(batchResp.RegionError)
+			if err != nil {
+				return nil, err
+			}
+			batchResp.Locked, err = c.decodeLockInfo(batchResp.Locked)
+			if err != nil {
+				return nil, err
+			}
+		}
 		r.Range, err = c.decodeCopRange(r.Range)
 		if err != nil {
 			return nil, err
--- a/internal/apicodec/codec_v2.go
+++ b/internal/apicodec/codec_v2.go
@@ -1014,6 +1014,18 @@ func (c *codecV2) decodeRegionError(regionError *errorpb.Error) (*errorpb.Error,
 		errInfo.CurrentRegions = decodedRegions
 	}
 
+	if errInfo := regionError.BucketVersionNotMatch; errInfo != nil {
+		// The region cache compares bucket boundaries with decoded user keys.
+		// BucketVersionNotMatch.Keys still uses API v2's mem-comparable,
+		// keyspace-prefixed region-key format, so caching it as-is would mix key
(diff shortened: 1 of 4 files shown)


0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback)

- [x] Code compiles successfully

- [x] Unit tests added

- [x] No AI-generated elegant nonsense in PR.

- [x] Comments added where necessary

- [x] PR title and description updated

- [x] Documentation PR created (or confirmed not needed)

- [x] PR size is reasonable


/cc @JmPotato @cfzjywxk

JmPotato (approved):
LGTM. The fix is reasonable and is placed at the right boundary: API v2 keys should be decoded in client-go before the response is exposed to callers. The tests also cover the exact double-prefix failure path by decoding the primary key and feeding it back into `CheckTxnStatus`.

One concern is the longer-term approach. We have fixed several similar gaps before, including #1986 and the “complete API v2 key-bearing command coverage” work in #1994/#1997, but another set of nested fields still escaped the codec. The current implementation relies on manually enumerating every key-bearing field in each request and response, so a new command or protobuf field can silently bypass the codec until a real execution path exposes it.

Could we follow up with a more fundamental safeguard to eliminate this class of issue? For example, we could maintain a declarative inventory of key-bearing field paths and generate the transformations, or at least add a descriptor-based completeness test that requires every key/range/`LockInfo`/`RegionError` path to be explicitly classified as decoded or intentionally opaque. Then a kvproto change would fail CI instead of relying on us to discover the omission later and manually fill the gap.

This concern does not block the current regression fix, but I think it is worth addressing as a follow-up so we do not keep repeating the same pattern.

internal/apicodec/codec_v2.go
```diff
@@ -1004,6 +1018,13 @@ func (c *codecV2) decodeRegionError(regionError *errorpb.Error) (*errorpb.Error,
 		errInfo.CurrentRegions = decodedRegions
 	}
 
+	if errInfo := regionError.BucketVersionNotMatch; errInfo != nil {
+		errInfo.Keys, err = c.DecodeBucketKeys(errInfo.Keys)
```
cfzjywxk, line 1027: Please add a malformed `BucketVersionNotMatch.Keys` case asserting that `DecodeResponse` returns an error, documenting that invalid bucket metadata is treated as a protocol error rather than silently retried.

0xPoe: Added. Thanks!

cfzjywxk (approved):

PR #2061, opened 2026-09-09
txnkv: prevent kill signals from aborting pessimistic rollback

King-Dylan:

TiDB can set a query kill signal after detecting that a client disconnected before an autocommit DML is committed. Pessimistic lock rollback currently reuses the transaction variables that contain that signal. If cleanup encounters a retryable region error, `Backoffer.CheckKilled` can abort the retry before the locks are released, potentially leaving them to subsequent cleanup or TTL-based lock resolution.

The rollback backoffer's coupling to query-kill state predates pingcap/tidb#68481. That change introduced additional interruption points for disconnected autocommit DML, which can expose this pre-existing cleanup weakness: the kill signal used to stop the statement can also abort its rollback retries. This PR addresses that interaction, as reported in pingcap/tidb#70974.

Related to pingcap/tidb#70974.

- Add a cleanup backoffer that copies the transaction retry settings but removes both the legacy `Killed` pointer and the newer `KillSignalHandler`.
- Use it for synchronous transaction rollback, asynchronous statement-level pessimistic rollback, and failed-commit cleanup (`BatchRollback` for 2PC/async commit and pessimistic lock rollback for failed 1PC).
- Keep cleanup bounded by the original context and backoff limit. Normal lock, prewrite, and commit requests remain interruptible.

### Scope and limitations

This PR prevents query-kill signals from aborting rollback retries on the covered pessimistic-rollback and failed-commit cleanup paths. It does not change SQL cancellation, disconnect handling, or commit semantics, and does not make canceled statements resume execution or commit.

The context supplied to each rollback path is unchanged. In particular, asynchronous pessimistic rollback still inherits its caller's context and may fail if that context is canceled or its deadline expires. The full-transaction `rollbackPessimisticLocks` path already creates a context based on `context.Background()`; that existing behavior is also unchanged. Failed-commit cleanup already uses a store-derived context carrying `TxnStartKey`, which is preserved as well. Removing the query-kill signal therefore does not guarantee successful cleanup in every cancellation or failure scenario.

The canceled-request-context limitation is separate from the kill-signal interference addressed here. This PR does not establish it as a regression introduced or worsened by pingcap/tidb#68481, nor claim that a context change is required to fix the specific mechanism reported in pingcap/tidb#70974.

### Tests

- `go test ./txnkv/transaction -count=1`
- `go test -race ./txnkv/transaction -run '^TestPessimisticRollbackIgnoresKillSignal$' -count=1`

`TestPessimisticRollbackIgnoresKillSignal` covers synchronous and asynchronous rollback with both kill mechanisms. It injects a retryable `NotLeader` response and verifies that cleanup sends the second rollback request. With the original rollback backoffers, all four cases stop after the first request and fail.

That original test does not cover canceled request contexts and is not an end-to-end reproduction of the full concurrent workload in pingcap/tidb#70974.

The review follow-up adds `TestCommitterCleanupIgnoresKillSignal` for optimistic 2PC, pessimistic 2PC, async commit, and pessimistic 1PC, with both kill mechanisms and live/canceled store contexts (16 subcases). The first rollback RPC returns `NotLeader`. Before the follow-up fix, all 8 live-store cases fail after only one request; after the fix, all 16 subcases pass. The test also checks that an already-canceled caller context does not replace the existing store-context boundary, store cancellation still stops retries, `TxnStartKey` and rollback key/version fields are preserved, and the original transaction kill state remains intact.

Additional local validation for the review follow-up (Go 1.25.12, macOS/arm64):

- `go test -race -tags=intest ./txnkv/transaction ./config/retry -count=1`
- `go test -race -tags=intest ./txnkv/transaction -run '^(TestCommitterCleanupIgnoresKillSignal|TestPessimisticRollbackIgnoresKillSignal)

### Possible follow-up — not implemented in this PR

A separate fix could give asynchronous pessimistic rollback a bounded cleanup context independent of request cancellation, while respecting store shutdown and preserving required request metadata. This is a candidate approach, not an implemented or validated fix. It would require separate validation of canceled-context cleanup, transaction/key/lock-version scope, interaction with statement retries, shutdown behavior, and cleanup resource usage under cancellation-heavy workloads. No zero-regression or guaranteed-cleanup claim is made for that proposal.

## Summary by CodeRabbit

- **Bug Fixes**
  - Pessimistic transaction cleanup now completes reliably even when a query kill signal is active.
  - Rollback retries continue after transient region errors, ensuring all required cleanup requests are issued.
  - Commit-failure cleanup now continues within the store’s cancellation boundary, even if the original request is canceled.
  - Cleanup operations consistently complete across supported transaction modes, including asynchronous commit and one-phase commit scenarios.

### DIFF
--- a/txnkv/transaction/txn.go
+++ b/txnkv/transaction/txn.go
@@ -1062,7 +1062,7 @@ func (txn *KVTxn) rollbackPessimisticLocks() error {
 		return nil
 	}
 	ctx := context.WithValue(context.Background(), util.RequestSourceKey, *txn.RequestSource)
-	bo := retry.NewBackofferWithVars(ctx, cleanupMaxBackoff, txn.vars)
+	bo := newCleanupBackoffer(ctx, cleanupMaxBackoff, txn.vars)
 	if txn.interceptor != nil {
 		// User has called txn.SetRPCInterceptor() to explicitly set an interceptor, we
 		// need to bind it to ctx so that the internal client can perceive and execute
@@ -2123,6 +2123,19 @@ func deduplicateKeys(keys [][]byte) [][]byte {
 
 const pessimisticRollbackMaxBackoff = 20000
 
+// newCleanupBackoffer preserves the transaction's retry settings while preventing
+// a query kill signal from aborting compensating cleanup such as pessimistic lock
+// rollback. Cleanup is still bounded by the backoff limit and context.
+func newCleanupBackoffer(ctx context.Context, maxSleep int, vars *tikv.Variables) *retry.Backoffer {
+	if vars == nil {
+		return retry.NewBackofferWithVars(ctx, maxSleep, nil)
+	}
+	cleanupVars := *vars
+	cleanupVars.Killed = nil
+	cleanupVars.KillSignalHandler = nil
+	return retry.NewBackofferWithVars(ctx, maxSleep, &cleanupVars)
+}
+
 // asyncPessimisticRollback rollbacks pessimistic locks of the current transaction on the specified keys asynchronously.
 // Pessimistic locks on specified keys with its forUpdateTS <= specifiedForUpdateTS will be unlocked. If 0 is passed
 // to specifiedForUpdateTS, the current forUpdateTS of the current transaction will be used.
@@ -2161,7 +2174,7 @@ func (txn *KVTxn) asyncPessimisticRollback(ctx context.Context, keys [][]byte, s
 			}
 		}
 
-		err := committer.pessimisticRollbackMutations(retry.NewBackofferWithVars(ctx, pessimisticRollbackMaxBackoff, txn.vars), &PlainMutations{keys: keys})
+		err := committer.pessimisticRollbackMutations(newCleanupBackoffer(ctx, pessimisticRollbackMaxBackoff, txn.vars), &PlainMutations{keys: keys})
 		if err != nil {
 			logutil.Logger(ctx).Warn("[kv] pessimisticRollback failed.", zap.Error(err))
 		}
--- a/txnkv/transaction/txn_test.go
+++ b/txnkv/transaction/txn_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pingcap/kvproto/pkg/errorpb"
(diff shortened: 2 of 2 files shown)


txnkv/transaction/txn.go
```diff
@@ -2161,7 +2174,7 @@ func (txn *KVTxn) asyncPessimisticRollback(ctx context.Context, keys [][]byte, s
 			}
 		}
 
-		err := committer.pessimisticRollbackMutations(retry.NewBackofferWithVars(ctx, pessimisticRollbackMaxBackoff, txn.vars), &PlainMutations{keys: keys})
+		err := committer.pessimisticRollbackMutations(newCleanupBackoffer(ctx, pessimisticRollbackMaxBackoff, txn.vars), &PlainMutations{keys: keys})
```
zyguan, line 2177: `twoPhaseCommitter.cleanup` still creates backoffers with `c.txn.vars` at `txnkv/transaction/2pc.go`, should it also use `newCleanupBackoffer`?

King-Dylan: Good catch — yes, those two rollback branches have the same issue. The non-interruptible action flag skips the action-level kill check, but retry backoff still calls `CheckKilled`.

Fixed in a05dfeb2cab40ea99d4fbb4d0b8c779e8ffbbc56:

- Use `newCleanupBackoffer` for `cleanupMutations` (`BatchRollback` for 2PC/failed async commit) and the failed pessimistic 1PC rollback branch.
- Preserve the existing store-derived context, `TxnStartKey`, backoff budget, and the committed/undetermined-result guards. This does not implement the separate statement-cleanup context proposal.
- Add a 16-case regression covering both kill mechanisms across optimistic/pessimistic 2PC, async commit, and pessimistic 1PC. Before the fix, all 8 live-store cases stop after the injected `NotLeader` response; after the fix, they retry successfully. The 8 store-cancellation controls still stop, and the original transaction kill state remains intact.

I also checked the pipelined branch: `broadcastToAllStores` only uses the supplied backoffer's context, and the ResolveLock workers create vars-free backoffers internally. I left those call sites unchanged because they do not use this backoffer's kill state for retrying.

zyguan (approved):

AndreMouche (approved):
LGTM

PR #2063, opened 2026-09-11
txnkv: preserve undetermined commit outcomes across retries and cancellation

### LINKED ISSUE #2062: Transactions may return ordinary errors when the commit result is undetermined
If a request that can commit a transaction loses its response, the transaction may already have committed. Several paths can lose or overwrite this uncertainty:

- **Pipelined DML:** [commitFlushedMutations](https://github.com/tikv/client-go/blob/f7a805b640fed900247fec2f95cee1971b78da21/txnkv/transaction/pipelined_flush.go#L328-L330) returns the underlying error without converting a recorded undetermined result to `ErrResultUndetermined`.
- **Async commit / 1PC retries:** after a lost prewrite response and a Region split, cancellation of retry batches or local 1PC fallback can prevent the original uncertainty from being recorded. Successful ordinary 2PC prewrites after 1PC fallback also do not establish whether the original 1PC request committed; later schema or timestamp errors can still hide that uncertainty.
- **Primary commit retries:** a generic `KeyError` clears the previous undetermined state in [commit.go](https://github.com/tikv/client-go/blob/f7a805b640fed900247fec2f95cee1971b78da21/txnkv/transaction/commit.go#L176-L184) and [txn_file.go](https://github.com/tikv/client-go/blob/f7a805b640fed900247fec2f95cee1971b78da21/txnkv/transaction/txn_file.go#L569-L573), even though it does not necessarily resolve the earlier request's outcome.
- **Server-reported `UndeterminedResult`:** cancellation can overwrite the response, or transaction handling can return the error without recording the internal undetermined state. This can allow other batch errors to mask it or trigger failure cleanup.

Returning an ordinary error may cause TiDB to skip its undetermined-result connection-close handling. Applications could retry an operation that has already committed.

While the outcome remains unknown, client-go should return `ErrResultUndetermined`, retain the internal undetermined state, and skip failure cleanup. Confirmed success and definitive failure should retain their respective results.

gengliqi:
Fix https://github.com/tikv/client-go/issues/2062
## Description
Prevent ordinary errors from masking undetermined commit outcomes across retries, protocol fallback, and cancellation, including pipelined DML.

Preserve the undetermined state, return `ErrResultUndetermined`, and skip failure cleanup while the outcome remains unknown. Definitive failures retain their existing behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction commit handling when the final outcome is unknown, preserving the correct undetermined-result status.
  * Prevented later RPC or region errors from incorrectly overwriting an already undetermined transaction state.
  * Correctly resolves previously uncertain outcomes when a definitive commit response or retry succeeds.
  * Preserved server-reported transaction results when requests are interrupted or sessions are terminated.
  * Improved async-commit fallback behavior when RPC errors occur during prewriting.

### DIFF
--- a/integration_tests/pipelined_memdb_test.go
+++ b/integration_tests/pipelined_memdb_test.go
@@ -36,6 +36,7 @@ import (
 	"github.com/tikv/client-go/v2/testutils"
 	"github.com/tikv/client-go/v2/tikv"
 	"github.com/tikv/client-go/v2/tikvrpc"
+	"github.com/tikv/client-go/v2/tikvrpc/interceptor"
 	"github.com/tikv/client-go/v2/txnkv"
 	"github.com/tikv/client-go/v2/txnkv/transaction"
 	"github.com/tikv/client-go/v2/txnkv/txnlock"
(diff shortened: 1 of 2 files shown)


ekexium (approved):

txnkv/transaction/prewrite.go
```diff
@@ -352,11 +352,10 @@ func (action actionPrewrite) newSingleBatchPrewriteReqHandler(c *twoPhaseCommitt
...
 	if err != nil {
-		// If we fail to receive response for async commit prewrite, it will be undetermined whether this
-		// transaction has been successfully committed.
-		// If prewrite has been cancelled, all ongoing prewrite RPCs will become errors, we needn't set undetermined
-		// errors.
-		if (handler.committer.isAsyncCommit() || handler.committer.isOnePC()) && handler.sender.GetRPCError() != nil && atomic.LoadUint32(&handler.committer.prewriteCancelled) == 0 {
+		// A lost async/1PC response can leave the result unknown. Ignore RPC errors caused
+		// by another failed batch's cancellation. A split can disable 1PC locally, whereas
+		// disabling async commit requires TiKV to confirm fallback.
+		if (handler.committer.isAsyncCommit() || handler.req.Prewrite().TryOnePc) && handler.sender.GetRPCError() != nil && atomic.LoadUint32(&handler.committer.prewriteCancelled) == 0 {
```
zyguan, line 358: Is the following concurrent ordering expected?

1. Starts two async-prewrite batches in separate regions.
2. Obtains a definitive async-fallback response for one batch but delays delivering it.
3. Lets the other batch's RPC error populate `undeterminedErr`.
4. Delivers the already-obtained fallback response.

Because the failed prewrite prevents any commit request, the transaction appears definitively uncommitted, but the recorded error remains and becomes `ErrResultUndetermined`, also skipping cleanup. The existing test `TestConfirmedFallbackWithRPCError` covers the opposite ordering; should previously recorded async uncertainty be reconciled when fallback is later confirmed?

zyguan (approved):

txnkv/transaction/prewrite.go
gengliqi: Good catch. I reproduced this ordering on both the PR base and the current head, so this is a pre-existing issue.

PR #2074, opened 2026-09-18
txnkv: wait for async BatchGet retry processing on cancel

### LINKED ISSUE #2073: txnkv: async BatchGet can return before retry workers exit on cancellation
## Summary

`asyncBatchGetByRegions` can return on context cancellation while retry workers it submitted through `cb.Executor().Go(...)` are still running. Region-error and lock-error responses both spawn such workers. The coordinator stops driving `runloop.Exec` when it returns a context error, but does not wait for workers that already captured the result collector and `KVSnapshot`.

This is a request-lifecycle problem, not just a `SnapshotRuntimeStats` getter problem.

## Impact

1. **Late result collection can crash the process.** `BatchGetWithTier` captures its named result map `m` in `collectF`. On the error path, `return nil, err` sets `m` to nil. A retry worker that processes a response afterward can call `collectF`, which writes `m[string(k)] = v` and panics with `assignment to entry in nil map`. This does not require runtime stats to be enabled.
2. **A canceled statement can contaminate a later statement's stats.** Response accounting looks up `s.mu.stats` when the late worker processes the response. In an explicit TiDB transaction, consecutive statements can reuse the same `KVSnapshot` while attaching different stats objects. The canceled statement's scan/time details may then be recorded in the next statement's stats. Since #2048, point-response payload bytes can also be attributed to the next statement and affect its RU accounting. `KVSnapshot.mu` serializes these writes but does not preserve request ownership.
3. **Returning does not establish stats quiescence for generic callers.** A caller that keeps the stats object attached and reads it after a canceled `BatchGet` returns can overlap with a late worker's write. TiDB's normal point-get teardown detaches stats before reading them, so this is an API-level hazard rather than an observed TiDB getter race.

## Reproduction and evidence

The [investigation on #2069](https://github.com/tikv/client-go/pull/2069#issuecomment-5727618571) reports a deterministic TiDB reproduction using an embedded UniStore and a client wrapper that pauses a retry worker after receiving its response and before collecting it. With async BatchGet enabled, canceling the request lets `BatchGet` return while that worker is paused; releasing it then triggers the nil-map panic on unmodified client-go, even with runtime stats disabled. A synchronous multi-region control drains its workers before returning. The same investigation reports cross-statement payload misattribution when the panic is contained for observation. No real TiKV cluster reproduction or production frequency measurement was reported.

A reachable SQL shape is a multi-Region `Batch_Point_Get` (`WHERE id IN (...)`) that encounters a lock or Region error, followed by query cancellation (`KILL QUERY` or an execution timeout). Cross-statement attribution additionally requires the same session to continue an explicit transaction on the reused snapshot.

Relevant code:

- [`asyncBatchGetByRegions`: cancellation exits the `runloop.Exec` loop](https://github.com/tikv/client-go/blob/f7a805b640fed900247fec2f95cee1971b78da21/txnkv/txnsnapshot/snapshot_async.go#L45-L87)
- [Retry workers submitted for Region and lock errors](https://github.com/tikv/client-go/blob/f7a805b640fed900247fec2f95cee1971b78da21/txnkv/txnsnapshot/snapshot_async.go#L155-L185)
- [`BatchGetWithTier`'s captured result map and error return](https://github.com/tikv/client-go/blob/f7a805b640fed900247fec2f95cee1971b78da21/txnkv/txnsnapshot/snapshot.go#L305-L324)
- [`mergePointResponse` writes the currently attached stats](https://github.com/tikv/client-go/blob/f7a805b640fed900247fec2f95cee1971b78da21/txnkv/txnsnapshot/snapshot.go#L1003-L1016)

## Cross-statement stats timeline

Statements A and B below run **sequentially in the same TiDB session and explicit transaction**. They reuse one `KVSnapshot`, but each statement has its own `SnapshotRuntimeStats`.

```mermaid
sequenceDiagram
    participant S as TiDB session (one explicit transaction)
    participant K as Shared KVSnapshot
    participant W as Retry worker from statement A
    participant A as Stats A
    participant B as Stats B

    S->>K: Statement A attaches Stats A
    K->>A: s.mu.stats = Stats A
    S->>K: Async BatchGet A (multiple Regions)
    K->>W: Start retry after lock or Region error
    W->>W: Receive response, then pause before collecting it
    S->>K: Cancel statement A
    K-->>S: BatchGet A returns context error
    S->>K: Close A and detach Stats A
    Note over S,K: Same session and transaction now starts statement B
    S->>K: Statement B attaches Stats B
    K->>B: s.mu.stats = Stats B
    W->>K: Process A's late response via mergePointResponse
    K->>B: Write A's scan/time detail and payload bytes
    Note over B: Existing scan/time contamination, payload bytes added by #2048
```

The snapshot mutex makes each write atomic, but the late worker selects the stats object currently attached to the snapshot, so it can charge A's work to B.

## Expected behavior

After an async BatchGet returns, no worker from that request should still be able to collect results or write response stats. The cancellation path should stop or join outstanding retry work before returning, with regression coverage for cancellation after a retry response arrives but before it is collected. Cover both the nil-map panic and cross-statement stats attribution.

Coordinate this with #2033, which reports a different cancellation race where `RunLoop.Append` can block after `Exec` exits. A plain worker wait must not become stuck on callback scheduling.

The missing worker join predates #2048/#2069 and dates back to async BatchGet (#1646); #2048 adds a new payload/RU manifestation of the same lifecycle gap.

zyguan:
Close #2073

## What changed

Fix async `BatchGet` cancellation returning before lock- or Region-error retry processing finishes.

The async path now waits until retry processing can no longer collect results or update snapshot runtime stats. Completion scheduling remains outside that wait to avoid coupling this fix to the separate `RunLoop.Append` cancellation race (#2033).

## Test

- `go test ./txnkv/txnsnapshot`
- `go test -race ./txnkv/txnsnapshot -run '^TestAsyncBatchGetCancellationWaitsForRetryWorker$' -count=20`

Includes a deterministic regression test that pauses a lock-retry worker, cancels `BatchGet`, and verifies it does not return until the retry processing is released.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cancellation handling for asynchronous batch retrievals.
  - Operations now wait for active retry work to finish before returning, preventing premature completion and inconsistent results or statistics.
  - Canceled requests now reliably return the cancellation status after in-progress retry processing exits.

- **Tests**
  - Added coverage for cancellation while an asynchronous retry is still running.
  - Strengthened multi-region asynchronous batch retrieval test coverage.


txnkv/txnsnapshot/snapshot_async_test.go
```diff
@@ -0,0 +1,151 @@
...
+	// The initial requests use SendRequestAsync. A synchronous BatchGet here is
+	// therefore the retry worker spawned after the injected lock response.
+	resp, err := c.Client.SendRequest(ctx, addr, req, timeout)
+	if req.Type != tikvrpc.CmdBatchGet {
+		return resp, err
+	}
+
+	close(c.retryStarted)
+	<-c.releaseRetry
+	if ctx.Err() != nil {
```
cfzjywxk, line 57: Could we also cover Region-error retries and let the paused retry return a successful response after cancellation, then reuse the snapshot with fresh runtime stats to verify that collection finishes before BatchGet returns and stats do not leak into the next request?

cfzjywxk (approved):

wfxr (approved):

PR #2075, opened 2026-09-20
txnkv: back off when TiKV ignores lock hints

lcwangchao:
## Related issue

Ref pingcap/tidb#71344

When a read RPC returns a lock that was already included in that request's `ResolvedLocks` or `CommittedLocks`, back off with `BoTxnLockFast` before resolving and retrying it. This avoids a tight retry loop if TiKV ignores the lock hints.

The request-specific hints are passed through `ResolveLocksOptions`, so concurrent PointGet and BatchGet requests use the hints from the exact RPC response rather than the helper's current shared state.

The async BatchGet path also keeps the request that produced the response when falling back to the synchronous retry path.

## Tests

- `go test -mod=readonly ./txnkv/txnlock ./txnkv/txnsnapshot ./tikv -run 'TestBackoffOnLockHintsInRequestOnlyForRead|TestKV/(TestResolveLocksWithOptsBacksOffOnlyForRead|TestSnapshotReadsBackOffWhenServerReturnsHintedLock)$'`
- `go vet -mod=readonly ./txnkv/txnlock ./txnkv/txnsnapshot ./tikv`

## Release note

None

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved snapshot reads when previously resolved or committed transaction locks are encountered.
  - Read requests now use faster retry handling for recognized locks, reducing unnecessary lock-resolution work.
  - Point gets and batch gets preserve lock information across retries for more consistent results.
  - Autocommit point reads continue resolving unexpected locks correctly instead of ignoring them.
  - Batch reads now stop retrying and return a timeout when a lock cannot be resolved within the retry limit.

- **Tests**
  - Added coverage for lock-hint handling across synchronous and asynchronous snapshot reads.


wuhuizuo:
/retest 

for diag the ci failure

tikv/kv_test.go
```diff
@@ -115,6 +117,36 @@ type splitRegionKeyErrorMockClient struct {
...
+	cmd           tikvrpc.CmdType
+	lockInfo      *kvrpcpb.LockInfo
+	requestCount  atomic.Int32
+	hintedRequest atomic.Bool
+}
+
+func (c *lockHintErrorMockClient) SendRequest(ctx context.Context, addr string, req *tikvrpc.Request, timeout time.Duration) (*tikvrpc.Response, error) {
+	if req.Type == c.cmd {
+		call := c.requestCount.Add(1)
+		if call == 1 {
```
wfxr: Could we add a test where the server repeatedly returns the same lock despite `committed_locks`, and verify retries stop when the backoff budget is exhausted? Covering async BatchGet would also help validate that each response uses its corresponding request’s hints. The rest LGTM.

wfxr (approved):

txnkv/txnsnapshot/snapshot.go
```diff
@@ -598,11 +598,12 @@ func (s *KVSnapshot) handleBatchGetRegionError(bo *retry.Backoffer, batch *batch
...
-func (s *KVSnapshot) handleBatchGetLocks(bo *retry.Backoffer, lockInfo *batchGetLockInfo, cli *ClientHelper) error {
+func (s *KVSnapshot) handleBatchGetLocks(bo *retry.Backoffer, lockInfo *batchGetLockInfo, cli *ClientHelper, req *tikvrpc.Request) error {
 	resolveLocksOpts := txnlock.ResolveLocksOptions{
-		CallerStartTS: s.version,
-		Locks:         lockInfo.locks,
-		Detail:        s.GetResolveLockDetail(),
+		CallerStartTS:      s.version,
+		Locks:              lockInfo.locks,
+		Detail:             s.GetResolveLockDetail(),
+		LockHintsInRequest: txnlock.NewLockHintsInRequest(req.ResolvedLocks, req.CommittedLocks),
```
zyguan: Any reason for passing `req *tikvrpc.Request` to `handleBatchGetLocks`? Can we juse pass the hints here?

zyguan (approved):

PR #2076, opened 2026-09-21
locate: ignore write-path ServerIsBusy for store read state

LykxSassinator:

Ref tikv/tikv#20076, related to tikv/client-go#2028.

`ServerIsBusy` can be produced by either the read path (read pool, coprocessor, raftstore local reader) or the write path (write flow control, e.g. `pending_compaction_bytes` exceeding the soft limit). In `onServerIsBusy`, the store's read-side state was updated regardless of the request type:

- `updateServerLoadStats` was called unconditionally in the `estimated_wait_ms != 0` branch;
- `markAlreadySlow` was called unconditionally in the `estimated_wait_ms == 0` branch, which is the branch write flow-control errors take.

So a write-only rejection pinned the shared `Store` client-side slow score to its maximum. That score only gates **non-leader** replica selection, so non-leader reads started avoiding a store that serves reads normally, collapsing its measured read load. PD's hot-read scheduler ranks stores by measured read load, so the store looked like the coldest one and became a destination for read hotspots (including creating replicas on it). When compaction caught up and throttling stopped, the store was saturated by the redirected read traffic.

Only update the store's read load/health state when the rejected request is a read request that can actually observe read-side pressure:

- `isReadReqForBusy` is a new predicate for read requests that can be rejected by TiKV's read path. It is intentionally broader than `isReadReq`: raw KV reads (`CmdRawGet`, `CmdRawBatchGet`, `CmdRawScan`, `CmdRawGetKeyTTL`) and `CmdBufferBatchGet` also go through `Storage::read_pool_spawn_with_busy_check` and can carry a read-side `ServerIsBusy`, so they must keep updating the read state.
- The read-state updates in `onServerIsBusy` (`updateServerLoadStats` and `markAlreadySlow`) are gated by that predicate.
- The `#2028` leader-busy probe and all retry/backoff behavior are unchanged and stay request-type agnostic — an existing test (`TestReplicaReadAccessPathByTryIdleReplicaCase`) asserts the probe fires for a write request as well.

### Tests

- New `TestRegionRequestToThreeStores/TestServerIsBusyOnlyTaintsReadStateForReadRequests`: fails before the fix (`CmdPrewrite`/`CmdRawPut` marked the store read-slow and polluted `EstimatedWaitTime`), passes after; it also asserts reads (`CmdGet`, `CmdRawGet`, `CmdBufferBatchGet`) still update the read state.
- `go test ./internal/locate/...`
- `go test --tags=intest ./internal/locate/...`
- `go test ./...`
- `golangci-lint run ./internal/locate/...`

### Compatibility

No public API change. The behavior change is limited to internal store health/load bookkeeping on write-path `ServerIsBusy`; leader reads are unaffected because leader selection does not consult the slow flag.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of busy-server responses so write requests no longer affect read-specific store health or load measurements.
  * Expanded accurate read-health and wait-time tracking for additional read operations, including raw key-value reads, batch reads, lock scans, and checksum requests.
  * Helps maintain better store selection and read performance when servers experience backpressure.


internal/locate/replica_selector.go
```diff
@@ -594,8 +594,18 @@ func (s *replicaSelector) onServerIsBusy(
...
+		// control, e.g. pending compaction bytes exceeding the soft limit).
+		// Only a read-path rejection reflects the store's read health, so a
+		// write-path rejection must not update the shared read load/health
+		// state. Otherwise write backpressure would divert non-leader reads
+		// away from a store that serves reads normally, collapse its measured
+		// read load and turn it into a hot-read destination (tikv/tikv#20076).
 		if serverIsBusy.EstimatedWaitMs != 0 {
-			ctx.Store.updateServerLoadStats(serverIsBusy.EstimatedWaitMs)
+			if isReadReqForBusy(req.Type) {
+				ctx.Store.updateServerLoadStats(serverIsBusy.EstimatedWaitMs)
```
ekexium, line 607: Can we add a comment to `loadStats` to explicitly state that it is only for "read pool" to avoid confusion.

internal/locate/replica_selector.go
```diff
@@ -694,6 +706,29 @@ func isReadReq(tp tikvrpc.CmdType) bool {
...
+// (Storage::read_pool_spawn_with_busy_check), the coprocessor read pool, or the
+// raftstore local reader. Only such requests can carry a read-side ServerIsBusy,
+// so only they should update the store's read load/health state in
+// onServerIsBusy.
+//
+// It is intentionally broader than isReadReq: raw KV reads and BufferBatchGet
+// also go through the read-pool busy check, and their ServerIsBusy must not be
+// mistaken for write-side backpressure (tikv/tikv#20076).
+func isReadReqForBusy(tp tikvrpc.CmdType) bool {
+	switch tp {
```
ekexium, line 729: How about RawChecksum and ScanLock? Can we somehow ensure this function is properly maintained as the codebase evolve..?

LykxSassinator: Good catch. TiKV's `scan_lock` and `raw_checksum` call `ReadPool::spawn_handle` directly (they skip `read_pool_spawn_with_busy_check`), so they can still carry a read-side (deadline exceeded) `ServerIsBusy`; added both in 466e063.

For maintenance, `TestIsReadReqForBusyCoversAllCmdTypes` now enumerates every command type known to `tikvrpc.CmdType.String()` and fails unless it is classified as read or non-read, so adding a new command type forces an explicit decision. A TiKV-side direction field would remove the heuristic entirely, but this keeps the client-side classification from silently drifting.

internal/locate/replica_selector.go
LykxSassinator: Done in 466e063: documented `Store.loadStats` and `updateServerLoadStats` as read-pool only, so it is explicit that the estimated wait must not come from a write-path rejection.

ekexium (approved):

cfzjywxk (approved):

